PR #948, opened 2014-02-19
Highlight online+active in results

stilez:
"class=grey" isn't really useful with a significant number of temporary leases, because you can't quickly pick out by eye, the few current or active leases from a long list of inactive or past leases.

This highlights the "online" and "active" leases that currently exist, so it's much easier to see the active or online devices in the leases table

### DIFF
--- a/usr/local/www/status_dhcp_leases.php
+++ b/usr/local/www/status_dhcp_leases.php
@@ -343,9 +343,17 @@ foreach ($leases as $data) {
 		if ($data['act'] != "active" && $data['act'] != "static") {
 			$fspans = "<span class=\"gray\">";
 			$fspane = "</span>";
-		} else {
+		} else
 			$fspans = $fspane = "";
-		}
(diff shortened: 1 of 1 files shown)


rbgarga:
@stilez Do you have plans to make changes suggested by @wagonza? Otherwise this can be closed.

stilez:
I'd like to, but not confident enough about how pfSense css formating is done (which files, at what level, etc) to be sure of it.  

If someone can reply here with css suitable for highlighting online/active items in a table, and which file/s to add it to, I can do the rest happily.  

(I'm assuming it would ideally be an extra css class in the theme css file/s, not just in the leases page, since the same improvement could conceptually apply to other tables with active/inactive members, and should be re-themable by user themes)

rbgarga:
pfSense is moving to bootstrap, as you can see in https://github.com/SjonHortensius/pfsense. I'm not sure if it's worth it to make this change to all themes at this point.

PR #955, opened 2014-02-20
ip2ulong() error trapping & resubmit ip_compare()

stilez:
Resubmit (https://github.com/pfsense/pfsense/pull/952) which tightened ipcmp(), ip_greater_than() and ip_less_than(), and extended them to IPv6 and bad data trap. 

2 modifications: error trapping in ip2ulong() and avoiding inet_pton() as required.

--- Info ---

1) ip_less_than() and ip_greater_than() both fail for IPv6 as they rely on 32 bit ip2ulong

2) ip_less_than() and ip_greater_than() both also fail for data validation as they do not trap bad data, or detect the malformed case where one arg is IPv4 and the other IPv6 due to improper user input. (One will always be "greater" or "less" than the other so one of these functions will attempt to return True if they don't error out first)

3) ip2long32() and ip2ulong() don't detect ip2long() returning an error, fix this so calling code can know if the returned value is good if not. (Example: code might accept IPv4 + v6 but mistakenly use ip2ulong() at some point for both)

4) It's easier and cleaner if ip_less_than() and ip_greater_than() call ip_compare() than ipcompare() calls the two separate sub-functions. Effectively they are aliases for it, it saves a function call.

4) Previously there was no way to distinguish "equal" from "bad data" (except "bad data" wasn't trapped anyway); now it returns '' for bad data and 0 for equal.

5) Switches to a more obvious and 'canonical' name ip_compare() for the main function, ipcmp() remains an alias to ensure nothing breaks

6) Avoid inet_pton() as stated by Ermal. 

Testing two IPv6 addresses as < > = has extra if() clauses since Net_IPv6 is 'expensive' and IPv6 has multiple valid formats.

### DIFF
--- a/etc/inc/util.inc
+++ b/etc/inc/util.inc
@@ -369,14 +369,18 @@ function long2ip32($ip) {
 	return long2ip($ip & 0xFFFFFFFF);
 }
 
-/* Convert IP address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms. */
+/* Convert IP address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms. 
+   Returns '' if not valid IPv4. */
 function ip2long32($ip) {
-	return ( ip2long($ip) & 0xFFFFFFFF );
+	$a = ip2long($ip);
+	return ($a === False ? '' : $a & 0xFFFFFFFF);
 }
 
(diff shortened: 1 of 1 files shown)


stilez:
WIll resubmit this shortly (see last comment(s) on closed PR #972)

PR #973, opened 2014-02-25
Tighten is_linklocal() [now rebased: see #1786]

stilez:
is_linklocal has a few issues, including validating as linklocal, addresses that aren't linklocal according to RFC 4291, and doing no validation of reasonableness on any %(interface) part that may exist, allowing this to contain arbitrary text.

Changes made:

1) Tests for IPv4 linklocal addresses - while not much needed, they should probably be recognised as code handling linklocal may reasonably expect is_linklocal() to be IPv4/IPv6 agnostic.

2) For IPv6, it tests at least, that the purported interface is [0-9a-z]+ otherwise user input or other text such as "fe80::%\n;ARBIRARYTEXT;" would be validated as a linklocal address and inserted into pf and perhaps other places without further detection, leading to possible vulnerabilities. But it doesn't test more than this (and probably should test for valid interface name).

### DIFF
--- a/etc/inc/util.inc
+++ b/etc/inc/util.inc
@@ -369,9 +369,11 @@ function long2ip32($ip) {
 	return long2ip($ip & 0xFFFFFFFF);
 }
 
-/* Convert IP address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms. */
+/* Convert IP address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms. 
+   Returns '' if not valid IPv4. */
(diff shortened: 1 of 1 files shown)


rbgarga:

stilez:
@rbgarga - rebased/resubmitted almost identically, at PR #1786.
Two minor improvements added, otherwise identical code: - scope/interface part length ensured not pathologically long, and a weakness in Net_IPv6 that allows it to return shorter strings rather than "not valid" if the data isn't IPv6, is tested.

PR #982, opened 2014-02-27
Avoid @unlink() in unlink_if_exists()

stilez:
Error suppression is generally a bad idea, especially if there could be several reasons unlink() fails and only one reason (file not existing) is to be ignored.

Use file_exists() to test existence first.


stilez:
resubmitted as #2153

PR #1081, opened 2014-04-13
Certificate Manager, call packages plugin function to show used certificates

PiBa-NL:
Certificate Manager, call packages plugin function to show used certificates


PiBa-NL:
@ermal what is the proper way to 'integrate' packages into pfSense base functionality?

ermal:
I merged this on master since it is good to have the functionality.
Though i think such implementation need to be done with a longer though in terms of providing API for other scenarios similar to this.

PR #1197, opened 2014-05-18
Hook for user shutdown script - "/etc/rc.custom_shutdown"

dreamcat4:
Solves https://redmine.pfsense.org/issues/3667.


rbgarga:
My understanding it patch won't be improved, then I'm closing the pull request.

PR #1221, opened 2014-05-31
Update functions.inc.php

gh0stshell:
Corrected math and sysctl commands to show correct memory, if using original code if you add it up it does not match the total mem shown pulled using hw.phymem in the system information widget


rbgarga:
No answers almost an year later, closing. It can be re-opened if you want to discuss. Thanks!

PR #1227, opened 2014-06-04
Added options to configure multiple instances of dnsmasq.

rabisuz:
The easiest (and sometimes only) way to allow for more complex dnsmasq configurations is to run multiple instances. I needed some overrides for one interface, but not for the others. Moreover, having the same hostname on multiple networks is quite troublesome with only one instance. The solution was to filter the hosts file based on the interfaces dnsmasq is configured to listen to.

Note: A more elegant solution to filter the hosts file is to modify dhcpleases. But that requires changing a binary, which is in the tools repository.


rbgarga:
You mentioned it was not working, did it change since then? Are you still working on it?

rabisuz:
I have not updated it in a while, but it was working again when I reopened this pull request. I have to merge the changes since then and update to 2.2 on both my production and test machine to verify it's still working after that.

rbgarga:

cbuechler:
This would need a good deal of updating for Bootstrap. But beyond that, I don't think this is enough in demand to justify adding that level of complexity. I can see the occasional circumstance like originally described here where it could be helpful, but that's a very rare scenario. It's a lot of added technical debt for a feature in very little demand.

PR #1231, opened 2014-06-08
RRD Graph Fixes

N0YB:
Replace the current date on graphs with graph begin and end date range.
Current time is not useful in some views.  Particularly Previous Period views.

Correct graph period ranges (start & end) in get_dates function.

Fix graph updates to use updated start and end times.
Rather than a static start time and "now" for end time.
Which:
# Causes Absolute Timespan, Current Period and Previous Period graph lengths to grow indefinitely .
# Makes Previous Period views turn in to previous period plus current period.
# Prevents the period of both Current and Previous period views from updating to the new period when a period time boundary is crossed. 
Consolidate if not previous period checks into a single check.

Set the 'average' granularity to be used for each graph length.

For 'Current Period' view, leave graph end variable set to period end rather than forcing it to current time (now).

Add Graph Update Interval Setting.

### DIFF
--- a/usr/local/www/status_rrd_graph.inc
+++ b/usr/local/www/status_rrd_graph.inc
@@ -1,0 +1,129 @@
+<?php
+/* $Id$ */
+/*
+	status_rrd_graph.inc
+	Part of pfSense
+	Copyright (C) 2009 Seth Mos <seth.mos@dds.nl>
+	All rights reserved.
+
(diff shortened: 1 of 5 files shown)


rbgarga:
No rebase or re-submit after almost an year. Closing. It can be reopened at any time. Thanks!

PR #1248, opened 2014-07-02
Explicitly set cipher list for BEAST-mitigated option and prefer the AES...

b:
...-GCM which were disabled previously. Note that RC4 suites should be removed entirely in the near future, but are left in place for now to avoid scaring people.


rbgarga:
No answers for a long time, also changes were made in this regard by commit 3a0a59c8

PR #1264, opened 2014-08-21
-fix different processing of url alias between gui and background updates. (could create a invalid pf rule)

PiBa-NL:
-fix different processing of url alias between gui and background updates. (could create a invalid pf rule)
-ignore text after space character (which allows to use the spamhaus drop/edrop lists)


PiBa-NL:
Most was commited by: https://github.com/pfsense/pfsense/commit/6d1907a3d2d9729f37fe15d0291d3a0f9a85a25e

The improvement for is_port() i send now in a separate pullrequest:
https://github.com/pfsense/pfsense/pull/1459

PR #1291, opened 2014-09-25
php, log errors and warnings, declare static functions static / remove new object by reference

PiBa-NL:
php, log errors and warnings, declare static functions static / remove new object by reference


PiBa-NL:
For some not completely clear reason this turned out to be a bad idea.( 
https://github.com/pfsense/pfsense/commit/e8e494f30d8d913b88b521cf33540aacd4b4627f )

PR #1333, opened 2014-11-10
Network Ups Tools widget

SunStroke74:
Added NUT widget and code to autoupdate widget content through ajax calls.
Discussion at https://forum.pfsense.org/index.php?topic=80981.0

### DIFF
--- a/usr/local/www/includes/functions.inc.php
+++ b/usr/local/www/includes/functions.inc.php
@@ -355,17 +355,17 @@ function get_nutstats() {
 		$nut_config = $config['installedpackages']['nut']['config'][0];
 	} else {
 		// No NUT package installed, return
-		return "No NUT installed!";
+		return gettext("No NUT installed!");
 	}
 	// "Monitoring" field and get command to fetch ups data - upsdata_array[0]
 	if ($nut_config['monitor'] == "local") {
-		$data = "Local UPS";
+		$data = gettext("Local UPS");
 		$cmd  = "upsc {$nut_config['name']}@localhost";
 	} elseif ($nut_config['monitor'] == "remote") {
-		$data = "Remote UPS";
+		$data = gettext("Remote UPS");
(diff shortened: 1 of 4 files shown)


usr/local/www/includes/functions.inc.php
```diff
@@ -346,4 +347,136 @@ function get_interfacestatus() {
 	return $data;
 }
 
+function get_nutstats() {
+	global $config;
+	$data = "";
+	$cmd = "";
+	$nut_config = $config['installedpackages']['nut']['config'][0];
```
rbgarga: You should check isset($config['installedpackages']['nut']['config'][0]) to avoid possible PHP warnings when package was never installed on the system.

usr/local/www/includes/functions.inc.php
```diff
@@ -346,4 +347,136 @@ function get_interfacestatus() {
...
+				foreach ($lines as $line) {
+					$line = explode(':', $line);
+					$ups[$line[0]] = trim($line[1]);
+				}
+			}
+		}
+	} elseif (isset($nut_config)) {
+		// No service running
+		if ($nut_config['monitor'] == "snmp") {
+			$condition = "NUT enabled but service not running!\nSNMP UPS may be unreachable.";
```
rbgarga: Please add gettext() calls to strings

rbgarga:
It would make more sense to add this code on pfSense-packages repo, to be installed together with nut package.

dld-r00f:
How it can be added to pfSense-packages repo? It is contained ajax updater part located in functions.inc.php and ajax.js.

rbgarga:
@dld-r00f You can see an example here: https://github.com/pfsense/pfsense-packages/tree/master/config/widget-havp

doktornotor:
Yeah indeed, add this to nut package instead and create your own .inc and .js for the required stuff, instead of patching. Really don't see how this belongs to /usr/local/www/includes/functions.inc.php

cbuechler:
This should be part of the package, please submit there.

PR #1337, opened 2014-11-13
Allow "From" field in RFC822 format

thiagoc:
This allow notification email "From" field with a custom format rather the default e-mail address.


rbgarga:
Sorry about the long delay. Can you please rebase your fork? Patch doesn't apply clean anymore

rbgarga:
PHP also provides a function to deal with it, it would be preferred over this regex.

http://php.net/manual/pt_BR/function.mailparse-rfc822-parse-addresses.php

thiagoc:
Hi @rbgarga, this PHP function requires the mailparse module, I'm trying to figure out how to install it. Do you think this is better than the regex solution? If so, how can I add the mailparse module?

thiagoc:
I have simplified the code on the pull #1726, please review it.

PR #1375, opened 2014-12-10
Update interfaces.php - Allow /31 Subnets

electro47h:
Removed exclusion of 31 from subnet list boxes. Starting in FreeBSD 9.x and above, the operating system supports /31 (255.255.255.254) subnets. This allows the use of point-to-point addresses for interfaces. Some ISP's (Cox Communications) are implementing IP-VPN services that use these 31-bit subnet masks. Customers are provided a point-to-point IP and a Gateway. There is no need for the 2 extra IP addresses that a /30 will give you, so a /31 is used instead.


cbuechler:
Thanks much for the help! I committed functionally the same thing but fixing the white space as well, which was easier to do myself than merging and fixing. This will be in 2.2.

PR #1380, opened 2014-12-11
pkg_edit.php fix interfaces_selection option while using row_helpers

marcelloc:
current code does not show interface_selection properly while using row_helpers

### DIFF
--- a/usr/local/www/pkg_edit.php
+++ b/usr/local/www/pkg_edit.php
@@ -115,7 +115,7 @@ if($config['installedpackages'] && !is_array($config['installedpackages'][xml_sa
 	$config['installedpackages'][xml_safe_fieldname($pkg['name'])]['config'] = array();
 
 // If the first entry in the array is an empty <config/> tag, kill it.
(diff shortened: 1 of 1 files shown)


doktornotor:
Resubmitted as #1920

doktornotor:
@marcelloc Since my resubmit was not acceptable due to authorship, you can either take that code from my PR referenced above and resubmit this yourself (it will probably only keep working for a couple of days at best before there's another bunch of conflicts), or live with the bug in pfSense code.

Have a nice day.

marcelloc:
@doktornotor, You're not wasting my time in any way. Thank's for all work you're doing to fix a lot of bugs on pfSense packages and core. In the past, contributions was something like collaboration with main devs but IMHO these days, free code and time we waste to get pfSense better does not get any attention and good code get invalidated by minor issues that can be fixed by reviewer it it's receives it as collaboration instead of junk code.

rbgarga:
@sbeaver-netgate can you please check this issue when you are converting pkg_edit.php?

PR #1389, opened 2014-12-15
Allow option to specify just 1 of user and pass in OpenVPN .up file

phil-davis:
As per comment in https://redmine.pfsense.org/issues/3633 sometimes the server end only requires a password, no username. Usually 1 long string that serves as the hard-to-guess authentication. OpenVPN expects something to be on the first line of the ".up" file - traditionally called the username. It also insists on the second line being present, but is happy with it being empty - this is the authentication information traditionally called "password".
Let the user put the single piece of authentication information in either the Username or Password field on the web GUI - whichever they feel comfortable calling it. In the ".up" file it has to always be the first line to keep OpenVPN happy.

### DIFF
--- a/etc/inc/openvpn.inc
+++ b/etc/inc/openvpn.inc
@@ -712,11 +712,15 @@ function openvpn_reconfigure($mode, $settings) {
 				$conf .= "ifconfig-ipv6 {$ipv6_2} {$prefix}\n";
 		}
 
-		if ($settings['auth_user'] && $settings['auth_pass']) {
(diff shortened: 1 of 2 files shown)


phil-davis:
I guess this "feature" is being left for 2.2.1?

phil-davis:
IMHO this should be good to go now for 2.2.1

mmxe:
Can we get this into 2.2.2?

rbgarga:
@phil-davis looks like original fork/branch is gone, can you please re-submit?

phil-davis:
Closed in favor of #1701

PR #1403, opened 2014-12-26
php error logging, helps identify when php 'crashes' due to for example a double function declaration

PiBa-NL:
php error logging, helps identify when php 'crashes' due to for example a double function declaration, redmine: https://redmine.pfsense.org/issues/4143

### DIFF
--- a/etc/inc/config.lib.inc
+++ b/etc/inc/config.lib.inc
@@ -924,13 +924,12 @@ function pfSense_clear_globals() {
 		$errtype = $error["type"];
 		if ($errtype != E_NOTICE) {
 			// there are a lot of 'innocent' errors.. so skip reporting E_NOTICE
-		$errfile = $error["file"];
+			$errfile = $error["file"];
 			$errline = $error["line"];
 			$errstr  = $error["message"];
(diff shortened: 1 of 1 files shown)


etc/inc/config.lib.inc
```diff
@@ -919,6 +919,20 @@ function get_config_backup_count() {
...
 
+	$error = error_get_last();
+	if( $error !== NULL) {
+		$errtype = $error["type"];
+		if ($errtype != E_NOTICE) {
+			// there are a lot of 'innocent' errors.. so skip reporting E_NOTICE
+			$errfile = $error["file"];
+			$errline = $error["line"];
+			$errstr  = $error["message"];
+			$str = "PHP ERROR Type:$errtype, File:$errfile, Line: $errline, Message: $errstr";
```
rbgarga: These variables are unecessary ($errfile, $errline and $errstr), you can print $error["file"], $error["line"] and $error["message"] direct when setting $str, just enclose them in {}.
```
$str = "PHP ERROR Type:{$error['type']}, File:{$error['file']}, Line: {$error['line']}, Message: {$error['message']}";
```

cbuechler:
modified version of this committed

PR #1409, opened 2014-12-29
fixes captive portal bug where long ipfw rules fail to apply bug#4150

ehayon:
When captive portal creates ipfw rules, too many interface addresses can cause ipfw to return "rule too long"

See iussue #4150: https://redmine.pfsense.org/issues/4150

This is a problem when running a captive portal zone on > ~120 interfaces. This pull request breaks up the rules into chunks of 100 interface addresses. 

$rulenum will overlap 65532 when ~20,000 interfaces are added to CP zone. This patch assumes you will not approach 20K interfaces on a CP zone, a fairly reasonable limitation.

### DIFF
--- a/etc/inc/captiveportal.inc
+++ b/etc/inc/captiveportal.inc
@@ -526,25 +526,21 @@ add 65307 deny layer2 not mac-type ip,ipv6
 EOD;
 
 	$rulenum = 65310;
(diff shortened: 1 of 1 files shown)


loos-br:
Fixed in 2.2.5 and 2.3 with the use of tables.

Thanks!

PR #1415, opened 2015-01-01
Remove duplicate copyright

phil-davis:
Noticed these had the copyright twice


phil-davis:
and I was not sure whether to leave the one at the top or bottom of the list - in other files the ESF copyright has been put at top or bottom, it varied - seemed to be no standard implemented there.

PR #1417, opened 2015-01-02
Adding boot delay to alleviate overwritten boot loader configs

lalanza808:
Auto-upgrades on ALIX hardware for nanoBSD image results in an overwritten /boot/loader.conf. Upon rebooting, the image will not properly boot until the kernel boot delay is manually done via console and webConfig. Hopefully this will alleviate this issue, so I can upgrade all my pfSense nodes without having to travel to remote locations. Please let me know if this is the improper file to modify, or if this is not a feasible solution: lrallen03@gmail.com


cbuechler:
It's never been in there, and adding a delay for everyone where most all systems don't need it isn't good. Where you need to make such changes, make them in loader.conf.local which will survive upgrades.

lalanza808:
Gotcha, I didn't realize that appending .local to the modified loader.conf
would cause it to still be read.

This is a much easier fix. Thank you, sorry for any inconvenience.

PR #1420, opened 2015-01-04
Limit unbound so-rcvbuf: 8m

phil-davis:
Issue reported here: https://forum.pfsense.org/index.php?topic=78356.msg472781#msg472781
Most unbound doc places mention setting it at up to 8m. I'm sure it would be possible to investigate more and find a way to make unbound+FreeBSD be able to go higher than 8m. But probably 8m is sufficient for everyone anyway (judging by what the unbound docs seem to assume will be a good value on a busy system).
Anyway, here is my easy fix for this. Someone else feel free to investigate more if they really need to set so-rcvbuf higher.


wagonza:
As mentioned on the forum I have seen values up to 32m, so it is definitely possible to configure higher values - why it is not working I am not sure. 

Setting it to a maximum of 8m is more than likely sufficient for everyone though.

PR #1423, opened 2015-01-07
Dynamic DNS Status widget replace deprecated split call

phil-davis:
Forum report: https://forum.pfsense.org/index.php?topic=86417.0


rbgarga:
@ermal already pushed a fix for that, this one can be closed. Thanks!

PR #1424, opened 2015-01-07
Replace deprecated split with explode

phil-davis:
A few more of these I spotted.


rbgarga:
@ermal already changed these too

phil-davis:
Oh OK - I guess there is a delay in seeing his changes on the public-facing GitHUb.

PR #1425, opened 2015-01-07
"Like with like" - move a few functions to better places in the code (they are placed strangely)

stilez:
A few basic functions for ip comparison and subnet overlap testing are in an illogical place - 1000 lines away from all the other basic IP/subnet functions, in the middle of alias handling, CARP interface, and sig checking library functions. 

Moving to earlier in the file (next to other IP compare and subnet functions) for ease of future contributors.


cbuechler:
merged, thanks

PR #1426, opened 2015-01-08
Deprecated and non-static method messages

phil-davis:
Fix various files that can emit messages like:
PHP Strict Standards:  Non-static method SimplePie_Misc::array_unique()
should not be called statically, assuming $this from incompatible
context in /etc/inc/simplepie/simplepie.inc on line 5508
php-fpm[16262]: /rc.newipsecdns: PHP ERROR: Type: 8192, File:
/etc/inc/shaper.inc, Line: 4365, Message: Assigning the return value of
new by reference is deprecated

Some of these style messages have been reported on the forum, e.g.
https://forum.pfsense.org/index.php?topic=86448.0

I had RSS widget on, and simplepie sent the system beserk telling about
all this stuff.

### DIFF
--- a/etc/inc/shaper.inc
+++ b/etc/inc/shaper.inc
@@ -405,16 +405,16 @@ class altq_root_queue {
 
 		switch ($this->GetScheduler()) {
 		case "PRIQ":
-			$q =& new priq_queue();
+			$q = new priq_queue();
 			break;
 		case "HFSC":
-			$q =& new hfsc_queue();
+			$q = new hfsc_queue();
 			break;
 		case "CBQ":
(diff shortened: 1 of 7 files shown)


cbuechler:
I reverted this after seeing multiple reports of breakage that were apparently caused by these changes. The PHP errors are harmless (and their logging has since been disabled), so we'll just review this again later.

phil-davis:
Yes, I agree. I did some basic unit tests, and internet research on these and thought it was all simple to change to meet coding rules for latest PHP. But I didn't notice the difference it made - sorry about that, I tricked myself into thinking it was all good and easy!

PR #1438, opened 2015-01-14
make page note more consistent / multiple entries for UPnP

xbipin:
- make page note more consistent

- patch will allow the web GUI for UPnP to enter more user specified entries rather than just 4, I replaced the 4 boxes with one single one which would allow to add more entries separated by a comma


xbipin:
can some1 check and apply, it mostly has fixes in page formatting etc where on some pages the section about Notes etc dont match with other pages, actually there are many such pages but these are the first few i found and fixed, will do so to the rest once i get some time.

PR #1442, opened 2015-01-18
Fix inconsistent handling of seperator in easyrule, should fix #4233

oliwel:
Change method how seperators are created for easyrule scripts.
Should fix https://redmine.pfsense.org/issues/4233


rbgarga:

oliwel:
CLA signed

oliwel:
This is included in #1444

PR #1443, opened 2015-01-18
Correct a few typos that made these files unparsable

SjonHortensius:
services_unbound - added missing $-sign to variable
xmlrcp - removed redundant closing brackets. Some trailing whitespace
also got removed


rbgarga:
Looks line your fork is not in sync with upstream because the diff shows a lot more changes than you mentioned

SjonHortensius:
Aargh; github attached all other commits on the same branch... What you're actually seeing are the merges I did afterwards.

Not sure how to correct this; only  1eb66a8 was supposted to be attached

rbgarga:
You can rebase your fork with this commit on top and push it to github, the pull request will be automatically updated

SjonHortensius:
@rbgarga I've been reading a lot of git documentation; but I seem to be unable to do what you describe. I think that resetting my master branch to be equal to the upstream/master branch; and then re-applying my single commit should work, so I've tried ```git rebase --onto refs/remotes/upstream/master master``` but it won't work unfortunately. I'm afraid I need some help with this.

rbgarga:
Usually something like this should work:

# git fetch upstream
# git rebase upstream/master

If it doesn't work, you can save your commit using:

# git show COMMIT_HASH > patch

And then reset your repo using

# git fetch upstream
# git reset --hard upstream/master

Then re-apply the patch

# git apply patch

And finally push it

# git push -f origin

SjonHortensius:
```git fetch upstream``` dies immediately saying ```fatal: Refusing to fetch into current branch refs/heads/master of non-bare repository``` Is this what --update-head-ok is meant for?

PR #1447, opened 2015-01-21
new feature: captiveportal 3M sip2 connectivity plus captive portal focus()

nilesingalls:
This modification will allow the captive portal to authenticate against an ILS SIP2 server using your library card credentials.  This has been tested with Evergreen and SirsiDynix and should work with any SIP2 capable ILS. Additionally, a threshold can be defined which will prevent authentication if the patron exceeds a specified value in fines.  I also set focus to the auth_user form input on the captive portal page.

### DIFF
--- a/etc/inc/captiveportal.inc
+++ b/etc/inc/captiveportal.inc
@@ -47,6 +47,7 @@ require_once("config.inc");
 require_once("functions.inc");
 require_once("filter.inc");
 require_once("radius.inc");
+require_once("sip2.inc");
 require_once("voucher.inc");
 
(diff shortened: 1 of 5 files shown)


nilesingalls:
I attempted to do another pull for an unrelated change, and it added it here.  ignore the changes to interfaces.php (or include it - it fixed the condition where you are unable to revert back to value none for option IPv6 Configuration Type on any given interface)

rbgarga:
3M SIP2 is not a protocol we are willing to support at pfSense, at least by now. You can always add your custom PHP files to captiveportal and use them to authenticate users.

PR #1449, opened 2015-01-25
Initial support for disabling sync of the skew value that will allow adm...

rmiddle:
...in to control who is master and who is backup.

This add feature # 2834 ( https://redmine.pfsense.org/issues/2834 )

### DIFF
--- a/etc/rc.filter_synchronize
+++ b/etc/rc.filter_synchronize
@@ -59,7 +59,7 @@ function backup_vip_config_section() {
 		    (strpos($section['interface'], '_vip') === FALSE) &&
 		    (strpos($section['interface'], 'lo0') === FALSE))
 			continue;
-		if ($section['advskew'] <> "") {
+		if (($section['advskew'] <> "") && ($section['dontadvskew'] <> "ignore")) {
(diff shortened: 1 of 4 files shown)


cbuechler:
Appreciate the submission, Robert. Unfortunately, this has so many edge case caveats that it's really not something we can accept at all (we should probably have something other than "needs patch" to close out such feature requests). It's almost never desirable to do that, and adding a button tempting people to press it that'll break virtually every use case in sometimes difficult to troubleshoot ways to accommodate something maybe 2 people all time want, just isn't a good idea. 

As a system patch to your own system, this probably isn't bad, as you understand what you're doing with it and I'm sure it works for your specific use case. To accommodate every use case, there are a variety of bugs it would create. It's one of those things that adds technical debt without adding benefit to the project as a whole.

rmiddle:
I really need this feature and I have way to many firewall to do it alone.  What approach will you accept for this?

rmiddle:
What if I made the checkbox a simple Make this the Backup connection.  That would subtract 100 instead of adding 100.

PR #1451, opened 2015-01-27
multiple allow/deny entries for UPnP

xbipin:
This patch will allow the web GUI for UPnP to enter more user specified
entries rather than just 4, I replaced the 4 boxes with one single one
which would allow to add more entries separated by a comma

### DIFF
--- a/usr/local/pkg/miniupnpd.inc
+++ b/usr/local/pkg/miniupnpd.inc
@@ -98,26 +98,29 @@
 			$input_errors[] = 'You must specify a valid traffic shaping queue.';
 
 		/* user permissions validation */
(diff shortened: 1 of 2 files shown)


jim-p:
Using a rowhelper control would be better for this, easier to read/manage and still allows for an arbitrary number of entries.

xbipin:
ok ill do that but can some1 guide me on how to create branches and create separate pull requests for each pair of commits, i tried forking the project, then cloning on desktop, created new branch and pushed first pair of commits and created pull request but then the next patch i submit it still does udner the initial pull request.
sorry im new to github

PR #1452, opened 2015-01-27
fix for Bug #4238

xbipin:
fix for Bug #4238

### DIFF
--- a/usr/local/www/firewall_rules_edit.php
+++ b/usr/local/www/firewall_rules_edit.php
@@ -1098,7 +1098,7 @@ include("head.inc");
 				</div>
 			</td>
 		</tr>
-		<tr style="display:none" id="sprtable">
+		<tr id="sprtable">
 			<td width="22%" valign="top" class="vncellreq"><?=gettext("Source port range");?></td>
 			<td width="78%" class="vtable">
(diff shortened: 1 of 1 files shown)


rbgarga:
I don't think change the behaviour is the right fix here, I'll investigate a bit

xbipin:
actually if u look at other pages where this port selection is that code doesnt appear there thats y it works fine there and not here, i compared it with 2.1.5 files as well

xbipin:
same page if u look at destination port field it doesnt have that code

rbgarga:
I pushed a different fix for it, changing javascript to set width just before show the popup. Could you please test it and close this one?

xbipin:
i tried ur fix but still broken, try on the firewall_rules_edit.php which was actually reported where the issue was

xbipin:
to test do this, create 2 port aliases with names as a1 and a2, then goto firewall rules and add a new rule and in source port type a and ull see its still broken

rbgarga:
I was able to reproduce and to test the fix. I added a ?rev=1 to all places that load autosuggest.js to avoid browser cache, but, just in case, can you try to clear your browser's cache and try again?

xbipin:
ys works now

PR #1453, opened 2015-01-27
Traffic Shaper Wizard Upstream SIP Server

phil-davis:
Not being remembered and actioned.
Bug #4314
The corresponding *.inc files had the field name changed to upstream_sip_server but they had not been changed in the XML files.


ermal:
Thx Phil merged.

PR #1454, opened 2015-01-27
remove pointless filter on dhcp static mappings table

wrboyce:
The filtering on the static mappings table (services_dhcp.php) seems backwards; this page should show all configured static mappings regardless of how they are configured.

It does, however, make sense to hide some from status_dhcp_leases.php, as any static mappings without an IP  configured will appear in the list twice.

### DIFF
--- a/usr/local/www/services_dhcp.php
+++ b/usr/local/www/services_dhcp.php
@@ -107,7 +107,7 @@ if($config['installedpackages']['olsrd']) {
 				break;
 			}
 	}
-}	
+}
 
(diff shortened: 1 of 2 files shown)


rbgarga:

wrboyce:
Is this new? I've had patches merged in the past (#671).

rbgarga:
It's been required for almost a year

wrboyce:
Gotcha, I'll get the appropriate forms sorted this week sometime.

rbgarga:
Manually merged. Thanks!

PR #1455, opened 2015-01-28
multiple allow/deny entries for UPnP (rowhelper)

xbipin:
This patch will allow the web GUI for UPnP to enter more user specified
entries rather than just 4, I replaced the 4 boxes with one rowhelper
control


xbipin:
now atleast we have an option to add multiple entries, next im planning to add schedules to it and so far im able to add the config for that to the config xml file and read it out on the GUI but how do i trigger events based on the schedule? i mean what file do i need to be looking at to edit so i can trigger upnp daemon on and off based on schedules?

xbipin:
the other issue is the pkg_edit.php doesnt allow schedules dropdown menu control so are there any plans to add that or else without that using schedules wont be possible, i could add that but if there are no plans to merge it then it would seem like waste of time and effort.

xbipin:
i added schedule selection to pkg_edit.php in another pull request so other packages can also use schedules in them, once that is merged then can add the feature to upnp as well

xbipin:
Bump
i have tried the patch on multiple systems and upnp works perfectly fine from past many days

afbholly72:
Does this patch work on the 2.2 Release?

xbipin:
Yes tried and tested on 2.1.5 as well as 2.2, all good

afbholly72:
thanks...works great...great patch!

On Sun, Feb 1, 2015 at 12:05 PM, xbipin <notifications@github.com> wrote:

> Yes tried and tested on 2.1.5 as well as 2.2, all good
>
> —
> <https://github.com/pfsense/pfsense/pull/1455#issuecomment-72381391>.
>



-- 
Antonio Vitti
+1 415 710 9111 (USA)
antoniovitti@gmail.com

rbgarga:
Doesn't it need an upgrade config code to avoid breaking current config?

xbipin:
well it doesnt because the miniupnp.inc file rewrites the config everytime u press save

rbgarga:
What happens if user has a config with all those 4 fields filled, and upgrade? the rules user had there will get lost, no?

xbipin:
actually it will remain in config but as soon as he activated upnp and clicks save they r overwritten

rbgarga:
Then it needs an upgrade config code, to convert old values into new one, so user don't need to click on activate to have uPnP working again, it's going to simply work.

xbipin:
i can code that if u tell me which file handles the upgrade of config

rbgarga:
Sure, you need to bump config version in conf.default/config.xml, from 11.6 to 11.7. Also change 'latest_config' in etc/inc/globals.inc.

After that you need to write a function upgrade_116_to_117() in etc/inc/upgrade_config.inc

Let me know if you have questions

xbipin:
ok thanks, doing it now

rbgarga:
Thanks!

xbipin:
im having a hard time writing the upgrade config code, i used this but the problem is values from the old permission fields dont get fed in the new fields

	$oldperm = $config['installedpackages']['miniupnpd']['config'][0];
	$config['installedpackages']['miniupnpd']['config'][0]['row'] = array();
	for ($i = 1; $i <= 4; $i++) {
		$config['installedpackages']['miniupnpd']['config'][0]['row'][$i-1]['permuser'] = $oldperm['permuser{$i}'];
	}

xbipin:
i think the issue happens at the below line, any suggestion on whats wrong

$config['installedpackages']['miniupnpd']['config'][0]['row'] = array();

xbipin:
any help?

rbgarga:
You need to check if
```
$config['installedpackages']['miniupnpd']['config'][0]['permuser1']
```
is defined and is not empty, then take its value and save on the new array called:
```
$config['installedpackages']['miniupnpd']['config'][0]['permuser']
```
After, you need to make the same check for permuser2, permuser3 and permuser4. And you also need to unset them.

xbipin:
Well the checks I ignored for a start because by default I checked it's set, I was going to do those steps later but the issue is I did a vardump and it returns the array with old values but when I try to assign them to new fields it just remains blank, I tried like a 100 different methods but failed so it's out of my league I guess or there is some other bug that's causing all this issue

xbipin:
And by the way the new array u mentioned actually that needs to be in a sub array called row as the rowhelper control needs that as that's how it works

xbipin:
Unsettling the old fields also doesn't work for me, no idea if it's due to the fact that  I'm working on a virtual box virtual machine for development

rbgarga:
Something like this should work (untested):
```
function upgrade_116_to_117() {
	global $config;

	if (!isset($config['installedpackages']['miniupnpd']['config'][0]))
		return;

	$miniupnpd =& $config['installedpackages']['miniupnpd']['config'][0];

	$miniupnpd['permuser'] = array();

	for ($i = 1; $i <= 4; $i++) {
		if (isset($miniupnpd["permuser{$i}"]) && !empty($miniupnpd["permuser{$i}"])) {
			$miniupnpd['permuser'][] = $miniupnpd["permuser{$i}"];
			unset($miniupnpd["permuser{$i}"]);
		}
	}
}
```

rbgarga:
Oh yeah, I forgot about the rowhelper format, do you have a block of config.xml with rowhelper set?

rbgarga:
What about this one:
```
function upgrade_116_to_117() {
    global $config;

    if (!isset($config['installedpackages']['miniupnpd']['config'][0]))
        return;

    $miniupnpd =& $config['installedpackages']['miniupnpd']['config'][0];

    $miniupnpd['row'] = array();

    for ($i = 1; $i <= 4; $i++) {
        if (isset($miniupnpd["permuser{$i}"]) && !empty($miniupnpd["permuser{$i}"])) {
            $miniupnpd['row'][] = array('permuser' => $miniupnpd["permuser{$i}"]);
            unset($miniupnpd["permuser{$i}"]);
        }
    }
}
```

xbipin:
with a row helper it should be stored this way

				<row>
					<permuser>allow 1024-65535 192.168.0.2 1024-65535</permuser>
				</row>
				<row>
					<permuser>allow 1024-65535 192.168.0.3 1024-65535</permuser>
				</row>

xbipin:
the last one worked, it seems the issue was i was using single quote 

['permuser{$i}']

whereas double were required

["permuser{$i}"]

xbipin:
i have committed the last code u gave me, works well and it seems complete now to merge it :)

etc/inc/upgrade_config.inc
```diff
@@ -3552,4 +3552,21 @@ function upgrade_115_to_116() {
...
+        return;
+
+    $miniupnpd =& $config['installedpackages']['miniupnpd']['config'][0];
+
+    $miniupnpd['row'] = array();
+
+    for ($i = 1; $i <= 4; $i++) {
+        if (isset($miniupnpd["permuser{$i}"]) && !empty($miniupnpd["permuser{$i}"])) {
+            $miniupnpd['row'][] = array('permuser' => $miniupnpd["permuser{$i}"]);
+            unset($miniupnpd["permuser{$i}"]);
```
rbgarga: Almost there, the only missing thing is it does not unset empty parameter. This unset should be out of this if

xbipin:
done
it wouldnt make much difference with empty values as they would anyways go the next time upnp settings would be saved

PR #1458, opened 2015-01-31
add schedule selection to pkg_edit.php

xbipin:
this allows schedule selection to pkg_edit.php, this can then be
utilized by packages to set schedules in many other places. This is the
first set of changes, will further modify other files to make full use
of schedules.

### DIFF
--- a/usr/local/www/pkg_edit.php
+++ b/usr/local/www/pkg_edit.php
@@ -816,22 +816,21 @@ if ($pkg['tabs'] <> "") {
 					foreach ($config['schedules']['schedule'] as $schedule) {
 						if ($schedule['name'] <> "")
 							$schedules[] = $schedule['name'];
-						}
 					}
+				}
 				foreach($schedules as $schedule) {
 					$selected = ($value == $schedule ? "selected=\"selected\"" : "");
 					if($schedule == "none")
 						$input .= "<option value=\"\" {$selected}>{$schedule}</option>\n";
 					else
 						$input .= "<option value=\"{$schedule}\" {$selected}>{$schedule}</option>\n";
-					}
+				}
 				$input .= "</select>\n<br />\n" . fixup_string($pkga['description']) . "\n";
 				if(isset($pkga['advancedfield']) && isset($adv_filed_count)) {
 					$js_array[] = $pkga['fieldname'];
 					$advanced .= display_advanced_field($pkga['fieldname']).$input;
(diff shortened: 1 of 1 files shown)


usr/local/www/pkg_edit.php
```diff
@@ -808,6 +808,33 @@ function enablechange() {
...
+			case "schedule_selection":
+				$input = "<select id='{$pkga['fieldname']}' name='{$pkga['fieldname']}'>\n";
+				$schedules = array();
+				$schedules[] = "none";
+				if(is_array($config['schedules']['schedule'])) {
+					foreach ($config['schedules']['schedule'] as $schedule) {
+						if ($schedule['name'] <> "")
+							$schedules[] = $schedule['name'];
+						}
+					}
```
rbgarga: Looks like it's wrong indented here

usr/local/www/pkg_edit.php
```diff
@@ -808,6 +808,33 @@ function enablechange() {
...
+				foreach($schedules as $schedule) {
+					$selected = ($value == $schedule ? "selected=\"selected\"" : "");
+					if($schedule == "none")
+						$input .= "<option value=\"\" {$selected}>{$schedule}</option>\n";
+					else
+						$input .= "<option value=\"{$schedule}\" {$selected}>{$schedule}</option>\n";
+					}
```
rbgarga: And also here

usr/local/www/pkg_edit.php
```diff
@@ -808,6 +808,33 @@ function enablechange() {
...
+				$input .= "</select>\n<br />\n" . fixup_string($pkga['description']) . "\n";
+				if(isset($pkga['advancedfield']) && isset($adv_filed_count)) {
+					$js_array[] = $pkga['fieldname'];
+					$advanced .= display_advanced_field($pkga['fieldname']).$input;
+					$advanced .= "</div>\n";
+					}
```
rbgarga: And also here

xbipin:
actually i knew it was wrong but rest of the file was such so i though to follow that only

xbipin:
fixed it

xbipin:
can u check this and commit if all good?

xbipin:
bump
also kindly let me know what files do i need to look at to trigger services on and off, for a start i would like to use this schedule thing in upnp and then later create something like a framework where other packages can also use schedules to turn things off and on

PR #1467, opened 2015-02-02
php error logging should 're-fix' 4143 with less side effects for now..

PiBa-NL:
php error logging should 're-fix' https://redmine.pfsense.org/issues/4143 with less side effects of 'to much' logging..


PiBa-NL:
rebased on current master.. perhaps it could be pulled now?

PR #1472, opened 2015-02-05
Use web-gui setting for pap or chap instead of having it hard-coded

bassebaba:
I have no idea if this is a bug or by design.

This is the thread describing my problem:
https://forum.pfsense.org/index.php?topic=88071.0

What I found is that the "l2tp_standard" config is hard-coded to "chap" and does not care about the substitution beeing done at line 1592 under vpn_l2tp_configure().


rbgarga:
Good catch! Thanks.

PR #1474, opened 2015-02-09
fix Net_IPv6::compress() to properly handle all-zeros address

razzfazz:
The existing implementation of Net_IPv6::compress produces an empty
string when compressing the all-zeros ("::") address; fix this by
checking for empty return values and replacing them with "::".


razzfazz:
Created a pull request against master; closing this one.

PR #1476, opened 2015-02-09
add dhcp6.name-servers option with DHCPD-PD regardless of PD length

razzfazz:
The existing code only includes a v6 name server IP in the
automatically generated dhcpdv6 configuration for tracking interfaces if
there are additional prefixes that can be delegated on to the next
router (i.e., if PD length > 2). The correct behavior is for this option
to be included regardless of the PD length, as implemented in this
change.


razzfazz:
While in there, I also fixed the leading white space for this option to be in line with all the others (tab instead of spaces).

razzfazz:
Opened [issue #4433](https://redmine.pfsense.org/issues/4433) to track this.

razzfazz:
Anything I can do to help get this merged?

PR #1481, opened 2015-02-11
Adding description to 'Descriptive name' field for Authentication Server...

exotime:
Using the string "local" anywhere within the descriptive name whilst using any remote auth server (for example, ldap.localdomain, or localradius.mydomain.com that resolves to an IP other than the pfSense box) will cause the selected authentication mechanism to fail.

I believe line 131 of ./etc/inc/ipsec.auth-user.php is relevant. I don't really have a better suggestion of how to approach/fix this, so at least notifying the user before they waste too much time would be great.

Arguably, using the term 'local' in a network hostname is asking for trouble.

Also, fixing some minor spelling and indentation issues that I found while searching through the code.

### DIFF
--- a/etc/inc/interfaces.inc
+++ b/etc/inc/interfaces.inc
@@ -1513,7 +1513,7 @@ function interface_ppps_configure($interface) {
 				}
 				break;
 			default:
-				log_error(sprintf(gettext("Unkown %s configured as ppp interface."), $type));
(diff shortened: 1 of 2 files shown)


rbgarga:
The real issue should be fixed. I opened a ticket on redmine to keep track of it:

https://redmine.pfsense.org/issues/4469

This pull request can be closed.

PR #1482, opened 2015-02-11
Random text typos

phil-davis:
that I noticed.
I think automagically should be an English word.


phil-davis:
"automagically" is in the Oxford dictionary: http://www.oxforddictionaries.com/definition/english/automagically
Great to know, I have been using it for some years, so nice to know it is official.

PR #1486, opened 2015-02-14
Use `none` instead of a whitespace in sshd_config

jlduran:
Use the `none` keyword instead of a whitespace to disable the FreeBSD version in sshd_config.

### DIFF
--- a/etc/sshd
+++ b/etc/sshd
@@ -125,7 +125,7 @@
 	/* Run the server on another port if we have one defined */
 	$sshconf .= "Port $sshport\n";
 	/* Hide FreeBSD version */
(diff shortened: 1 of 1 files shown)


rbgarga:
Do you see problems with current empty value?

jlduran:
I can only think of two things:

1. It's not the right way to set (or to unset) this value:

  http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man5/sshd_config.5?query=sshd%5fconfig&sec=5

2. If I strip trailing spaces out of my config files, the value will not be set. There is no trivial way of detecting that a space is missing/present.

PR #1487, opened 2015-02-15
BearShare

francisuk:
BearShare


rbgarga:

cbuechler:
This functionality hadn't worked in quite some time and was removed, so this is no longer applicable.

PR #1488, opened 2015-02-15
Bittorrent tracker passkey support

francisuk:
Added tracker passkey support for Bittorrent and a few tiny updates to make this work better.


rbgarga:

cbuechler:
This functionality hadn't worked in some time and was removed so this is no longer applicable.

PR #1490, opened 2015-02-15
Netflix Video Stream

francisuk:
When the user streams a video from Netflix.


rbgarga:
You can add custom patterns to your pfSense installation using https://IP/diag_patterns.php.

This pattern looks too generic, and I'm afraid it could create false-positives.

PR #1493, opened 2015-02-15
pkg-utils, make dynamic package functions more robust to 'errors'

PiBa-NL:
pkg-utils, make dynamic package functions more robust to 'errors' 
 (fixes problems re-installing a partially installed package)


PiBa-NL:
doesnt work..

PR #1494, opened 2015-02-16
Fixing issues with NTP RRD graph state changes

k-paulius:
fixes ticket #4434

 - only call enable_rrd_graphing() after $config['ntpd']['statsgraph']
   is set
 - fix if condition; empty and isset are each other's opposites

### DIFF
--- a/usr/local/www/services_ntpd.php
+++ b/usr/local/www/services_ntpd.php
@@ -148,12 +148,14 @@ if ($_POST) {
 		elseif (isset($config['ntpd']['notrap']))
 			unset($config['ntpd']['notrap']);
 
-		if ((empty($_POST['statsgraph'])) != (isset($config['ntpd']['statsgraph'])));
-			enable_rrd_graphing();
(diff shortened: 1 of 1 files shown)


usr/local/www/services_ntpd.php
```diff
@@ -148,12 +148,14 @@
...
 
-		if ((empty($_POST['statsgraph'])) != (isset($config['ntpd']['statsgraph'])));
-			enable_rrd_graphing();
+		if ((empty($_POST['statsgraph'])) == (isset($config['ntpd']['statsgraph'])))
+			$enable_rrd_graphing = true;
 		if (!empty($_POST['statsgraph']))
 			$config['ntpd']['statsgraph'] = $_POST['statsgraph'];
 		elseif (isset($config['ntpd']['statsgraph']))
 			unset($config['ntpd']['statsgraph']);
+		if (!empty($enable_rrd_graphing))
```
rbgarga: It would be better to check using isset() here to avoid PHP complain when variable is not set

rbgarga:
I'll merge it and fix the ckeck

PR #1495, opened 2015-02-17
Do not request prefix delegation if no tracking interfaces are setup to use it. Ticket #4436

k-paulius:


etc/inc/interfaces.inc
```diff
@@ -3459,10 +3459,16 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$dhcp6cconf .= "	script \"{$g['varetc_path']}/dhcp6c_{$interface}_script.sh\"; # we'd like some nameservers please\n";
 		$dhcp6cconf .= "};\n";
 	} else {
+		$iflist = link_interface_to_track6($interface);
+		foreach ($iflist as $ifname => $ifcfg) {
+			if (is_numeric($ifcfg['track6-prefix-id']))
+				$trackiflist[$ifname] = $ifcfg;
```
rbgarga: $tracklist should be initialized

etc/inc/interfaces.inc
```diff
@@ -3459,10 +3459,16 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$dhcp6cconf .= "	script \"{$g['varetc_path']}/dhcp6c_{$interface}_script.sh\"; # we'd like some nameservers please\n";
 		$dhcp6cconf .= "};\n";
 	} else {
+		$iflist = link_interface_to_track6($interface);
+		foreach ($iflist as $ifname => $ifcfg) {
+			if (is_numeric($ifcfg['track6-prefix-id']))
```
rbgarga: Just to be sure, check if $ifcfg['track6-prefix-id'] is set before call is_numeric()

rbgarga:
I'll merge it and implement the fixes

PR #1497, opened 2015-02-18
diag_arp allow underscore in resolved host names

phil-davis:
is_hostname() and is_domain() allow underscore in the names. So it is possible to have underscore in host names, for example in DHCP server static mapped entries I have some things like:
10.42.3.4 client-pc-01_LAN
10.42.3.5 client-pc-01_WIFI
These reverse-resolve fine - 10.42.3.4 becomes client-pc-01_LAN
But the preg_match here misses such names that have an underscore in them.
I noticed this when looking into forum post: https://forum.pfsense.org/index.php?topic=88956.0

### DIFF
--- a/usr/local/www/diag_arp.php
+++ b/usr/local/www/diag_arp.php
@@ -251,7 +251,7 @@ function _getHostName($mac,$ip) {
 		return $dhcpip[$ip];
 	else{
 		exec("host -W 1 " . escapeshellarg($ip), $output);
-		if (preg_match('/.*pointer ([A-Za-z0-9.-]+)\..*/',$output[0],$matches)) {
+		if (preg_match('/.*pointer ([A-Za-z_0-9.-]+)\..*/',$output[0],$matches)) {
 			if ($matches[1] <> $ip)
 				return $matches[1]; 
 		}


rbgarga:
It sounds like a good chance to replace this exec(host...) by gethostbyaddr(), don't you agree?

phil-davis:
I guess so. But you might as well commit this little fix, and then do the gethostbyaddr() change yourself to the line above.

rbgarga:
hold on, @jim-p remembered that gethostbyaddr() was not used because of timeout, -W1 fix the problem.

phil-davis:
I made https://github.com/pfsense/pfsense/pull/1512 also since the code was easy.
It works for me on a small system at home.
I guess gethostbyaddr() might take longer than the 1 second wait specified here for "host".
Vommit whichever code you like and close the other.

rbgarga:
The problem will happen if DNS server doesn't answer, client timeout is much longer than 1s defined with host -W 1. Please close this one.

rbgarga:
Ops, sorry, I commented in the wrong PR. Please close #1512

PR #1498, opened 2015-02-18
Handle reverse lookup domain overrides

phil-davis:
that match exactly a whole block of private address space.
e.g. if the user has checked "Do not forward private reverse lookups" and also adds adds a domain override that matches a whole block of private address space, such as:
10.in-addr.arpa -> 10.42.1.1
then we want all reverse lookups in the "10" network to be referred to their DNS server at 10.42.1.1 and reverse lookups of other private addresses to immediately return NXDOMAIN.
Without this change, those referrals were not happening, because there was a "-server=10.in-addr.arpa" to nowhere put on the command line and that was effectively overriding the user-specified domain override!

### DIFF
--- a/etc/inc/services.inc
+++ b/etc/inc/services.inc
@@ -1825,18 +1825,34 @@ function services_dnsmasq_configure() {
 		}
 
 		/* If selected, then first forward reverse lookups for private IPv4 addresses to nowhere. */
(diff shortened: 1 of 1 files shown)


phil-davis:
Forum: https://forum.pfsense.org/index.php?topic=88956.msg491905#msg491905

PR #1501, opened 2015-02-20
Tweak aes-gcm options, add ecp for DH

bobveznat:
These changes allow configuring pfsense/strongswan according to RFC 6379
(Suite B Cryptographic Suites for IPsec). The change to the gcm options
I think simplifies the UI (prior you could choose aes128gcm with a
256bit key, which didn't make any sense).

I tested this by patching a running pfsense box and configuring two site to site VPNs, one using aes in cbc mode and one in gcm with a suite b configuration.

If this moves closer to "yes, we want this patch" I'll happily sign the contributor agreement.

### DIFF
--- a/etc/inc/ipsec.inc
+++ b/etc/inc/ipsec.inc
@@ -77,9 +77,9 @@ $p1_ealgos = array(
 global $p2_ealgos;
 $p2_ealgos = array(
 	'aes' => array( 'name' => 'AES', 'keysel' => array( 'lo' => 128, 'hi' => 256, 'step' => 64 ) ),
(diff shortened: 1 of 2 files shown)


rbgarga:

bobveznat:
I've signed electronically.

bobveznat:
I see 2.2.2 came out today and this isn't in it. Any hope of getting this change in the next release?

ermal:
Merged manually and partially.
Can be closed.

bobveznat:
I'm curious why you kept the keysize selection on the GCM options? It doesn't have the effect that the UI implies it has.

ermal:
It will be fixed moving forward with other changes.
Thx for the pull request.

PR #1503, opened 2015-02-21
Status_dhcp_leases fix edit button for static entries with no IP address

phil-davis:
The edit button for static entries always has an index id=0 and thus pressing the edit button goes to (mostly) edit the wrong entry.
We need to check for a match on the IP or MAC, whichever is defined. Probably could just check for match on MAC, since I think there has to be a MAC for every static mapped entry. But I have left the IP match check also to make sure.
Forum: https://forum.pfsense.org/index.php?topic=89072.msg493241#msg493241


phil-davis:
This does not consider the case when the same MAC address has a static mapped entry on multiple interfaces. It keeps matching the first occurrence of the MAC address, usually in LAN, for the edit button. Better solution coming.

PR #1504, opened 2015-02-22
Status_dhcp_leases fix edit button for static entries with no IP address

phil-davis:
The edit button for static entries with no IP address always has an index id=0 and thus pressing the edit button goes to (mostly) edit the wrong entry.
It was easier and made the code cleaner to get rid of the looping through the staticmap array every time trying to match IP, MAC address... That data ('if' and 'staticmap_array_index') is easily gathered further up, the first time the staticmap array for each DHCP interface is traversed. I think this should also be more efficient when there are many static mapping entries.
Forum: https://forum.pfsense.org/index.php?topic=89072.msg493241#msg493241

### DIFF
--- a/usr/local/www/status_dhcp_leases.php
+++ b/usr/local/www/status_dhcp_leases.php
@@ -274,17 +274,21 @@ if(count($pools) > 0) {
 foreach($config['interfaces'] as $ifname => $ifarr) {
 	if (is_array($config['dhcpd'][$ifname]) &&
 		is_array($config['dhcpd'][$ifname]['staticmap'])) {
+		$staticmap_array_index = 0;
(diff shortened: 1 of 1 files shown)


phil-davis:
To test;
a) Put various static mapped entries on different interfaces, some with an IP address, some without.
b) Put some entries using the same MAC address on multiple interfaces, both with and without an IP address. Make sure you have some that have the same MAC address on multiple interfaces and no IP address specified on any of the entries.
c) Bring up status_dhcp_leases - hover over each "edit" button and confirm the link address has various different "if=" and "id=" that vary over the interfaces. Click on some and confirm that the correct entry comes up for editing.

PR #1508, opened 2015-02-23
Modified DynDns -> Eurodns url

johnbyronent:


Gertjanpfsense:
Please read https://forum.pfsense.org/index.php?topic=88692.0 - It seems that just changing the URL isn’t enough.

rbgarga:
According documentation at https://www.eurodns.com/products/plugins/documentation/ the URL is correct:

The URL to perform the client update requests is 
http://eurodyndns.org/update/?hostname=%HOSTNAME%&myip=%NEW_IP%

rbgarga:
The certificate is for domain update.eurodyndns.org, so anyway this change makes sense, I will merge it

PR #1512, opened 2015-02-23
diag_arp allow in a host name whatever can be returned by gethostbyaddr(...

phil-davis:
...)

Valid chars like underscore were missing from the regex at line 254.
Using gethostbyaddr() avoids having to get the regex right for whatever host name permutations might be valid, now it displays whatever host names are there.


phil-davis:
Actually the code in https://github.com/pfsense/pfsense/pull/1497 is necessary - see discussion there.

PR #1513, opened 2015-02-24
Fix PTR records in host overrides

znerol:
Follow up for redmine issue [4447](https://redmine.pfsense.org/issues/4447) for branch master


znerol:
Thanks for merging, should also be cherry-picked into `RELENG_2_2`

PR #1514, opened 2015-02-24
Fixes ticket #4471 - stf tunnel interface is not destroyed when 6rd or 6to4 tunnel is disabled

k-paulius:
Destroy stf tunnel when 6rd or 6to4 tunnel is being disabled or type switches between 6rd <-> 6to4


cbuechler:
merged, thanks!

PR #1517, opened 2015-02-26
Fixes range displayed by the 6rd prefix validation message ticket #4475

k-paulius:
Range always starts from 0
$ipv6_delegation_length is not set


rbgarga:
I pushed a commit with a different fix, printing range in hex. This one can be closed. Thanks!

PR #1521, opened 2015-02-26
Code style guide changes for filter.inc

phil-davis:


rbgarga:
At line 164, it ended up with a block like this (I'm commenting here because github is hiding the diff)
```
        if (!file_exists("{$g['varrun_path']}/filterlog.pid") ||
                !isvalidpid("{$g['varrun_path']}/filterlog.pid")) {
                mwexec("/usr/local/sbin/filterlog -i pflog0 -p {$g['varrun_path']}/filterlog.pid");
        }
```
It's hard to figure out second line is still part of the if statement, in this case, we follow the FreeBSD style rules - https://www.freebsd.org/cgi/man.cgi?query=style&sektion=9 - that says to use 4 spaces on second and subsequent lines, as noted in this particular section:
```
﻿for (;;) {
	z = a + really + long + statement + that + needs +
	    two + lines + gets + indented + four + spaces +
	    on + the + second + and + subsequent + lines;
}
```

phil-davis:
Github will not show big diffs. I reviewed my changes using Meld, which did a really good job of showing me the bits and pieces.
I have found those multi-line if statements with regex:
.*if ?\(.*[^{]$
and fixed them up.
I explained what I did on the commit. Let me know if you disagree with any of the formatting conventions I used for laying out all the clauses.

phil-davis:
I was merging the branch up to the latest master. Git lists a silly extra commit in there that looks like it has a whole pile of changes. I will make a new pull request with just my finalised filter.inc against the current master.

PR #1522, opened 2015-02-26
Code Style Guide etc inc f to g

phil-davis:


etc/inc/filter_log.inc
```diff
@@ -45,33 +45,36 @@ function conv_log_filter($logfile, $nentries, $tail = 50, $filtertext = "", $fil
...
+		}
 
 		$flent = parse_filter_line($logent);
-		if (!$filterinterface || ($filterinterface == $flent['interface']))
-		{
-			if ( ( ($flent != "") && (!is_array($filtertext)) && (match_filter_line ($flent, $filtertext))) || 
-			     ( ($flent != "") && ( is_array($filtertext)) && (match_filter_field($flent, $filtertext)) ) ) {
+		if (!$filterinterface || ($filterinterface == $flent['interface'])) {
+			if ((($flent != "") && (!is_array($filtertext)) && (match_filter_line ($flent, $filtertext))) ||
+				(($flent != "") && ( is_array($filtertext)) && (match_filter_field($flent, $filtertext)))) {
```
rbgarga: Here is another case as I commented out on filter.inc. I'll re-review #1515, #1519 and #1520 and look for similar cases

phil-davis:
Yeh, when I did a global "convert leading spaces to tabs" in notepad++ it converts every 4 spaces to a tab. Then it displays (at least for PHP for me) a tab as 4 spaces. So in places like this there is no difference in the appearance in the editor.
Of course in an editor or diff visualizer (Github GUI diff display) that is rendering a tab as 8 spaces, it looks obvious that there is a difference.
I will have a look through these 2 pull requests and fixup...

jim-p:
@phil-davis In notepad++, activate View > Show Symbol > Show Whitespace and Tab, it renders tabs as a red arrow and spaces as a red dot.
I use a similar mode in Ultraedit, it makes spotting those sorts of things easy. Also using 8 for a tab stop helps. At one point we had that listed somewhere as the recommended tab size.

rbgarga:
Since pfSense uses K&R style, the best is to set editor to show tabs in 8-spaces size

rbgarga:
There are tools like squizlabs/PHP_CodeSniffer that could be useful too, but looks like there is not any standard like K&R there

phil-davis:
I checked this for multi-line if statements and fixed it up. Also found a couple more single-line "if" that I had missed. It should be good to go now.

BBcan177:
I use Ultraedit also and prefer it over Notepad++

What about changing  "<>" to "!=" to keep everything consistent?

PR #1536, opened 2015-03-03
Standardise power mode selection boxes

phil-davis:
I noticed that this looked a little odd, the boxes for AC and Battery Power Mode are 1 after the other on the same line, then the 3rd box for unknown power mode is on a new line.
Maybe make them all on their own line, like this?


phil-davis:
With the second commit IMHO it is even better - the dropdown selection boxes all line up vertically.

PR #1543, opened 2015-03-06
Be nicer when checking if alias is numeric

phil-davis:
Because an ordinary port can be numeric here.
Forum https://forum.pfsense.org/index.php?topic=89906.0


cbuechler:
merged, thanks!

PR #1551, opened 2015-03-10
Setup serial gps port for raw output and speed

rnoland:
We need to at least setup the serial port before we try to blast config data to it.  My system was hanging during boot because cat couldn't output gps.init to the port.

### DIFF
--- a/etc/inc/system.inc
+++ b/etc/inc/system.inc
@@ -1575,6 +1575,30 @@ function system_ntp_setup_gps($serialport) {
 	unlink_if_exists($gps_device);
 	@symlink($serialport, $gps_device);
 
+	$gpsbaud = '4800';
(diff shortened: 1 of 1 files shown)


etc/inc/system.inc
```diff
@@ -1575,6 +1575,30 @@ function system_ntp_setup_gps($serialport) {
...
+				$gpsbaud = '57600';
+				break;
+			case '80':
+				$gpsbaud = '115200';
+				break;
+		}
+	}
+
+	/* Configure the serial port for raw IO and set the speed */
+	`stty -f {$serialport}.init raw speed {$gpsbaud}`
```
rbgarga: Please do not use back tickles, use mwexec() instead

PR #1552, opened 2015-03-11
Support for Radius OTP login and fetching groups from Radius

kdbo:
Hi,

I have succesfully setup pfSense to work with Radius for TOTP instead of static passwords.
Tested setup:
- LinOTP for TOTP management and Auth to Radius. Users are fetched from SQL Database and Active Directory.

in LinOTP  you can define if a valid passcode consists of:
Pin + Token (otppin=0)
or 
User Password + Token (otppin=1)
or 
Token only (otppin=2)

This has to be setup in the linOTP authentication policies ( action = "otppin=0|1|2" )

- FreeRadius fetches user information from Active Directory ( Group membership ) and adds, besides the auth_accept message, also a pfSense-Group-Name parameter to pfSense
- pfSense maps a user, either to a local account, and groups OR to an Active Directory user and Group.

This way, we can authenticate and map users into different groups, for example for GUI access, or VPN.
I have tested this with both OpenVPN and IPSec, as well as the admin access to the pfSense web GUI.

For freeradius (2.x), a few configration changes need to be done.
I am assuming that freeradius is setup correctly for both both Active Directory and LinOTP 
So the changes documented here only apply for the pfSense part.

First create /usr/share/freeradius/dictionary.pfsense

````
# -*- text -*-
##############################################################################
#
#       pfSense's VSA's
#
#       $Id$
#
##############################################################################

#
#       pfSense VSA's
#

VENDOR          pfSense                 15000

BEGIN-VENDOR    pfSense
ATTRIBUTE       pfSense-Group-Name                      1       string

# Add more ATTRIBUTES for pfSense here

#
# Integer Translations
#

END-VENDOR pfSense
````

enable the pfsense dictionary in /usr/share/freeradius/dictionary

````
$INCLUDES dictionary.pfsense
````

create a client entry for pfSense in
/etc/raddb/client.conf

````
client pfsenseadmin {
        ipaddr = <ipaddress of pfsense lan>
        netmask = 32
        secret = 'YOURSECRET'
        shortname = 'pfsenseadmin'
        nastype = pfsense
}
````

I have created two groups on the pfSense box, PFSense Admins with all pfSense admin privileges,
and a VPNusers group with User - VPN - IPsec xauth Dialin access.
When freeradius gets a request from the pfsense box, it should return pfSense-Group-Name attributes.
So we will set that up in /etc/raddb/sites-enabled/default.
This should go in the post-auth section:

````
# Map AD/LDAP groups Domain Admins and Network Admins to the PFSense Admins group , make sure this group exists on pfSense and has admin privs
        if ("%{client:shortname}" =~ /^pfsenseadmin/ && ( LDAP-Group == "Domain Admins" || LDAP-Group == "Network Admins")){
                        update reply {
                                pfSense-Group-Name += "PFSense Admins"
                        }
       }

        # Map AD/LDAP group VPNusers to pfSense Group VPNusers
        if ("%{client:shortname}" =~ /^pfsenseadmin/ &&  LDAP-Group == "VPNusers" ){
                        update reply {
                                pfSense-Group-Name += "VPNusers"
                        }
        }
        # Add more group mappings for pfSense here
}
````

in freeradius 3.x it is possible  to use a foreach loop, to add all groups  a user is member of , automatically 
to the pfSense-Group-Name attribute.
But I havent played around with 3.x yet

### DIFF
--- a/etc/inc/auth.inc
+++ b/etc/inc/auth.inc
@@ -1336,7 +1336,11 @@ function radius_backed($username, $passwd, $authcfg, &$attributes = array()) {
 		if ($debug) {
 			printf(gettext("Radius Auth succeeded")."<br />\n");
 		}
-		$ret = true;
+		if($attributes) {
+			$ret = $attributes;
+		} else {
+			$ret = true;
+		}
 	} else {
 		$retvalue['auth_val'] = 3;
 		if ($debug) {
@@ -1409,7 +1413,7 @@ function auth_get_authserver_list() {
 }
 
 function getUserGroups($username, $authcfg) {
-	global $config;
+	global $config,$radiusgroups;
 
 	$allowed_groups = array();
 
@@ -1418,6 +1422,7 @@ function getUserGroups($username, $authcfg) {
 			$allowed_groups = @ldap_get_groups($username, $authcfg);
 			break;
 		case 'radius':
+			 $allowed_groups = $radiusgroups;
 			break;
 		default:
 			$user = getUserEntry($username);
@@ -1438,6 +1443,7 @@ function getUserGroups($username, $authcfg) {
 }
 
 function authenticate_user($username, $password, $authcfg = NULL, &$attributes = array()) {
+	global $radiusgroups;
 
 	if (!$authcfg) {
 		return local_backed($username, $password);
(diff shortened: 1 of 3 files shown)


rbgarga:

kdbo:
Hi Renato,

I did this already a while ago :-)

rbgarga:
I couldn't find your CLA, what is your portal username?

rbgarga:
Found it!

jim-p:
The code in master has diverged significantly here. We have had support for obtaining groups from the Class reply attribute in place for several months now, which could be leveraged here. If the new attribute is required, you can resubmit this as a change to just radius.inc to add the attribute and it should only take a couple lines added to radius_get_groups() in auth.inc to have it check the other attribute in addition to the Class attribute.

PR #1553, opened 2015-03-11
Status DHCP Leases handle expire never

phil-davis:
Note: We can let the code pass "never" (or any other unexpected stuff) to adjust_gmt()
adjust_gmt() should anyway handle the case when strtotime() cannot understand the input string and thus returns false. In that case we return the input string as-is so it will be displayed as the time. That way the user will see it and can report easily whatever other unexpected char data was in the leases file.
It also prevents "false" (zero) being converted to the date-time string and thus becoming the Unix epoch 1 Jan 1970 on the display.
Latest forum report of this kind of thing: https://forum.pfsense.org/index.php?topic=90083.0

### DIFF
--- a/usr/local/www/status_dhcp_leases.php
+++ b/usr/local/www/status_dhcp_leases.php
@@ -192,7 +192,7 @@ foreach($leases_content as $lease) {
 				if ($data[$f+1] == "never") {
 					// Quote from dhcpd.leases(5) man page:
 					// If a lease will never expire, date is never instead of an actual date.
-					$leases[$l]['end'] = $data[$f+1];
+					$leases[$l]['end'] = gettext("Never");
 					$f = $f+1;
 				} else {
 					$leases[$l]['end'] = $data[$f+2];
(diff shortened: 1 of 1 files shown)


usr/local/www/status_dhcp_leases.php
```diff
@@ -186,9 +189,16 @@ function remove_duplicate($array, $field)
 				$f = $f+3;
 				break;
 			case "ends":
-				$leases[$l]['end'] = $data[$f+2];
-				$leases[$l]['end'] .= " " . $data[$f+3];
-				$f = $f+3;
+				if ($data[$f+1] == "never") {
+					// Quote from dhcpd.leases(5) man page:
+					// If a lease will never expire, date is never instead of an actual date.
+					$leases[$l]['end'] = $data[$f+1];
```
rbgarga: It would be good to have a intl text for "Never" here, don't you think? gettext("Never")

phil-davis:
Yes, I agree about the gettext for "Never" and did that just now. So you can commit this.
I could also internationalize the various keywords use in the Online and Lease Type columns - online, offline, expired, dynamic, static - in a separate pull request. What is the policy on that sort of thing? Do we internationalize "computer technology standard words" like that? Or in the "computer technology" world are they defined to be no longer really English, but "world standard Computer Language"?

phil-davis:
I made a newer pull request #1559 which has the same code as this in 1 commit and merged up to date with master.

PR #1555, opened 2015-03-12
Update get_possible_traffic_source_addresses returned array format

phil-davis:
With this change it looks to me like the way it is intended to be, based on what was done to get_possible_listen_ips()
Please review and check if this is what was intended for the code. With this change it should make the returned $sourceips array be in a consistent format the same way as get_possible_listen_ips()
This is a fix for code that was only committed in master. So at this time the fix is only needed in master.


phil-davis:
Updated version in #1560 merged with recent changes. Closing this one.

PR #1557, opened 2015-03-13
Missing double equals in captiveportal.inc

phil-davis:
Looking at where this is nested inside various if statements, I do not think this error did too much harm - only to the $mac['descr'] - in this particular code flow $username is not used for important stuff after this point.


cbuechler:
merged, thanks

PR #1558, opened 2015-03-13
Missing double equals in ipsec.inc

phil-davis:
This would have done no harm as long as ipsec_find_id() is always called with $side = empty, local or peer


cbuechler:
merged, thanks

phil-davis:
Oh - I see I accidentally made this change against RELENG_2_2 - but you noticed and have also made the fix in master.

PR #1559, opened 2015-03-13
Status DHCP Leases handle expire never

phil-davis:
Note: We can let the code pass "never" (or any other unexpected stuff) to adjust_gmt()
adjust_gmt() should anyway handle the case when strtotime() cannot understand the input string and thus returns false. In that case we return the input string as-is so it will be displayed as the time. That way the user will see it and can report easily whatever other unexpected char data was in the leases file.
It also prevents "false" (zero) being converted to the date-time string and thus becoming the Unix epoch 1 Jan 1970 on the display.
Latest forum report of this kind of thing: https://forum.pfsense.org/index.php?topic=90083.0


phil-davis:
Does this one count as a bug fix?
The existing code does not handle the (documented in the man page) case of "ends" being "never" in the DHCP leases file.

PR #1563, opened 2015-03-14
Small cleanup fixes for ipsec.conf

gitbensons:
When ipsec.conf is generated I noticed 2 issues
- left/rightsubnet, when using IPKv2, has duplicate entries
- AH/ESP algorithms section also has duplicate entries

This patch just uniques the entries for readability; I dont think the functionality is affected by those duplicates.

### DIFF
--- a/usr/local/www/wizards/openvpn_wizard.xml
+++ b/usr/local/www/wizards/openvpn_wizard.xml
@@ -410,7 +410,7 @@
 		<field>
 			<name>state</name>
 			<displayname>State or Province</displayname>
-			<description>Full State of Province name, not abbreviated (e.g. Kentucky, Indiana, Ontario).</description>
+			<description>Full State or Province name, not abbreviated (e.g. Kentucky, Indiana, Ontario).</description>
 			<type>input</type>
 			<size>30</size>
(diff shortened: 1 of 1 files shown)


rbgarga:
You need to sign Individual or Corporate CLA as described at https://www.pfsense.org/about-pfsense/#legal in order to have your patches applied.

gitbensons:
Just signed the Individual CLA

etc/inc/vpn.inc
```diff
@@ -1083,10 +1083,10 @@ EOD;
 				$ipsecconnect .= "\t{$ealgosp1}\n";
 			}
 			if (!empty($ealgoAHsp2arr)) {
-				$ipsecconnect .= "\tah = " . join(',', $ealgoAHsp2arr) . "!\n";
```
rbgarga: It would be better to prevent adding duplicate items, so if you need to use variable in multiple places you don't need to take care of it

gitbensons:
So I adjusted the patch a bit. In case for IKEv1 its now just using the selected algorithm for phase1. 
For IKEv2 its still just removing duplicates from the array, as I don't see a better way to handle it.

PR #1565, opened 2015-03-15
Do not allow VLAN tag zero

phil-davis:
At the moment you can make a VLAN with tag 0. The input validation does not catch it because when $_POST['tag'] = "0" that evaluates to false by PHP.
There is no need for that check here, because do_input_validation already checks that $_POST['tag'] is non-empty, and we really always want the other numericint and limit checks here.

### DIFF
--- a/usr/local/www/interfaces_vlan_edit.php
+++ b/usr/local/www/interfaces_vlan_edit.php
@@ -80,7 +80,7 @@ if ($_POST) {
 
 	do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
 
-	if ($_POST['tag'] && (!is_numericint($_POST['tag']) || ($_POST['tag'] < '1') || ($_POST['tag'] > '4094'))) {
+	if (!is_numericint($_POST['tag']) || ($_POST['tag'] < '1') || ($_POST['tag'] > '4094')) {
 		$input_errors[] = gettext("The VLAN tag must be an integer between 1 and 4094.");
 	}
 


usr/local/www/interfaces_vlan_edit.php
```diff
@@ -80,7 +80,7 @@
 
 	do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
 
-	if ($_POST['tag'] && (!is_numericint($_POST['tag']) || ($_POST['tag'] < '1') || ($_POST['tag'] > '4094'))) {
+	if (!is_numericint($_POST['tag']) || ($_POST['tag'] < '1') || ($_POST['tag'] > '4094')) {
```
rbgarga: Please keep a isset($_POST['tag']) here just to be sure the key exist

phil-davis: This is the issue here - when the user puts VLAN tag "0" it gets $_POST['tag'] false - and so the "if" statement never gets even executed. The code thinks that "tag" was false, but actually the value "0" was supplied and gets written to the config.
Because this parameter is required anyway, we can skip $_POST['tag'] boolean check - it has to be numericint from 1 to 4094 in all cases.

rbgarga: yeah, but keep a isset() call is good, and it doesn't depend of the value. pfSense PHP needed to be patched when we moved to 5.5 because of there are too many places where a non-defined variable or array key is accessed. On recent PHP versions it throws an error, what we changed to a notice. We need to remove all those access without checks to be able to remove PHP patch.

phil-davis:
Yes, understand now. Will make another pull request with a different change.

PR #1566, opened 2015-03-15
Support checking overlapping subnets

phil-davis:
This allows the caller to specify $check_subnets and also provide a $cidrprefix so that the code can check if the proposed IPaddress/CIDRprefix will make a subnet that overlaps with any other subnet defined on other interfaces.
is_ipaddr_configured() returns a simple true/false answer, like it always has.
where_is_ipaddr_configured() returns an array of the conflicting interfaces and subnets. This is handy for front-end validation so that the caller can be returned information about what addresses conflicted and display it to the user.
This should be helpful for stopping stupidity by users that put overlapping subnets on various interfaces.


phil-davis:
Hmmm - I was too hasty, needs a few more checks when looping around interfaces that might or might not have an IPv4 or IPv6 address defined.

PR #1570, opened 2015-03-17
Update dyndns.class

johnbyronent:
Fix Certificate validation issue

### DIFF
--- a/etc/inc/dyndns.class
+++ b/etc/inc/dyndns.class
@@ -676,6 +676,7 @@
 						log_error("EuroDynDns ({$this->_dnsHost}) DNS update() starting.");
 					}
 					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser.':'.$this->_dnsPass);
+					curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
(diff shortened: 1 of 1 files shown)


cbuechler:
I'd really rather not disable cert validation, especially since if they fix their F graded HTTPS config it would work as is. 
https://www.ssllabs.com/ssltest/analyze.html?d=update.eurodyndns.org

They're missing part of the cert chain. Chrome figures it out, which is why I didn't notice when I enabled cert verification there. curl, fetch, wget and similar tools fail because of the incomplete chain. 

I'm guessing you must be a customer of theirs, could you instead open a support request with them requesting they fix their HTTPS setup?

johnbyronent:
I will try to see if I can have them fix it.
But indeed I don't have any control on it.

Gertjanpfsense:
johnbyronent   :: "But indeed I don't have any control on it."
I guess that mentioning this test : https://www.ssllabs.com/ssltest/analyze.html?d=update.eurodyndns.org&latest will motivate them.
It will take them a couple of minutes to get their certificate straight.

rbgarga:
Disable cert validation is not the way to go here.

PR #1572, opened 2015-03-18
Disable lighttpd server header

jlduran:
Set the `server.tag` to an empty string to prevent lighttpd from
displaying the version number in the header.

### DIFF
--- a/etc/inc/system.inc
+++ b/etc/inc/system.inc
@@ -1365,6 +1365,9 @@ url.access-deny             = ( "~", ".inc" )
 
 ######### Options that are good to be but not necessary to be changed #######
 
+## disable server header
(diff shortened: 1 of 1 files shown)


rbgarga:
What are you trying to fix with this?

jlduran:
To hide the lighttpd server header. For example, a guest user in the captive portal may be able to obtain this information:

    $ curl -Ik https://portal.pfsense
    HTTP/1.1 200 OK
    ...
    Server: lighttpd/1.4.35

For aesthetic reasons (some may argue security as well), it's not good to disclose this information.

Once the patch is applied, the Server header is not displayed. Another option could be just to display the product name, however it's better not to have one.

PR #1573, opened 2015-03-18

phil-davis:
on factory default boot.
This allows the system to switch interfaces from the newer ones in the default config (e.g. em0 em1) back to the interfaces used by:
Alix - vr1 vr0
APU - re1 re2
It means these devices can boot the default config and this will automatically detect that there is no em0/em1 and will instead select whatever exists out of vr1/vr0 or re1/re2. This avoids the user having to use the serial cable to do interface assignment when starting a brand new image, or when resetting to factory defaults. It could easily be extended to other common interface combinations.
For me, this (or similar) would be very beneficial. At remote sites it is really good if it is possible to do reset to factory defaults, or put a fresh CF/SD card in, and the system boots without needing to connect a serial cable and do interface assignment that way.

### DIFF
--- a/etc/inc/config.console.inc
+++ b/etc/inc/config.console.inc
@@ -556,4 +556,34 @@ EOD;
 	}
 }
 
+function check_for_alternate_interfaces() {
+	global $config;
+
+	// If the factory default configured WAN and/or LAN devices do not exist,
+	// then use alternate devices if they exist.
+	// These are devices on Alix (vr) and APU (re) systems, and this lets them boot a
+	// factory default config without being forced to do interface assignment
+	// over the serial port.
+
+	$other_devices_arr['wan'] = array("vr1", "re1");
+	$other_devices_arr['lan'] = array("vr0", "re2");
+	$interface_assignment_changed = false;
+
+	foreach ($other_devices_arr as $ifname => $other_devices) {
+		if (!does_interface_exist($config['interfaces'][$ifname]['if'])) {
+			foreach ($other_devices as $other_device) {
+				if (does_interface_exist($other_device)) {
+					$config['interfaces'][$ifname]['if'] = $other_device;
+					$interface_assignment_changed = true;
+					break;
+				}
+			}
+		}
+	}
+
+	if ($interface_assignment_changed) {
+		write_config("Factory default boot detected WAN " . $config['interfaces']['wan']['if'] . " and LAN " . $config['interfaces']['lan']['if']);
+	}
+}
+
 ?>
--- a/etc/rc.bootup
+++ b/etc/rc.bootup
@@ -163,6 +163,10 @@ echo "Loading configuration...";
 parse_config_bootup();
 echo "done.\n";
 
+if (file_exists("/conf/trigger_initial_wizard")) {
+	check_for_alternate_interfaces();
+}
+
 if ($g['platform'] == "jail") {
 	/* We must determine what network settings have been configured for us */
 	$wanif = "lo0";	/* defaults, if the jail admin hasn't set us up */


phil-davis:
I will vote for this being classified as a regression in 2.2 for Alix boxes:)
Prior to 2.2 you could write a standard image to CF card, or reset to factory defaults an Alix 2D13 and it would boot up without needing a serial cable to do interface assignment. In 2.2 the default config changed to be emo/em1.
This change brings back the old behavior for Alix, em0/em1 still works as in 2.2, and as a bonus the APU device names also come up to automagically.

phil-davis:
But I also understand that a change like this really needs to be tested in a few scenarios to make sure nothing dumb happens that breaks interface assignment, which is extremely important to be correct from a user and a security point of view!

ermal:
@phil-davis i do not think this will be merged.
Another trick can be that when the reset-default settings you can change make it smarter to detect the platform.

Though obsolescence management on the devices will not be fixed in a generic software!

phil-davis:
It can be done many different ways. I am happy to do it in whatever way is suggested and acceptable - this is just an example of 1 way to do it.
It seems to me that we can support a variety of default device names em* vr* re* and others so that many devices can directly boot a default image without need for the console interface reassign. This makes it easier for user setup and disaster recovery.
It applies to old hardware, new hardware and "just plain old" hardware - not just to obsolete hardware.
Anyway, if the project wants some flexibility in factory default interface assignment then let me know how you would like to design it and I am very happy to code it.

phil-davis:
I made a Redmine issue so that what is good to have can be discussed:
https://redmine.pfsense.org/issues/4535

doktornotor:
Works great on Alix. (On a semi-unrelated note: lately, I managed to screw up by disabling (but not removing the assignment) of WLAN, removing the card and booting (which of course never finished.) Had to disassemble the box again and stick the wifi card back temporarily. Not sure if this is something that could be handled more intelligently.

phil-davis:
Closing in favor of #1602 - I had messed up some things in my forked pfsense repo. It was easier to start again with a new repo, make a new branch for this and make a new pull request.

cypher386:
This needs comparison with the factory image, as we did some work on it for
auto detection. It's still a good idea we should just try and make sure
factory and community editions don't duplicate functionally.

phil-davis:
I only closed it for technical reasons when I got my forked repo in a mess. It can continue on as #1602 
It would be a shame if some "factory image" that has closed changes in it becomes significantly different from the public repo and publicly released images. IMHO:
a) Things like this, picking some default interfaces at factory default boot, can be coded in a flexible way that supports a good range of hardware. This sort of thing should be in the public repo and thus everything "just works" for everybody.
b) Special settings in loader.conf.local or similar files might be needed to make certain hardware boot.In that case it might be more difficult to have some script that automagically works out what is needed based on the hardware and writes a suitable loader.conf.local or... and so it is convenient to make available some "factory image" variants that have this stuff already done, as a customer service.
But we should try to do (a) first before having to resort to more "closed" tricks via (b)

cypher386:
On 4/10/2015 8:51 AM, Phil Davis wrote:
>
> I only closed it for technical reasons when I got my forked repo in a
> mess. It can continue on as #1602
> <https://github.com/pfsense/pfsense/pull/1602>
> It would be a shame if some "factory image" that has closed changes in
> it becomes significantly different from the public repo and publicly
> released images. IMHO:
> a) Things like this, picking some default interfaces at factory
> default boot, can be coded in a flexible way that supports a good
> range of hardware. This sort of thing should be in the public repo and
> thus everything "just works" for everybody.
> b) Special settings in loader.conf.local or similar files might be
> needed to make certain hardware boot.In that case it might be more
> difficult to have some script that automagically works out what is
> needed based on the hardware and writes a suitable loader.conf.local
> or... and so it is convenient to make available some "factory image"
> variants that have this stuff already done, as a customer service.
> But we should try to do (a) first before having to resort to more
> "closed" tricks via (b)
>
> —
> <https://github.com/pfsense/pfsense/pull/1573#issuecomment-91564213>.
>
There are a number of changes in the factory image that don't make sense
for the community image, as they only improve "manfuacturability" of the
product, usually be reducing choices available in the installer.  For
what its worth the factory installer defaults to going into
"easy_install" and easy_install installs to the first hard disk type
device in the unit.  We had a need to have the network enabled before
launching the installer, and ran into an issue with the interface
assignments, so Renato wrote code to auto-detect interfaces based on
product system_identify_specific_platform, to avoid triggering vlan
assignment/interface mismatch in our workflow.  Seeing your pull request
made me think that code would be good in the in the community eddition
as well, which is why I posted.   I'll see what I can do to get that
code comitted to the main pfsense code base.  (We like having the
factory codebase deviate as small as possible from the community
edition, for validation, testing, and general maintenance).
 Some of what we do only makes sense if you are doing hundreds of the
same system, those changes don't always fit into the community edition
that can be installed on any platform.   (That's why it took a little
longer for the community edition of the SG-2xxx installer to be
released, it requires specific changes that won't work with "generic"
pfsenes community edition).  (Those changes a described on the forum and
on reddit, the community edition for SG-2XXX is a convenience for the
community.)

This isn't a reject of the pull request, only a "that's a really good
idea, we did some of the work for that internally, and we'd like to
release that as well, which may require some changes to your pull request".

phil-davis:
Yeh, there are plenty of implementation designs for how this could be done. I am very happy for you to choose some other design than this one, no problem. The end result is that I would like for my guys to be able to flash a standard image to CF/SD card and that the factory default boots up on Alix 2D13 and APU.* systems without any need to connect serial console. That's my selfish minimal hardware list, but the more systems that boot the factory default without needing to do anything at the console, the better.

PR #1574, opened 2015-03-20
Adding ability to run dhcp6c in debug mode #4534

k-paulius:
Adding checkbox under DHCP6 configuration on interface page and backend code. dhcp6c only supports debug mode when running in foreground so it needs to be piped to awk to remove date and time stamp and then to logger to be logged to syslog. AWK needs to flush stdout buffer else logger only logs in infrequent intervals or when dhcp6c is closed.

### DIFF
--- a/etc/inc/interfaces.inc
+++ b/etc/inc/interfaces.inc
@@ -3864,13 +3864,8 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 	$rtsoldscript .= "\t/bin/pkill -F {$g['varrun_path']}/dhcp6c_{$wanif}.pid\n";
 	$rtsoldscript .= "\t/bin/sleep 1\n";
 	$rtsoldscript .= "fi\n";
-
-	if (!isset($wancfg['dhcp6debug'])) {
-		$rtsoldscript .= "/usr/local/sbin/dhcp6c -d -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
-	} else {
-		// dhcp6c only supports -D debug option in foreground mode -f and outputs messages to stderr instead of syslog
-		$rtsoldscript .= "/usr/local/sbin/dhcp6c -f -d -D -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif} 2>&1 | /usr/bin/awk '{ print substr($0,23); fflush(stdout) }' | /usr/bin/logger -t dhcp6c &\n";
-	}
+	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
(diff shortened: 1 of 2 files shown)


ermal:
Really you cannot send dhcp6c to syslog.
This seems rather cumbersome!

k-paulius:
I agree that this is not a very pretty solution, but I can't think of any other way to accomplish this without patching dhcp6c. Not sure how feasible that would be.

Here is the relevent code from dhcp6 project:
- dhcp6c.c - processing command line options:

    		case 'f':
    			foreground++;
    			break;
    		case 'd':
    			debug = 1;
    			break;
    		case 'D':
    			debug = 2;
    			break;
    	setloglevel(debug);
- common.c - it only assigns log level 2 if running in foreground

        void
        setloglevel(debuglevel)
        	int debuglevel;
        {
        	if (foreground) {
        		switch(debuglevel) {
        		case 0:
        			debug_thresh = LOG_ERR;
        			break;
        		case 1:
        			debug_thresh = LOG_INFO;
        			break;
        		default:
        			debug_thresh = LOG_DEBUG;
        			break;
        		}
        	} else {
        		switch(debuglevel) {
        		case 0:
        			setlogmask(LOG_UPTO(LOG_ERR));
        			break;
        		case 1:
        			setlogmask(LOG_UPTO(LOG_INFO));
        			break;
        		}
        	}
        }

- common.c - logging procedure that they use; including here for reference

        void
        dprintf(int level, const char *fname, const char *fmt, ...)
        {
        	va_list ap;
        	char logbuf[LINE_MAX];
        	int printfname = 1;
        
        	va_start(ap, fmt);
        	vsnprintf(logbuf, sizeof(logbuf), fmt, ap);
        
        	if (*fname == '\0')
        		printfname = 0;
        
        	if (foreground && debug_thresh >= level) {
        		time_t now;
        		struct tm *tm_now;
        		const char *month[] = {
        			"Jan", "Feb", "Mar", "Apr", "May", "Jun",
        			"Jul", "Aug", "Sep", "Oct", "Nov", "Dec",
        		};
        
        		if ((now = time(NULL)) < 0)
        			exit(1); /* XXX */
        		tm_now = localtime(&now);
        		fprintf(stderr, "%3s/%02d/%04d %02d:%02d:%02d: %s%s%s\n",
        		    month[tm_now->tm_mon], tm_now->tm_mday,
        		    tm_now->tm_year + 1900,
        		    tm_now->tm_hour, tm_now->tm_min, tm_now->tm_sec,
        		    fname, printfname ? ": " : "",
        		    logbuf);
        	} else
        		syslog(level, "%s%s%s", fname, printfname ? ": " : "", logbuf);
        }

k-paulius:
Simplified the command, but this requires dhcp6 to be patched. See RM ticket #4534.

rbgarga:
Patch submitted to FreeBSD - https://reviews.freebsd.org/D5105

rbgarga:
@k-paulius Patch is applied on FreeBSD and also on pfSense/FreeBSD-ports. Can you please rework this patch to apply on current code?

k-paulius:
@rbgarga, I created new pull request #2585. Closing this one.

PR #1575, opened 2015-03-20
Misc DHCP6 changes

k-paulius:
- DHCP6 config file override, advanced and basic settings override each other so put them in single if/else statement. Currently code for all three setting types runs no matter which one is actually used.
 - A mix of literal tabs, spaces and \t is used in dhcp6c config file code. Convert evertyhing to use \t. A bit of housekeeping to keep things nice and tidy.
 - Supress errors when opening custom DHCP6 config file and check if content was successfully retrieved. Currently PHP throws error in case file does not exist.
 - Remove old dhcp6c and rtsold config scripts when bringing down interface. Currently only dhcp6c config file gets removed, but script files are left on file system.

### DIFF
--- a/etc/inc/interfaces.inc
+++ b/etc/inc/interfaces.inc
@@ -1301,6 +1301,8 @@ function interface_bring_down($interface = "wan", $destroy = false, $ifacecfg = 
 			}
 			sleep(3);
 			unlink_if_exists("{$g['varetc_path']}/dhcp6c_{$interface}.conf");
+			unlink_if_exists("{$g['varetc_path']}/dhcp6c_{$interface}_script.sh");
+			unlink_if_exists("{$g['varetc_path']}/rtsold_{$realifv6}_script.sh");
 			if (does_interface_exist($realifv6)) {
 				$ip6 = find_interface_ipv6($realifv6);
 				if (is_ipaddrv6($ip6) && $ip6 != "::") {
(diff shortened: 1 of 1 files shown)


etc/inc/interfaces.inc
```diff
@@ -4052,10 +4053,15 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif) {
 
 function DHCP6_Config_File_Override($wancfg, $wanif) {
 
-	$dhcp6cconf = file_get_contents($wancfg['adv_dhcp6_config_file_override_path']);
-	$dhcp6cconf = DHCP6_Config_File_Substitutions($wancfg, $wanif, $dhcp6cconf);
+	$dhcp6cconf = @file_get_contents($wancfg['adv_dhcp6_config_file_override_path']);
 
-	return $dhcp6cconf;
+	if ($dhcp6cconf === false) {
+		printf("Error: cannot open {$wancfg['adv_dhcp6_config_file_override_path']} in DHCP6_Config_File_Override() for reading.\n");
```
rbgarga: It's better to use log_error() here instead of printf()

etc/inc/interfaces.inc
```diff
@@ -4052,10 +4053,15 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif) {
...
+		return '';
+	} else {
+		$dhcp6cconf = DHCP6_Config_File_Substitutions($wancfg, $wanif, $dhcp6cconf);
```
rbgarga: You don't need to set a variable and return it on next line, just return function result directly

k-paulius:
Thanks for you feedback Renato. I have made the adjustments and also applied same fix to DHCP config file override function.

PR #1578, opened 2015-03-22
Add option to not register IPv4 and/or IPv6 addresses.

Robert-Nelson:
This change adds options to prevent registering IPv4 or IPv6 addresses.

### DIFF
--- a/etc/inc/services.inc
+++ b/etc/inc/services.inc
@@ -2422,7 +2422,7 @@ EOD;
 
 				conf_mount_rw();
 				/* Update IPv4 if we have it. */
-				if (is_ipaddrv4($wanip) && $dnsupdate['addresses'] != "ipv6") {
+				if (is_ipaddrv4($wanip) && $dnsupdate['addresstypes'] != "ipv6") {
 					if (($wanip != $cachedipv4) || (($currentTime - $cacheTimev4) > $maxCacheAgeSecs) || $forced) {
 						$upinst .= "update delete {$dnsupdate['host']}. A\n";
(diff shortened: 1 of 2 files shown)


rbgarga:
IMO it would be better if you add a single field to let user choose which field(s) he wants to update, A, AAAA or both, both being the default option.

Robert-Nelson:
Fixed to use radio buttons to select between IPv4, IPv6 or Both.

rbgarga:
Field name would make more sense to be called "protocol" or "ip_protocol" instead of addresses

Robert-Nelson:
I disagree, the purpose of dynamic DNS is to register addresses not protocols.  This option controls which type of addresses, addresses used by the IP4 or IP6 protocols.  If you don't like addresses then perhaps addresstypes but I believe calling it protocols would be misleading and incorrect.

rbgarga:
Yeah, you are right, they are address types and not protocols

Robert-Nelson:
Darn you were too quick, I just added a change to rename it to addresstypes which I think is better based on your feedback.

rbgarga:
yeah, it sounds great. I'll test and merge

rbgarga:
I'm just wondering about another option, call it 'Record type' and options 'A', 'AAAA'. Maybe it would make more sense for a DNS page

Robert-Nelson:
I thought of that too, but I think it is too much of a technical implementation detail and might not be as obvious to most administrators.

rbgarga:
I also discussed it with @jim-p and he suggested to change it to 'Record type' and use in the description A (IPv4) and AAAA (IPv6). Could you do the changes?

Robert-Nelson:
Okay I've made the changes.

PR #1579, opened 2015-03-23
Add option to not register IPv4 and/or IPv6 addresses.

Robert-Nelson:
This change adds options to prevent registering IPv4 or IPv6 addresses.


rbgarga:
You just need to send PR to branch master, this one can be closed.

PR #1583, opened 2015-03-24
Iterate array and append each line instead of using array_merge()

cbernander:
The array_merge() operation takes a long time to complete (especially for large input arrays), and with a large enough DHCP scope the dhcp_clean_leases() function will result in a PHP timeout.

This code runs significantly faster and should produce identical output in all cases since there will never be any string keys in either the $templease or the $newleases_contents arrays.


jim-p:
Truth be told I've been considering removing the dhcp_clean_leases code entirely. It seems to cause some grief with failover. That would be even faster. :-)

cbernander:
I'm OK with that; any fix that prevents the DHCP service from terminating every time a static lease is (re)configured is welcome!

PR #1585, opened 2015-03-25
Add support for DNS Made Easy

jlduran:
Documentation:

http://www.dnsmadeeasy.com/dynamic-dns/

### DIFF
--- a/etc/inc/dyndns.class
+++ b/etc/inc/dyndns.class
@@ -33,6 +33,7 @@
 	 *    - GleSYS (glesys.com)
 	 *    - DNSimple (dnsimple.com)
 	 *    - Google Domains (domains.google.com)
+	 *    - DNS Made Easy (www.dnsmadeeasy.com)
 	 * +----------------------------------------------------+
 	 *  Requirements:
 	 *    - PHP version 4.0.2 or higher with the CURL Library and the PCRE Library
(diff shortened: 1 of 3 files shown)


etc/inc/dyndns.class
```diff
@@ -1288,6 +1299,33 @@
 						$this->_debug($body);
 					}
 					break;
+				case 'dnsmadeeasy':
+					if (preg_match('/success/i', $data)) {
```
rbgarga: All these preg_match() calls are an overhead. IMO all these if/else blocks should be replace by a switch case block

etc/inc/dyndns.class
```diff
@@ -1288,6 +1299,33 @@
...
+						$status = "phpDynDNS({$this->_dnsHost}): (Success) IP Address Changed Successfully! (".$this->_dnsIP.")";
```
rbgarga: Error messages should have gettext() calls to be able to be translated to other languages

etc/inc/dyndns.class
jlduran: I agree that switch cases are more readable, however this seems to be the coding style for this file.

etc/inc/dyndns.class
jlduran: In that case, all the other entries should use `gettext()` as well. I am reluctant to change them in this PR.

etc/inc/dyndns.class
jlduran: Done!

PR #1587, opened 2015-03-26
Voucher messages using wrong config field name

Gertjanpfsense:
https://forum.pfsense.org/index.php?topic=91168.msg505273#msg505273
$config['voucher'][$cpzone]['msgnoaccess']
and 
$config['voucher'][$cpzone]['msgexpired']
do not exist.
These should be 
$config['voucher'][$cpzone]['descrmsgnoaccess']
and
$config['voucher'][$cpzone]['descrmsgexpired']


rbgarga:
Good catch! Thanks!

PR #1591, opened 2015-03-30
Change the default PHP session ID cookie name

jlduran:
By default PHP uses `PHPSESSID` as the name of the cookie that stores the session Id.

For example:

    $ curl -Ik https://portal.pfsense
    HTTP/1.1 200 OK
    ...
    Set-Cookie: PHPSESSID=00000000000000000000000000000000; path=/; secure; HttpOnly
    ...

This commit changes the name to `sessionId`.

References:

http://php.net/manual/en/session.configuration.php#ini.session.name

### DIFF
--- a/etc/inc/functions.inc
+++ b/etc/inc/functions.inc
@@ -85,7 +85,7 @@ if (!function_exists("get_menu_messages")) {
 
 			## Get Query Arguments from URL ###
 			foreach ($_REQUEST as $key => $value) {
-				if ($key != "PHPSESSID") {
+				if ($key != session_name()) {
 					$requests[] = $key.'='.$value;
 				}
 			}
--- a/etc/rc.php_ini_setup
+++ b/etc/rc.php_ini_setup
@@ -177,6 +177,7 @@ TIMEZONE=`cat /conf/config.xml | egrep -E '<timezone>(.*?)</timezone>' | awk -F'
 ; File generated from /etc/rc.php_ini_setup
 output_buffering = "0"
 expose_php = Off
+session.name = "sessionId"
 implicit_flush = true
 magic_quotes_gpc = Off
 max_execution_time = 900


cbuechler:
is there a reason I'm missing to change this from the default here?

jlduran:
Aesthetic reasons olny. A guest using the captive portal can look at the headers and find the following:

    $ curl -Ik https://portal.pfsense
    HTTP/1.1 200 OK
    [1] Expires: Tue, 31 Mar 2015 12:00:00 GMT
    [1] Expires: Thu, 19 Nov 1981 08:52:00 GMT
    [2] Cache-Control: max-age=180000
    [2] Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
    Last-Modified: Sun, 29 Mar 2015 12:00:00 GMT
    X-Frame-Options: SAMEORIGIN
    [3] Set-Cookie: PHPSESSID=00000000000000000000000000000000; path=/; secure; HttpOnly
    Pragma: no-cache
    Content-type: text/html
    Date: Sun, 29 Mar 2015 12:00:00 GMT
    [4] Server: lighttpd/1.4.35

Legend:

`1`: Remove duplication. 19 Nov 1981, 8:52 is hardcoded in `session.c` (@saschaschumann's birthdate)
`2`: Remove duplication.
`3`: Use a different name (other than the default). #1591
`4`: Has been addressed in #1572

Only `4` was really annoying. `1`, `2` and `3` are purely cosmetic changes.

cbuechler:
Now they can look in the header and see a different name instead? I fail to see where this improves anything. It's not necessarily purely cosmetic either since it could have negative impacts on already-existing cookies.

jlduran:
Well, this is the session cookie, so it will get invalidated once the pfsense is restarted (for example).
Yes, it was only a cosmetic suggestion, MS ASP uses `ASPSESSIONID`, JEE uses `JSESSIONID`, etc. It is customary to rename it for cosmetic reasons.

PR #1592, opened 2015-03-31
Small cleanup fixes for ipsec.conf.

gitbensons:
This is cleaner version of #1563 

- In case for IKEv1, just use the UI selected algorithm for phase2
- For IKEv2, just removing duplicates from the array. This is the best way to do that, otherwise we would need to copy&paste code above

### DIFF
--- a/etc/inc/vpn.inc
+++ b/etc/inc/vpn.inc
@@ -1082,12 +1082,6 @@ EOD;
 			if (!empty($ealgosp1)) {
 				$ipsecconnect .= "\t{$ealgosp1}\n";
 			}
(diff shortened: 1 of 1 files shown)


rbgarga:

cbuechler:
@gitbensons I'm not seeing any way of ending up with duplicates here, or why that change might otherwise be desirable. Could you provide the <ipsec> XML for an affected circumstance?

cbuechler:
Still don't see any way you can end up with duplicates there. Guessing something elsewhere here changed in the mean time. 

@gitbensons if there's something here we're missing, please let us know how to replicate a problem circumstance and we'll get it promptly fixed. Thanks!

PR #1593, opened 2015-03-31
Fix IPsec Advanced Settings uniqueids

gitbensons:
It was neither set in strongswan ipsec.conf, nor picked up correctly in the UI.


ermal:
Merged, thx.

PR #1596, opened 2015-04-04
Typo of processed

phil-davis:


cbuechler:
Thanks Phil. I didn't notice this pull request until after I'd already fixed that in the process of revamping those bits as a whole, so that's already fixed.

PR #1602, opened 2015-04-10

phil-davis:
on factory default boot.
Alix - vr1 vr0
APU - re1 re2
serial cable and do interface assignment that way.

### DIFF
--- a/etc/inc/config.console.inc
+++ b/etc/inc/config.console.inc
@@ -559,23 +559,52 @@ EOD;
 function check_for_alternate_interfaces() {
 	global $config;
 
-	// If the factory default configured WAN and/or LAN devices do not exist,
-	// then use alternate devices if they exist.
-	// These are devices on Alix (vr) and APU (re) systems, and this lets them boot a
-	// factory default config without being forced to do interface assignment
-	// over the serial port.
+	// If the WAN and/or LAN devices in the factory default config do not exist,
+	// then look for alternate devices.
+	// This lets many systems boot a factory default config without being
+	// forced to do interface assignment on the console.
 
-	$other_devices_arr['wan'] = array("vr1", "re1");
-	$other_devices_arr['lan'] = array("vr0", "re2");
+	$specplatform = system_identify_specific_platform();
+	$default_device = array();
+
+	// If we recognise the platform, then specify the devices directly.
+	switch ($specplatform['name']) {
+		case 'alix':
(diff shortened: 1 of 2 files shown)


rbgarga:
What about use system_identify_specific_platform() to detect some pre-defined platforms?

phil-davis:
I won't have access to real hardware to test on until I get to work on Monday. But yes, a mix of the 2 approaches seems good - on a factory defaults boot detect particular known hardware and set the interface assignments based on that, then if nothing matched look for some common network devices and use those fro WAN and LAN.
If you already have a workable solution then feel free to commit it and close this.
If this is still open next week then I will work on enhancing it as you suggest.

phil-davis:
I added code to use system_identify_specific_platform()
For the general case of other boxes that are not one of the known platforms, I think we still want lists of common device names that are reasonable to default to. e.g. some bit of hardware that has "vr" devices but is not seen as "alix" - it would be reasonable to set vr0/1 for the user...
So that makes the code look a bit redundant actually. But at least it clearly shows support for the various specific platforms in the switch statement.
I am not familiar with device names for oter platforms. So when you are happy with the code structure, you could then add cases for FW7541 and so on.

phil-davis:
Any chance that this or a functionally similar solution can go into a release in the near future?
It really is nice if, on as much hardware as possible, a virgin system and/or reset to factory defaults results in a system with a sensible functioning LAN (and WAN even) that can be used to continue with configuration without needing console access. And if WAN DHCP works to some upstream ISP device, then actually an ordinary user on LAN will get internet access, and remote support can connect to an ordinary user client on LAN side with TeamViewer, then browser to pfSense... and proceed with settings.

nagyrobi:
How about dropping this stupid BSD-style interface naming and using something more universal, like eth0, eth1, eth2 and so on.
Just take the interfaces and enumerate them by the MAC address number increasing. So the lowest MAC address would always be eth0, and up.
Installing a new interface may cause of course the need to re-assign, as it is now. But it would save a lot of headaches in other aspects.

Various Linux flavorus have implemented UDEV, which is a bit more complex, it assigns the names by MAC address at system installation, but when adding a new interface it doesn't mess up, it follows enumeration upwards.

phil-davis:
This is integrated with the recent repo folder structure changes and resubmitted as https://github.com/pfsense/pfsense/pull/1819

PR #1604, opened 2015-04-11
When deleting a VIP check it is not used by OpenVPN

phil-davis:
I noticed this when cleaning up VIPs and OpenVPN server when testing for this forum post https://forum.pfsense.org/index.php?topic=92174.0
The system let me delete my test VIP before I deleted the OpenVPN server using, probably not a good thing.

### DIFF
--- a/usr/local/www/firewall_virtual_ip.php
+++ b/usr/local/www/firewall_virtual_ip.php
@@ -108,6 +108,30 @@ if ($_GET['act'] == "del") {
 			}
 		}
 
+		/* make sure no OpenVPN server references this entry */
+		if (is_array($config['openvpn']['openvpn-server'])) {
+			foreach ($config['openvpn']['openvpn-server'] as $openvpn) {
+				if ($openvpn['ipaddr'] <> "") {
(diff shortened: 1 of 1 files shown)


usr/local/www/firewall_virtual_ip.php
```diff
@@ -108,6 +108,30 @@
 			}
 		}
 
+		/* make sure no OpenVPN server references this entry */
+		if (is_array($config['openvpn']['openvpn-server'])) {
+			foreach ($config['openvpn']['openvpn-server'] as $openvpn) {
+				if ($openvpn['ipaddr'] <> "") {
+					if ($openvpn['ipaddr'] == $a_vip[$_GET['id']]['subnet']) {
+						$input_errors[] = gettext("This entry cannot be deleted because it is still referenced by at least one OpenVPN server.");
```
rbgarga: It would be useful to point which tunnel is using this IP

phil-davis: Good idea - I was too lazy.

phil-davis:
I combined the server and client loops as well as adding the OpenVPN instance description to the error message. Easier to make another pull request, since the new-and-better code is significantly different to this anyway. See #1608

PR #1606, opened 2015-04-13
Domain override with multiple authoritative DNS servers

phil-davis:
Tell users that this is possible in DNS Resolver and how to achieve it. The code in unbound.inc already supports it and works.
I had asked for this in Redmine feature request #4350 and when I went to look at coding to implement it I found code that already did it.

### DIFF
--- a/usr/local/www/services_unbound.php
+++ b/usr/local/www/services_unbound.php
@@ -498,7 +498,8 @@ function show_advanced_dns() {
 </tr>
 <tr>
 	<td><p><?=gettext("Entries in this area override an entire domain by specifying an".
-	" authoritative DNS server to be queried for that domain.");?></p></td>
+	" authoritative DNS server to be queried for that domain.");?>
+	<?=gettext("If there are multiple authoritative DNS servers available for a domain then make a separate entry for each, using the same domain name.");?></p></td>
 </tr>
 </table>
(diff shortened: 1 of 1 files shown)


ocafebabe:
@phil-davis it seems that your fix that was explaining that it was possible to have multiple authoritative DNS servers for the same domain override isn't there anymore? I'm using 2.4.3-p1 and the description under domain overrides doesn't include your comment!

phil-davis:
The code was there when the whole source code tree was renamed under ``src`` folder:
https://github.com/pfsense/pfsense/blob/46bc6e545a17e77202aaf01ec0cd8d5a46567525/src/usr/local/www/services_unbound.php

But then the bootstrap work was merged:
https://github.com/pfsense/pfsense/commit/03b19a93f4d8d870507ee96121cee4acd748dd2a#diff-2b0179476e287738228504f858dc0cc2

That commit claims to have renamed the file without changes??? But that is what happened in the previous commit.

But at that point the code here is gone.

phil-davis:
I think bootstrap removed various of those "on-screen" chunks of help text. Maybe this can go in the docs, which are reached by pressing the help "?" from ``services_unbound.php``

docs PR https://github.com/pfsense/docs/pull/39

PR #1607, opened 2015-04-13
Expiring vouchers doesn't disconnect portal client - it should.

Gertjanpfsense:
As found here : https://forum.pfsense.org/index.php?topic=92157.0
Solution : https://forum.pfsense.org/index.php?topic=92157.msg510872#msg510872.


Gertjanpfsense:
Closed.
Proposed code 'works' but isn't perfect.

PR #1614, opened 2015-04-16
Edit gateway max weight

nmors:
Useful for load balancing, where you have more than a 5:1 difference in gateway speed.

In my case, this is required for balancing CABLE & ADSL WAN connections properly, where I need to set the CABLE weight to 8, which is higher than the current maximum (5)

### DIFF
--- a/usr/local/www/system_gateways_edit.php
+++ b/usr/local/www/system_gateways_edit.php
@@ -754,7 +754,7 @@ function enable_change() {
 								<td width="78%" class="vtable">
 									<select name='weight' class='formfldselect' id='weight'>
 									<?php
-										for ($i = 1; $i < 6; $i++) {
+										for ($i = 1; $i < 21; $i++) {
 											$selected = "";
 											if ($pconfig['weight'] == $i)
 												$selected = "selected='selected'";


rbgarga:

nmors:
CLA completed

nmors:
Just wondering; are there any technical limitations that I have failed to become aware of, which gives reason to have gateway's max. weight set to 5? 

I'd like to understand why this limit exists, or how the max. weight of 5 was decided upon in the first place.

Thanks,
Nathan

rbgarga:
Maybe @jim-p or @cbuechler remember why it was limited to 5 in the past

jim-p:
IIRC it was to keep the calculated line length for pf to a sane value. The "weight" is accomplished by repeating the gateway address in the pf line and if we allow that to have too many gateways with a high weight, pf will fail to parse the rule. At least that is how it was back when the code first went in, there have been changes to pf in the meantime, might be worth trying again.

cbuechler:
Yeah it's for the reason @jim-p noted. I just checked 2.3 to see where that limit is currently. It can have 384 interfaces/gateways specified in the route-to round-robin now. I think it was significantly less than that when the weights were first implemented. It'd be possible, though difficult, to reach that limit with a max gateway weight of 20. 20 WANs with weight 20 and one with weight 1 would suffice to exceed that limit. 

If we bump the limit, we'll also want to add something to filter.inc to prevent creating an invalid ruleset.

cbuechler:
I committed a change to allow weights up to 30, and added check in filter.inc so it won't create too long of a route-to line.

nmors:
That's great. Thanks for implementing this.

PR #1622, opened 2015-04-19
Code style usr-local-www back-end

phil-davis:
files that mostly do stuff in the background.

### DIFF
--- a/usr/local/www/fbegin.inc
+++ b/usr/local/www/fbegin.inc
@@ -28,8 +28,7 @@ if (empty($pagename)) {
 	$pagename = "index.php";
 }
 
-/* If the filename is pkg_edit.php or wizard.php, reparse looking
-	for the .xml filename */
+/* If the filename is pkg_edit.php or wizard.php, reparse looking for the .xml filename */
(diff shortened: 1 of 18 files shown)


usr/local/www/fbegin.inc
```diff
@@ -245,11 +256,12 @@ $gold_menu = array();
...
 
-if(! $g['disablehelpmenu']) {
+if (!$g['disablehelpmenu']) {
 	$help_menu = array();
 	$help_menu[] = array(gettext("About this Page"), $helpurl);
-	if($g['product_name'] == "pfSense")
-	$help_menu[] = array(gettext("Bug Database"), "https://www.pfsense.org/j.php?jumpto=redmine");
+	if ($g['product_name'] == "pfSense") {
+		$help_menu[] = array(gettext("Bug Database"), "https://www.pfsense.org/j.php?jumpto=redmine");
+	}
```
phil-davis: Note: this "if" statement was not tabbed correctly. I put the curlies and tabbing to make it clear what it is doing, without changing the current behavior..
But there are more things on the help menu, like User Forum... directly below this "if" test - when product_name is not pfSense, should they be in the Help menu or not?

PR #1627, opened 2015-04-21
Clarify switch case statement

phil-davis:
Ermal made these semi-colons into colons on master some time ago with https://github.com/pfsense/pfsense/commit/83b8ed6b2bec13d3b60acd9bd4786e5a7df4de90 which also references bug 4202. For some reason that change did not get into RELENG_2_2.
Actually semi-colon is legal syntax (surprised me) so I do not see how that commit really changed anything.
In any case, using semi-colon here is so unusual that it seems good to fix it up in RELENG_2_2, hence this pull request.
If https://github.com/pfsense/pfsense/pull/1626 is also cherry-picked into RELENG_2_2 that will get rid of the remaining places where switch-case-semicolon appears in the code.

### DIFF
--- a/etc/inc/ipsec.inc
+++ b/etc/inc/ipsec.inc
@@ -675,16 +675,16 @@ function ipsec_find_id(& $ph1ent, $side = "local", $rgmap = array()) {
 		$thisid_data = $rgmap[$ph1ent['remote-gateway']];
 		break;
 
-	case "address";
+	case "address":
 		$thisid_data = $id_data;
 		break;
 
-	case "fqdn";
-	case "keyid tag";
-	case "user_fqdn";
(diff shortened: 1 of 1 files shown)


phil-davis:
Another slightly-related note:
https://github.com/pfsense/pfsense/commit/5324ea38a1783200f3d8c1eec1229505c19968ef
Move to specifically specifying the ID type...
That commit, which is around the code here also, does not appear in RELENG_2_2. But it looks like "a good thing". Is it really supposed to be waiting for 2.3? Just checking in case it got missed for RELENG_2_2.

ermal:
Can be closed since i restored the original commit.
It was reverted after rollback of strongswan 5.2.1 update.

PR #1633, opened 2015-04-30
Independent Dashboard Theme

BBcan177:
With the current **'Theme'** design, the chosen theme is used in all pages.

This work-around will allow the setting of a 'theme' for the Dashboard, Independent of the Theme for the balance of the pfSense Pages.

Use case would be setting the "Dashboard" to a Wide-Screen theme, and leaving all other pages in a Standard Width theme.

When the   **index.php**   file is loaded, if the "Dashboard Theme" is defined with a theme that is different than the "Default Theme", the "Default Theme" setting is captured into    **$g['theme_revert']** . The new "Dashboard Theme" is now used for the Dashboard. All other pages will use the "Default Theme" setting.

If the user applies any changes in any of the Dashboard Widgets, it will save the "Dashboard Theme" as the "Default Theme" and this will cause all future pages to use the "Dashboard Theme". To over come this issue, code is added to the php  **function    write_config()**    to revert the "Theme" back to its "Default" before saving the changes to the config.xml file.

Future designs of Themes should allow for independent Theme customization per page to better utilize
Wide-Screen/Standard-Width-Screen settings.


usr/local/www/system.php
```diff
@@ -511,6 +512,32 @@ function is_timezone($elt) {
...
+				<td width="78%" class="vtable">
+					<select name="theme_dashboard">
+						<?php
+							$files = return_dir_as_array("/usr/local/www/themes/");
+							foreach($files as $f):
+								if ((substr($f, 0, 1) == "_") && !isset($config['system']['developer']))
+									continue;
+								if ($f == "CVS")
+									continue;
+								$curtheme = "pfsense";
```
phil-davis: When I tried this, I immediately got "pfsense" theme on the dashboard, when my overall theme was set to pfsense_ng. I think that was even before selecting a them here and saving.
Perhaps $curtheme here needs to default to $config['theme']
 And the code up in index.php should only happen if isset($config['theme_dashboard']) - at the moment that code executes when 'theme_dashboard' has not been set, which I think is why on first use $config['theme'] ends up null, and then I guess something later uses 'pfsense' as the theme when the theme is blank.

BBcan177:
Thanks Phil, I have incorporated those changes.

rbgarga:
pfSense is moving to bootstrap, as you can see in https://github.com/SjonHortensius/pfsense. I'm not sure it's worth it to add this right now.

PR #1634, opened 2015-05-05
Fix DHCPv6 Dynamic DNS

Robert-Nelson:
There are three issues:
- The wrong domain was being used, domain instead of ddnsdomain.
- The option "deny client-updates" wasn't being set so forward entries weren't being added.
- No zone information was provided for reverse DNS.


Robert-Nelson:
Doesn't look like anyone is going to get to this any time soon and these fixes are also in PR #1638.

Since I don't feel like maintaining both PRs for an indeterminate time I'm closing this one.

PR #1635, opened 2015-05-05
Call clear_subsystem_dirty('staticmaps') if using Unbound

Robert-Nelson:


Robert-Nelson:
This fixes issue # 4678 - DHCPv6 with static entries, Apply configuration button never goes away

Gertjanpfsense:
Thanks !
I tried your proposal. Looking good.
Finally the "Your have to apply" button goes away.

PR #1636, opened 2015-05-07
Unmatched td in firewall_nat

phil-davis:
This file seems to have an unmatched "td" ending. Adding the line here matches the "td" at line 320 and this embraces the little table that has the 4 icons in it in a square that comes at the right hand end of each port-forward entry in the main table.
I can't see any difference in the rendering of the page, at least on Firefox, with and without this fix.
The tabbing of this file is woeful. I am fixing that up with a code-style review. But thought I should do a separate pull request for this kind-of-functional fix.


phil-davis:
This commit https://github.com/pfsense/pfsense/commit/00bcbdd0855a880bc251e82e979d7ab8c30859ba seems to have introduced the "td" at line 216 that is now at line 320 and was never matched by "/td" after the table of stuff that is inside it. That was back in 2005!

PR #1637, opened 2015-05-07
Fix uniqeness 

alex-berger:
Currently pfsense enforces unique unqualified hostnames for static dhcp leases, which is not correct as only the fully qualified hostname (hostname + domainname) must be unique. With this commit the old validation logic for uniqeness is modified such that hostnames no longer need to be unique and at the same time the fully qualified hostname hast to be unique.

This change makes it possible to have host with identical hostnames in different (sub)domains. For example myhost.sales.acme.com and myhost.support.acme.com will now be possible.

### DIFF
--- a/usr/local/www/services_dhcp_edit.php
+++ b/usr/local/www/services_dhcp_edit.php
@@ -187,19 +187,19 @@ if ($_POST) {
 	}
 
 	/* check for overlaps */
-  foreach ($a_maps as $mapent) {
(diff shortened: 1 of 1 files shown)


rbgarga:
Please respect style guide - https://doc.pfsense.org/index.php/Developer_Style_Guide

This commit broke indent

rbgarga:

alex-berger:
OK, I signed an ICLA.

PR #1638, opened 2015-05-07
Add support for DHCP when interface is set to track a delegated prefix

Robert-Nelson:
This fixes issue # 3029.


rbgarga:
@Robert-Nelson sorry for the long delay. Since this patch was submitted before code was moved to src/ subdir it doesn't apply cleanly anymore. Do you plan to merge it to recent version of pfSense repo?

rbgarga:
@Robert-Nelson I've merged DDNS changes, as you can see at 391d63da8fa0c46daa7275c225222dff5ec2522b, with a fix on PTR zones.

rbgarga:
I've imported this patch to my local tree, but it only works with PD with length = 64. I'll push it as soon as I finish to fix it to work with other lengths

PR #1646, opened 2015-05-11
Code style Firewall Rules

phil-davis:


rbgarga:
This one cannot be automatically merged, can you please rebase your fork @phil-davis ?

phil-davis:
Should be integrated now.

phil-davis:
I have a small fix for the change to show selected interfaces on the Floating Rules tab.
When there are no floating rules defined, the box that says this only extends over 10 columns, but now that tab has 11 columns.
The fix is on top of this change, so it will be easy to submit a pull request for it once this pull request is committed.

PR #1653, opened 2015-05-13
Code style Load Balancer

phil-davis:
I noticed that load_balancer_relay*.php are not really used. They seem to be code that was in development at some time but never quite finished. Anyway, those files are also now formatted to the code style guide.


rbgarga:
Already merged

PR #1660, opened 2015-05-15
Code style WWW Status RRD

phil-davis:


usr/local/www/status_rrd_graph.php
```diff
@@ -166,7 +167,7 @@
 }
 
 if (is_numeric($_GET['end'])) {
-        $end = $_GET['end'];
+		$end = $_GET['end'];
```
rbgarga: There is an extra tab here

usr/local/www/status_rrd_graph.php
```diff
@@ -174,29 +175,30 @@
 		$end = $now;
 	}
 } else {
-        $end = $now;
+		$end = $now;
```
rbgarga: and also here

phil-davis:
Well spotted - my editor converts blocks of 4 spaces into a tab, then I often have to rub out extra tabs while reviewing. So I missed a few here.

PR #1661, opened 2015-05-15
Fix startingat var name typo in pkg.php

phil-davis:
Even with this fix, the code does not make sense. The first test is:
if ($startingat > -1)

if it gets into the else, then $startingat must be negative. But the test in the "else" tests for >1 ???

Someone who knows what it really is supposed to do could fix it!


rbgarga:
The second test should be:
```
if ($startdisplayingat > 1)
```
And this will be used to add href when there are at least one item on previous page. If you want to see it working you just need to set a URL with:
```
startdisplayingat=2&display_maximum_rows=5
```
Code won't set startdisplayingat to 2, it's a safe belt.

rbgarga:
I'll merge your patch and push a fix after it. Thanks!

phil-davis:
Oops - I deleted 1 too many branches.

PR #1664, opened 2015-05-16
Trim trailing whitespace

jlduran:
Perhaps this commit is too big to merge at once. Those extra spaces are everywhere! :scissors:
Let me know if I should split it.


rbgarga:
It would be better to send it in small chunks. And also, there are files that we import from vendors, for instance .pat files for l7 protocol patterns. Remove spaces from those files will make it harder to us to compare with new versions of the same files.

I'm not sure if you are aware, but @phil-davis is doing a great job fixing trailing spaces, indent and code style on all files, so it probably is going to conflict with hist work. If you would like to help, please co-ordinate with him.

Thanks!

phil-davis:
The remaining files I have to do are in usr/local/www
services*
system*
vpn*
and there is a pull request already done for pkg* and waiting to coordinate with @rbgarga working on pkg stuff for 2.3
So remove those files because I will get to them anyway (hopefully I will finish this week - I have spare evenings). And as Renato says, do not change files that are modules provided by other places - we put up with things in those that do not meet the style guide.

jlduran:
The PR is now in chunks of smaller commits, just in case you want to cherry pick from it.

As a suggestion, is it viable to create an [`.editorconfig`](http://editorconfig.org) file?

rbgarga:
I don't see a problem to create a .editorconfig file

jlduran:
To start with just `.php` files. According to the [Developer Style Guide](https://doc.pfsense.org/index.php/Developer_Style_Guide):

- Use tabs for indentation (not spaces or a mixture of both)
- Use a tab stop of 8, rather than 4
- Ensure there is NO trailing whitespace at the end of a line, for example spaces or tabs when there is no more text afterward
- Ensure there is NO whitespace on empty lines. For example, a line must not contain only spaces or only tabs

Additionally, I've added:

- UTF-8 character encoding
- LF line endings
- Final newline

Proposed `.editorconfig` file (`.php` files for now):

```ini
; This file is for unifying the coding style for different editors and IDEs.
; More information at http://editorconfig.org

root = true

[*.php]
indent_style = tab
tab_width = 8
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

rbgarga:
I never used editorconfig and have a question. Can it deal with multiple line conditionals indent, that uses 4 spaces after tab?

```
if ($this_big_var_name == "this_big_string_value" ||
    $x == true) {
        foo();
}
```

jlduran:
EditorConfig defines just basic settings. For this case, it tells your editor that when <kbd>Tab ↹</kbd> is pressed, a tab (`\t`) should be inserted and displayed as 8 spaces. You should press the space key 4 times for that second line anyway.

From the Style Guide:

- If a conditional statement must span multiple lines, indent using four spaces to align with the start of the conditional above it:

```php
if ($foo1 && $foo2 && $foo3 && $foo4 && $foo5 && $foo6 &&
    $foo7 && $foo8 && $foo9) {
        something();
}
```

> **Note**: In the code above I used only spaces for displaying purposes.

If we "show invisibles":

```php
if ($foo1 && $foo2 && $foo3 && $foo4 && $foo5 && $foo6 &&
␣␣␣␣$foo7 && $foo8 && $foo9) {
↹       something();
}
```

Some people prefer to avoid having this type of mixed indentation and use only spaces. In this case, change to `indent_style = space` and add `indent_size = 8`. Now, when you press <kbd>Tab ↹</kbd>, the editor will insert 8 spaces.

PR #1665, opened 2015-05-18
Code style vpn_l2tp

phil-davis:


rbgarga:
Just a note, PPTP is going to be removed on 2.3, then please do not waste your time touching those files. You are doing an awesome job here, thanks!

PR #1669, opened 2015-05-20
Fix tr use for combinedfields in pkg xml

phil-davis:
When specifying combinedfields begin and end in a package XML file, IE reports some unexpected start tag messages. This is because some "td" tags are not enclosed in "tr" tags in these cases, both at the begin and end of the combinedfields table that is generated. Firefox seems to be forgiving of this.
This has been investigated because BBscan177 who is developing pfBlockerNG is trying to use the combinedfields feature and it was not working so well. I will ask him to try these changes for RELENG_2_2 and confirm that they work.
When combinedfields is not set, then the code path with this change results in the same HTML being output as previously, so it should not change the HTML generated for packages that do not use combinedfields - i.e. I claim not to have broken anything else :)

### DIFF
--- a/usr/local/www/pkg_edit.php
+++ b/usr/local/www/pkg_edit.php
@@ -509,7 +509,7 @@ if ($pkg['tabs'] <> "") {
 		$size = "";
 		if (isset($pkga['dontdisplayname'])) {
 			$input="";
-			if (!isset($pkga['combinefields'])) {
+			if ($pkga['combinefields'] != "begin") {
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: see https://github.com/pfsense/pfsense/pull/1668 for a version of this that I have used on a 2.2.2 system with success - if there are any tricky bits in cherry-picking this back to RELENG_2_2 then use that pull request as a guide to what works.

PR #1708, opened 2015-06-06
Pass-through MAC for certain users

paulxb:
Small modification to allow the add pass-through MACs only for certain users. 
eg we have 2 users: staff & guest and want to add all the staff MAC addresses.


rbgarga:

cbuechler:
This is of very limited use, don't see this as something attractive to add. Adding technical debt for things 1 person in the world wants isn't good. Most of the time the staff and guest access will be on separate interfaces and hence separate captive portal instances. 

Appreciate the contribution, but not something we'll merge.

PR #1711, opened 2015-06-09
Bug fix for #4065. After base64 encoding username/password, properly escape characters =,+,/ before submitting auth details

the-real-ed:
For full details regarding the patch, please see: https://redmine.pfsense.org/issues/4605#note-1


the-real-ed:
Sorry, my commit message/pull request references the incorrect bug (#4065). This is a fix for #4605.

For the fix for #4605 please see this new pull request: https://github.com/pfsense/pfsense/pull/1712

PR #1715, opened 2015-06-12
Set correct noshuntlaninterfaces boolean after save

phil-davis:
The truth value of noshuntlaninterfaces is displayed the opposite to what is actually stored in the config - a bit tricky, but it should work. The logic is:
1) read value from config into $pconfig array
2) if $pconfig value false then display box checked, otherwise display box unchecked
3) when saving, if $POST value (from checkbox state) is true then unset noshuntlaninterfaces in config, otherwise set it. (change/fix by Ermal just a few hours ago)

The remaining problem here is that the display code uses the value in $pconfig array to decide what to display in the checkbox. During the $POST processing, $pconfig array is getting set to what was sent in $POST. So after a save the checkbox was getting displayed in the wrong state. A full refresh of the page would display it correctly based on the "opposite of what is in $config".
For most settings the $POST value is the same as what ends up in $config so it works. But for noshuntlaninterfaces the value in $POST is the opposite of what goes into $config.

Solution: when processing the $POST (and save if no input_errors...) set the $pconfig value based on the opposite of what was in $POST.
I take from $POST rather than $config because this preserves whatever the user might have changed the checkbox too, for redisplay, particularly in the case of input_errors (from other inputs).

### DIFF
--- a/usr/local/www/vpn_ipsec_settings.php
+++ b/usr/local/www/vpn_ipsec_settings.php
@@ -216,6 +216,14 @@ if ($_POST) {
 //		header("Location: vpn_ipsec_settings.php");
 //		return;
 	}
+
+	// The logic value sent by $POST is opposite to the way it is stored in the config.
+	// Reset the $pconfig value so it reflects the opposite of what was $POSTed.
+	if ($_POST['noshuntlaninterfaces'] == "yes") {
+		$pconfig['noshuntlaninterfaces'] = false;
+	} else {
+		$pconfig['noshuntlaninterfaces'] = true;
+	}
 }
 
 $pgtitle = array(gettext("VPN"), gettext("IPsec"), gettext("Settings"));


ermal:
I solved it differently since i do not like having to scratch my head on things later on ;)

Thanks though Phil

phil-davis:
@ermal I don't see any commit that makes the fix equivalent to this - is it still coming?
Without the fix here:
1) Go to the settings page.
2) Press save without changing anything.
3) After the save, the "noshuntlaninterfaces" checkbox is displayed the opposite way.
4) Refresh the whole page - now it shows the same as (1).
Actually config.xml is fine underneath - the problem is just at step (3) when the GUI page is redisplayed directly after a save.

ermal:
is this one basically https://github.com/pfsense/pfsense/commit/b5d102e779383af35af435e321691f3df2c58c5b

phil-davis:
Enabling the redirect fixes it for the case when !$input_errors - it effectively refreshes the page.
But if there are input errors, then the page needs to be redisplayed with the errors reported and the user settings re-displayed as they were. For example:
1) Check the box "	Enable MSS clamping on VPN traffic"
2) Enter an invalid value (e.g. 555) in "Maximum MSS" box
3) Press Save
The page is redisplayed with the error reported, but the displayed state of the "Enable bypass for LAN interface IP" checkbox is changed - problem.
4) Now fix the invalid value - change 555 to 576 - and press Save
Now the user has accidentally also changed the state of "Enable bypass for LAN interface IP".

ermal:
Hrm you are right let me fix that case as well.
Thx for spotting it.

phil-davis:
@ermal With your latest fix (moving the redirect code further out).
Now do this:
1) Open the IPsec settings page
2) Select "Enable MSS clamping on VPN traffic"
3) Enter 555 in the box
4) Press Save
The page refreshes, there is no message to the user about the invalid data, the data the user attempted to enter simply vanishes.
You could just use the code in this pull request - it puts the data back the correctly in $pconfig for display to the user, leaving any $input_errors to also be displayed :)

ermal:
Ok the last commit similar to this i did hopefully closes the comedy :)

phil-davis:
Sorry, no. In the UI forms, when there is an input error, we usually try to redisplay the form with all the changes that the user made. That way the user can fix whatever value/s caused the input error/s and does not have to re-input other changes they made.
That is not working here:
1) Display the IPsec Advanced Settings page - e.g. with "Enable MSS clamping on VPN traffic" off and "Enable bypass for LAN interface IP" off.
2) Select "Enable bypass for LAN interface IP"
3) Select "Enable MSS clamping on VPN traffic" and enter 555 in the box.
4) Press Save
The form is re-display with the input error message about "555" being invalid - good.
But "Enable bypass for LAN interface IP" is now off.
That breaks the convention that works on all the other input fields - that changed-pending values are redisplayed to the user after there are input errors.
The code needs to set $pconfig['noshuntlaninterfaces'] based on what the user had last selected - that information is in the $POST value, but in this case has to be the opposite of the $POST value.
I want to find some way to say nicely that my pull request here does what is required for this processing - why not just use it, or I am happy if you copy functionally the same thing.

ermal:
Merged manually.

PR #1719, opened 2015-06-15
XHTML Compliance - Diagnostics Tables

N0YB:
html id's not permitted to begin with a number.
html id's not permitted to contain '/'
add prefix (entry_) and replace slash with hyphen.
table entry id format becomes: entry_<ip address>-<cidr>
replacing the format: <ip address>/<cidr>
does not change the displayed format.

### DIFF
--- a/usr/local/www/diag_tables.php
+++ b/usr/local/www/diag_tables.php
@@ -114,7 +114,8 @@ include("head.inc");
 			if (200 == response.status) {
 				// Escape all dots to not confuse jQuery selectors
 				name = response.responseText.replace(/\./g,'\\.');
-				name = name.replace(/\//g,'\\/');
+				name = name.replace(/\//g,'\\-');
+				name = "entry_" + name;
 				jQuery('#' + name).fadeOut(1000);
 			}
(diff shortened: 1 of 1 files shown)


usr/local/www/diag_tables.php
```diff
@@ -147,7 +147,7 @@ function del_entry(entry) {
 	foreach ($entries as $entryA):
 		$entry = trim($entryA);
 ?>
-	<tr id="<?=$entry?>">
+	<tr id="entry_<?=str_replace("/", "-", $entry)?>">
```
rbgarga: Missing semicolon

rbgarga:
It breaks javascript function that fade out the item when you delete it

rbgarga:
Something wrong happened here, for some reason commit bf2dba7 is missing from Pull Request, if you check https://github.com/pfsense/pfsense/pull/1719/files you will see it only shows 7d8dd0b

N0YB:
Re-pushed and it seem to be there now.  Please let me know if it's not.  Thanks.

PR #1722, opened 2015-06-15
Fixing PD size selection for HughesNet

chapmajs:
HughesNet, a popular US satellite Internet provider, provides native IPv6 via DHCPv6 prefix delegation. They provide a /61 prefix, which pfSense does not currently support. Using an incorrect prefix size will result in intermittent IPv6 connectivity as the prefix delegation changes with time. I've forced a prefix size of 3 for the /61 using Firefox's developer tools. This PR integrates that change, it's just another key => value pair for the select list.


ermal:
Have you signed the CLA?

rbgarga:

chapmajs:
Just signed the CLA.

PR #1723, opened 2015-06-16
Why is break missing for reinstallxml

phil-davis:
I thought that "reinstallxml" should do less than "reinstallpkg" but actually it was getting stuff here, then falling through "reinstalpkg" which did delete_package_xml and then install_pkg, which got the files a 2nd time and...
Maybe that was supposed to happen?
Anyway, I thought I would point this out and someone can either commit this pull request if the "break" should be there, or explain to me why "reinstallxml" is supposed to fall through executing all this code.

### DIFF
--- a/usr/local/www/pkg_mgr_install.php
+++ b/usr/local/www/pkg_mgr_install.php
@@ -238,6 +238,7 @@ if ($_GET) {
 		case 'reinstallxml':
 			pkg_fetch_config_file($pkgid);
 			pkg_fetch_additional_files($pkgid);
+			break;
 		case 'reinstallpkg':
 			delete_package_xml($pkgid);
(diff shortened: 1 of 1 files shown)


ermal:
This was intentional behaviour.

phil-davis:
then unmerge it - I was just asking because it seemed odd to me that "reinstall XML" did even more than "reinstall package".

cbuechler:
Its entire existence is screwy. Phil's right on the original intent of it from way back in the day, it should only do those two things. It's fine to have merged. 

That entire button is going away in 2.3, something @rbgarga and I discussed at BSDCan. The package and the GUI won't be separately reinstallable in 2.3 and forward.

PR #1726, opened 2015-06-19
Allow "From" field in RFC822 format

thiagoc:


etc/inc/notices.inc
```diff
@@ -350,6 +350,10 @@ function send_smtp_message($message, $subject = "(no subject)", $force = false)
 
 	if ($config['notifications']['smtp']['fromaddress']) {
 		$from = $config['notifications']['smtp']['fromaddress'];
+		preg_match('/[\w\s]+/', $from, $matches);
+		$from_display = trim($matches[0]);
+		if(preg_match('/(?:<)(.+)(?:>)$/', $from, $matches))
+			$from = $matches[1];
```
rbgarga: Based on this logic, if user fills 'fromaddress' with "renato@pfsense.org" it will end up with:
```
From: renato <renato@pfsense.org>
```
Where is the current code preventing you to fill the fromaddress field with an address in RFC822 format? This change is only forcing and changing user input.

thiagoc: Sorry, I don't know if I got it right. This change allow users to input "
renato@pfsense.org" or "Renato Botelho <renato@pfsense.org>", either way
will work.

You think it's better to have another field for $from_display?

rbgarga: What I meant is, without this change, you already are able to add both forms, there is no check preventing you to do it. And with the change, if you add a single email address (e.g. renato@pfsense.org) it's going to be modified to "renato <renato@pfsense.org>" because preg_match at line 353 is going to match every char before '@'.

thiagoc: Yes, you're right. What do you think about add another field on the page to
fill $from_display?

rbgarga: I don't think it's a good idea, it would just add more complication with no gain. You can improve the text that describe the field in that form so users will be aware they can use both formats. I'm going to close this PR. Thanks!

thiagoc:
@rbgarga can you reopen this pull request? Actually, by default, it doesn't accept the field in RFC822 format, it doesn't work.

rbgarga:
What is the error you get when you try to use it?

thiagoc:
> A mensagem não pôde ser enviada para meu@email.com.br -- Erro: 501 >: "@" or "." expected after "CLIENTE"

Tried with the following source address field:

`CLIENTE <servidor@email.com.br>`

PR #1727, opened 2015-06-21
Fix typo 

devnullity:
It looks like get_bandwidthtype_scale can only return default of "1", being that queue[bandwidthtype] will not return b, Kb, Gb, etc.


devnullity:
It looks like get_bandwidthtype_scale can only return default of "1", being that queue[bandwidthtype] will not return b, Kb, Gb, etc.

PR #1728, opened 2015-06-21
Fix var name typo in shaper.inc

devnullity:
Fix typo so get_bandwidthtype_scale will return more than exclusively the default "1", because of a non-existant var that will not return the expected 'b, Kb, Gb, %'.

### DIFF
--- a/etc/inc/shaper.inc
+++ b/etc/inc/shaper.inc
@@ -491,7 +491,7 @@ class altq_root_queue {
 					$myBw = $this->GetAvailableBandwidth() * $queue['bandwidth'] / 100;
 					break;
 				default:
-					$myBw = $queue['bandwidth'] * get_bandwidthtype_scale($queue['bandwdithtype']);
+					$myBw = $queue['bandwidth'] * get_bandwidthtype_scale($queue['bandwidthtype']);
 					break;
 			}
 		}


phil-davis:
Just for fun I looked back to see how long this typo has been there: https://github.com/pfsense/pfsense/commit/40de74f5c7207077a012ec95124329ceb9da8e1e - Jan 23, 2008.
Seems that people don't complain much about bugs:)
And it shows how annoying are languages where vars... are not declared. So easy to make a typo in a var name or array key and the silly interpreter just says "Oh well, that doesn't exist so I'll give back a blank". It would be nice to write code that always sets everything first. Then at least some "pre-compiler parser" thing could have a go at detecting some of this. But its a bit hard with array keys - the code might read in a whole array then refer to keys in that array. So the parser is going to have a tough time detecting any errors in that.
I kept an eye open for this sort of stuff when doing the code style changes. But there were so many mindless formatting things to change that I ended up with a lot of automated grep/replace with only very quick looking. I wonder how we can find the rest of the var/key name typos that are probably still there?

devnullity:
Maybe a script to search for rarely occuring permutations of a repeating word. 

Hmm... I am surprised that simple logic could find bugs in code I do not fully understand. I thought the "with enough eyes, all bugs are shallow" only applied to experienced programmers.

phil-davis:
From my observation without doing any statistical analysis, a number of these bugs that have been noticed in the last year or so originate from 2008-2009 era and often in commits that had a lot of lines/files changed in a single commit. I guess at that time there were not enough reviewer eyes and tired eyes after the first 1,000 lines of code to spot accidental typos like this or...
I guess with V3 the backend processing code and the data validation routines will be reworked and separated from the front-end GUI. At that time it should be "easy" to write test cases that perform all the exposed object manipulations and verify the results. That would catch errors like this straight away - an automated script of test cases should fail somewhere.

devnullity:
I have my own selfish goals of lessening the QoS learning curve for networking-newbies, streamlining the traffic-shaping GUI, and demystifying as much of HFSC as I can.

Beyond that, I owe many working hours to any Free and Open Software project.

ermal:
@devnullity for sure writing wizards is the best way to perform that.

PR #1733, opened 2015-06-26
XHTML_Compliance_Interfaces_Menu

NOYB:
Correct various XHTML compliance issues in the Interfaces menu


rbgarga:
I'm not sure if you are aware, but pfSense is moving to Bootstrap on 2.3, and I'm afraid you will waste time fixing this and your work will be lost when bootstrap conversion is merged back here.

You can follow the conversion, and also submit PRs, on https://github.com/SjonHortensius/pfsense

cbuechler:
This isn't relevant to 2.3 and should all be OK there. If not, please adapt as needed and re-submit off of 2.3. Thanks!

PR #1734, opened 2015-07-02
changes to allow a new user to login on the same computer

nmccurdy:
I'm not clear on what the ramifications of tearing down a connection and creating a new connection are, so I made this an option so that my fix doesn't break existing functionality.  I suspect that my usage scenario is different from that of the hotels and coffee shops that are trying to make sure that a user doesn't cheat by logging in as a different user from the same computer.

In my scenario, I don't care about usage caps, but I do need to have an accurate accounting of who is actually logged in.  In a school setting, if a shared computer is checked out to a student, that student should be able to log in using his/her own credentials even if someone else is already logged in.  In most situations the fact that the previous user's session is retained probably goes unnoticed, but in my case I am relying on the pfsense captive portal to tell me the user/ip address mapping so that I can filter traffic appropriately in dansguardian.  I imagine others would at least want the database to reflect who is actually logged in.  The fact that the user's attempt to login fails silently sounds like a bug to me.

This is my first pull request, so please let me know if I haven't done this correctly.  In order to have the patch work for my production box I had to branch off of RELENG_2_2.  I apologize if the standard is to branch from master.

Thanks.

Neil

### DIFF
--- a/etc/inc/captiveportal.inc
+++ b/etc/inc/captiveportal.inc
@@ -1913,7 +1913,13 @@ function portal_allow($clientip,$clientmac,$username,$password = null, $attribut
 		}
 		/* on the same ip */
 		if ($cpentry[2] == $clientip) {
-			if (isset($config['captiveportal'][$cpzone]['nomacfilter']) || $cpentry[3] == $clientmac)
+			if(isset($config['captiveportal'][$cpzone]['newloginwins']) && strcasecmp($cpentry[4], $username) != 0) {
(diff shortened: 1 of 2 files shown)


rbgarga:
@nmccurdy In order to have patches accepted you must sign a CLA as described at https://www.pfsense.org/about-pfsense/#legal

rbgarga:
CLA not signed after months, closing it for now

PR #1737, opened 2015-07-05
Make Cancel button work after input error

phil-davis:
If you add or edit a route, input some invalid data, click Save, then you get presented with the input error message/s and the input screen again, then if you click Cancel you are taken to an empty System Routes Edit screen.
The problem is that after first clicking Save, the form is submitted and (if there are input errors) the new form sees the old System Routes Edit as HTTP_REFERER.
This seems to be a generic issue across the current GUI - e.g. it happens in System GW Groups Edit also and... If the data has input errors then after display of the input errors the Cancel button becomes wrong, pointing to the existing page, not to where the user really came from.
With the new "bootstrapped" webGUI coming, maybe this will all be handled correctly anyway, so maybe nobody will care to fix the existing GUI.
Anyway, I got bugged by this just now so thought it worth pointing out the issue and 1 way to fix it. If you think it is worth fixing then let me know if you agree with the method here, or... and I am happy to find and fix the various webGUI edit screens that have this Cancel button issue.

### DIFF
--- a/usr/local/www/system_routes_edit.php
+++ b/usr/local/www/system_routes_edit.php
@@ -45,7 +45,11 @@ require_once("filter.inc");
 require_once("util.inc");
 require_once("gwlb.inc");
 
-$referer = (isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : '/system_routes.php');
+if (isset($_POST['referer'])) {
+	$referer = $_POST['referer'];
+} else {
+	$referer = (isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : '/system_routes.php');
+}
 
 if (!is_array($config['staticroutes']['route'])) {
 	$config['staticroutes']['route'] = array();
@@ -354,6 +358,7 @@ include("head.inc");
 					<?php if (isset($id) && $a_routes[$id]): ?>
 						<input name="id" type="hidden" value="<?=htmlspecialchars($id);?>" />
 					<?php endif; ?>
+					<input name="referer" type="hidden" value="<?=$referer;?>" />
 				</td>
 			</tr>
 		</table>


rbgarga:
thanks @phil-davis. If you submit patches, of course they can be applied, specially for RELENG_2_2. But your assumption is correct, it's not supposed to be a problem on bootstraped version. Can you confirm this @sbeaver-netgate ?

phil-davis:
And to anyone who has expertise:
Is the code here a reasonable (or even good!) way to do this? Or is there a better way?
Are there any issues with somebody/something somehow faking $_POST['referer'] and thus hijacking the Cancel button so it will "return" the user to an unwanted/nasty place out on the internet?

sbeaver-netgate:
Interesting this comes up. The Bootstrap team were just discussing the whole ‘Cancel button’ thing. Why do we have cancel buttons? Why do some pages NOT have cancel buttons? Don’t all browsers have a “Back” button? Doesn’t refreshing the page essentially cancel all input?

The current “Cancel” functionality is not consistent. Some pages simply reload the page, others point to $referer, which value is composed on page load, others link to a static parent page.

I need to solicit everyone’s opinion on this. Do we need “Cancel” at all or should we simplify the UI and remove it from all pages?

Steve

> On Jul 7, 2015, at 12:18 PM, Phil Davis <notifications@github.com> wrote:
> 
> And to anyone who has expertise:
> Is the code here a reasonable (or even good!) way to do this? Or is there a better way?
> Are there any issues with somebody/something somehow faking $_POST['referer'] and thus hijacking the Cancel button so it will "return" the user to an unwanted/nasty place out on the internet?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/pfsense/pfsense/pull/1737#issuecomment-119257993>.
> 

Steve
--
Stephen Beaver | +1 (512) 646-4100;239 | sbeaver@netgate.com <mailto:sbeaver@netgate.com> | www.netgate.com <http://www.netgate.com/>

phil-davis:
The input validation after pressing Save needs to be done in some way so that validation errors can be shown to the user without doing a redraw of the page that causes browsers to think it is a new page to add to the browser history.
If that is so, then the back button is consistently useful to "abort" the current data entry and go back to where the user last came from. Refreshing the page should reset any in-progress user-entered data back to the way it was when the page was first displayed (default/empty fields for entry of a new item, the current underlying data values for an edit) the equivalent of Cancel/Back followed by Add or Edit again.
Then the Cancel button is really just a Back button, and so I guess users can realize that in the modern browser-based software environment "<- Back" is the new "Cancel".
Conclusion: remove the Cancel buttons.
Other opinions welcome - there might be gaping holes in my analysis :)

phil-davis:
I did all the changes to fix up 2.2.n in #1749 
In bootstrap you can decide what to do with Cancel buttons in the new UI. For now it seemed useful to at least give 2.2.n UI consistency for the existing Cancel button behavior.
Closing this in favor of #1749

PR #1738, opened 2015-07-05
Fix #4813 validation of enable/disable of gateways and static routes

phil-davis:
1) A disabled gateway can always be enabled - no extra validation
needed.
2) When disabling an enabled gateway, check to see that the gateway is
not used in any gateway group or enabled static route (similar tests to
what is already checked before deleting a gateway).
3) A static route can always be disabled - no extra checks needed.
4) When enabling a static route, check that the selected gateway is
enabled - you cannot have a static route enabled on a disabled gateway.
5) Do the address family cross-check between static route and gateway
even when the static route is disabled - we do not want to save
mismatched   IP address families in any case.
This covers all the cases I can see to ensure that the enable/disable
status combinations of Gateways and Static Routes is always valid.

### DIFF
--- a/usr/local/www/system_gateways.php
+++ b/usr/local/www/system_gateways.php
@@ -78,7 +78,7 @@ if ($_POST) {
 	}
 }
 
-function can_delete_gateway_item($id) {
+function can_delete_disable_gateway_item($id, $disable = false) {
 	global $config, $input_errors, $a_gateways;
 
 	if (!isset($a_gateways[$id])) {
@@ -90,8 +90,11 @@ function can_delete_gateway_item($id) {
 			foreach ($group['item'] as $item) {
 				$items = explode("|", $item);
 				if ($items[0] == $a_gateways[$id]['name']) {
-					$input_errors[] = sprintf(gettext("Gateway '%s' cannot be deleted because it is in use on Gateway Group '%s'"), $a_gateways[$id]['name'], $group['name']);
-					break;
+					if ($disable) {
+						$input_errors[] = sprintf(gettext("Gateway '%s' cannot be disabled because it is in use on Gateway Group '%s'"), $a_gateways[$id]['name'], $group['name']);
+					} else {
+						$input_errors[] = sprintf(gettext("Gateway '%s' cannot be deleted because it is in use on Gateway Group '%s'"), $a_gateways[$id]['name'], $group['name']);
+					}
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


usr/local/www/system_gateways.php
```diff
@@ -100,8 +103,16 @@ function can_delete_gateway_item($id) {
...
+				if ($disable) {
+					// The user wants to disable this gateway.
+					if (!isset($route['disabled'])) {
+						// But there is a static route that uses this gateway and is enabled (not disabled).
+						$input_errors[] = sprintf(gettext("Gateway '%s' cannot be disabled because it is in use on Static Route '%s'"), $a_gateways[$id]['name'], $route['network']);
+					}
+				} else {
+					// The user wants to delete this gateway, but there is a static route (enabled or disabled) that refers to the gateway.
+					$input_errors[] = sprintf(gettext("Gateway '%s' cannot be deleted because it is in use on Static Route '%s'"), $a_gateways[$id]['name'], $route['network']);
+				}
```
rbgarga: IMO this block would be more readable if the check is reversed, like:
```
if (!disable) {
        // the user wants to delete
} else if (!isset($route['disabled'])) {
        // The user want to disable
}
```

usr/local/www/system_gateways_edit.php
```diff
@@ -126,6 +126,33 @@
 	}
 	if (!is_validaliasname($_POST['name'])) {
 		$input_errors[] = gettext("The gateway name must not contain invalid characters.");
+	} else {
+		if (isset($_POST['disabled'])) {
```
rbgarga: A simple elseif can be used in this case

usr/local/www/system_routes_edit.php
```diff
@@ -102,9 +102,15 @@
...
 			$input_errors[] = gettext("A valid gateway must be specified.");
-		}
-		if (!validate_address_family($_POST['network'], $_POST['gateway'])) {
-			$input_errors[] = gettext("The gateway '{$a_gateways[$_POST['gateway']]['gateway']}' is a different Address Family as network '{$_POST['network']}'.");
+		} else {
+			if (isset($a_gateways[$_POST['gateway']]['disabled']) && !$_POST['disabled']) {
+				$input_errors[] = gettext("The gateway is disabled but the route is not. You must disable the route in order to choose a disabled gateway.");
+			} else {
+				// Note that the 3rd parameter "disabled" must be passed as explicitly true or false.
+				if (!validate_address_family($_POST['network'], $_POST['gateway'], $_POST['disabled'] ? true : false)) {
```
rbgarga: A simple elseif can be used here as well

usr/local/www/system_gateways.php
phil-davis: There is also a test at line 93 using $disable - that reads OK to me. It would seem odd if the test here at line 106 was based on the reverse logic.
Should I reverse both tests? Or?

phil-davis: @rbgarga Any comment on what you would like me to do here?
I have the commit ready for the other comments you made, and can easily make a change here and/or at line 93, whatever you suggest, and then immediately make a single commit for all the changes.

rbgarga: IMO it's fine to reverse both just to keep it sane

phil-davis:
That should do it. I can get all 4 error messages in the changed tests to appear at the appropriate times still.

PR #1740, opened 2015-07-11
Services DNS Resolver - System Domain Local Zone Type Option

NOYB:
Make the Unbound local-zone type configurable.

### DIFF
--- a/etc/inc/system.inc
+++ b/etc/inc/system.inc
@@ -2270,18 +2270,4 @@ function system_get_dmesg_boot() {
 	return file_get_contents("{$g['varlog_path']}/dmesg.boot");
 }
 
-function get_possible_unbound_local_zone_types () {
-
-	$unbound_local_zone_types_array = array("deny" => "Deny", "refuse" => "Refuse", "static" => "Static", "transparent" => "Transparent", "typetransparent" => "Type Transparent", "redirect" => "Redirect", "inform" => "Inform", "inform_deny" => "Inform Deny", "nodefault" => "No Default");
-
-	$unbound_local_zone_types = array();
-	foreach ($unbound_local_zone_types_array as $type => $typename) {
-		$tmp["name"] = $typename;
-		$tmp["value"] = $type;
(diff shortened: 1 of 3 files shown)


etc/inc/system.inc
```diff
@@ -2270,4 +2270,18 @@ function system_get_dmesg_boot() {
 	return file_get_contents("{$g['varlog_path']}/dmesg.boot");
 }
 
+function get_posible_unbound_local_zone_types () {
+
+	$unbound_local_zone_types_array = array("deny" => "Deny", "refuse" => "Refuse", "static" => "Static", "transparent" => "Transparent", "typetransparent" => "Type Transparent", "redirect" => "Redirect", "inform" => "Inform", "nodefault" => "Node Fault");
```
phil-davis: nodefault needs description "No Default"
https://www.unbound.net/documentation/unbound.conf.html

usr/local/www/services_unbound.php
```diff
@@ -325,6 +338,33 @@ function show_advanced_dns() {
 								</td>
 							</tr>
 							<tr>
+								<td width="22%" valign="top" class="vncellreq"><?=gettext("System Domain Local-Zone Type"); ?></td>
+								<td width="78%" class="vtable">
+									<?php
+										$unbound_local_zone_types = get_posible_unbound_local_zone_types(false);
+										$size=count($unbound_local_zone_types);
```
phil-davis: We usually put space around equals signs on ordinary lines like this. It looks inconsistent with the line above :)
$size = count(...

etc/inc/system.inc
```diff
@@ -2270,4 +2270,18 @@ function system_get_dmesg_boot() {
 	return file_get_contents("{$g['varlog_path']}/dmesg.boot");
 }
 
+function get_posible_unbound_local_zone_types () {
```
phil-davis: s/posible/possible
And same change where the function is used.
Otherwise the mis-spelling will trip up English speakers for years to come, like HTTP "referer" became entrenched in a standard :)

PR #1743, opened 2015-07-12
Firewall Aliases Edit ensure input_addresses array exists

phil-davis:
If you click "+" to add an alias, then press Save without entering anything, you get:
Warning: Invalid argument supplied for foreach() in /usr/local/www/firewall_aliases_edit.php on line 402
as well as the various messages related to $input_errors.
This change ensures that $input_addresses array always exists (even if it has no real entries) so that the foreach() warning does not happen.

### DIFF
--- a/usr/local/www/firewall_aliases_edit.php
+++ b/usr/local/www/firewall_aliases_edit.php
@@ -273,6 +273,7 @@ if ($_POST) {
 		$wrongaliases = "";
 		$desc_fmt_err_found = false;
 		$alias_address_count = 0;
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: If you enter just a valid Alias Name (and optional description) and press save, without entering any rows of actual alias data, then you get an empty alias.
That seems to work OK, and results in an empty table in pf. I can use it in rules and nothing breaks. Of course it does not match anything, but then maybe someone might use this "feature" so they can then add to the empty pf table on-the-fly by other means.

PR #1749, opened 2015-07-13
Cancel button after input error

phil-davis:
If there is an input error then the edit page is redrawn showing the
input errors. The HTTP_REFERER becomes the current page, rather than the
true original referer. Then if you click Cancel the current page is just
redrawn.
This change makes the code remember the original referer, so if the user
enters some invalid data, presses Save, reads the input error messages
then presses Cancel they are taken back out to the original page - the
same as if Cancel was used before Save.

### DIFF
--- a/usr/local/www/firewall_aliases_edit.php
+++ b/usr/local/www/firewall_aliases_edit.php
@@ -52,7 +52,11 @@ require_once("shaper.inc");
 
 $pgtitle = array(gettext("Firewall"), gettext("Aliases"), gettext("Edit"));
 
-$referer = (isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : '/firewall_aliases.php');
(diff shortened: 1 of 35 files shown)


phil-davis:
I have tested all of these on top of a 2.2.4-DEVELOPMENT system. IMHO it would be useful to make this all nice in 2.2.4 since it is not difficult. If Git does not want to easily back-port the changes then let me know and I can make a specific pull request for RELENG_2_2 branch.
Along the way I was pressing Save on all the edit screens with empty data to get input_errors and came across a few error-checking bugs which I fixed and those are all separate little pull requests made over the last day. Those would be good to go into 2.2.4 also. Might get 2.2.n perfect just in time to change the whole underlying UI code base :)

rbgarga:
Please submit a separate Pull Request to RELENG_2_2 @phil-davis. Thank you!

phil-davis:
Yes, will do in a little while...

PR #1750, opened 2015-07-13
Add L2TP server's interface to mpd.conf (#4830)

TarasSavchuk:
https://redmine.pfsense.org/issues/4830
https://forum.pfsense.org/index.php?topic=95908.0

### DIFF
--- a/etc/inc/vpn.inc
+++ b/etc/inc/vpn.inc
@@ -1850,6 +1850,11 @@ function vpn_l2tp_configure() {
 	switch ($l2tpcfg['mode']) {
 
 		case 'server':
+                        $l2tp_listen="";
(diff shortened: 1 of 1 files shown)


doktornotor:
Resubmitted as noted above, with a backport for RELENG_2_2. This can be closed.

doktornotor:
@TarasSavchuk Since my resubmit was not acceptable due to authorship, you can either take that code from my PR referenced above and resubmit this yourself (it will probably only keep working for a couple of days at best before there's another bunch of conflicts), or live with the bug in pfSense code.

There is also a backport of the code to RELENG_2_2 in PR #1911, feel free to resubmit that as well.

Have a nice day.

PR #1752, opened 2015-07-14
Fix glob for serial device names

phil-davis:
Removing the "." that was in {,.[0-9]} allows it to match /dev/cuau10 and onward.
I added lots of comments on the glob expression, because the format of the glob expression is not the same as an ordinary regex.


phil-davis:
Forum: https://forum.pfsense.org/index.php?topic=96466.0
This fixes a general glob of this format for me when creating some junk files in /tmp - to be seen if it fixes it for the user on a real system with 10 or more cua* devices.

PR #1753, opened 2015-07-14
Fix GratisDNS support

mortencombat:
The current implementation is not working for me, maybe the interface was changed by GratisDNS? I tested the update URL manually with the IP added ("&i=<ip>") which fixed the issue for me. Code adjusted to include IP.


cbuechler:
They did change that, you were the first to notice apparently. 
https://larsendata.wiki/gratisdns:ddns
Merged, thanks! Also manually merged to RELENG_2_2 so it'll be in 2.2.4.

PR #1758, opened 2015-07-16
Make serial ports glob cope with many more possibilities

phil-davis:
It originally coped with things like cuau1 cuau1.1
Then I made it cope with things like cuau1 cuau11 but it stopped working for cuau1.1
This one copes with:
cuau1
cuau1.1
cuau1.11
cuau11
cuau11.1
cuau11.11
That should allow for all sorts of reasonable device name files without matching other stuff in /dev (like cuau1.init cuau1.lock) that we need to ignore.
Please think if I have covered the bases here.

### DIFF
--- a/usr/local/www/interfaces_ppps_edit.php
+++ b/usr/local/www/interfaces_ppps_edit.php
@@ -493,9 +493,9 @@ $types = array("select" => gettext("Select"), "ppp" => "PPP", "pppoe" => "PPPoE"
 				// Match files in /dev starting with "cua" then:
 				// ? = any single character e.g. like "cuau"
 				// [0-9] = a digit from 0 to 9
-				// {,[0-9]} = nothing (the empty before the comma) or a digit from 0 to 9
-				// This supports up to 100 devices (0 to 99), e.g. cuau0 cuau1 ... cuau10 cuau11 ... cuau99
-				$serialports = glob("/dev/cua?[0-9]{,[0-9]}", GLOB_BRACE);
+				// stuff in {} = the various possible digit and dot combinations to allow an optional 2nd digit, dot, followed by 1 or 2 optional digits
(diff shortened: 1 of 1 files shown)


rbgarga:
I thought about a simpler one that should cover all necessary cases:
```
glob("/dev/cua[a-zA-Z]{,[0-9.]*}[0-9]", GLOB_BRACE)

/dev/cua - fixed part
[a-zA-Z] - non-number char (u, U, d)
{,[0-9.]*} - nothing or numbers and dots
[0-9] - Make sure it ends with a number, this will make it ignore .lock and .init
```

Can you give it a try? It passed all my tests

phil-davis:
Your suggestion would work for the /dev files that are actually there in normal systems. But it also will match stuff like:
/dev/cuau.abc1
/dev/cuau1a1
which seems odd and potentially open to problems one day.
But the [a-zA-Z] thing seems better than just the "?". I expect it could be just [a-z] because I can't imagine if the first 3 chars are "cua" in lowercase that the next letter would be uppercase.
It could be:
glob("/dev/cua[a-z][0-9]{,.[0-9],.[0-9][0-9],[0-9],[0-9].[0-9],[0-9].[0-9][0-9]}", GLOB_BRACE);

rbgarga:
I need to check FreeBSD src code, but IIRC there are cases with capital U, /dev/cuaU1. So please change it to use [a-zA-Z]

phil-davis:
I replaced the "?" with [a-zA-Z] and adjusted the comments.

jim-p:
USB serial devices are cuaU<number>, traditional serial ports are cuau<number>

phil-davis:
Good to know - we really do need lower and upper case here.
@jim-p Does this long-winded glob expression look like it is going to cover all the reasonable possibilities of serial device numbering?

jim-p:
I'd have to test it to see... too complicated to go by sight. At a glance it seems reasonable but seeing as I've been on vacation this week up until today, my brain isn't exactly in pattern matching mode. :-)

PR #1763, opened 2015-07-18
Add labels to some default firewall rules

doktornotor:
... so that people can get useful descriptions in the System Logs - Firewall GUI, instead of useless tracker numbers. This is for master branch.


rbgarga:
Can you please sign the CLA as described on https://www.pfsense.org/about-pfsense/#legal ?

doktornotor:
@rbgarga - Done.

PR #1765, opened 2015-07-19
Provide separate checkboxes for tmp and var RAM disks

phil-davis:
As suggested in forum: https://forum.pfsense.org/index.php?topic=96705.0
It is not hard to do - so here it is if someone thinks it is worth having. e.g. on my SG-2440 full install that has plenty of free RAM I could enable just /tmp memory disk and leave /var where it is on the eMMC.

### DIFF
--- a/etc/inc/rrd.inc
+++ b/etc/inc/rrd.inc
@@ -55,7 +55,7 @@ function restore_rrd() {
 
 	$rrdrestore = "";
 	$rrdreturn = "";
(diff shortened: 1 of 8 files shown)


phil-davis:
Leave this sitting here for now. @jimp made some suggestions in that forum thread about maybe having a dropdown that has the 4 combinations all in a single dropdown list to choose from. It needs some design thought before implementing. A change like this would not go in 2.2.* anyway, so there is plenty of time to decide what to do before 2.3.

phil-davis:
Closing in favour of PR #1820 which is integrated with later master.

PR #1766, opened 2015-07-19
Add ability to hide drives from SMART Status widget

doktornotor:
Useful for machines with USB card readers which display as daX, or system with media that do not support SMART in general.


doktornotor:
Resubmitted as #1841 for RELENG_2_2. Anything else is way to hectic these days.

PR #1767, opened 2015-07-19
Bug #4569 - system_set_harddisk_standby() needs rewritten for current…

doktornotor:
… CAM-based ATA stack

### DIFF
--- a/etc/inc/system.inc
+++ b/etc/inc/system.inc
@@ -2066,12 +2066,10 @@ function system_set_harddisk_standby() {
 		$mt = microtime();
 		echo "system_set_harddisk_standby() being called $mt\n";
 	}
-
 	if (isset($config['system']['harddiskstandby'])) {
 		if (platform_booting()) {
 			echo gettext('Setting hard disk standby... ');
 		}
-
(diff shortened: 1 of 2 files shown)


etc/inc/system.inc
```diff
@@ -2070,13 +2075,15 @@ function system_set_harddisk_standby() {
...
-				if (platform_booting()) {
-					echo gettext("done.") . "\n";
+			// Since get_smart_drive_list() only matches ad|da|ada; lets put the check below 
+			// just in case of some weird pfSense platform installs.
+			if (count($harddisks) > 0)  {
+				// Iterate disks and run the camcontrol command for each
+				foreach($harddisks as $harddisk)  {
+					mwexec("/sbin/camcontrol standby {$harddisk} -t {$standby}");
+					if (platform_booting()) {
+						echo gettext("done.") . "\n";
```
rbgarga: It's going to print "done" multiple times if $harddisks > 1

doktornotor:
Yeah, kinda expected. You prefer to print it just once, or rather prefix it with the device name for each line?

doktornotor:
Resubmitted as #1830

PR #1768, opened 2015-07-20
wol widget, dont show static-arp addresses as online

PiBa-NL:


rbgarga:
Committed a similar patch on d237b4447adc41b0c05bc7d0d0a74211b70d7d2c

PR #1770, opened 2015-07-21
Unset old CA and Cert in system config

phil-davis:
This looked odd. Why would we leave behind the old "ca" and "cert" section in $config["system"]?
I guess it would do no harm, but seems confusing for the future to have some unused entries like this remaining in the config.
Should a piece of code be put into the latest upgrade function to clean out these in any current config?


jim-p:
You're right, those should be unset. I could swear the old ones did get unset at some point but I don't see it now.

cbuechler:
merged, thanks!

PR #1773, opened 2015-07-21
Protect unset() with isset() in upgrade_config

phil-davis:
Forum: https://forum.pfsense.org/index.php?topic=96827.0
Sometimes there can be unexpected stuff in a config. To avoid errors like "PHP Fatal error:  Cannot unset string offsets in /etc/inc/upgrade_config.inc on line 291" it seems good to use belts-and-braces checks with any unset() of associative array elements, just in case the thing above is an empty string (instead of an array or non-existent thing). An unnecessary error in processing the upgrade code can mean missing other important upgrade stuff.
These are the places I could see in upgrade_config.inc where unset() is not already protected by isset(). A lot of them are "that could not ever go wrong", but IMHO it is worth protecting them all just to be sure.

### DIFF
--- a/etc/inc/upgrade_config.inc
+++ b/etc/inc/upgrade_config.inc
@@ -258,8 +258,8 @@ function upgrade_014_to_015() {
 		if ($config['interfaces']['wan']['gateway'] <> "") {
 			$config['system']['gateway'] = $config['interfaces']['wan']['gateway'];
 		}
+		unset($config['interfaces']['wan']['gateway']);
 	}
-	unset($config['interfaces']['wan']['gateway']);
 
 	/* Queues are no longer interface specific */
 	if (isset($config['interfaces']['lan']['schedulertype'])) {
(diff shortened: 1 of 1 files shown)


etc/inc/upgrade_config.inc
```diff
@@ -2976,10 +3001,10 @@ function upgrade_083_to_084() {
 			$config['hasync'] = $config['installedpackages']['carpsettings']['config'][0];
 			unset($config['installedpackages']['carpsettings']);
 		}
-		if (empty($config['installedpackages']['carpsettings'])) {
+		if (empty($config['installedpackages']['carpsettings']) && isset($config['installedpackages']['carpsettings'])) {
 			unset($config['installedpackages']['carpsettings']);
 		}
-		if (empty($config['installedpackages'])) {
+		if (empty($config['installedpackages']) && isset($config['installedpackages'])) {
```
rbgarga: In this case, if $config['installedpackages'] is not set, the call for empty() will throw an error, no?

phil-davis: var_dump(empty($config['installedpackages']));
bool(false)
var_dump(empty($config['crud']));
bool(true)

empty() on something that is not set returns true, just the same as if it is set to the blank string, or an empty array, or the number 0.
So just empty()=true does not mean it is set.

phil-davis:
Closing in favour of #1821 which is integrated with the current master.

PR #1776, opened 2015-07-22
Backup IPsec: MSS clamping. Wrong group location used.

laped83:
MSS value of IPsec isn't backed up when selecting IPsec, because its in the wrong config group.

Now uses "IPsec" instead of "System" group.

Signed-off by: Lars Alex Pedersen <thacaleb@gmail.com>

### DIFF
--- a/etc/inc/filter.inc
+++ b/etc/inc/filter.inc
@@ -554,10 +554,10 @@ function filter_generate_scrubing() {
 	global $config, $FilterIflist;
 	$scrubrules = "";
 
-	if (isset($config['system']['maxmss_enable'])) {
+	if (isset($config['ipsec']['maxmss_enable'])) {
 		$maxmss = 1400;
-		if (!empty($config['system']['maxmss'])) {
-			$maxmss = $config['system']['maxmss'];
+		if (!empty($config['ipsec']['maxmss'])) {
+			$maxmss = $config['ipsec']['maxmss'];
 		}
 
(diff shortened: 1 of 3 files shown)


laped83:
To solve the following issue with backup of IPsec configuration values.

https://redmine.pfsense.org/issues/4864

cbuechler:
This will break MSS clamping on every existing install. If we're going to move it, it needs config upgrade code as well.

laped83:
Added config upgrade code.

doktornotor:
Pardon me, but... This is used for ```<vpn_networks>``` table:

```
        if (isset($config['system']['maxmss_enable'])) {
                $maxmss = 1400;
                if (!empty($config['system']['maxmss']))
                        $maxmss = $config['system']['maxmss'];

                $scrubrules .= "scrub from any to <vpn_networks> max-mss {$maxmss}\n";
                $scrubrules .= "scrub from <vpn_networks> to any max-mss {$maxmss}\n";
        }
```

which boils down to filter_get_vpns_list() in filter.inc. There are things such as OpenVPN there, plus also PPPoE. How on earth is this ipsec-specific setting? This is actually breaking things, this is system-wide setting and should stay where it is. If there's supposed to be something IPSec-specific as well, then you'd better start fixing different parts of code.

rbgarga:
Max MSS parameter is used by IPsec, OpenVPN and PPPoE server, that is the reason it's a global setting in System -> Advanced.

If you want to have specific values for all kinds of VPNs a different patch should be done considering all the cases

PR #1780, opened 2015-07-23
Redmine #4568 Preserve MLPPP settings when saving interface settings

phil-davis:
The interface.php UI only displays and allows the user to edit the first (0) element of the "localip" "subnet" and "gateway". I changed the name of some POST fields so that it is clearer that they just contain the "0" element of the list.
If there are other elements (i.e. if there is MLPPP happening) then these are preserved across save, even though they are not displayed to the user for editing. The "Advanced" message gets the extra text "There are additional Local and Remote IP addresses defined for MLPPP." displayed so the user can know that there is more data and where to click to see and edit it.
IMHO this is a reasonable way to fix this bug. There does not seem much to be gained by putting a whole lot of code here to edit a variable number of Local-Remote IP Address sets when that is already easily available in interfaces_ppps_edit.


phil-davis:
Closing in favour of #1822 which is integrated with the current master.

PR #1784, opened 2015-07-25
filter.inc - Remove absolutely useless logspam

doktornotor:
See Bug #4102 and others. Getting completely pointless "errors" noise for completely normal situations in logs is NOT cool. Other annoyed users:
https://forum.pfsense.org/index.php?topic=91455.0
https://forum.pfsense.org/index.php?topic=96918.0


cbuechler:
it is excessive. There was a reason for it initially (IIRC it should only be triggered where a gateway is expected but one doesn't exist), but it's just spam at this point. I'll review closer to make sure post-2.2.4 and merge this assuming I don't find some reason not to.

doktornotor:
Alternatively, I can see some undocumented developerspew var used for similar debugging noise...

doktornotor:
Resubmitted as #1831

PR #1785, opened 2015-07-26
Inform on dashboard when on a later version than the official release

phil-davis:
For people that are running a development or RC version that is later than the official release, this will display that on the dashboard version check.


phil-davis:
Closing in favour of #1825 which is integrated with the latest master.

PR #1786, opened 2015-07-27
Tighten is_linklocal() [rebase of PR #973]

stilez:
As requested by @rbgarga, rebase/resubmit of PR #973 which is superseded.

ORIGINAL DESCRIPTION AT #973:

is_linklocal has a few issues, including validating as linklocal, addresses that aren't linklocal according to RFC 4291, and doing no validation of reasonableness on any %(scope/interface) present, allowing this to contain arbitrary text.

1) IPv4/6 agnostic: while IPv4 linklocal testing isn't much needed, it should probably be recognised because some code handling linklocal may reasonably expect is_linklocal() to be IPv4/IPv6 agnostic.

2) For IPv6, it tests at least, that the purported scope/interface is [0-9a-z]+ otherwise user input or other text such as "fe80::%\n;ARBIRARYTEXT;" would be validated as a linklocal address and inserted into pf and perhaps other places without further detection, leading to possible vulnerabilities. But it doesn't test more than this (and probably should test for valid scope/interface if present).

To which I add: 
- ip2long32() now a bit more robust:  this now returns FALSE for non-valid IPv4
- scope/interface part still not tested other than [0-9a-z], but a non-pathological max length of 64 has been added "just in case"
- Net_IPv6::_Ip2Bin() can return shorter binary strings for IPv4 or "junk" input. So this resubmit tests at lines 642-643 that it returned a 128 bit length, which should be sufficient for input sanitising that it was meaningful for IPv6.

### DIFF
--- a/etc/inc/util.inc
+++ b/etc/inc/util.inc
@@ -406,9 +406,11 @@ function long2ip32($ip) {
 	return long2ip($ip & 0xFFFFFFFF);
 }
 
-/* Convert IP address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms. */
+/* Convert IP address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms.
+   Returns '' (not numeric) if not valid IPv4. */
 function ip2long32($ip) {
-	return (ip2long($ip) & 0xFFFFFFFF);
+	$a = ip2long($ip);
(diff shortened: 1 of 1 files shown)


stilez:
Superseded by PR #2320 - can be closed

PR #1787, opened 2015-07-27
Pkg related cleanups and fixes

k-paulius:
Removing some unused variables and initializing variables that are not being initialized. Also fixing issue with default value assignment. The code was only checking if $a_pkg[$id] existed instead of $a_pkg[$id][$fieldname] before assigning value when loading the form. This caused default value to never be displayed for any new fields added to the package xml.

### DIFF
--- a/usr/local/www/pkg.php
+++ b/usr/local/www/pkg.php
@@ -73,11 +73,6 @@ if ($pkg['include_file'] != "") {
 	require_once($pkg['include_file']);
 }
 
-$package_name = $pkg['menu'][0]['name'];
-$section      = $pkg['menu'][0]['section'];
-$config_path  = $pkg['configpath'];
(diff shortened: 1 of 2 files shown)


k-paulius:
Reverted the offending commit. Totally missed the fact that those lines were wrapped conditional statements.

doktornotor:
@k-paulius I resubmitted this as noted above due to conflicts. Can be closed.

doktornotor:
@k-paulius Since my resubmit in #1906 was not acceptable due to authorship, you can either take that code from my PR referenced above and resubmit this yourself (it will probably only keep working for a couple of days at best before there's another bunch of conflicts), or live with the bug in pfSense code.

There is also a backport of the code to RELENG_2_2 in PR #1907, feel free to resubmit that as well.

Have a nice day.

PR #1788, opened 2015-07-27
Simplify is_literalipaddrv6 & is_hostnamewithport (resubmit/rebase of PR #962)

stilez:
rbgarga asked for PR #962 to be rebased/resubmitted.

On doing so I find that 
- is_ipaddrwithport() has been sorted out effectively already, so that part is redundant
- is_hostnamewithport() can be simplified (we don't need to pop() or assign a new variable just to test 2nd member of the array)
- is_literalipaddrv6() can be much simplified - we don't need an expensive preg_match() to test if it's a valid IPv6 wrapped in "[" ... "]"

If this is ok then the original PR can be closed as it's superseded.

### DIFF
--- a/etc/inc/util.inc
+++ b/etc/inc/util.inc
@@ -637,16 +637,14 @@ function get_ll_scope($addr) {
 
 /* returns true if $ipaddr is a valid literal IPv6 address */
 function is_literalipaddrv6($ipaddr) {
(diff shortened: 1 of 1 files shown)


stilez:
Superseded by PR #2320 - can be closed

PR #1789, opened 2015-07-27
data sanitising: ip2long32, ip2ulong, long2ip32

stilez:
Self explanatory.  If these functions find themselves trying to convert non-int data (or an x64 int with non-zeros in any bits >32) to dotted IPv4, or non-dotted IPv4 to integer IPv4 values, something's wrong and they shouldn't return a value that looks like they succeeded.


stilez:
Resubmitted as #2152

PR #1790, opened 2015-07-27
Pkg install error handling and connect timeout RELENG_2_2

phil-davis:
Fixes Redmine #4884
1) Line 778-780 - If the fetch of any of the package additional files fails then bail out. This prevents half-installed packages that look like they had a successful install.
2) Line 1458 - use the return boolean value from download_file_with_progress_bar() to determine success or failure here, like is done in the other places in this file. I had a case of installing a package with an error (timeout) and the download (I presume it was the download code) had left an empty file /usr/local/pkg/autoconfigbackup.xml - it passed the file_exists() check and the rest of the code went on to happily install the "nothing" in the package and then claim the package was successfully installed :(
After the above 2 changes I could get reliable indication of success/failure of the package install and the code would abort nicely if a download went wrong.
3) Package installs happen either:
i) On the end of a boot after upgrade or config restore, or;
ii) Online while the main system is running (happily)
Therefore there is no need to rush to abort if the download of a package file is taking some time to get started. It seems better to me to wait a decent amount of time rather than abort.
Thus I have increased the connect timeout for this from the default (5) to 30 seconds.
This makes my crap sites load packages much better :)
I will make a version for master also.

### DIFF
--- a/etc/inc/pkg-utils.inc
+++ b/etc/inc/pkg-utils.inc
@@ -87,6 +87,11 @@ if(!is_dir("/usr/local/pkg") or !is_dir("/usr/local/pkg/pf")) {
 	conf_mount_ro();
 }
 
+// Use a longer (30 second) connect timeout when trying to download each package file.
+// Package installs are not holding up general boot and routing/firewall functionality,
+// so it is nicer to wait a while if necessary rather than abort.
+define('PKG_FILE_DOWNLOAD_CONNECT_TIMEOUT', '30');
+
 /****f* pkg-utils/remove_package
  * NAME
  *   remove_package - Removes package from FreeBSD if it exists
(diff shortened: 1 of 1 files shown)


phil-davis:
Actually the whole package thing is changed in master for 2.3. There are no calls to download_file_with_progress_bar(). I guess the PHP and other extra things for a package are going to be bundled up in a single installer or...?
So I will leave it to whoever is doing the new packages stuff for 2.3 to work out if any of the changes here are needed at all in master/2.3

phil-davis:
This is a PR against RELENG_2_2 so it is still mergeable after the recent changes to folder structure in master - good.

doktornotor:
Due to the recent slew of complaints about half-assed broken pkg installs on the forums - can someone merge this? Could live without the timeout cosmetics, but going on after a failed fetch as if nothing happened is just a complete brainfart.

PR #1791, opened 2015-07-27
Update index.php

yakar:
Widget names can translation with gettext..


rbgarga:
gettext doesn't deal with variable names, you need to call gettext in the place variable is being set

yakar:
im tested, works fine for me :)

rbgarga:
But it's wrong, gettext calls should have only text as parameters to keep the sanity. If a variable will contain a text that will need to be translated, it must call gettext when the text is being attributed to the variable, not when it's being used.

PR #1792, opened 2015-07-27
Consider url_port alias type when checking port-type aliases

phil-davis:
Hopefully will fix forum thread https://forum.pfsense.org/index.php?topic=97101.0
This is the version for RELENG_2_2


phil-davis:
Redmine bug logged https://redmine.pfsense.org/issues/4888

phil-davis:
See https://github.com/pfsense/pfsense/pull/1794
Hopefully that one is better.

PR #1793, opened 2015-07-27

mortencombat:
Previous GratisDNS implementation did not work with hosts without subdomains (www.bob.dk works, bob.dk does not).

### DIFF
--- a/etc/inc/dyndns.class
+++ b/etc/inc/dyndns.class
@@ -695,8 +695,14 @@
 						log_error("GratisDNS.dk ({$this->_dnsHost}): DNS update() starting.");
 					}
 					$server = "https://ssl.gratisdns.dk/ddns.phtml";
-					list($hostname, $domain) = explode(".", $this->_dnsHost, 2);
-					curl_setopt($ch, CURLOPT_URL, $server . '?u=' . $this->_dnsUser . '&p=' . $this->_dnsPass . '&h=' . $this->_dnsHost . '&d=' . $domain . '&i=' . $this->_dnsIP);
+					$host = trim($this->_dnsHost);
+					$hostnames = explode(".", $host);
+					$hostnames_count = count($hostnames);
+					if ($hostnames_count > 1)
+						$domain = $hostnames[$hostnames_count-2] . "." . $hostnames[$hostnames_count-1];
+					else
+						$domain = $host;
+					curl_setopt($ch, CURLOPT_URL, $server . '?u=' . $this->_dnsUser . '&p=' . $this->_dnsPass . '&h=' . $host . '&d=' . $domain . '&i=' . $this->_dnsIP);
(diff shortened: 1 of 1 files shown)


rbgarga:
Please sign CLA as described at https://www.pfsense.org/about-pfsense/#legal

mortencombat:
Thanks, done.

etc/inc/dyndns.class
```diff
@@ -695,8 +695,14 @@
 						log_error("GratisDNS.dk ({$this->_dnsHost}): DNS update() starting.");
 					}
 					$server = "https://ssl.gratisdns.dk/ddns.phtml";
-					list($hostname, $domain) = explode(".", $this->_dnsHost, 2);
-					curl_setopt($ch, CURLOPT_URL, $server . '?u=' . $this->_dnsUser . '&p=' . $this->_dnsPass . '&h=' . $this->_dnsHost . '&d=' . $domain . '&i=' . $this->_dnsIP);
+					$host = trim($this->_dnsHost);
```
rbgarga: Why set a variable just to trim possible extra whitespaces? You can call trim direct on explode() call.

etc/inc/dyndns.class
```diff
@@ -695,8 +695,14 @@
...
+					$hostnames = explode(".", $host);
+					$hostnames_count = count($hostnames);
```
rbgarga: $hostnames_count is going to be > 1 for "bob.dk" and for "www.bob.dk", this check looks wrong to me.

What does gratisdns API expect to receive on "d" and "h" parameters for both cases? I imagine for "www.bob.dk" it's supposed to be "h=www" and d="bob.dk", right? What about when it's just "bob.dk"? Is h supposed to be blank?

etc/inc/dyndns.class
mortencombat: $host is used twice, in the explode() call and on line 705.

etc/inc/dyndns.class
mortencombat: You're right, it should be > 2.

GratisDNS expects:
www.bob.dk : h=www.bob.dk d=bob.dk
bob.dk : h=bob.dk d=bob.dk

mortencombat:
@rbgarga - is this one okay to merge or do you require further changes?

doktornotor:
As noted above, resubmitted as #1917 due to conflicts; this can be closed. Please, comment there if required.

doktornotor:
@mortencombat - Since my resubmit was not acceptable due to authorship, you can either take that code from my PR referenced above and resubmit this yourself (it will probably only keep working for a couple of days at best before there's another bunch of conflicts), or live with the bug in pfSense code.

Have a nice day.

mortencombat:
Resubmitted as #1927.

PR #1794, opened 2015-07-27
Consider url_port alias type when checking port-type aliases V2

phil-davis:
This time I have typed url_ports correctly.
Hopefully will fix forum thread https://forum.pfsense.org/index.php?topic=97101.0
This is the version for RELENG_2_2
Redmine bug logged https://redmine.pfsense.org/issues/4888


phil-davis:
I have tried making a urltable_ports alias to test if that (still) works. But even using the code before any of this stuff was changed, I get errors loading the rules for that. I get an empty txt file in /var/db/aliastables for the urltable_ports alias and then that ends up as an empty alias in /tmp/rules.debug and then any rule that uses the alias gets an error.
Is urltable_ports supposed to work? Or is it known to be not working?

PR #1795, opened 2015-07-27
Var name typo processing URL table alias

phil-davis:
I just noticed this while looking into other stuff. It must be of some consequence. Maybe it will help with loading the contents of URL table aliases at first startup?

### DIFF
--- a/etc/inc/filter.inc
+++ b/etc/inc/filter.inc
@@ -11,7 +11,7 @@
 	originally part of m0n0wall (http://m0n0.ch/wall)
 	Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>.
 	All rights reserved.
-
+ne
 	Redistribution and use in source and binary forms, with or without
 	modification, are permitted provided that the following conditions are met:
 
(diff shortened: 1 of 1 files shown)


etc/inc/filter.inc
```diff
@@ -11,7 +11,7 @@
 	originally part of m0n0wall (http://m0n0.ch/wall)
 	Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>.
 	All rights reserved.
-
+ne
```
rbgarga: Looks like some accidental leftover

rbgarga: But nvm, I'll push a fix to the typo quick

rbgarga:
Done. Good catch! Thanks Phil!

etc/inc/filter.inc
phil-davis: Yeh, I guess I was pressing random keys at the time, and just kept going through to submit pull request without looking.
I see you put the correct fix in yourself - good.

PR #1796, opened 2015-07-27
Adds 'ue' interface as ALTQ capable

timello:
Even though 'axe' is being listed as ALTQ capable, pfSense web
interface does not allow traffic shapper wizzard to create queues
in case 'ue' is the associated interface.


doktornotor:
Well, this has merge conflicts now, but first of all it is a wrong fix. This will mark all ueX interfaces as ALTQ capable, even those which have drivers attached that don't support any such thing.

rbgarga:
@timello In order to have patches accepted you must sign a CLA as described at https://www.pfsense.org/about-pfsense/#legal

PR #1798, opened 2015-07-28
Fix pkg removal with pkg logging enabled

k-paulius:


doktornotor:
@k-paulius I resubmitted this as noted above due to conflicts. I do not understand your second change being done there. AFAICT, the whole $config['installedpackages']['package'][$pkgid] array is unset on package uninstall. Unsetting that is not needed, nor does it do anything useful regarding syslog configuration.

PR #1800, opened 2015-07-28
Handle invalid negative temperature reads

nagyrobi:
Handle the case when sysctl errorneously returns negative temperature values.


rbgarga:
Applied a similar fix on 8822be67ce3085c30a1127960d6c8ae1d7eb1876

PR #1801, opened 2015-07-28
Exclude negative temperature reads

nagyrobi:
Handle the case when sysctl errorneously returns negative temperature values by excluding them from the widget.


rbgarga:
Please sign CLA as described at https://www.pfsense.org/about-pfsense/#legal

nagyrobi:
It says:
An Error has occurred
"License Agreement" can be ordered along with these products/subscripton(s) only: Individual CLA,Corporate CLA
:(

rbgarga:
Qre you trying it on this page?
https://portal.pfsense.org/members/signup/ICLA

nagyrobi:
Yes, but this time from your link it worked.

rbgarga:
Applied a similar fix on 8822be67ce3085c30a1127960d6c8ae1d7eb1876

PR #1803, opened 2015-07-29
 run early shell commands earlier

nagyrobi:
Early shell commands are not run early enough. Running at boot "ifconfig emX name ethX" for all interfaces in order to obtain a less driver-specific, hardware-independent, consistent network interface naming interferes with interface re-assignment and results in an unusable state.

Putting this before interface assignment check allows for keeping interface renaming commands within the config file, and not have to use manual rc.conf.local on the filesystem.

### DIFF
--- a/etc/rc.bootup
+++ b/etc/rc.bootup
@@ -160,6 +160,9 @@ echo "Loading configuration...";
 parse_config_bootup();
 echo "done.\n";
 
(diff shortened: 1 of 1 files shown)


nagyrobi:
Discussion here: https://forum.pfsense.org/index.php?topic=97204.msg541593#msg541593

PR #1805, opened 2015-07-31
fix interface assignment menus running off VGA screen

nagyrobi:
When using VGA console, interface assignment can be a real pain in the ass because of the standard 80 columns width.

Dmesg reports the many interface description names in very long strings that don't fit in a row, this breaks the nice appearance of the interface list in the assignment menu.
The aestethics is one thing, but the real pain is that the interface list goes off screen by the time the menu asks for the WAN interface name, if there are many interfaces present. It's a real problem to choose from a list which is not visible anymore.

One fix is to maximize the length of the interface description to 48 chars.

The second fix (and also improvement for better overlooking when the list really goes off) is to print a small list of the interface names at each question. This is necessary because when somebody wants to assign the first interface to the last Optional port, the big list will be way off screen by then, and the name of it won't be visible.

This also makes it nice clean and straightforward.

The problem:
assign_interfaces_list_goes_off_the_screen

The fixes:
assign_interfaces_list_maxwidth

assign_interfaces_smallist_every_question

assign_interfaces_smallist_vlan_question

### DIFF
--- a/etc/inc/config.console.inc
+++ b/etc/inc/config.console.inc
@@ -86,8 +86,9 @@ EOD;
 		$iflist = array();
 	} else {
 		foreach ($iflist as $iface => $ifa) {
+			$ifsmallist = trim($ifsmallist . " " . $iface);
 			echo sprintf("% -7s%s %s %s\n", $iface, $ifa['mac'],
-				$ifa['up'] ? "  (up)" : "(down)", $ifa['dmesg']);
+				$ifa['up'] ? "  (up)" : "(down)", substr($ifa['dmesg'], 0, 48));
 		}
 	}
 
@@ -190,6 +191,7 @@ EOD;
 					"VLAN tag {$vlan['tag']}, parent interface {$vlan['if']}");
 
 				$iflist[$vlan['if'] . '_vlan' . $vlan['tag']] = array();
+				$ifsmallist = trim($ifsmallist . " " . $vlan['if'] . '_vlan' . $vlan['tag']);
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


rbgarga:
That looks good, but It would be good if you remove already selected interfaces from $ifsmalllist. If user choose em3 for WAN, it should not be present as an option for LAN and other ifs

nagyrobi:
updated patch to fulfill your request, now the list presented to the user only contains interfaces which are not assigned so far.

etc/inc/config.console.inc
```diff
@@ -202,7 +204,8 @@ hitting 'a' to initiate auto detection.
 EOD;
 
 		do {
-			echo "\n" . gettext("Enter the WAN interface name or 'a' for auto-detection:") . " ";
+			echo "\n" . gettext("Enter the WAN interface name or 'a' for auto-detection") . " ";
+			echo "\n" . gettext("(") . $ifsmallist . gettext(" or a): ");
```
rbgarga: In cases like this, you are better off using a single gettext with printf(), like this:
```
printf(gettext("Your message with %s parameter"), $var);
```
This way who is translating the message can modify it according the language structure, and the message makes sense

doktornotor:
Instead of discussing non-existent localization (noone updates the gettext template, this won't ever get translated) -- could the installer get fixed (read - made usable)?! This yet again has conflicts now. 10 line fix. Sigh.

doktornotor:
Resubmitted with ```printf()``` as #1925. Have a nice day.

PR #1807, opened 2015-08-04
Redmine #4918 DH groups 22-24 do not function

miken32:
When selecting DH group 22-24, the entry is not created properly in ipsec.conf. This resolves the issue.


miken32:
This error was introduced last year in commit b0cbebeb351fbfe50cd164a87f466f91e7aea4f3

Should be a pretty straightforward fix...

PR #1809, opened 2015-08-06
Fix file encoding for #1806

jlduran:
Make the file encoding UTF-8 again. Partially reverting #1806.


jlduran:
Fixed in 46bc6e545a17e77202aaf01ec0cd8d5a46567525

PR #1811, opened 2015-08-09
Redmine #4924 Improve pkg version comparison RELENG_2_2

phil-davis:
This compares 2 package version strings by spliting the strings into pieces that are digits and pieces that are anything else. e.g.
"utility-24.9_5 pkg v4.5"
becomes
"utility-"
24
"."
9
"_"
5
" pkg v"
4
"."
5

A new version might be:
"utility-24.10_5 pkg v4.5"

Each component is compared.
If both corresponding components are "integer chunks" they are compared using "<" and ">" - that means that 10 becomes greater than 9.
If either component is not "integer" then string comparison is done. This makes the ordinary string changes work as expected and also stuff like "1.05" and "1.1" - a change like this will be seen as an upgrade (previously it might have been seen that ("05" = 5) and (5 > 1)).

Note that I have left the string comparison case-sensitive. If someone making their version string changes the "fixed text" they have put in the version string then anything is possible :) It is a bit difficult to guess all of that.

Also if people use alphabetic versioning that does not follow strcmp() sort order then there is always trouble - it works by luck with "-BETA" then "-RELEASE" because the English words happen to sort as expected.

I think this provides a reasonably generic way to compare version strings which should work for all the combinations used by reasonably imagined packages.

### DIFF
--- a/etc/inc/pkg-utils.inc
+++ b/etc/inc/pkg-utils.inc
@@ -237,9 +237,9 @@ function compare_pkg_versions($installed_pkg_ver, $available_pkg_ver) {
 			$val2_as_int = filter_var($available_arr[$n], FILTER_VALIDATE_INT);
 			if (($val1_as_int === false) || ($val2_as_int === false)) {
 				// One of them does not look like a genuine integer so use string comparison.
-				if (strcmp($val, $available_arr[$n]) > 0) {
+				if (strcasecmp($val, $available_arr[$n]) > 0) {
 					return 1;
-				} elseif (strcmp($val, $available_arr[$n]) < 0) {
+				} elseif (strcasecmp($val, $available_arr[$n]) < 0) {
 					return -1;
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


doktornotor:
Should this really be case sensitive? IMHO just another can of worms. Why should Zabbix be newer than zabbix, or Apache than apache, etc. Better make it strcasecmp().

And yeah, the whole concept of greater or less as a result of string comparison is IMHO flawed (also, locale dependent is bad), but that's outside the scope here.

phil-davis:
Yeh, I changed it to ``strcasecmp()`` - that does seem more sensible. Now these thingscompare the same and do not trigger upgrade (or downgrade):

"utility-1.2.3" and "Utility-1.2.3"
"thing-1.2.3 pkg v4.5" and "thing-1.2.3 pkg V4.5" and...

phil-davis:
Note: I have done this against RELENG_2_2.
If and when it is reviewed and committed I am happy to look at what is relevant and how to do it in master for 2.3. I know there are big changes to the package install system coming for 2.3 so it seemed easier to work directly against RELENG_2_2 and get a working solution then port it forward to master.

doktornotor:
:+1:

phil-davis:
Another PR against RELENG_2_2 - merging of this is not affected by the recent folder structure changes in master - good.

PR #1812, opened 2015-08-10
Fix Cloudflare support for Dynamic DNS Updates

rohankapoorcom:
The current implementation isn't working. Looks like cloudflare changed their API. Using the following solution from the forum fixed the problem for me. All credit to det0nat3 on the forums https://forum.pfsense.org/index.php?topic=87436.msg534817#msg534817. I'm just submitting a pull request since I don't see it in the repos yet.

### DIFF
--- a/etc/inc/dyndns.class
+++ b/etc/inc/dyndns.class
@@ -671,10 +671,43 @@
 					break;
 				case 'cloudflare':
 					$needsIP = TRUE;
-					$dnsServer ='www.cloudflare.com';
-					$dnsHost = str_replace(' ', '', $this->_dnsHost);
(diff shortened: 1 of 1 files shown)


doktornotor:
Resubmitted as #1915 due to conflicts. This can be closed. Thanks!

rohankapoorcom:
Nice, thanks.

doktornotor:
@rohankapoorcom  - Since my resubmit was not acceptable due to authorship, you can either take that code from my PR referenced above and resubmit this yourself (it will probably only keep working for a couple of days at best before there's another bunch of conflicts), or live with the bug in pfSense code. There is also a backport of the code to RELENG_2_2 in PR #1916, feel free to resubmit that as well.

Have a nice day.

PR #1814, opened 2015-08-12
make sure that cron is running before reconfiguring it

doktornotor:
Please, see https://github.com/pfsense/pfsense-packages/pull/997 for rationale. However, this is not limited to a single broken package, when cron goes away for whatever reason, you can keep reconfiguring it till blue in face but nothing will happen.


doktornotor:
Resubmitted as #1828.

PR #1816, opened 2015-08-15
reduce from 10000 to 100 the maximum Y axis of quality graphs.

zxvv:
A large peak in ping delay often completely obscure the average trends in both delay and packet loss. 

This change reduces the maximum value for the Y axis to 100 (from 100000).

In the presense of intermittent large ping delays, this makes it easier to see the general trend in both ping delay and packet loss.

### DIFF
--- a/etc/inc/rrd.inc
+++ b/etc/inc/rrd.inc
@@ -957,7 +957,7 @@ function create_gateway_quality_rrd($rrd_file) {
 	if (!file_exists("$rrd_file")) {
 		$rrdcreate = "$rrdtool create $rrd_file --step $rrdinterval ";
 		$rrdcreate .= "DS:loss:GAUGE:$valid:0:100 ";
-		$rrdcreate .= "DS:delay:GAUGE:$valid:0:100000 ";
+		$rrdcreate .= "DS:delay:GAUGE:$valid:0:100 ";
 		$rrdcreate .= "RRA:AVERAGE:0.5:1:1200 ";
 		$rrdcreate .= "RRA:AVERAGE:0.5:5:720 ";
 		$rrdcreate .= "RRA:AVERAGE:0.5:60:1860 ";


rbgarga:
@zxvv In order to have patches accepted you must sign a CLA as described at https://www.pfsense.org/about-pfsense/#legal

zxvv:
I submitted the ICLA.

Sorry for the delay. Gmail inbox buried it, and I found ti today.

rbgarga:
@jdillard can you check this while you work on rrdtool changes please?

davidjwood:
To my mind, this change is unwise. There are plenty of scenarios where first hop delay can be greater than 100ms - satellite and some cellular scenarios come to mind.

10000ms is maybe too high, but I wonder whether any change is needed now that dpinger is in use and the latency figures should be more reliable.

smos:
Op 8-2-2016 om 11:15 schreef David Wood:
> To my mind, this change is unwise. There are plenty of scenarios where
> first hop delay can be greater than 100ms - satellite and some cellular
> scenarios come to mind.
> 
> 10000ms is maybe too high, but I wonder whether any change is needed now
> that dpinger is in use and the latency figures should be more reliable.
> 
> —
> <https://github.com/pfsense/pfsense/pull/1816#issuecomment-181290866>.
> 

3G internet is regularly in the 100-300ms region, so I'd recommend
atleast using 1 second. Although I've seen in the 2 second ranges if you
have wifi somewhere in between or switches with buffer bloat.

That also applies to other wireless solutions. 3G under load is often in
the second range, 4G is quite a lot better suited and well behaved.

Regards,

Seth

cbuechler:
10000 isn't unreasonable as an upper limit. 100 is way too low. 5000 maybe if it were to be dropped at all, though that doesn't seem like a good idea at this point now that we have dpinger in 2.3 which always returns sane values. The only significant need for this before was attributable to apinger's bad math.

PR #1817, opened 2015-08-16
remove routed service, should be handled by the package

doktornotor:
See https://github.com/pfsense/pfsense-packages/pull/1006


doktornotor:
Resubmitted as #1829

PR #1819, opened 2015-08-26

phil-davis:
on factory default boot.
Alix - vr1 vr0
APU - re1 re2
serial cable and do interface assignment that way.
This is a resubmit of PR #1602 and is integrated with yesterday's change
to the pfSense repo folder naming.


phil-davis:
Closing in favor of integrated newer version in https://github.com/pfsense/pfsense/pull/2185

PR #1820, opened 2015-08-26
Provide separate checkboxes for tmp and var RAM disks

phil-davis:
As suggested in forum: https://forum.pfsense.org/index.php?topic=96705.0
It is not hard to do - so here it is if someone thinks it is worth
having. e.g. on my SG-2440 full install that has plenty of free RAM I
could enable just /tmp memory disk and leave /var where it is on the
eMMC.
This is an integration and resubmit of PR #1765
Note: Leave this sitting here for now. @jimp made some suggestions in
that forum thread about maybe having a dropdown that has the 4
combinations all in a single dropdown list to choose from. It needs some
design thought before implementing. A change like this would not go in
2.2.* anyway, so there is plenty of time to decide what to do before
2.3.

### DIFF
--- a/src/etc/inc/rrd.inc
+++ b/src/etc/inc/rrd.inc
@@ -55,7 +55,7 @@ function restore_rrd() {
 
 	$rrdrestore = "";
 	$rrdreturn = "";
-	if (file_exists("{$g['cf_conf_path']}/rrd.tgz") && (isset($config['system']['use_mfs_tmpvar']) || $g['platform'] != "pfSense")) {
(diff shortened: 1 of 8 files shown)


rbgarga:
@phil-davis since you submitted it, code has changed a lot. Do you still have interest to see it committed? If yes, can you review all changes? If not, please close it

phil-davis:
My care-factor is quite low for this, and I think that @jim-p suggestion https://forum.pfsense.org/index.php?topic=96705.msg539158#msg539158 to have a dropdown with the 4 possibilities as options is the best solution.
As he says, with separate checkboxes for /tmp and /var there is the problem that mostly users want both checked together, but some will accidentallt check just 1 of the boxes.
If someone cares enough they could code it :)

PR #1821, opened 2015-08-26
Protect unset() with isset() in upgrade_config

phil-davis:
Forum: https://forum.pfsense.org/index.php?topic=96827.0
Sometimes there can be unexpected stuff in a config. To avoid errors
like "PHP Fatal error: Cannot unset string offsets in
/etc/inc/upgrade_config.inc on line 291" it seems good to use
belts-and-braces checks with any unset() of associative array elements,
just in case the thing above is an empty string (instead of an array or
non-existent thing). An unnecessary error in processing the upgrade code
can mean missing other important upgrade stuff.
These are the places I could see in upgrade_config.inc where unset() is
not already protected by isset(). A lot of them are "that could not ever
go wrong", but IMHO it is worth protecting them all just to be sure.
This is a resubmit of PR #1773 after integrating with current master.

### DIFF
--- a/src/etc/inc/upgrade_config.inc
+++ b/src/etc/inc/upgrade_config.inc
@@ -258,8 +258,8 @@ function upgrade_014_to_015() {
 		if ($config['interfaces']['wan']['gateway'] <> "") {
 			$config['system']['gateway'] = $config['interfaces']['wan']['gateway'];
 		}
+		unset($config['interfaces']['wan']['gateway']);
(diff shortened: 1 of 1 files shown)


phil-davis:
Should I backport this to RELENG_2_2 so it becomes part of the code if another 2.2.* release becomes necessary?

rbgarga:
Please! Thanks!

PR #1826, opened 2015-08-26
Redmine #4925 Fix version comparison to know 10 is bigger than 9

phil-davis:
This rtrim of ".0" is stripping any "0" from the end of the passed-in version strings. That makes "2.3.10" become "2.3.1" which then removes any chance of the following nice comparison logic working.
Just removing the "0" seems fine. It keeps the supplied version data untouched, just getting rid of any trailing dots.
Apart from fixing the bug here, this change has the side-effect that a version change from "2.3" to "2.3.0" will now be seen as an upgrade.
What is the requirement for that?
Do you want to have extra logic that checks for "bare" zeroes on the end and make "2.3", "2.3.0", "2.3.0.0"... all be considered the same version?
This is a resubmit of PR #1810 after integrating with the current master.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -2239,8 +2239,8 @@ function version_compare_string($a, $b) {
 	}
 }
 function version_compare_numeric($a, $b) {
-	$a_arr = explode('.', rtrim($a, '.0'));
-	$b_arr = explode('.', rtrim($b, '.0'));
+	$a_arr = explode('.', rtrim($a, '.'));
+	$b_arr = explode('.', rtrim($b, '.'));
 
 	foreach ($a_arr as $n => $val) {
 		if (array_key_exists($n, $b_arr)) {


rbgarga:
We do not use .0 versions, our first release of a new major version is 2.2, 2.3, ...

Other than that, I cannot se a reason to call that rtrim('.'), and I was wondering we can use PHP version_compare() function instead of having our one. Then we could ditch this function completely and replace all calls by version_compare. What do you think?

http://php.net/version_compare

jim-p:
Seems like that gets most of the way there but does not understand some of our release names (-RELEASE, -DEVELOPMENT, -PRERELEASE, etc). Would need testing, but if it works, great.

rbgarga:
version_compare_numeric() doesn't care about -RELEASE, -DEVELOPMENT, ...

rbgarga:
Check this out @jim-p - https://github.com/pfsense/pfsense/blob/master/src/etc/inc/pfsense-utils.inc#L2266

jim-p:
Ah, yeah, I missed that it was just the one part of it. Should be fine for that then if the behavior is desired.

phil-davis:
I was just fixing the bug here, not trying to re-engineer the whole thing :) Actually I think the whole suite of *version_compare* functions already in pfsense-utils.inc is doing all the stuff that PHP version_compare does and handles various pfSense-specific keywords used in version strings.

rbgarga:
I'll merge the fix then, and add an item to my TODO list to clean this up. Thanks!

PR #1830, opened 2015-08-28
Bug #4569 - system_set_harddisk_standby() rewrite for current CAM-based ATA stack

doktornotor:
Resubmit of #1767

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -2042,6 +2042,11 @@ function system_dmesg_save() {
 
 function system_set_harddisk_standby() {
 	global $g, $config;
+	// Get only suitable candidates for standby; using get_smart_drive_list()
+	// from utils.inc to get the list of drives.
+	$harddisks = array();
+	$harddisks = get_smart_drive_list();
+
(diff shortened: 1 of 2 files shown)


src/etc/inc/system.inc
```diff
@@ -2042,6 +2042,11 @@ function system_dmesg_save() {
 
 function system_set_harddisk_standby() {
 	global $g, $config;
+	// Get only suitable candidates for standby; using get_smart_drive_list()
+	// from utils.inc to get the list of drives.
+	$harddisks = array();
```
rbgarga: Why initialize this as array() if you overwrite its value on next line?

src/etc/inc/system.inc
```diff
@@ -2042,6 +2042,11 @@ function system_dmesg_save() {
...
+	$harddisks = get_smart_drive_list();
```
rbgarga: This variable is only used if $config['system']['harddiskstandby'] is set, then this declaration can be moved inside that if()

doktornotor: Indeed a better idea. Thanks.

PR #1831, opened 2015-08-28
filter.inc - Remove useless "could not find gateway" logspam

doktornotor:
Resubmit of #1784.

See Bug #4102 and others. Getting completely pointless "errors" noise for completely normal situations in logs is NOT cool. Other annoyed users:
https://forum.pfsense.org/index.php?topic=91455.0
https://forum.pfsense.org/index.php?topic=96918.0


cbuechler:
had conflicts so I just made the changes manually both RELENG_2_2 and master. Thanks!

doktornotor:
Thanks as well... :)

PR #1832, opened 2015-08-28
filter.inc - Remove useless "could not find gateway" logspam (2.2.x)

doktornotor:
Same as #1831 - this is for RELENG_2_2 branch.


doktornotor:
Was merged manually in cb7c4631b70e6698443a0deb5033c8929d87ef44 - closing and thanks.

PR #1837, opened 2015-08-28
Redmine #5046 UEFI network booting arch 00:09 RELENG_2_2

phil-davis:
This PR against RELENG_2_2 so the bug reporter can easily test. Then can also be applied to master.


sofakng:
Works great for me.  Thanks!

PR #1839, opened 2015-08-28
This fix the issues #3900 and #5051 to make No-ip round robin DNS works.

tiagobar:
The paramater "&hostname" should be changed to "&h[]"
I think this affect only NO-IP DNS that uses round robin. Anyway the no-ip official client uses the "&h[]" parameter.

There's also a change to Dynamic DNS hostname allow '@' character (bug #3900).

### DIFF
--- a/etc/inc/dyndns.class
+++ b/etc/inc/dyndns.class
@@ -340,7 +340,7 @@
 					} else {
 						$iptoset = $this->_dnsIP;
 					}
-					curl_setopt($ch, CURLOPT_URL, $server . $port . '?username=' . urlencode($this->_dnsUser) . '&pass=' . urlencode($this->_dnsPass) . '&hostname=' . $this->_dnsHost.'&ip=' . $iptoset);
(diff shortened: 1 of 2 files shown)


doktornotor:
PR #1914 has a generic fix for '@.' and '*.' in hostnames... As noted there, these overly restrictive input checks cause just periodic breakage, maintainance burder and are no good with this perpetually changing stuff. E.g., the NameCheap thing is yet again broken with wildcards, see https://redmine.pfsense.org/issues/3568#note-6

doktornotor:
@tiagobar Since my resubmit was not acceptable due to authorship, you can either take that code from my PR referenced above and resubmit this yourself (it will probably only keep working for a couple of days at best before there's another bunch of conflicts), or live with the bug in pfSense code.

There is also a forwardport of the code to master in PR #1914, feel free to resubmit that as well.

Have a nice day.

rbgarga:
@tiagobar in order to have patches accepted, you must sign a CLA as described at https://www.pfsense.org/about-pfsense/#legal

cbuechler:
@tiagobar could you please update for master for 2.3 and re-submit? Then I'll make sure this gets included for 2.3.

PR #1841, opened 2015-08-29
Add ability to hide drives from SMART Status widget

doktornotor:
This is a resubmit of #1766 for RELENG_2_2. No point in doing this for master/bootstrap until things get somewhat stable there.

Useful for machines with USB card readers which display as daX, or system with media that do not support SMART in general.


cbuechler:
limiting changes on 2_2 at this point. Bootstrap things are stable now, please re-submit for master.

PR #1842, opened 2015-08-29
Display monitor IP on Gateways widget RELENG_2_2

phil-davis:
This version is for RELENG_2_2 on top of 2.2.4
This code was committed to master a while ago. If there is a 2.2.5 then it would be really handy for me if it comes as part of 2.2.5. I have it installed with System Patches on every pfSense I have. For me it is a useful enhancement to the Gateways Widget that would have been nice to have ever since the Gateways Widget existed.

### DIFF
--- a/usr/local/www/includes/functions.inc.php
+++ b/usr/local/www/includes/functions.inc.php
@@ -5,7 +5,7 @@
 	Copyright (C) 2013-2015 Electric Sheep Fencing, LP
 */
 
-if(Connection_Aborted()) {
+if (Connection_Aborted()) {
 	exit;
 }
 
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: This PR also includes the developer style guide changes for these files, which are no change to functionality.

phil-davis:
I think that most of the code here is real functional change. The included style guide changes are smaller and obvious in the diffs. So maybe it will be easy to review and take all of this as-is.
Let me know if I need to make a minimal-change version without any style-guide stuff.

phil-davis:
Note for reference: This is Redmine #4782

PR #1843, opened 2015-08-29
Add description as a display option on Traffic Graph RELENG_2_2

phil-davis:
This is for patching to RELENG_2_2 on top of 2.2.4
This code was committed to master a while ago. If there is going to be a 2.2.5 released then it would be handy for me if this was included. That will save me a System Patches entry :)
I realise this one might be a bit of an ask, because it is an enhancement that few people might care about, rather than a bug-fix or deficiency-fix.
This also contains the developer style guide changes, which are no change to functionality.

### DIFF
--- a/usr/local/www/bandwidth_by_ip.php
+++ b/usr/local/www/bandwidth_by_ip.php
@@ -1,8 +1,6 @@
 <?php
-/* 
- * To change this template, choose Tools | Templates
- * and open the template in the editor.
- *
+/*
+	bandwidth_by_ip.php
  */
 
 /*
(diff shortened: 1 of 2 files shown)


rbgarga:
Doing style changes combined with functional changes in the same commit makes it harder to review. Specially for RELENG_2_2, that is a maintenance branch. Lets leave these style changes only for master or bootstrap branches. Can you re-send it only with functional changes?

phil-davis:
Yes, I will have a go at working out exactly which lines were the functional changes for this...

phil-davis:
Resubmited as https://github.com/pfsense/pfsense/pull/1888

PR #1847, opened 2015-08-30
Reword text in etc/rc

phil-davis:
The current text is not good English. A suggested re-wording is here.
There is a discussion that could be had about what actual action to take
here, and how good fsck is at automatically recovering from xyz. This PR
is not about that. I just thought I would suggest some better English
for this message.


doktornotor:
```s/tried/attempted/``` perhaps? (Should rather say failed as far as the final result of that action is concerned, though :angry: )

phil-davis:
Closed in favor of newer integrated version in PR https://github.com/pfsense/pfsense/pull/1898

PR #1848, opened 2015-08-30
Integrate bootstrap etc/inc with master

phil-davis:
This applies the little changes/differences in etc/inc master to the bootstrap
branch so that etc/inc in bootstrap will now just have the real
differences that are due to real bootstrap changes.
This should make it easier to compare diffs between bootstrap and master, and easier to eventually merge bootstrap into master.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -309,7 +309,7 @@ function get_user_privileges(& $user) {
 	foreach ($names as $name) {
 		$group = getGroupEntry($name);
 		if (is_array($group['priv'])) {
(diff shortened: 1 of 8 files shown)


phil-davis:
@cbuechler @sbeaver-netgate
Any interest in stuff like this, and other later pull requests for other little things that I noticed.
If these get left too long to review... then conflicts will come as you guys also work on the same code modules.
Others also have a few pull requests for bootstrap and it would be good to comment on or commit those also.

jim-p:
@phil-davis Most of them are good we just haven't had a chance to approve them all properly yet. @sbeaver-netgate looked most of them over and I think all but one were OK.

PR #1855, opened 2015-08-31
Update copyright-master.txt

phil-davis:
to match license.txt
So that whichever file someone uses as a template to paste in all files, it will be the same.
Or else delete copyright-master.txt
1) copyright-master.txt has a copyright line for Scott Ullrich - is that supposed to be there? Is it meant to go in the header of every file? What is the policy about putting/keeping the names of other previous contributors as copyright lines in the various source files?
2) At point 3. there is a special copyright symbol used. Does that work OK to go in ordinary text source code files?
3) The double-quotes at the start of "This product includes... on lines 18 and 33 are never closed - it bothered me, my brain-resident-text-parser routine failed :)

It would be good to have a really final version of this before someone re-pastes it into every code header!

### DIFF
--- a/src/usr/local/www/copyright-master.txt
+++ b/src/usr/local/www/copyright-master.txt
@@ -1,34 +1,34 @@
 /* ====================================================================
- *  Copyright (c)  2004-2015  Electric Sheep Fencing, LLC. All rights reserved. 
- *  Copyright (c)  2004, 2005 Scott Ullrich
+ *  Copyright (c) 2004-2015 Electric Sheep Fencing, LLC. All rights reserved.
+ *  Copyright (c) 2004, 2005 Scott Ullrich
  *
- *  Redistribution and use in source and binary forms, with or without modification, 
- *  are permitted provided that the following conditions are met: 
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions are met:
  *
  *  1. Redistributions of source code must retain the above copyright notice,
- *      this list of conditions and the following disclaimer.
+ *     this list of conditions and the following disclaimer.
  *
  *  2. Redistributions in binary form must reproduce the above copyright
- *      notice, this list of conditions and the following disclaimer in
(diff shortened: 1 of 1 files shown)


gonzopancho:
We have a full assignment from Ullrich.  In fact, we own all rights, and the code should have been Copyright <the project, or BSDP, or any other entity along the way>

gonzopancho:
http://www.copyright.gov/circs/circ03.pdf

phil-davis:
@gonzopancho Fun reading of the intricacies of how and when to put what copyright marks where.
Anyway, I am just pointing out inconsistencies that are in the various copyright template files that are here in the repo, and suggesting that somebody who knows what ESF wants/requires as the correct end result should make a final decision and then make the necessary changes.
It becomes a pest in diffs if there are so many files that have a bit different spacing of this stuff and...
If you are going to retain some copyright notices of past individual contributors (I am not giving any opinion as to whether that is required, good practice or just nice to do), then it will have to be someone from ESF who takes the time to do the proper research and work out which copyright attributions are genuine (from where the mentioned party actually contributed that module, or significant chunks of it) and which just got in the copyright list because it happened to come along in a copy/paste when someone was creating a new file or was copy/pasting general formatting of the license notice and...
Please feel free to close this pull request and do something completely different in resolving the inconsistencies of this stuff in the current repo/s, if you wish.

doktornotor:
To shorten what phil correctly noted - all we are asking for is **consistency**. It's a damn mess across the repos.

phil-davis:
This format seems to have been superseded by now pasting in a block that is formatted with tabs.
I put some sample small edits to the tabbed format in https://github.com/pfsense/pfsense/pull/1863
I guess what people think should be the standard format for this copyright block changes every day.

doktornotor:
Plus, those extra tabs look like POS.

phil-davis:
Point Of Sale? :)

phil-davis:
Closing this since it is a general matter for ESF to sort out. This was just meant to point out the issue, which it has done.

PR #1859, opened 2015-09-01
Encoded the alias and the rule description to avoid stored XSS 

itshari:


jim-p:
I fixed these in slightly different ways on the RELENG_2_2 and bootstrap branches. Fixes will come to master once we merge the bootstrap branch back into master.

PR #1860, opened 2015-09-01
ipsec: set retransmission parameters for IKEv2 auth and dpd.

laped83:
Added strongswan charon retranmission values (tries, timeout and base) to advanced ipsec settings.

Signed-off by: Lars Alex Pedersen <thacaleb@gmail.com>

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -329,6 +329,11 @@ function is_numericint($arg) {
 	return (((is_int($arg) && $arg >= 0) || (is_string($arg) && strlen($arg) > 0 && ctype_digit($arg))) ? true : false);
 }
 
+/* validate non-negative numeric decimal string, or equivalent numeric variable */
+function is_numericfloat($arg) {
+	return (((is_float($arg) && $arg >= 0) || (is_string($arg) && strlen($arg) > 0 && ctype_digit(preg_replace('/\./', '', $arg, 1)))) ? true : false);
+}
+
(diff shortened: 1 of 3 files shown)


laped83:
Update pfsense doc:
https://doc.pfsense.org/index.php/Advanced_IPsec_Settings

with retransmission information at:
https://wiki.strongswan.org/projects/strongswan/wiki/Retransmission

laped83:
Okay seems like this branch it kinda hard to get merged after master has changed to bootstrap :)

I am considering starting over and maybe changing the way to add change settings in strongswan.conf. Strongswan.conf has a lot of configuration values so what do you guys think about making something that looks like the "Advanced System Tunables" with "section, description, value"? A set of default values is presented but the user can also add additional settings. This way the user will have access to all settings in https://wiki.strongswan.org/projects/1/wiki/StrongswanConf.

rbgarga:
What do you think about this change and about the ide from last comment @cbuechler ?

cbuechler:
I've been contemplating this and other options, there are a bunch of other configurable knobs in strongswan that the vast majority don't need, but some small portion do. I agree with @laped83 's last comment re: adding a tunables list of sorts like the sysctl one now, that's the thing I was leaning towards for options like the ones covered here, plus that would cover all the other ones. If we kept adding to the current advanced tab it'd end up being unwieldy. That should also discourage people from mucking with things they shouldn't be.

PR #1861, opened 2015-09-01
Lint build.sh using ShellCheck

jlduran:
[SC2006] Use $(..) instead of legacy \`..\`
[SC2086] Double quote to prevent globbing and word splitting
[SC2155] Declare and assign separately to avoid masking return values
[SC2166] Prefer [ p ] && [ q ] as [ p -a q ] is not well defined

[SC2006]: https://github.com/koalaman/shellcheck/wiki/SC2006
[SC2086]: https://github.com/koalaman/shellcheck/wiki/SC2086
[SC2155]: https://github.com/koalaman/shellcheck/wiki/SC2155
[SC2166]: https://github.com/koalaman/shellcheck/wiki/SC2166

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -77,7 +77,8 @@ usage() {
 	exit 1
 }
 
-export BUILDER_ROOT=$(realpath $(dirname ${0}))
+export BUILDER_ROOT
+BUILDER_ROOT=$(realpath "$(dirname "${0}")")
 export BUILDER_TOOLS="${BUILDER_ROOT}/tools"
 
 unset _SKIP_REBUILD_PRESTAGE
@@ -200,13 +201,13 @@ while test "$1" != ""; do
 done
 
 # Suck in local vars
-. ${BUILDER_TOOLS}/builder_defaults.sh
+. "${BUILDER_TOOLS}"/builder_defaults.sh
 
 # Suck in script helper functions
-. ${BUILDER_TOOLS}/builder_common.sh
+. "${BUILDER_TOOLS}"/builder_common.sh
 
 # Print var required with -V and exit
-if [ -n "${var_to_print}"  ]; then
+if [ -n "${var_to_print}" ]; then
 	eval "echo \$${var_to_print}"
 	exit 0
 fi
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -77,7 +77,8 @@ usage() {
 	exit 1
 }
 
-export BUILDER_ROOT=$(realpath $(dirname ${0}))
+export BUILDER_ROOT
+BUILDER_ROOT=$(realpath "$(dirname "${0}")")
```
rbgarga: what is the reason to split it in 2 lines? Also, I cannot think about a reason to have empty $0, then these extra quotes are unnecessary IMO

build.sh
```diff
@@ -200,13 +201,13 @@ while test "$1" != ""; do
 done
 
 # Suck in local vars
-. ${BUILDER_TOOLS}/builder_defaults.sh
+. "${BUILDER_TOOLS}"/builder_defaults.sh
```
rbgarga: Since BUILDER_ROOT is mandatory and BUILDER_TOOLS is BUILDER_ROOT/tools, it will never be empty, these extra quotes are unnecessary

build.sh
```diff
@@ -292,10 +293,10 @@ fi
 echo ">>> Building image type(s): ${_IMAGESTOBUILD}"
 
 if [ -z "${_SKIP_REBUILD_PRESTAGE}" ]; then
-	[ -n "${CORE_PKG_TMP}" -a -d "${CORE_PKG_TMP}" ] \
-		&& rm -rf ${CORE_PKG_TMP}
-	[ -n "${CORE_PKG_PATH}" -a -d "${CORE_PKG_PATH}" ] \
-		&& rm -rf ${CORE_PKG_PATH}
+	[ -n "${CORE_PKG_TMP}" ] && [ -d "${CORE_PKG_TMP}" ] \
```
rbgarga: test (aka [) has an 'and' option, -a, I don't think it's a good idea to convert one test call in 2 test calls for no other reason

build.sh
jlduran: I agree, but it does not harm either. Do you prefer `"${BUILDER_TOOLS}/builder_defaults.sh"` instead?

build.sh
jlduran: > The XSI extensions specifying the -a and -o binary primaries and the '(' and ')' operators have been marked obsolescent. (Many expressions using them are ambiguously defined by the grammar depending on the specific expressions being evaluated.) 

Source:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

jlduran:
I'll wait until the dust settles and resubmit, if applicable.

PR #1863, opened 2015-09-03
Suggested format changes to copyright

phil-davis:
Those spaces in the lines of section 4 are still being pasted in.
I suggest closing the double-quotes of the acknowledgment text.
Of course every copyright section has variations on the formatting. It still needs someone to declare a final correct version, and then someone to paste it everywhere.

### DIFF
--- a/src/usr/local/www/system_advanced_sysctl.php
+++ b/src/usr/local/www/system_advanced_sysctl.php
@@ -22,12 +22,12 @@
  *	3. All advertising materials mentioning features or use of this software
  *		must display the following acknowledgment:
  *		"This product includes software developed by the pfSense Project
- *		 for use in the pfSense software distribution. (http://www.pfsense.org/).
+ *		 for use in the pfSense software distribution. (http://www.pfsense.org/)."
  *
  *	4. The names "pfSense" and "pfSense Project" must not be used to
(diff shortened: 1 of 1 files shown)


doktornotor:
WTH is with indentation there? Noone writes paragraphs like this, completely backwards- if anything, should be the first list that has the extra tab, if anything.

phil-davis:
I am surprised, but actually I do not care any more. If everyone is going to paste in a differently formatted copyright section every day, then so be it. I just have to get some therapy for my OCD tendencies :)

doktornotor:
I probably have some OCD as well.... Or, it's just that the header is the thing that immediately strikes you since it's the very first thing you see in the file. This looks like it was "formatted" by someone who keeps using space key instead of styles in MS Word. (Also, the ShrewSoft copyright line in this one file, god knows where this got pasted from from.)

jlduran:
That's the purpose of adding an [`.editorconfig`](http://editorconfig.org) file to the project. Apparently some developers are not taking advantage of it.

phil-davis:
Closing since this is a general issue in all files for ESF to sort out. Nothing to be gained by having this particular example sitting in the open pull requests list. The matter has been brought to people's attention.

PR #1883, opened 2015-09-06
fix sending interim update accounting info

mhsenpc:
when this conditions happens:
send RADIUS accounting packets: on
Reauthenticate connected users every minute: off
Accounting updates: interim update
captive portal doesn't send accounting info to radius server.because of that if statement


rbgarga:
This will not fix the problem. The if statement is checking if !isset($cpcfg['reauthenticate']), and this is the option disabled when you uncheck "Reauthenticate connected users every minute".

PR #1890, opened 2015-09-09
Openvpn pam auth

pccowboy:
This adds a new auth server type to the gui, which relies on setup of the openvpn-plugin-auth-pam and local PAM.  When this is used to create an auth server, it can then be selected as the backend for authentication in the openvpn page, and the 'Client Export' page will properly show a client install package for 'Certificate with External Auth'.

I am happy to writeup a howto on this, just need to know where to get a login.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1466,6 +1466,11 @@ function authenticate_user($username, $password, $authcfg = NULL, &$attributes =
 				$authenticated = true;
 			}
 			break;
+		case 'openvpn_PAM':
+			// In the openvpn flow, authenticate_user does not get called until after the
+			// openvpn-plugin-auth-pam.so is executed, so if PAM saz the user is OK, trust that
+			$authenticated = true;
+			break;
 		case 'radius':
 			if (radius_backed($username, $password, $authcfg, $attributes)) {
 				$authenticated = true;
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -135,6 +135,7 @@ $firewall_rules_dscp_types = array(
 
 $auth_server_types = array(
 	'ldap' => "LDAP",
+	'openvpn_PAM' => "OpenVPN PAM Plugin",
 	'radius' => "Radius");
 
 $ldap_urltypes = array(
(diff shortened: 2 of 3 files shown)


rbgarga:
@pccowboy In order to have patches accepcted, you must sign a CLA as described at https://www.pfsense.org/about-pfsense/#legal

pccowboy:
Just signed the ICLA, thanks!

pccowboy:
ping - is there anything I can do to make this PR more palatable?

gonzopancho:
I'll ping garga about it.

rbgarga:
@pccowboy Is this patch completed? Is it working on 2.3? I couldn't see where it configures openvpn PAM module.

pccowboy:
I have not tested it on 2.3, I will try that out in the next few days.

I can provide a "default" openvpn PAM config, but the reason I wrote this patch was because I expect anyone going this route has a custom PAM configuration in place that they want to use to authorize VPN access.  In the case of my dayjob, we use a combination of LDAP and Kerberos via a custom shell script.  So, rather than making a default PAM config that might allow access that someone does not want, I figured it was safer to let the user supply details appropriate to their environment.

I have not started yet, but I hope to provide a PR that will also add a pfSense PAM module, which I also need at my dayjob.

pccowboy:
Hello, anything I can add to help this along?  I am moving on to a new company, and I'd like to be able to have my current company run from the main releases, rather than patching their installation.

cbuechler:
This won't work as-is, the PAM components don't exist and it doesn't configure OpenVPN to use it. 

@pccowboy if you'd still like to submit it, probably should discuss approach and end result goal first on the development board on the forum.

pccowboy:
@cbuechler bummer, then - but it has been working fine in a production environment since I submitted it.

After thinking about it, trying to work the configuration of PAM into pfsense makes little sense to me, configuration of PAM should be done by admins that are very familiar with the local environment and local security concerns.  While a howto could give guidelines, trying to make a pfsense GUI component to control this seems like a nightmare.

This patch as-is gives admins with a custom PAM authorization setup an easy way to integrate their existing bits with pfSense and OpenVPN.  The setup required to use it with openvpn is the same setup mentioned in a couple of other howto docs, and I was willing to write up instructions that would enable an organization using PAM in a custom manner to integrate pfSense/OpenVPN.

Since I have left the company I wrote this for, I really have no interest in pursuing it - it seems to have been really low priority.

PR #1895, opened 2015-09-12
Fix incorrect netmask sent to client with static IP set in RADIUS (Bug #5129)

doktornotor:
https://redmine.pfsense.org/issues/5129

### DIFF
--- a/src/etc/inc/openvpn.auth-user.php
+++ b/src/etc/inc/openvpn.auth-user.php
@@ -186,15 +186,13 @@ if (is_array($attributes['routes'])) {
 }
 
 if (isset($attributes['framed_ip'])) {
-/* XXX: only use when TAP windows driver >= 8.2.x */
-/*	if (isset($attributes['framed_mask'])) {
+	if (isset($attributes['framed_mask'])) {
 		$content .= "topology subnet\n";
 		$content .= "ifconfig-push {$attributes['framed_ip']} {$attributes['framed_mask']}";
 	} else {
-*/
-	$content .= "topology net30\n";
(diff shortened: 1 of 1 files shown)


src/etc/inc/openvpn.auth-user.php
```diff
@@ -186,15 +186,13 @@ function getNasIP() {
...
+	if (isset($attributes['framed_mask'])) {
 		$content .= "topology subnet\n";
 		$content .= "ifconfig-push {$attributes['framed_ip']} {$attributes['framed_mask']}";
 	} else {
-*/
-	$content .= "topology net30\n";
-	$content .= "ifconfig-push {$attributes['framed_ip']} ". long2ip((ip2long($attributes['framed_ip']) + 1));
-//	}
+		$content .= "topology net30\n";
+		$content .= "ifconfig-push {$attributes['framed_ip']} ". long2ip((ip2long($attributes['framed_ip']) + 1));
```
jim-p: As long as this is being fixed, we may as well fix the consistency in client/server address as well. In client-specific config/overrides the server address is made one less than the client address, not one higher, which is consistent with OpenVPN's own internal address assignment.  See https://github.com/pfsense/pfsense/blob/master/src/etc/inc/openvpn.inc#L1053 - thoughts?

Should be as easy and changing that +1 to -1.

doktornotor: Yeah, definitely makes sense.

@ermal - any particular reason here why's it done the other way round for RADIUS?

jim-p: Most likely it was done at the request of the original customer that sponsored the work, and probably should have been flipped (or made an option) when imported. Given how little it is used the impact should be fairly low.

doktornotor:
Yeah I don't think there's actually much to break here. :)

PR #1901, opened 2015-09-16
implement rule hit count/show/kill

marcelloc:
hit_count_23
hit_count_23_02


marcelloc:
At this point on 2.3. hit count is ok, rule dele is ok, rule title info is ok but rule current states gets on html form ajax but I't does not popup on mouse over.
This is something I'll need some help to adjust on bootstrap.

src/etc/inc/rule_count.inc
```diff
@@ -0,0 +1,25 @@
...
+                }
+                if ($cr != "" && preg_match("/Evaluations:\s+(\d+)\s+Packets:\s+(\d+)\s+Bytes:\s+(\d+)\s+States:\s+(\d+)\s+/",$line,$m2)) {
+                        if (isset($rules_count_array[$cr][Packets])) {
+                                $rules_count_array[$cr]['Packets'] += $m2[2];
+                                $rules_count_array[$cr]['Evaluations'] += $m2[1];
+                                $rules_count_array[$cr]['Bytes'] += $m2[3];
+                                $rules_count_array[$cr]['States'] += $m2[4];
+                                $rules_count_array[$cr]['RuleId'] .= "|$rid";
+                        } else {
+                           $rules_count_array[$cr]=array('Evaluations'=>$m2[1],'Packets'=>$m2[2],'Bytes'=>$m2[3],'States'=>$m2[4],'Label'=>$rl,'RuleId'=>$rid);
```
phil-davis: While looking at this, I might as well comment that the indent here seems to have spaces not tab.

phil-davis:
I made a couple of PRs to your repo/branch for code style here. That does not help you with ajax, bootstrap and popup on mouse over, sorry :(

src/etc/inc/filter.inc
```diff
@@ -2313,7 +2313,9 @@ function filter_generate_user_rule_arr($rule) {
 	$line = filter_generate_user_rule($rule);
 	$ret['rule'] = $line;
 	$ret['interface'] = $rule['interface'];
-	if ($rule['descr'] != "" and $line != "") {
+	if ($rule['tracker'] != "" and $line != "") {
+		$ret['descr'] = "label \"" . fix_rule_label("USER_RULE: {$rule['tracker']}") . "\"";
```
rbgarga: The idea of adding rule description to USER_RULE: label is to make it easy to identify what rule it is when you are reading /tmp/rules.debug. Since every rule has a tracker, it will replace all USER_RULE: labels by the tracker, what is not a good idea

src/etc/inc/rule_count.inc
```diff
@@ -0,0 +1,25 @@
...
+				$rules_count_array[$cr]['Evaluations'] += $m2[1];
+				$rules_count_array[$cr]['Bytes'] += $m2[3];
+				$rules_count_array[$cr]['States'] += $m2[4];
+				$rules_count_array[$cr]['RuleId'] .= "|$rid";
+			} else {
+				$rules_count_array[$cr] = array('Evaluations' => $m2[1], 'Packets' => $m2[2], 'Bytes' => $m2[3], 'States' => $m2[4], 'Label' => $rl,'RuleId' => $rid);
+			}
+		}
+	}
+}
```
rbgarga: @loos-br What do you think about it? I was considering to implement it in pfSense PHP module [1]. What do you think?

[1] https://github.com/pfsense/FreeBSD-ports/blob/devel/devel/php56-pfSense-module/files/pfSense.c

src/etc/inc/filter.inc
marcelloc: Including the tracker id in rule description is the only way I found to track it back to count hits. what do you think if I concat description after tracker id? This way it can get counters on gui and still get human readable on /tmp/rules.debug.

rbgarga: @loos-br is checking the possibility of implementing it in C, as part of pfSense PHP module. This way we would kill that code you added that calls exec(pfctl -vvsr) and will end up calling pfSense_get_stats($rule_tracker) or something like this.

loos-br: Yes, I'll extend the pfSense PHP module to export the firewall rules information, it will simplify the implementation of this feature.
Stay tuned...

marcelloc:
I've updated the code to current 2.3 beta but it looks like has some conflicts yet.

How can I find where they are?

rbgarga:
Conflict means your tree is out of date comparing with upstream. Usually what you need to do is (considering you have a remote called upstream pointing to pfSense main repo):

# git fetch upstream
# git merge upstream/master

If there are conflicts, resolve them and run 'git commit'

# git push origin

src/etc/inc/filter.inc
```diff
@@ -4218,4 +4218,4 @@ function filter_get_antilockout_ports($wantarray = false) {
 
 }
 
-?>
+?>
```
rbgarga: These changes on filter.inc looks unnecessary

src/etc/inc/rule_count.inc
```diff
@@ -0,0 +1,42 @@
...
+	foreach ($line as $key=>$value){
+		switch ($key){
+			case 'bytes':
+			case 'states':
+			case 'packets':
+			case 'evaluations':
+				$rules_count_array[$line['tracker']][$key] += $value;
+				break;
+			case 'id':
+				if (isset($rules_count_array[$line['tracker']][RuleId])) {
```
rbgarga: What is RuleId? Shouldn't it be $value?

src/etc/inc/rule_count.inc
```diff
@@ -0,0 +1,42 @@
...
+					<h3 class="modal-title" id="myModalLabel"><?=gettext("Rule States")?></h3>
+				</div>
+
+				<div class="modal-body" id="RuleStates-modal-body">
+				</div>
+			</div>
+		</div>
+	</div>
+<?php }?>
+
```
rbgarga: Why add a new .inc, without any function? this code should go to firewall_rules.php instead

rbgarga:
@marcelloc, after discuss it with @sbeaver-netgate, and check that it increases time to load page. We believe that it would be better if this feature is conditional, off by default.

Can you please add a new option, off by default, on System -> Advanced and add a conditional to show this data only if this option is enabled?

marcelloc:
@garga, it loads firewall states (mini tcpdump modal) only on hitcount mouse over then click. 

the ovehead of kernel function pfSense_get_pf_rules() is minimal. 

"..$html  = "<a href=\"#\" onmouseover=\"showState('{$ruleid}','{$hitcount}','{$label}');.."

sbeaver-netgate:
Its a matter of screen real-estate too, If you would please wrap something like if ($showcount ) {} around the proposed code, I would be pleased to add the control on the system admin page. You can just set it to "true" and I'll take it from there. I will be making a few cosmetic changes anyway.

rbgarga:
@sbeaver-netgate I had a different idea, what about if we have a function to get data only from one rule, and this would be called on mouseover through ajax.

The way it's implemented today, all data is collected, from all rules, when firewall_rules.php is loaded, then data is static. Changing it to do a call per-rule would make data dynamic

sbeaver-netgate:
I'm not really concerned about page load time. When I profiled the PR I found the it had very little impact. Not even measurable.

If we were going to make it dynamic, I would make the whole thing dynamic and reload the hit counters every few seconds. That should probably be a separate task though.

rbgarga:
Yeah, nevermind my last message. I was thinking it would only show on mouseover.

May I manually merge it as is and add necessary isolation to an option called "$config['system']['show_rule_hits'] ?

sbeaver-netgate:
I think that would work :)

rbgarga:
I'll handle that

loos-br:
I'm committing my version of this feature, it fixes the issues in this pull request.

rbgarga:
@loos-br committed a different version. Thanks!!

PR #1904, opened 2015-09-18
Redmine #4924 Improve pkg version comparison

phil-davis:
Improved way to compare 2 version strings.
This is the code that is already merged in RELENG_2_2 from https://github.com/pfsense/pfsense/pull/1811 now integrated with the current master.

### DIFF
--- a/src/etc/inc/pkg-utils.inc
+++ b/src/etc/inc/pkg-utils.inc
@@ -347,6 +347,66 @@ function get_pkg_info($pkgs = 'all', $info = 'all') {
 	return $result;
 }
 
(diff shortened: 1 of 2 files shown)


rbgarga:
For master, since we are moving to pkg(8), the best option is to use pkg to check versions.
```
pkg version -t v1 v2
```
It echoes '<', '=' or '>'. It shoud be called using pkg_exec() from pkg-utils.inc

rbgarga:
It's going to be handled using pkg, I'm now working on it. Thanks!

phil-davis:
Great - saves me learning for myself, I can cheat ny seeing what you do :)

PR #1906, opened 2015-09-19
php.php/pkg_edit.php - cleanups and fixes (resubmit of #1787)

doktornotor:
Kindly merge this before conflicts arise again... Really wasting time here.

- Remove unused variables
- Initialize variables
- Check if the actual $fieldname element is present in the $a_pkg[$id] array before trying to assign its value. Do same with default_value. Fixes issue where default value was not being populated for newly added fields.


doktornotor:

PR #1907, opened 2015-09-19
pkg_edit.php - fix issue where default value was not being populated for newly added fields (RELENG_2_2)

doktornotor:
Backport from PR #1906 / PR #1787 for RELENG_2_2. 

Check if the actual $fieldname element is present in the $a_pkg[$id] array before trying to assign its value. Do same with default_value. Fixes issue where default value was not being populated for newly added fields.


doktornotor:

PR #1908, opened 2015-09-20
DH groups 22-24 do not function (Bug #4918)

doktornotor:
Resubmit of #1808


doktornotor:

PR #1912, opened 2015-09-20
Add L2TP server's interface to mpd.conf (Bug #4830)

doktornotor:
Resubmit of #1750

The interface gets saved OK in the config, but in /etc/inc.vpn.inc function vpn_l2tp_configure() there is no mention of 'interface'. There is simply no code to implement the selected interface.


doktornotor:

PR #1914, opened 2015-09-20
Allow '@' at hostname (Bug #3900, Bug #5051), allow '@.' and '*.' wildcards in hostname

doktornotor:
Forward-port of #1839 

Also, allow '@.' and '*.' in DynDNS hostnames regardless of provider. These overly restrictive and provider-specific checks only cause maintenance burden and are good for nothing.


doktornotor:

PR #1915, opened 2015-09-20
Fix Cloudflare support for Dynamic DNS Updates (Todo #4737)

doktornotor:
Resubmit of #1812

The current implementation isn't working due to API change. Credits: det0nat3 @ https://forum.pfsense.org/index.php?topic=87436.msg534817#msg534817


doktornotor:

PR #1917, opened 2015-09-20

doktornotor:
Resubmit of #1793

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -695,8 +695,15 @@
 						log_error("GratisDNS.dk ({$this->_dnsHost}): DNS update() starting.");
 					}
 					$server = "https://ssl.gratisdns.dk/ddns.phtml";
-					list($hostname, $domain) = explode(".", $this->_dnsHost, 2);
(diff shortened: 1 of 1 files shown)


rbgarga:
@doktornotor I was thinking about those PRs you are re-submitting to fix conflicts, and I believe we have a problem. Original author is getting lost, and this is not good.
I would like to ask you to use --author parameter on git commit, adding original author, this would preserve history and people that originally wrote the changes.

doktornotor:
Yeah, the original author is getting lost. Because, normally, the original author is either gone or annoyed like hell by wasting time.

Closed all the offending comments. You are welcome to either merge those manually or ignore those forever. I seriously do NOT have time for similar bullshit.

Have a nice day.

gonzopancho:
Jacob, throttle back some.

doktornotor:
@gonzopancho I am pretty much throttled. Original commiters have been notified and politely asked to recycle my wasted time and resubmit their contributions from my PRs, until you break it again. I have even apologized on your behalf for wasting their time. Afraid, they just won't do the legwork regardless. Why should they after all? Not their fault.

That, or you can do that manually with whatever git magic commands you wish.

shamanime:
Please guide me if this is not the place to post about this... I got here because of #1916.

I am having trouble using CloudFlare and a domain like "test.example.com.br".
The log says "ERROR - Zone or Host ID was not found, check your hostname.".

Seems like it's not parsing the TLD correctly.

Changing to another domain with just ".com" works.

PR #1918, opened 2015-09-20
GratisDNS support for hosts without subdomains - RELENG_2_2

doktornotor:
Backport of #1917 / #1793 for RELENG_2_2


doktornotor:

PR #1920, opened 2015-09-20
pkg_edit.php - fix interfaces_selection option while using row_helpers

doktornotor:
Resubmit of #1380

@marcelloc - Please, recheck this.


doktornotor:

PR #1922, opened 2015-09-20
Move "Enable MSS clamping on VPN traffic" to System - Advanced - Firewall / NAT (Bug #4864)

doktornotor:
As noted on PR #1776 - this setting is not IPSec-specific and applies to IPSec, OpenVPN and PPPoE server, as such does not belong to IPSec Advanced Settings tab.

This PR is for RELENG_2_2 only, do whatever you want with these settings and code in 2.3/bootstrap.


rbgarga:
Move one setting to a different place on a maintenance branch is not a good practice. I would be glad to merge this on master

doktornotor:
Hmmm... The point was to stop pretending that it's IPSec-specific. For users who already set this there, nothing changes, the value will remain set. For anyone else, this change clears up the confusion. Cannot see what's the harm done here?

My thoughts regarding 2.3 were - first, it should be determined whether this really is supposed to be such a global setting, or if that should be configurable at least per VPN type, or per VPN interface? This whole thing seems rather weird here. Plus, how did the PPPoE server get into the mix here?

PR #1925, opened 2015-09-21
fix interface assignment menus running off VGA screen

doktornotor:
Resubmit of #1805 with requested gettext handling changes. Be so kind and merge it before yet another round of conflicts. This ain't really funny any more.

Credits: @nagyrobi  - this code ain 't mine, merely resubmitting it due to conflicts.

Original description from previous PR:
---------
When using VGA console, interface assignment can be a real pain in the ass because of the standard 80 columns width.

Dmesg reports the many interface description names in very long strings that don't fit in a row, this breaks the nice appearance of the interface list in the assignment menu.
The aestethics is one thing, but the real pain is that the interface list goes off screen by the time the menu asks for the WAN interface name, if there are many interfaces present. It's a real problem to choose from a list which is not visible anymore.

One fix is to maximize the length of the interface description to 48 chars.

The second fix (and also improvement for better overlooking when the list really goes off) is to print a small list of the interface names at each question. This is necessary because when somebody wants to assign the first interface to the last Optional port, the big list will be way off screen by then, and the name of it won't be visible.

This also makes it nice clean and straightforward.

### DIFF
--- a/src/etc/inc/config.console.inc
+++ b/src/etc/inc/config.console.inc
@@ -86,8 +86,9 @@ EOD;
 		$iflist = array();
 	} else {
 		foreach ($iflist as $iface => $ifa) {
+			$ifsmallist = trim($ifsmallist . " " . $iface);
 			echo sprintf("% -7s%s %s %s\n", $iface, $ifa['mac'],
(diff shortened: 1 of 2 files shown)


nagyrobi:
Thanks for pulling it in.

PR #1927, opened 2015-09-21

mortencombat:
Resubmit of #1793


rbgarga:
Applied in commit bf213d1d10dc7452cc1587ec727fede2c58d82f3
Thanks!

PR #1929, opened 2015-09-22
Fix pptp_remote0 var name

phil-davis:
I missed one of these in all the manual integration.
Commit https://github.com/pfsense/pfsense/commit/66fd7b47679187ddcfaf5852a55b2af15394f341 had this error.


phil-davis:
@rbgarga @sbeaver-netgate

PR #1931, opened 2015-09-23
Clarify the use of noshuntlaninterfaces and autoexcludelanaddress

phil-davis:
The existing code here looks like it should be working correctly. Maybe it would help future maintainers if the field name used for gathering the data value in the UI and $POSTing it is changed to autoexcludelanaddress. Then the code reads a bit more easily where the logic value of noshuntlaninterfaces and autoexcludelanaddress is swapped back and forth.

### DIFF
--- a/src/usr/local/www/vpn_ipsec_settings.php
+++ b/src/usr/local/www/vpn_ipsec_settings.php
@@ -202,7 +202,9 @@ if ($_POST) {
 			unset($config['ipsec']['makebeforebreak']);
 		}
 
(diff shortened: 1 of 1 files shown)


phil-davis:
This has already been committed https://github.com/pfsense/pfsense/commit/f1e89d805a361a784e6a315073cf1bf6c43c643d
So I wonder why the PR is still in the open state?

rbgarga:
Thsnks for spotting Phil!

PR #1933, opened 2015-09-24
Redmine #5196 Remove incorrect text about DNS servers - RELENG_2_2

phil-davis:
The correct text is already displayed under the DNS server boxes at line 892.
This should also be done to master once the conversion of services_dhcp.php to bootstrap is stable.


rbgarga:
@sbeaver-netgate can you replicate this change on branch master when converting this page?

PR #1935, opened 2015-09-25
Redmine #5200 be less aggressive about DHCP Pool Notice V1

phil-davis:
The reload_all code that happens, for example, at the end of the setup wizard, seems to have multiple places down in the various calls to regenerate lots of stuff that end up calling services_dhcpd_configure multiple times at points where the new DHCP pool range is in $config[] but the new LAN interface settings are in $config[] but have not yet been applied to the interface.
Rather than try to sort all that out right now, I test if the DHCP pool range matches with the interface subnet in $config[] - if that is so then things are looking good - eventually the interface settings from $config[] will be applied to the actual interface and all will be well.
So in that case I don't bother to file_notice() - just silently skip the pool for now.


phil-davis:
If this is the solution, then it needs to also be cherry-picked forward to master.

phil-davis:
I also made another version https://github.com/pfsense/pfsense/pull/1938 which does a log_error() when this condition happens, so that at least the event(s) can be seen in the system log.
Pick whichever you like and close the other.

rbgarga:
Close in favor of #1938. Thanks!

PR #1938, opened 2015-09-26
Redmine #5200 be less aggressive about DHCP Pool Notice V2

phil-davis:
This one will log_error() the DHCP pool message when it detects the inconsistency at the end of the setup wizard during reload all.
That way it can still be seen in the system log that this happened, and one day someone might chase down all the steps in the "reload all" process.
Compare this with https://github.com/pfsense/pfsense/pull/1935 and choose which way you would like to go.


jim-p:
This looks good to me. With this one, it logs to the main system log if the error happens during the wizard, but does not file a notice. If you change the interface outside of the wizard, it files a notice. Seems like the most desirable fix to me. Will need to be pulled into 2.3 also.

phil-davis:
For the record, this has been cherry-picked correctly up to master laso in commit https://github.com/pfsense/pfsense/commit/f8cc55bfffc33ffe02c1cc06a702d27064d54b93

PR #1939, opened 2015-09-27
Remove syslog.conf entries on package uninstall (Bug #5210) - RELENG_2_2

doktornotor:
The (broken, since it doesn't match the syslog entry) remove_text_from_file() is not needed at all. However, system_syslogd_start() must be run **after** the package entries are gone from config.xml, otherwise system_syslogd_start() just re-adds the (now almost removed) package logging configuration from there.

### DIFF
--- a/etc/inc/pkg-utils.inc
+++ b/etc/inc/pkg-utils.inc
@@ -1187,17 +1187,16 @@ function delete_package_xml($pkg) {
 		update_output_window($static_output);
 	}
 	/* syslog */
+	$need_syslog_restart = false;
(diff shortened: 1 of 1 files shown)


etc/inc/pkg-utils.inc
```diff
@@ -1205,6 +1204,14 @@ function delete_package_xml($pkg) {
 	$static_output .= gettext("done.") . "\n";
 	update_output_window($static_output);
 	write_config("Removed {$pkg} package.\n");
+	
+	/* remove package entry from /etc/syslog.conf if needed */
+	/* this must to be done after removing the entries from config.xml */
```
phil-davis: You can remove "to" from this sentence :)

PR #1940, opened 2015-09-27
Remove syslog.conf entries on package uninstall (Bug #5210)

doktornotor:
The (broken, since it doesn't match the syslog entry) remove_text_from_file() is not needed at all. However, system_syslogd_start() must be run **after** the package entries are gone from config.xml, otherwise system_syslogd_start() just re-adds the (now almost removed) package logging configuration from there.

### DIFF
--- a/src/etc/inc/pkg-utils.inc
+++ b/src/etc/inc/pkg-utils.inc
@@ -863,14 +863,14 @@ function delete_package_xml($package_name, $when = "post-deinstall") {
 		}
 	}
 	/* syslog */
+	$need_syslog_restart = false;
 	if (is_array($pkg_info['logging']) && $pkg_info['logging']['logfilename'] <> "") {
 		$static_output .= "Syslog entries... ";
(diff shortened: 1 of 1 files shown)


src/etc/inc/pkg-utils.inc
```diff
@@ -880,6 +880,13 @@ function delete_package_xml($package_name, $when = "post-deinstall") {
 	$static_output .= gettext("done.") . "\n";
 	update_output_window($static_output);
 	write_config("Removed {$package_name} package.\n");
+
+	/* remove package entry from /etc/syslog.conf if needed */
+	/* this must to be done after removing the entries from config.xml */
```
phil-davis: You can remove "to" from this sentence :)

doktornotor: I couldn't decide between *must* and *needs to* so I made a compromise :stuck_out_tongue_winking_eye:

PR #1943, opened 2015-10-03
Patch 4

heper:
attempt to address https://forum.pfsense.org/index.php?topic=100232.0

alias_edit


phil-davis:
That looks good :+1:

PR #1945, opened 2015-10-04
Fix #5252 Preserve alias when editing an existing rule

phil-davis:
and default src and dst to any when creating a new rule.


pfSense-git-updates:

doktornotor:
1/ The CLA script is clearly still broken.
2/ The "we cannot be bothered with braces in singleline if/else is a disaster in itself.

sbeaver-netgate:
Merged. Thank you!

ariyako:
nice!

PR #1947, opened 2015-10-06
Show the delete button only if there is more than 1 row

phil-davis:
This is not properly working piece of code!
@sbeaver-netgate 
If the alias edit screen only displays the "Delete" button when there is more than 1 row, then the user will never be able to click "Delete" on the last row, and thus will never need to see the popup "You may not delete the last row!"
Also then, if the user chooses a different alias type (e.g. a URL Table type that should have only 1 row), but already has multiple rows on their screen, the "Delete" buttons will stay and they can delete the extra rows that are no longer wanted.
This code fragment does the test. But leaving it inside typechange() means it is only activated if the user selects a different alias type. I am not a JavaScript events and BootStrap guy - how do I make this show/hide test happen whenever the "Add" or "Delete" row button is pressed?

### DIFF
--- a/src/usr/local/www/firewall_aliases_edit.php
+++ b/src/usr/local/www/firewall_aliases_edit.php
@@ -763,13 +763,18 @@ events.push(function(){
 		var labelstr = <?php echo json_encode($label_str); ?>;
 		$('.repeatable:first').find('label').text(labelstr[tab]);
 
-		// The add button and delete buttons must not show on  URL Table IP or URL table ports
+		// The delete button at the end of each row is only shown when there is more than 1 row.
+		if($('[id^=deleterow]').length > 1) {
+			$('[id^=deleterow]').show();
+		} else {
+			$('[id^=deleterow]').hide();
+		}
+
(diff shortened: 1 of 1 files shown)


phil-davis:
I found pfSenseHelpers.js - seems the proper place to try and do this.

sbeaver-netgate:
In this file only, I would change the name of the add and delete buttons to something else. e.g.: mydelrow . $counter or something.

Then you could put the button selects ( $('[id^=mydelro]'). . . ) in a script on the page and add any code you like in there.

I suppose something like:

<script>
//<![CDATA[
events.push(function(){
// These are action buttons, not submit buttons
$('[id^=myaddrow]').prop('type','button');
$('[id^=mydelete]').prop('type','button');

// on click . .
$('[id^=myaddrow]').click(function() {
	add_row();
});

$('[id^=mydelete]').click(function(event) {
	if($('.repeatable').length > 1) {
		if((typeof retainhelp) == "undefined")
			moveHelpText(event.target.id);

		delete_row(event.target.id);
	}
	else
		alert('You may not delete the last row!');
});

PR #1948, opened 2015-10-06
Hide the delete row button when there is only 1 row remaining

phil-davis:
When deleting the 2nd-last row this happily causes the Delete button for the last remaining row to be hidden - that is the first thing I wanted to achieve. It gets rid of the button and thus the need for the message "You may not delete the last row!"to ever pop up.

When adding a row, I want to always show the delete row buttons (there is always at least 1 row already, so whenever "Add" is pressed then there will be at least 2 rows). I thought line 297 would do it easily - but no joy when testing it.

@sbeaver-netgate - what have I missed here?

If we do this, then "Delete" buttons at the end of rows should come and go automagically as required when rows are added and removed.

### DIFF
--- a/src/usr/local/www/jquery/pfSenseHelpers.js
+++ b/src/usr/local/www/jquery/pfSenseHelpers.js
@@ -292,6 +292,9 @@ function add_row() {
 	$('[id^=address]').autocomplete({
 		source: addressarray
 	});
+
+	// Show any deleterow buttons.
+	$('[id^=deleterow]').show();
 }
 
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
I’ll take a look.

Steve
--

phil-davis:
Not sure why, but moving the code:
$('[id^=deleterow]').show();
to the top of function add_row() makes it work.
https://github.com/pfsense/pfsense/pull/1966 works for me.

PR #1949, opened 2015-10-06
Fix #5261 last time I hope

phil-davis:
The previous version was considering "Single host or alias" to be "Network" because the !is_specialnet() test came first - and that matched both 'network' and 'single' types, which are both not a special net.
If have re-ordered the tests so that it is written in a positive order. I think that makes it easier to read.
1) If is_specialnet then pass the special name straight through to $ruleType.
2) Next see if it is a single IP address or alias.
3) The rest must (should, had better) be 'network'.

I tried editing all sorts of aliases with various types and this works for me.


sbeaver-netgate:
Thanks Phil

PR #1951, opened 2015-10-07
Another interfaces.php text change to IPv6

phil-davis:
This one was missed in recent fixes around this area.


phil-davis:
@sbeaver-netgate This should be any easy one to review and commit in 5 seconds :)

phil-davis:
Already committed by @sbeaver-netgate separately.

PR #1952, opened 2015-10-07
Add brackets around self rule type #5283

phil-davis:
It needs the brackets here - like in line 1271. The "(self)" with brackets is what is stored in config.xml in 2.2.* and needs to be the same here in 2.3. Otherwise other validation messages are spat out...


sbeaver-netgate:
Thanks. THought I copied that from 2.2.x but must have missed the parenthesis.

sbeaver-netgate:
Done. Thanks.

phil-davis:
Done by @sbeaver-netgate in commit https://github.com/pfsense/pfsense/commit/bda7ff6da0cfadbacf5f564fc56a632cc2fc8f99

PR #1953, opened 2015-10-08
Wording of alias_info_popup tip RELENG_2_2

phil-davis:
I noticed this while comparing alias popup behavior between 2.2.5-DEVELOPMENT and 2.3
Might as well fix the grammar here for 2.2.5
This tip does not exist in 2.3 because the popup works more nicely there and so this text is not needed.
Therefore this change does not need to be ported forward to master.


cbuechler:
merged, thanks!

PR #1955, opened 2015-10-09
Fix #5289 Available Packages version display

phil-davis:
I left the "View changelog" code that was commented out, for future reference I guess when that data becomes available.
$index['version'] is still happily available in the $index array - so it just needed to be uncommented.


sbeaver-netgate:
Merged.

Thank you for the fix.

PR #1960, opened 2015-10-11
proposed adjustment firewall_rules_edit.php

heper:
https://forum.pfsense.org/index.php?topic=100695.msg561326#msg561326


sbeaver-netgate:
Merged as submitted.

Thanks!

PR #1965, opened 2015-10-15
Tidy some text in package install code

phil-davis:
1) The "Beginning package removal" "Beginning package installation"... heading above the text box that displays the progress text output is never updated along the way, so it always says "Beginning...". It seems easy to just remove the word "Beginning" - the detailed progress is in the text box below, and the "please wait"/"completed"/"failed" appears below the text box already.
2) Replace remaining occurrences of add/delete with install/remove to make the words used on the UI consistent.
3) Make a custom sentence for the "reinstallall" case. (Note: I am just fixing text UI stuff with this PR. After pressing "Confirm" to reinstallall, actually nothing happened. So that needs looking at also.)


sbeaver-netgate:
Merged.

Thanks!

PR #1966, opened 2015-10-15
Hide the delete row button when there is only 1 row remaining

phil-davis:
When deleting the 2nd-last row this causes the Delete button for the last remaining row to be hidden. Having no delete button for a single-row alias means there is no temptation for the user to press it, and thus no need for them to get the message 'You may not delete the last row!'.
When adding the 2nd row the Delete buttons will appear again. In this case the user can delete any of the rows, leaving 1 or more remaining.

### DIFF
--- a/src/usr/local/www/jquery/pfSenseHelpers.js
+++ b/src/usr/local/www/jquery/pfSenseHelpers.js
@@ -243,6 +243,9 @@ function delete_row(row) {
 }
 
 function add_row() {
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
Phil,

I sort of Cheery-picked this. I implemented the check for < 2 rows in a function allowing it to be called from pfSenseHelpers, and from individual pages on page load.

Thanks for the help.

phil-davis:
Yep, works.

PR #1967, opened 2015-10-15
Do not show the delete row button if there is only 1 row

phil-davis:
The initial code and when switching between types (e.g. from URL table type to some other type) runs this typechange() function. Whenever we are displaying an alias that has only 1 row of data, we do not want to show the Delete row button.
That button is not usable when there is only 1 row of data, so there is not much point displaying it and tempting the user to press it.


sbeaver-netgate:
Phil,

I believe this has been accomplished the new Javascript in pfSenseHelpers.js. Would you confirm please?

phil-davis:
Yes, that works for me. Closing this, thanks.

PR #1969, opened 2015-10-16
Change INCLUDE lines to use full path instead of relative path.

bmeeks8:
Need to use the full path in INCLUDE_ONCE lines within the Bootstrap base Class files instead of a relative path so things work when the base Class files are used from PHP files in sub-directories underneath the www root directory.

### DIFF
--- a/src/usr/local/www/classes/Form.class.php
+++ b/src/usr/local/www/classes/Form.class.php
@@ -27,9 +27,9 @@
 	POSSIBILITY OF SUCH DAMAGE.
 */
 
-require_once('classes/Form/Element.class.php');
-require_once('classes/Form/Input.class.php');
-foreach (glob('classes/Form/*.class.php') as $file)
+require_once('/usr/local/www/classes/Form/Element.class.php');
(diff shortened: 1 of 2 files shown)


bmeeks8:
Without this change, packages such as Snort and Suricata that have the PHP files in a sub-directory under www root cannot use the Bootstrap PHP wrapper Classes because the INCLUDE_ONCE lines in those Class files will fail to find other needed class files due to referencing a relative path that is invalid within a sub-directory of www root.

rbgarga:
Maybe adding the classes directory to PHP include_path set? What do you think @sbeaver-netgate ?

bmeeks8:
I like Renato's suggestion.  Just add those directories to the base PHP include_path and remove the relative path from the include lines.  That's the cleanest way, I think.

sbeaver-netgate:
For sure. I have a few other organizational changes I’m planning to make next week and will certainly include this.

Steve
--

sbeaver-netgate:
PHP include path has been updated.

PR #1970, opened 2015-10-20
Point people to 'Clear Package Lock' if the reinstall of packages got…

doktornotor:
… stuck

Way too many threads on forums with people asking what's going on. The most recent one with user waiting for 9 days (!!!) - https://forum.pfsense.org/index.php?topic=101022.0


cbuechler:
merged (both), thanks!

doktornotor:
Thanks as well...

PR #1973, opened 2015-10-23
System Logs Consolidate Code and Add Advanced Filtering

N0YB:
diag_logs_type_system.php
Consolidate system logs code.
Split the columns in advanced mode view.
Add advanced mode filtering and raw mode simple filtering.

guiconfig.inc
word wrap non breakable lines to fit them within the system log page instead of expanding the page.

filter_log.inc
Add support for system logs advanced filtering.

Replace system diag log php files to use the consolidated php.

diag_logs_filter.php
Move raw mode simple filter to bottom (consistent with others)


pfSense-git-updates:

N0YB:
Had already signed the CLA about a year ago.  Thought it was supposed to be “lifetime”, so don’t know why being required to sign it again?  Went ahead and signed it again though.

rbgarga:
@N0YB When it started, CLA was valid for one year. It's now lifetime and you won't need to do it again. Thanks!

cbuechler:
This will have to go to master only. Too big of a change for a maintenance branch for something that isn't fixing any problems. 

Also, the license can't be stripped from any files, please leave it in place as-is.

N0YB:
I guess fixing problems is in the eye of the beholder.  I consider system logs with no search capability to be a broken feature.  This fixes that for many of the system logs.
 
The size of the change really is not as significant as it appears.  The code of the specific diag_log*.php files is rolled into a single parameterized file (diag_logs_type_system.php), but it is essentially the same code with addition of the same search code as is used in the firewall log through many version.
 
Don't really consider the license as having been stripped from the specific system log php files, as they are essentially removed and replaced with new files that are nothing more than the equivalent of a URL query string, and the original license of them followed the code to the diag_logs_type_system.php file.
 
Oh well, it's your project.  I'll just apply it to my installation when 2.2.5 is released (currently using it in my 2.2.4 installation).  Too bad others won't get the benefit.

PR #1980, opened 2015-10-24
Properly validate IPv4 for UPnP - RELENG_2_2

doktornotor:
Since https://redmine.pfsense.org/issues/1835 got exactly nowhere for the past 4 years, the input should be properly validated, instead of allowing people to configure nonfunctional/broken nonsense.

P.S. If you don't plan on fixing the above-linked feature for 2.3, let me know and I'll do the same for 2.3

### DIFF
--- a/usr/local/pkg/miniupnpd.inc
+++ b/usr/local/pkg/miniupnpd.inc
@@ -59,7 +59,7 @@
 			$ip_array[] = $ip;
 
 		/* validate ip */
-		if (!is_ipaddr($ip_array[0]))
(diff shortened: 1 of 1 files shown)


rbgarga:
I've enabled IPv6 option for 2.3 in 679c4ac73f8d5c5fe71a9edb3ccdb1f955d616cd

doktornotor:
OK, so this won't be needed there; good. :) BTW, the CHECK_PORTINUSE is something that'd be nice to have in 2.2.5 as well...

PR #1984, opened 2015-10-25
Fix multi-lingual save of NAT mode Redmine #5342 for RELENG_2_2

phil-davis:
and forum https://forum.pfsense.org/index.php?topic=101276.0

### DIFF
--- a/usr/local/www/firewall_nat_out.php
+++ b/usr/local/www/firewall_nat_out.php
@@ -78,7 +78,7 @@ if ($_POST['apply']) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


phil-davis:
This should be a simple and easy fix for RELENG_2_2. It makes this work in just the same way that this sort of thing is done in other places in the RELENG_2_2 code. The language-specific translation of the string is the thing that is passed as the $_POST variable value.
That can all be re-engineered in a different/better/more maintainable way in master with bootstrap. But there is little point in re-engineering for RELENG_2_2.

PR #1985, opened 2015-10-25
Fix multi-lingual save of NAT mode Redmine #5342 for master

phil-davis:


phil-davis:
Note: At the moment this is NOT broken in Master. That is because the "Save" button translation is not happening on the UI - switch to Portuguese and the button still says "Save" rather than "Salvar".

But if this change is made in Master, then the gettext("Save") does work and produces "Salvar" which then does not match the $_POST value which is the English "Save".

It seems that button text translation still needs to be sorted out in master. At that time it will have to be resolved if things like this put untranslated fixed text or translated text in $_POST when submitted for action. Actually to me it seems much more straightforward that the string submitted in $_POST should be a "code" indicating the thing to be done and that code should be language-neutral (could be an acronym, abbreviation or something entirely unrelated to what the text string happens to be in English). The displayed text in the UI (whether in English or another language) is something different and local to the UI only. It seems odd to me to pass around a language-specific string like that in this kind of context.

PR #1988, opened 2015-10-26
Remove bonus ) in pkg.php

phil-davis:
Reported in forum https://forum.pfsense.org/index.php?topic=101390.0


sbeaver-netgate:
Merged. Thanks

Steve
--

PR #1991, opened 2015-10-27
Add option select in combobox

marianamorim:
Add option select in the class Select.class.php

### DIFF
--- a/src/usr/local/www/classes/Form/Select.class.php
+++ b/src/usr/local/www/classes/Form/Select.class.php
@@ -33,7 +33,7 @@ class Form_Select extends Form_Input
 	protected $_values;
 	protected $_value;
 
-	public function __construct($name, $title, $value, array $values, $allowMultiple = false)
+	public function __construct($name, $title, $value, array $values, $allowMultiple = false, $optionDefault = true)
 	{
 		if ($allowMultiple)
 			$name .= '[]';
@@ -43,6 +43,7 @@ class Form_Select extends Form_Input
 		if ($allowMultiple)
 			$this->_attributes['multiple'] = 'multiple';
 
+		$this->option_default = $optionDefault;
 		$this->_value = $value;
 		$this->_values = $values;
 	}
@@ -52,6 +53,10 @@ class Form_Select extends Form_Input
 		$element = parent::_getInput();
 
 		$options = '';
+		if ($this->option_default) {
+			$options = "<option value=''>-- " . gettext('select') . " --</option>";
+		}
+
 		foreach ($this->_values as $value => $name)
 		{
 			// Things can get wierd if we have mixed types


pfSense-git-updates:

sbeaver-netgate:
Would you explain the purpose of this patch please? What would you like to do that the existing class cannot?

marianamorim:
Allow display the default value in fields as defined in the W3C of select tag

sbeaver-netgate:
Sorry to be stupid, but I don't understand. If we need a selector with options of red, green, and blue with red being the default option, we would write:

new Form-Select(
    'color',
    'Page color',
    '2'    // Make "Red" the default (selected) value,
    array('1' => 'Green', '2' => 'Red', '3' => 'Blue')
)->setHelp("Please choose your favorite color");

rbgarga:
@sbeaver-netgate If I understood correct, what @marianamorim is trying to do is a way to add a fixed text to be shown when none of the options are selected. Like a default text saying "please, select one of the items here". Is that correct @marianamorim?

marianamorim:
Yes, that's right.

sbeaver-netgate:
OK. Let me research that. Thanks.

Steve
--

sbeaver-netgate:
OK I get it. Sorry. The proposal would provide the ability to introduce an additional option with a value of "" and a display a string of "--select--". 

I think this might conflict with many existing pfSense select controls that already make use of the null value. For example selectors may use: $interfaces = array("" => "Any", "lan" => "LAN", "wan" => "WAN"); There are quite a few like this in pfSense.

It is also very easy to add the default text you suggest like this:

```php

    $interfaces = array("" => "Any", "lan" => "LAN", "wan" => "WAN");

    new Form-Select(
        'color',
        'Page color',
        ""    // Make "any" the default (selected) value,
        ["default" => "--select--"] + array('1' => 'Green', '2' => 'Red', '3' => 'Blue')
    )->setHelp("Please choose your favorite color");
```
See that I just added a default/--select-- array to my array of options.

Would this work for your needs do you think?

marianamorim:
Understood. It's a solution, but if the array have is empty, it'll reload this option with null value. It'll just show a default/ --select-- array without others values, it isn't smart. What do you think about it?

rbgarga:
@sbeaver-netgate can you check the last question?

sbeaver-netgate:
If this functionality is really useful in pfSense perhaps it would be possible to add it using the  'placeholder' mechanism. That way it would be compatible with the other input classes that display a message of this type. Form_Input has a "setPlaceholder()" method for example. This isn't technically a placeholder in the HTML sense, but it acts in the same way.

I found this helpful: http://stackoverflow.com/questions/5805059/how-do-i-make-a-placeholder-for-a-select-box

It would also be worth checking with Sjon Hortensius who designed the classes used in the GUI. Sjon usually has some very insightful input to ideas such as this.

https://github.com/SjonHortensius/pfsense

PR #1992, opened 2015-10-27
system_gateways font-awesome convert

heper:
https://forum.pfsense.org/index.php?topic=101343.15


sbeaver-netgate:
Merged. Thanks.

PR #2000, opened 2015-10-28
Fix old SNMP bindlan setting - Redmine #3883 RELENG_2_2

phil-davis:
Remove run-time code in services.inc that referenced "bindlan".
Provide upgrade code to convert bindlan to bindip="lan" if needed.
Update current config version


phil-davis:
If this seems a good thing for RELENG_2_2 2.2.5, then it will also need to be ported forward to master.

rbgarga:
We are on config version 12.7 on master now, then introduce a upgrade config on RELENG_2_2 would make us to move all upgrade versions on master and fit this one together.

Since it's not a 2.2.5 regression, we should put the fix and upgrade config only on 2.3 and set ticket target to 2.3 too.

Do you want to re-submit it? Otherwise I will work on the fix for 2.3

phil-davis:
It probably only effects 1 or 2 people in the universe anyway, so you are correct, at this stage of the release process there is not much point in applying it to RELENG_2_2 and then shuffling all the upgrade_config numbers in master (which would need a bit of thought to make sure not to miss out or double up for 2.3-ALPHA installs that have already used the numbers up to 12.7)
I will integrate and submit against master...

phil-davis:
See https://github.com/pfsense/pfsense/pull/2088

PR #2001, opened 2015-10-28
fix padding in select boxes

heper:
Todo #5228
found this solution ... seems a lot less time consuming to fix all option-padding without having to add classes to the option fields on every page. 
i've clicked around the GUI a bit and it seems to behave pretty nicely

### DIFF
--- a/src/usr/local/www/bootstrap/css/pfSense.css
+++ b/src/usr/local/www/bootstrap/css/pfSense.css
@@ -104,6 +104,9 @@ tr.disabled th {
     padding: 0 12px;
     margin-bottom: 5px;
 }
+.form-control option {
+    padding: inherit;
+}
 
 #login .form-control {
 	height: 34px;


sbeaver-netgate:
Testing this idea looks promising although I note there is still a small misalignment between the selected option and the rest of the list.Might be close enough though. Opinions?

heper:
i notice a small misalignment in firefox 38.2 (the selectbox is 3px bigger then the option itself, on my screen)
in chromium 44, it seems to be aligned perfectly.

i don't know which of these two browser are misbehaving ;)

sbeaver-netgate:
Ah. The joys of web development.

I think Phil Davis originally suggested the need to address this, so let’s see what he thinks. Phil?

Steve
--

sbeaver-netgate:
I also wonder whether this is really a fix for the problem as reported. I think the issues was reported as an unneeded space before the selected value. This fix basically adds an unneeded space before everything else so that it signs up. Not quite the same :)

Steve
--

heper:
formcontrol

well, remove the padding is also an option (as i stated on redmin). but imho, it looks worse overall

sbeaver-netgate:
Agreed. My preference would be to leave it alone. I find it better to let things default when possible. If there are people who feel it would be better with that leading space removed, however, I don’t mind trying to do that. 

Steve
--

phil-davis:
As long as the drop-down list entries line up then I don't care too much about having the padding at the front. Whatever is the default and accepted "modern" UI :)
In Chrome at the moment it is lining up, with the padding at the front of the current value and the dropdown list values.

sbeaver-netgate:
OK good. Thank you. I have been experimenting tonight and I think I have a compromise that will work. I’ok do some more testing Thursday morning (GMT -4) and then post something.

Thanks for everyone’s input. I really appreciate all the eyes on the project.

Steve
--

sbeaver-netgate:
Changing the select padding from 12 to 6 and adding the inherit CSS for the options seems to produce the best looking display. Please let me know what you think.

Steve
--

heper:
fixed by commit 467669b
closing this.

PR #2012, opened 2015-11-02
Fix display of reject rules RELENG_2_2

phil-davis:
on firewall log normal and dynamic views.
Firewall log entries that are for "reject" rules currently show the "block" red cross icon. That is because the data in filter.log just has the word "block", and so the downstream code that parses and displays the filter log on the UI does not get an easy chance to know if it is a "block" or "reject" rule.
The "block" vs "reject" information is in the config - we could look that up using the rule tracker number.  But that requires indexing through the config rules to match the tracker number.
There is already function buffer_rules_load(). Currently it is used to get the rule descriptions. Might as well call it all the time and then use the global array that it creates to work out if a block rule is really "block" or "reject". A "reject" rule starts with "block return".
There is already code in find_action_image() that returns the correct image for the "reject" case. It was just waiting to be passed "reject" instead of "block".
I noticed this in 2.3-ALPHA and when looking on 2.2.4 and 2.2.5 systems to see how it worked. That made me realise that it did not work on 2.2.* also.
Note: There is another bug with the firewall log dynamic view - new rows that are dynamically added do not get the rule tracker number added in. So they do not display the rule when you click the icon, and they also do not choose the "reject" icon. I will have a look at that separately, because it is a whole different issue with the JavaScript and how it passes stuff around.

### DIFF
--- a/etc/inc/filter_log.inc
+++ b/etc/inc/filter_log.inc
@@ -128,7 +128,7 @@ function in_arrayi($needle, $haystack) {
 }
 
 function parse_filter_line($line) {
-	global $config, $g;
+	global $config, $g, $buffer_rules_normal;
 
 	$flent = array();
 	$log_split = "";
@@ -149,6 +149,7 @@ function parse_filter_line($line) {
 	$flent['interface']  = convert_real_interface_to_friendly_descr($flent['realint']);
 	$flent['reason'] = $rule_data[$field++];
 	$flent['act'] = $rule_data[$field++];
+	$flent['acttype'] = $flent['act'];
 	$flent['direction'] = $rule_data[$field++];
 	$flent['version'] = $rule_data[$field++];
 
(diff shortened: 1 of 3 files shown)


phil-davis:
See Redmine https://redmine.pfsense.org/issues/5363

phil-davis:
Fixing the display of dynamically added rules on the firewall log dynamic display turned out to be not os difficult. So I have added it here as the 2nd commit.
Now log entries of the various types of new rule (pass/block/reject) can appear dynamically and:
1) The appropriate icon is shown (the "reject" icon is now displayed for "reject" rules)
2) Clicking on the icon displays a popup with the rule (previously that did not work)

jim-p:
The rule description display option is off by default because it doesn't scale well, and this would likely have similar problems. With large rule sets and logs set to display many lines, it would increase processing time dramatically. That isn't something I'd want to enable unconditionally. If it were an option, like the rule descriptions, it may be OK. Also it's not a change we'd want to bring into 2.2.x, we're trying to get a release out in the next couple days. Might be fine in 2.3 but it would need some performance testing on low-end platforms.

Changes to fix up dynamic logging unrelated to the block/reject part would be OK.

phil-davis:
See https://github.com/pfsense/pfsense/pull/2014 for just the changes to fix the firewall log dynamic view for rows that are dynamically added.

PR #2014, opened 2015-11-02
Fix firewall log dynamic rule lookup RELENG_2_2

phil-davis:
for rows that are dynamically added as time goes by.


phil-davis:
I see that the 2.2.5 release went well and that new bits of code for RELENG_2_2 are now being added. This is handy for those who might gitsync to RELENG_2_2 to pickup bits of code, even though there might or might not ever be a 2.2.6.
Maybe the few waiting little bits of stuff in PRs could also be checked and merged.

PR #2023, opened 2015-11-03
convert status_dhcp_leases to fa

heper:
haven't found a good icon to differentiate between "add static mapping" & "add WOL mapping"
used 2 different types of fa-plus-square for now. (until someone has a better idea)


sbeaver-netgate:
We need fonts-stupendous  :)

Merged.

Thanks.

Steve
--

PR #2024, opened 2015-11-03
convert status_dhcpv6_leases to fa

heper:
/*i don't have a working dhcpv6 setup to properly test, did basic testing by copy pasting some bogus data in config.xml*/
page needs some additional work: 
a) table-data is wrapped because it cannot fit inside the "main" container-class_div
perhaps one of the columns should be hidden and make it optional to show?

b) it's missing "edit static mapping" & "send wol packet"  (don't want to commit it, when unable to test)


sbeaver-netgate:
Merged. Thank you

PR #2025, opened 2015-11-04
Make DHCP deny action configurable

apilloud:
The default behavior of PFsense is to send a DHCP NAK to denied clients. This is problematic in situations where there are multiple DHCP servers on the same subnet serving different sets of clients. This patch adds a config option to allow denied clients to be ignored rather then explicitly rejected.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -776,20 +776,24 @@ EOPP;
 				}
 				$dhcpdconf .= "		allow members of \"" . str_replace(':', '', $mac) . "\";\n";
 			}
+			$deny_action = "deny";
+			if (isset($poolconf['nonak'])) {
+				$deny_action = "ignore";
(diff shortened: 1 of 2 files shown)


pfSense-git-updates:

apilloud:
CLA submitted by email. Patch updated with whitespace issues corrected.

rbgarga:
@jdillard Can you please check if CLA was received?

gonzopancho:
I have no idea where it might have been sent.  I haven't seen it.

apilloud:
@gonzopancho It is a Corporate agreement for Igneous Systems. I emailed it to the licensing email address on the form, but that bounced, so I mailed it to the address on McNeil Road in Austin, Texas.

apilloud:
Ping @rbgarga CLA is all set now.

PR #2026, opened 2015-11-05
DNS Forwarder Host override

phil-davis:


phil-davis:
There are more bits of inconsistent text. I will resubmit a PR with a bunch of consistency fixes for DNS Forwarder and Resolver pages.

PR #2027, opened 2015-11-05
sed inline style

heper:
@sbeaver-netgate 
https://forum.pfsense.org/index.php?topic=101865.0
seems ok, did a quick test.


sbeaver-netgate:
Another nice improvement. Thank you!

PR #2033, opened 2015-11-05
proposed fix for Bug #5375

heper:
value passed by checkbox was "yes" when checked, instead of "Block" / "Pass"


sbeaver-netgate:
Thank you!

PR #2038, opened 2015-11-08
services_dhcpv6 put back interface tabs

phil-davis:
https://forum.pfsense.org/index.php?topic=102063.0


sbeaver-netgate:
Sorry Phil, I was working on this at the same time as you and added a couple of other changes to format the tabs the same way we do on other pages. I have checked my changes against yours and committed them.

phil-davis:
Yep, good stuff.

PR #2042, opened 2015-11-09
interfaces_pps_edit use latest GLOB from RELENG_2_2

phil-davis:
Use the latest cua* glob from the code in RELENG_2_2 that was "recently" modified to get this right.
Remove the debug version that looked for "tty" so that people can use it on real systems now.
Forum thread: https://forum.pfsense.org/index.php?topic=102111.0


phil-davis:
e.g. from https://github.com/pfsense/pfsense/commit/693c13cb74fedd1e3b57927b755099f112f0b986 and other earlier commits to that file which finalized the glob.

PR #2044, opened 2015-11-09

NOYB:


rbgarga:
It looks wrong, maybe wrong branch? @NOYB

NOYB:
Don't know what happened there.  Looks like maybe it went to master instead of 2_2.  But I was not on master.  I'll retry.

NOYB:
Don't know why but every time I push the local branch to origin it goes to master.
"This branch is 1049 commits ahead, 3470 commits behind pfsense:master."

PR #2046, opened 2015-11-09

NOYB:


cbuechler:
I don't see where this has poor performance. What specifically are you seeing poor performance with there? That was added specifically to be faster, though I don't see any difference either way in the pfctl times. 

Regardless, this isn't something we're going to merge into RELENG_2_2 at this point, please submit against master with specifics as to why it's preferable/what is slow.

NOYB:
"I don't see where this has poor performance."
Two places.  Firewall log display with descriptions, also clicking the action icon to reveal the rule that triggered the action.
 
"What specifically are you seeing poor performance with there?"
	pfctl -vvPnf /tmp/rules.debug is slow due to parsing large table sources such as bogonsv6 and alias urltables.
	pfctl -vvPsr is very fast but is only used in place of the other if /tmp/rules.debug file does not exist.
 
"That was added specifically to be faster, ..."
It may be faster with some specific pristine /tmp/rules.debug file.  But add a fairly large real world table source or two for it to parse and it is suddenly much slower.
 
"...though I don't see any difference either way in the pfctl times."
Then the performance test setup scenario, dataset, etc. being used is insufficient to reveal the performance difference.

PR #2047, opened 2015-11-09
Partial fix Bug #5389

heper:
system_groupmanager seems OK

system_usermanager:
-added the fa-trash icon
-can't seem to get the onclick submit the form

### DIFF
--- a/src/usr/local/www/system_usermanager.php
+++ b/src/usr/local/www/system_usermanager.php
@@ -268,9 +268,11 @@ if ($_POST['save']) {
 		input_errors2Ajax($input_errors);
 		exit;
 	}
-
+	echo "privid: ";
+	echo $_POST['privid'];
 	if (!$input_errors) {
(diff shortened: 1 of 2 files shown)


src/usr/local/www/system_usermanager.php
```diff
@@ -893,6 +896,15 @@ function moveOptions(From, To)	{
 			$('form').submit();
 		}
 	});
+	$('[id^=delprivid]').click(function(event) {
+		if(confirm(event.target.title)) {
+			$('#privid).val(event.target.id.match(/\d+$/)[0]);
+			$('#userid').val('<?=$id;?>');
```
sbeaver-netgate: Missing a closing ' here I think. Should be:

$('#privid').val(event.target.id.match(/\d+$/)[0]);
              ^

heper:
@sbeaver-netgate seems to work now, thanks for pointing me towards the missing '

PR #2048, opened 2015-11-10
Add global so that unbound.conf gets written

phil-davis:
to the correct place based on $g setting.

### DIFF
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -119,6 +119,7 @@ function test_unbound_config($unboundcfg, &$output) {
 
 
 function unbound_generate_config() {
+	global $g;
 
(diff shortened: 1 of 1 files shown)


phil-davis:
The stuff generated by unbound_statistics() goes into unbound.conf
It had got separated out and was going nowhere. This 2nd commit puts the behavior back the way it was yesterday.

Note: There is a separate issue about unbound_statistics() because the UI has no place to actually select the statistics settings. There is some upgrade code that looks like it was converting statistics settings from the old Unbound package into custom commands in the integrated Unbound. So it looks like the unbound_statistics() function can/should be removed???

doktornotor:
Frankly I'd want to statistics checkbox back. The thing is fairly noisy and cannot be controlled in any way now.

PR #2049, opened 2015-11-10
Unbound advanced Incoming TCP Buffers validation RELENG_2_2

phil-davis:
There is a variable cut-paste error here. Actually it never makes any difference in practice because outgoing_num_tcp and incoming_num_tcp are both set, so the logic comes out the same.
I noticed this while reviewing stuff in master for 2.3-ALPHA - might as well back-port stuff like this so that nobody who notices will worry about it in RELENG_2_2.


phil-davis:
The PR for this in master is https://github.com/pfsense/pfsense/pull/2050

PR #2050, opened 2015-11-10
Unbound UI Review

phil-davis:
Few things found when looking at the Unbound UI code

### DIFF
--- a/src/usr/local/www/services_unbound_advanced.php
+++ b/src/usr/local/www/services_unbound_advanced.php
@@ -129,7 +129,7 @@ if ($_POST) {
 		if (isset($_POST['outgoing_num_tcp']) && !in_array($_POST['outgoing_num_tcp'], array('0', '10', '20', '30', '40', '50'), true)) {
 			$input_errors[] = "A valid value must be specified for Outgoing TCP Buffers.";
 		}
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_unbound_advanced.php
```diff
@@ -369,7 +369,7 @@
 	'disable_auto_added_access_control',
 	'Disable auto-added access control',
 	'disable the automatically-added access control entries',
-	$pconfig['hdisable_auto_added_access_control']
+	$pconfig['disable_auto_added_access_control']
```
phil-davis: Need to have a rant here. What in the *** happened to code review/test in the original repo where this bootstrap stuff was done??? There are bugs like this all through the code - hundreds have been got out so far and who knows how many more there are to be found.
2.2.5 is a good piece of software, and 2.3 will be a PoC unless every line of code is re-checked to find all the dumb bugs introduced by the bootstrap rush.
End of rant

sbeaver-netgate: Thanks Phil,

Rant very much understood. I have learned that “converted” in the context of this project often meant “Looks pretty but completely untested”. You know the issues as well as (perhaps better) than me. My goal is to have the functionality restored to all pages by the end of this week, and then to start a month-long code review/QC process. I can’t tell you how much I value your help in this. It’s priceless.

PR #2051, opened 2015-11-10
Delete Id comment lines from etc files

phil-davis:
More crud cleaned out.


phil-davis:
There are plenty of file in /etc that have old or no copyright stuff. I guess ESF will get around to standardizing all that some day.

PR #2059, opened 2015-11-11
No XMLRPC Sync checkbox layout

phil-davis:
If null is passed for the text to go after the checkbox, then the checkbox does not line up nicely on the page with the other items above and below it. If some text (or even the empty string) is passed then the lineup is all good.
In this case it seems reasonable to put some text here, like is done higher up for "Disable" and "No RDR" checkboxes.


phil-davis:
Got mixed in with the other one - I will resubmit the now-correct diff.

PR #2061, opened 2015-11-11
Limit alias info popup size #5415 RELENG_2_2

phil-davis:
This is code for RELENG_2_2 pfSense 2.2.5 that will limit the number of rows in the alias info popup to 100.


phil-davis:
Note: I expect that this works, but it is a quick demo of limiting the length of alias entries displayed in the popup - all the various types of long alias need to be tested. The final limit would need to be decided (50 suggested in the issue in Redmine). And of course necessary code engineered for however this works in 2.3

cbuechler:
merged, thanks!

PR #2062, opened 2015-11-13
proposed fix Bug #5433

heper:
set 'gui log entries' to 3000.
=> no horizontal scroll on my 2.3 system. (best to test on multiple systems/browsers)


heper:
this is based on https://github.com/pfsense/pfsense/pull/2002

PR #2068, opened 2015-11-15
gateway routing support for interface based gateway assignment

PiBa-NL:
Fixes #1847, and was checked with that provider to work properly for the IPv4 part. In test machine IPv6 also seems to insert the proper routes.

### DIFF
--- a/etc/inc/system.inc
+++ b/etc/inc/system.inc
@@ -540,6 +540,24 @@ function system_routing_configure($interface = "") {
 		}
 	}
 
+	$gateways_arr = return_gateways_array(false, true);
+	foreach($gateways_arr as $gateway) {
+		// setup static interface routes for nonlocal gateways
+		if (isset($gateway["nonlocalgateway"])) {
+			$srgatewayip = $gateway['gateway'];
+			$srinterfacegw = $gateway['interface'];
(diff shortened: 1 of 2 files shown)


rbgarga:
I would like to hear @cbuechler and/or @jim-p opinion about this

jim-p:
The actual redmine entry for this is #972 -- Not sure if there is a more elegant way to handle this. Having an extra option saves us from having to completely eliminate the other input validation, though it introduces more complexity to the gateway handling code. Also isn't clear if we should provide a means to set it on the small gateway adding form on interfaces.php. Might be overkill there.

PiBa-NL:
Ok, so whats the verdict? This can be merged like it is currently? Should i create the same request for 2.3 ? If it needs changing it rather know before spending the time on the 2.3 equivalent.

cbuechler:
I'd rather have it much more hidden than that since it's very rarely necessary. Though it also has to be specified on a per-gateway basis, so it would have to be configurable on the individual gateways. No need for it on the interfaces.php widget, lots of other advanced options that aren't exposed (and the input validation check wouldn't apply there anyway IIRC, at least if the interface wasn't yet configured). 

@PiBa-NL Put it under the advanced options so it's hidden by default, and submit against master, and we'll get it merged. Thanks!

We're not adding features to RELENG_2_2, so closing this.

PR #2069, opened 2015-11-16
Add missing stratum parameter

jean-m-cyr:
Add missing stratum parameter  in ntpd.conf when specified on Serial GPS
page.


jean-m-cyr:
Yes, I have signed the CLA... though it doesn't seem to be flagged here!

PR #2081, opened 2015-11-18
Fix tooltip text for associated filter rules

phil-davis:
1) Add a port forward with associated filter rule
2) Go to firewall_rules and hover over the icon on the left of the associated rule.
It says "traffic is ed".

The problem is that, for whatever reason, an associated rule does not have ['type'] specified, even in config.xml. It is that way in 2.2.5, so not a new thing or regression. I guess that the back-end rule generation works by assuming that a rule in config.xml that does not have 'type' defined should be a pass rule.

So we can't write code anywhere that assumes that every rule has a defined 'type'.

This was different fixed text in 2.2.5, so there was no problem there.

In fixing this I also made the text explicit for each type. This will help with the gettext translation. The way it did it before, putting "ed" on the end, would not have worked for translation.

### DIFF
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -361,19 +361,24 @@ for ($i = 0; isset($a_filter[$i]); $i++):
 							<input type="checkbox" id="frc<?=$nrules;?>" onClick="fr_toggle(<?=$nrules;?>)" name="rule[]" value="<?=$i;?>"/>
 						</td>
 
-						<td title="<?=gettext("traffic is ").$filterent['type']."ed"?>">
-
 	<?php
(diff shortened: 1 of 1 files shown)


rbgarga:
Could you please rebase your fork @phil-davis? Thanks!

phil-davis:
@rbgarga Somehow this was already committed by @sbeaver-netgate yesterday https://github.com/pfsense/pfsense/commit/d08aeac11a9be8a89a85341515063e98eee1943f
and all the changes here are correctly in the current master, plus the recent font-aewsome changes on top of that. Somehow the way Steve did it did not mark this PR as merged.
So do we just close this?

rbgarga:
It happens sometimes, closing this is OK. Thank you!

PR #2083, opened 2015-11-18
Make block bogon rule display consistent - RELENG_2_2

phil-davis:
The block private networks and anti-lockout rules display in the firewall rules display with the schedule column blank. But the block bogon networks has an asterisk. That looks odd. I noticed it just now when looking at firewall rules in 2.3.
Might as well fix it here in RELENG_2_2.


phil-davis:
Pull request on master is https://github.com/pfsense/pfsense/pull/2084

PR #2085, opened 2015-11-18
added support flash messages

brunostein:
Was added support the flash messages. It's util to show messages for the users when the page is redirected.


rbgarga:
Assigning to @sbeaver-netgate

PR #2090, opened 2015-11-19
add attribute 'id' for the Form_Section

brunostein:
Added the attribute 'id' for the Form_Section class.


sbeaver-netgate:
The element id and name are currently set from the first argument to Form_*  Why would we need to specify an ID that differs from the name?

PR #2094, opened 2015-11-20
If webguicss is not yet specified then default to pfSense

phil-davis:
On systems where webguicss (the theme) was not yet set in the config, the selection here was defaulting to other than pfSense.
Now that pfSense-dark has come along, it happened to be first in the list. When I changed something else in General Setup today, my theme went to pfSense-dark because I did not notice the (accidentally) selected value that was in the field.

I also added a check for if the current webguicss is not a valid one - e.g. if a theme is selected in the config that is no longer on the system. That should help future-proof it so it will default back to pfSense.css in that case also.

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -467,6 +467,10 @@ foreach ($css as $file) {
 	}
 }
 
+if (($pconfig['webguicss'] == "") || (!isset($csslist[$pconfig['webguicss']]))) {
+	$pconfig['webguicss'] = "pfSense.css";
+}
+
 $section = new Form_Section('Web configurator theme');
 
 $section->addInput(new Form_Select(


src/usr/local/www/system.php
```diff
@@ -467,6 +467,10 @@ function is_timezone($elt) {
 	}
 }
 
+if (($pconfig['webguicss'] == "") || (!isset($csslist[$pconfig['webguicss']]))) {
```
rbgarga: It would be better to check using isset() here

phil-davis: I was thinking that it was always set up above even if the $config value did not exist.
You are right - isset() will still return false here.
@sbeaver-netgate has committed this already https://github.com/pfsense/pfsense/commit/c040eeda3a9710869caf276a564207866497ee67
However he does that is different to you and the PR does not get closed.
I can submit another PR to make it better...

sbeaver-netgate: Please do :)

phil-davis:
Follow-on PR is https://github.com/pfsense/pfsense/pull/2096

PR #2095, opened 2015-11-20
Hide Firewall Log Filter Form

NOYB:
Status - System Logs - Firewall - Filter Form Hide Feature
 Form hidden by default, shown when filtering/selected.


sbeaver-netgate:
Incorporated in later PR and after file name changes

PR #2099, opened 2015-11-21
Add class autoloading to the GUI system

dougwollison:
I think it would be convenient to add use of class autoloading to the GUI scripts; this way any GUI-related class like the form elements can simply be called as needed without having to load all the classes needed.

This could also be done with a separate autoloader for the classes in etc/inc but because most of those files include multiple class (and function) definitions it'd probably require splitting them up into their own files which may be more trouble than it's worth.


sbeaver-netgate:
Thanks Doug. I'd like to take a couple of days to test it before I merge it. It might be particularly useful for package maintainers.

dougwollison:
No problem; I only briefly tested it but nothing seemed to blow up.

PR #2104, opened 2015-11-22
Remove layer7 components divert

phil-davis:
The setting of the 'divert' array key was only done by layer7, so these downstream bits of code are also now dead.

What happens to rules in old configs that had layer7 selected? Now that there is no code to react to any layer7 setting, those rules are probably going to apply whatever other settings they had, without the diversion to the layer7 processing? Will that make people's rulesets become automagically/unexpectedly more permissive? Should there be code in upgrade_config.inc to remove (or disable) any rules with layer7 specified?

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2910,10 +2910,6 @@ function filter_generate_user_rule($rule) {
 					$aline['flags'] .= "/" . $rule['max-src-conn-rates'] . ", overload <virusprot> flush global ";
 				}
 
-				if (!empty($aline['divert'])) {
-					$aline['flags'] .= "max-packets 8 ";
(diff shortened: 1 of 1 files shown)


cbuechler:
@phil-davis yeah I'd left that ticket there to finish some clean up. With 2.2.x systems, it'd pass a lot more post-upgrade because layer7 would /dev/null everything. :) Now 2.3 has config upgrade code to unset all of that from the config, and file a notice for those who had that in their config so users are definitely aware.

phil-davis:
Yes, I noticed your addition of config upgrade code. Once users are aware, then it is up to them to re-evaluate their rule set and decide what to do with traffic that they previously passed to layer7 for evaluation.

PR #2106, opened 2015-11-23
GitSync Diff Only Option

NOYB:
--diff
 SSD Friendlier


rbgarga:
Not adding more unnecessary changes to 2.2.x

PR #2107, opened 2015-11-23
GitSync Diff Only Option

NOYB:
--diff
 SSD Friendlier


rbgarga:
I like the idea but I don't think it's necessary to create a .inc file that will be only used by one consumer. Could you please revert this part of the change?

NOYB:
How can the function redefinition that prevents running gitsync again in the same dev shell session (command recall) be resolved?  That's why I moved them to an include file and used require once.
Is there a better way?  Thanks.

src/etc/phpshellsessions/gitsync
```diff
@@ -1,4 +1,4 @@
-/*   cvs_sync
+<?/*   cvs_sync
```
NewEraCracker: Is this change intended?

NOYB: No.

NewEraCracker:
I think this would be a good addition for 2.3.2.
Please include it. Thanks.

cbuechler:
@NOYB could you please re-submit as a single commit, it's cleaner and easier to cherry-pick that way. The end result here looks fine, I can get that merged promptly.

PR #2110, opened 2015-11-24
LT2P breadcrumbs

phil-davis:
The repetition of LT2P here seemed odd. To me it seems better that the last level matches the tab name below, like it does if you click the "Users" tab.

### DIFF
--- a/src/usr/local/www/vpn_l2tp.php
+++ b/src/usr/local/www/vpn_l2tp.php
@@ -60,7 +60,7 @@
 ##|*MATCH=vpn_l2tp.php*
 ##|-PRIV
 
-$pgtitle = array(gettext("VPN"), gettext("L2TP"), gettext("L2TP"));
+$pgtitle = array(gettext("VPN"), gettext("L2TP"), gettext("Configuration"));
 $shortcut_section = "l2tps";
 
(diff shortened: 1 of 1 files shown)


phil-davis:
But maybe it should have only 2 levels shown?
e.g. in many other places (Services, DNS Resolver is an example) when you select Services->DNS Resolver it goes to the first of multiple tabs (e.g. the General settings tab). The breadcrumbs does not include "General settings". But when you click on one of the other tabs the breadcrumbs show all levels, e.g. Services / DNS Resolver / Advanced.

What is the convention we want for this, when displaying the first (left) tab of a multi-tab thing?

sbeaver-netgate:
It seems better that we show all levels (adding the “General settings” in your example). One of my tasks for today is to review the page/file names and to propose some changes. e.g.: When you select Status-> IPSec, you are sent to a page called diag_ipsec.php Why isn’t that status_ipsec.php?

Seems like I could add another column to that spreadsheet to propose the breadcrumb text.

Steve
--

phil-davis:
I made issue https://redmine.pfsense.org/issues/5527 so it can be discussed and then final decided structure implemented consistently.

PR #2111, opened 2015-11-24
Status IPsec inconsistent Logs tab

phil-davis:
For some reason Status->IPsec has a "Logs" tab in addition to the
"Overview", "Leases", "SAD" and "SPD" tabs. The "Logs" tab takes you off
to the logs screens, and so when you click the tab all the other tabs
are gone and are replaced by all the "Logs" interface tabs.
That seems odd and inconsistent. Other Status displays do not have this
(e.g. Status->DHCP, Status->Load Balancer, Status->OpenVPN). With all
these, direct access to the associated log page is by the shortcut icon.
Status->IPsec also has the shortcut icon to go to the logs, and it works
fine.

So I am suggesting to remove the "Logs" tab here.

### DIFF
--- a/src/usr/local/www/diag_ipsec.php
+++ b/src/usr/local/www/diag_ipsec.php
@@ -124,7 +124,6 @@ $tab_array[] = array(gettext("Overview"), true, "diag_ipsec.php");
 $tab_array[] = array(gettext("Leases"), false, "diag_ipsec_leases.php");
 $tab_array[] = array(gettext("SAD"), false, "diag_ipsec_sad.php");
 $tab_array[] = array(gettext("SPD"), false, "diag_ipsec_spd.php");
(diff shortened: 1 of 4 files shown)


jim-p:
Sounds good to me, that log tab pre-dates the shortcuts. It was handy in the past but no longer necessary.

phil-davis:
Note: There is also the piece of history that the underlying files here are called diags_ipsec*.php
They could be renamed to status_ipsec*.php and all references to them be changed.
Does anyone care if that is done or not?

jim-p:
@sbeaver-netgate is working on that -- he's coming up with a list of incorrectly named pages and corrections. It will be nice to have all of that cleared up.

phil-davis:
Good - I will leave the file names alone then. This change here can go in first, then he can rename stuff later.

PR #2113, opened 2015-11-25
Status - System Logs - Firewall - Descriptions Performance

NOYB:
pfctl -vvPnf {$g['tmp_path']}/rules.debug performance suffers noticeably due to parsing table sources such as bogonsv6 and alias urltables.
grep the {$g['tmp_path']}/rules.debug file to exclude the "persist file" lines from pfctl input.


loos-br:
Committed a slightly different fix.

Thanks!

NOYB:
That's what I submitted against RELENG 2.2.5.
https://github.com/pfsense/pfsense/pull/2046

Also just to be sure you are aware.  JIMP expressed some concern about not using the rules.debug file as input.  I never saw any issues with it, though my system certainly doesn't encompass all use cases.
https://forum.pfsense.org/index.php?topic=102107.0

Thanks

PR #2116, opened 2015-11-25
Remove double full stops priv.defs.inc

phil-davis:


jim-p:
That file is automatically generated. If there are problems they need to be fixed in the metadata in the files referenced. Also I'm making tons of edits to things here -- hold off a bit.
I'll fix these manually though, thanks! :-)

PR #2117, opened 2015-11-25
firewall_aliases Bug #5527

heper:


src/usr/local/www/firewall_aliases.php
```diff
@@ -195,11 +201,6 @@ function find_alias_reference($section, $field, $origname, &$is_alias_referenced
...
 	print_info_box_np(gettext("The alias list has been changed.") . "<br />" . gettext("You must apply the changes in order for them to take effect."));
 
-$tab_array = array();
-$tab_array[] = array(gettext("IP"), ($tab == "ip" ? true : ($tab == "host" ? true : ($tab == "network" ? true : false))), "/firewall_aliases.php?tab=ip");
-$tab_array[] = array(gettext("Ports"), ($tab == "port"? true : false), "/firewall_aliases.php?tab=port");
-$tab_array[] = array(gettext("URLs"), ($tab == "url"? true : false), "/firewall_aliases.php?tab=url");
-$tab_array[] = array(gettext("All"), ($tab == "all"? true : false), "/firewall_aliases.php?tab=all");
 display_top_tabs($tab_array);
 
 ?>
```
sbeaver-netgate: Thanks for this fix. I tried to find a way to take your basic idea, but to leave the tab_array section with "$tab_array[] =" format. That is because one of the documented goals of original Sjon Hortensius project was to eliminate the tab keys, and someone might just come across this page in the future, think that we forgot and remove them :) The compromise I came up with was this:

$tab_array = array();
$tab_array[] = array(gettext("IP"),    ($tab == "ip" ? true : ($tab == "host" ? true : ($tab == "network" ? true : false))), "/firewall_aliases.php?tab=ip");
$tab_array[] = array(gettext("Ports"), ($tab == "port"? true : false), "/firewall_aliases.php?tab=port");
$tab_array[] = array(gettext("URLs"),  ($tab == "url"? true : false), "/firewall_aliases.php?tab=url");
$tab_array[] = array(gettext("All"),   ($tab == "all"? true : false), "/firewall_aliases.php?tab=all");

foreach ($tab_array as $dtab) {
	if($dtab[1] == true) {
		$bctab = $dtab[0];
		break;
	}
}

$pgtitle = array(gettext("Firewall"), gettext("Aliases"), $bctab);

heper:
@sbeaver-netgate i didn't know about that part of bootstrap history ;)
I figured that using the variables/values that were already present, would be better then adding an additional foreach.

PR #2118, opened 2015-11-25
firewall_rules Bug #5527

heper:


sbeaver-netgate:
Cheery-picked your solution. File now consistent with firewall_aliases.php

PR #2119, opened 2015-11-25
Allow gateway outside interface subnet. Fixes #972, #1847

PiBa-NL:
Allow gateway outside interface subnet. Fixes #972, #1847, and was checked (on 2.2.5) with that provider to work properly for the IPv4 part. In 2.3 test machine IPv6 also seems to insert the proper routes.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -586,6 +586,24 @@ function system_routing_configure($interface = "") {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cbuechler:
Merged, thanks @PiBa-NL 

The removal doesn't happen correctly though, when you delete such a gateway it needs to delete the route. Could you submit another pull request to fix deletion?

PR #2120, opened 2015-11-26
Services - DNS Resolver - System Domain Local Zone Type Option

NOYB:
Expose system domain unbound local-zone type to be configurable.

### DIFF
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -606,7 +606,7 @@ function unbound_add_host_entries($cfgsubdir = "") {
 		$system_domain_local_zone_type = "transparent";
 	}
 
-	$unbound_entries = "local-zone: \"{$config['system']['domain']}\" $system_domain_local_zone_type\n";
+	$unbound_entries = "local-zone: \"{$config['system']['domain']}\" {$system_domain_local_zone_type}\n";
 
 	$hosts = read_hosts();
 	$added_ptr = array();
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -599,10 +599,11 @@ function unbound_add_domain_overrides($pvt_rev="", $cfgsubdir = "") {
 function unbound_add_host_entries($cfgsubdir = "") {
 	global $config, $g;
 
-	if (empty($config['unbound']['system_domain_local_zone_type'])) {
-		$system_domain_local_zone_type = "transparent";
-	} else {
+	// Make sure the config setting is a valid unbound local zone type.  If not use "transparent".
+	if (array_key_exists($config['unbound']['system_domain_local_zone_type'], unbound_local_zone_types())) {
 		$system_domain_local_zone_type = $config['unbound']['system_domain_local_zone_type'];
+	} else {
+		$system_domain_local_zone_type = "transparent";
 	}
 
 	$unbound_entries = "local-zone: \"{$config['system']['domain']}\" $system_domain_local_zone_type\n";
@@ -791,6 +792,21 @@ function unbound_hosts_generate() {
 
 	// Reload our service to read the updates
 	unbound_control("reload");
+}
+
+// Array of valid unbound local zone types
+function unbound_local_zone_types() {
+	return array(
+		"deny" => gettext("Deny"),
+		"refuse" => gettext("Refuse"),
+		"static" => gettext("Static"),
+		"transparent" => gettext("Transparent"),
+		"typetransparent" => gettext("Type Transparent"),
+		"redirect" => gettext("Redirect"),
+		"inform" => gettext("Inform"),
+		"inform_deny" => gettext("Inform Deny"),
+		"nodefault" => gettext("No Default")
+	);
 }
 
 ?>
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_unbound.php
```diff
@@ -297,6 +314,15 @@ function build_if_list($selectedifs) {
 	true
 ))->setHelp('Utilize different network interface(s) that the DNS Resolver will use to send queries to authoritative servers and receive their replies. By default all interfaces are used.');
 
+$unbound_local_zone_types = array("deny" => gettext("Deny"), "refuse" => gettext("Refuse"), "static" => gettext("Static"), "transparent" => gettext("Transparent"), "typetransparent" => gettext("Type Transparent"), "redirect" => gettext("Redirect"), "inform" => gettext("Inform"), "inform_deny" => gettext("Inform Deny"), "nodefault" => gettext("No Default"));
```
rbgarga: Please break this long line

src/etc/inc/unbound.inc
```diff
@@ -559,7 +559,13 @@ function unbound_add_domain_overrides($pvt_rev="") {
 function unbound_add_host_entries() {
 	global $config, $g;
 
-	$unbound_entries = "local-zone: \"{$config['system']['domain']}\" transparent\n";
+	if (empty($config['unbound']['system_domain_local_zone_type'])) {
```
rbgarga: It's better to use isset() here to check

src/etc/inc/unbound.inc
```diff
@@ -559,7 +559,13 @@ function unbound_add_domain_overrides($pvt_rev="") {
...
+		$system_domain_local_zone_type = "transparent";
+	} else {
+		$system_domain_local_zone_type = $config['unbound']['system_domain_local_zone_type'];
+	}
+
+	$unbound_entries = "local-zone: \"{$config['system']['domain']}\" $system_domain_local_zone_type\n";
```
rbgarga: Please add variables inside quotes between { }

src/usr/local/www/services_unbound.php
```diff
@@ -120,6 +120,12 @@
 	$pconfig['outgoing_interface'] = explode(",", $a_unboundcfg['outgoing_interface']);
 }
 
+if (empty($a_unboundcfg['system_domain_local_zone_type'])) {
```
rbgarga: Use isset() here too

src/usr/local/www/services_unbound.php
```diff
@@ -151,6 +157,10 @@
 			$input_errors[] = "One or more Outgoing Network Interfaces must be selected.";
 		}
 
+		if (empty($pconfig['system_domain_local_zone_type'])) {
```
rbgarga: isset() here too

src/usr/local/www/services_unbound.php
```diff
@@ -168,6 +178,11 @@
 			$pconfig['outgoing_interface'] = implode(",", $pconfig['outgoing_interface']);
 		}
 
+		if (isset($pconfig['system_domain_local_zone_type']) && !empty($pconfig['system_domain_local_zone_type'])) {
+			$display_system_domain_local_zone_type = $pconfig['system_domain_local_zone_type'];
+			$pconfig['system_domain_local_zone_type'] = $pconfig['system_domain_local_zone_type'];
```
rbgarga: This looks wrong, set the variable to itself

rbgarga:
It would be good to have an upgrade code to set transparent as default value for all items

NOYB:
The use of empty checks is consistent with the other existing code and in this case I believe a better choice than isset.  If the variable were to get set as empty, isset would be ineffective in this case.  Would prefer to remain consistent with the other code using empty check.  Hope that is okay.
 
Local-Zone syntax I believe is correct and consistent with the unbound-conf(5) manual.  If I'm misunderstanding please let me know.
 
Didn't do an upgrade code because the default transparent is handled in the code.  Same as it was before.  That could be changed, but I think being in the code it is less prone to config issues.
 
Have to look into if that variable set to itself was supposed to be set to something and if not remove it.

src/etc/inc/unbound.inc
rbgarga: It's just a good practice to variables between { } inside strings, as done on previous var, instead of:
```
$unbound_entries = "local-zone: \"{$config['system']['domain']}\" $system_domain_local_zone_type\n";
```
you can have:
```
$unbound_entries = "local-zone: \"{$config['system']['domain']}\" {$system_domain_local_zone_type}\n";
```

rbgarga:
You are right, empty() is fine since it already check if variable is or not set

src/usr/local/www/services_unbound.php
```diff
@@ -168,6 +178,10 @@
 			$pconfig['outgoing_interface'] = implode(",", $pconfig['outgoing_interface']);
 		}
 
+		if (isset($pconfig['system_domain_local_zone_type']) && !empty($pconfig['system_domain_local_zone_type'])) {
```
rbgarga: Considering empty() already check for isset(), isset() is redundant here and can be removed

src/usr/local/www/services_unbound.php
```diff
@@ -168,6 +178,10 @@
...
+			$display_system_domain_local_zone_type = $pconfig['system_domain_local_zone_type'];
```
rbgarga: This variable is unnecessary, code makes sure $pconfig['system_domain_local_zone_type'] is never empty then you can use it at line 209 and just remove $display_system_domain_local_zone_type

NOYB:
Okay that makes sense.  Thanks for the clarification.  I was thinking you wanted the whole thing it quotes.  Something like this:
$unbound_entries = "local-zone: \"{$config['system']['domain']} {$system_domain_local_zone_type}\"\n";

PR #2132, opened 2015-11-29
Update Form_Input class

ExolonDX:
Set **$_tagSelfClosing** to false

Any HTML tag with the **class="form-control"** will be automatically closed, however, this causes problems for tags that should not be closed in this way, e.g. SELECT tags.  HTML5 does not require INPUT tags to be closed this way, so setting **$_tagSelfClosing** to false should not cause any issues.


sbeaver-netgate:
Thanks for your changes, and for the helpful explanations that you provided. Very much appreciated :)

PR #2134, opened 2015-11-29
Tidy up SCRIPT tags

ExolonDX:
Be more consistent with the SCRIPT tag
- add type
- add CDATA section


NOYB:
Was adding Thumbs.db and oops?

PR #2147, opened 2015-12-02
Status-System Logs-Firewall-Filter

NOYB:
Spruce it up with the cool buttons.
Add no logs to display notice.

### DIFF
--- a/src/usr/local/www/diag_logs_filter.php
+++ b/src/usr/local/www/diag_logs_filter.php
@@ -207,7 +207,7 @@ if ($filterfieldsarray['interface'] == "All")
 if (!isset($config['syslog']['rawfilter'])) { // Advanced log filter form
 	$form = new Form(false);
 
-	$section = new Form_Section(gettext('Advanced Log Filter'), 'adv-filter-panel', true);
+	$section = new Form_Section('Advanced Log Filter', 'adv-filter-panel', true);
 
 	$group = new Form_Group('');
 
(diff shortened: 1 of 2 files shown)


src/usr/local/www/diag_logs_filter.php
```diff
@@ -216,14 +216,14 @@ function build_if_list() {
...
-	))->setHelp('Source IP Address');
+	))->setHelp(gettext('Source IP Address'));
 
 	$group->add(new Form_Input(
 		'filterlogentries_destinationipaddress',
 		null,
 		'text',
 		$filterfieldsarray['dstip']
-	))->setHelp('Destination IP Address');
+	))->setHelp(gettext('Destination IP Address'));
```
sbeaver-netgate: Both gettext() is applied within the Form_* classes, so it is not necessary to use the here. 
Other changes in this PR are good, but please revert the gettext()s

rbgarga: > On Dec 3, 2015, at 22:07, Steve Beaver <notifications@github.com> wrote:
> 
> In src/usr/local/www/diag_logs_filter.php:
> 
> >  
> >  	$group->add(new Form_Input(
> >  		'filterlogentries_destinationipaddress',
> >  		null,
> >  		'text',
> >  		$filterfieldsarray['dstip']
> > -	))->setHelp('Destination IP Address');
> > +	))->setHelp(gettext('Destination IP Address'));
> >  
> 
> Both gettext() is applied within the Form_* classes, so it is not necessary to use the here. 
> Other changes in this PR are good, but please revert the gettext()s

gettext() calls must be added direct to the text, not to variables, otherwise xgettext will not be able to get text sentences and produce .pot file.

--
Renato Botelho do Couto
Software Engineer
Netgate, Inc.

sbeaver-netgate:
I have cherry-picked from this PR to avoid some display issues. I have emailed you with the details.
Thanks for the PR.

PR #2148, opened 2015-12-02
Status - System Logs - Filter

NOYB:
Add advanced log filter form.
Break out log listing fields in to columns (Time, Process, PID, Message).
Basic log filter form is still used in raw mode.

### DIFF
--- a/src/usr/local/www/diag_logs.php
+++ b/src/usr/local/www/diag_logs.php
@@ -199,7 +199,7 @@ if (in_array($logfile, array('system', 'gateways', 'routing', 'resolver', 'wirel
 if (!isset($config['syslog']['rawfilter'])) { // Advanced log filter form
 	$form = new Form(false);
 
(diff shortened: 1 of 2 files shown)


src/etc/inc/filter_log.inc
```diff
@@ -79,10 +79,60 @@ function conv_log_filter($logfile, $nentries, $tail = 50, $filtertext = "", $fil
 	/* Always do a reverse tail, to be sure we're grabbing the 'end' of the log. */
 	$logarr = "";
 
+	    if ($logfile == "{$g['varlog_path']}/system.log")		{ $logfile_type = "system"; }
+	elseif ($logfile == "{$g['varlog_path']}/gateways.log")		{ $logfile_type = "system"; }
```
sbeaver-netgate: I should check the style guide, but every other file I have worked on uses "else if" (with a space).

sbeaver-netgate:
Closed in favor of newer PR.

PR #2152, opened 2015-12-03
data sanitising: ip2long32, ip2ulong, long2ip32 (Resubmit of #1789)

stilez:
Self explanatory. If these functions find themselves trying to convert non-int data (or an x64 int with non-zeros in any bits >32) to dotted IPv4, or non-dotted IPv4 to integer IPv4 values, something's wrong and they shouldn't return a value that looks like they succeeded.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -427,19 +427,24 @@ function gen_subnet_mask($bits) {
 	return long2ip(gen_subnet_mask_long($bits));
 }
 
-/* Convert long int to IP address, truncating to 32-bits. */
+/* Convert long int to IPv4 address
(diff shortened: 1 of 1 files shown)


cbuechler:
Merged, thanks!

stilez:
Looks like we found the error:  the operation "$ip>>32" fails on x32 systems or systems with 32 bit integers, where one would expect it to return zero.  We're testing replacement code that avoids x32 issues on the forum thread, and hopefully resubmit today

stilez:
Resubmitted with fix as tested in forum, see PR #2317

sbeaver-netgate:
Thank you!

I'll do some testing Wednesday morning before merging.

PR #2161, opened 2015-12-05
proposed fix breadcrumb pkg xml #5527

heper:
For this to work every xml needs a title and optional tab[text]

### DIFF
--- a/src/usr/local/www/pkg_edit.php
+++ b/src/usr/local/www/pkg_edit.php
@@ -285,20 +285,6 @@ if ($_POST) {
 	}
 }
 
-if ($pkg['title'] != "") {
-	$edit = ($only_edit ? '' : ": " . gettext("Edit"));
-	$pgtitle = $pkg['title'] . $edit;
-} else {
-	$pgtitle = gettext("Package Editor");
-}
-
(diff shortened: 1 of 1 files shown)


src/usr/local/www/pkg_edit.php
```diff
@@ -529,12 +515,17 @@ function parse_package_templates() {
...
+//breadcrumb
+if ($pkg['title'] != "") {
+	if(!$only_edit) {
+		$pgtitle = array($pkg['title'], gettext("Edit"));
+	} else {
+		$pgtitle = array($pkg['title']);
+	}
+} else {
+	$pgtitle = array(gettext("Package Editor"));
+}
```
sbeaver-netgate: Tracing head.inc and the functions it uses shows that the $pgtitle array must have at least two elements if it is to be printed on the page, so I have committed this PR  but with:

$pgtitle = array(gettext("Package"), $pkg['title']);

Need to check if the >= 2 elements thing is still valid, but this works for now.

heper:
@sbeaver-netgate after some back & forth mailing with bbcan, i've noticed the same/similar code is missing from pkg.php (pfblocker has some pages that are rendered by pkg.php, instead of pkg_edit.php)

PR #2165, opened 2015-12-06
fix showing when a error happens parsing pf rules

PiBa-NL:
fix showing when a error happens parsing pf rules

For example a queue interface speed mismatch could go unnoticed. Should this also be applied to 2.2.5 ? Same happens there..


rbgarga:
Yes, please send a new patch against RELENG_2_2 too. Thanks!

PR #2167, opened 2015-12-07
Redmine #4279 Package reinstall displayed when shutting down before upgrade

phil-davis:
Use a different flag file to indicate that a package reinstall is
required after a reboot is done first. This avoids the possibility that
the user navigates in the webGUI during the time while the shutdown is
in progress and is accidentally presented with the reinstall all
packages GUI button.
Early in rc.bootup switch the flag file to use its ordinary name, so
that all subsequent code in boot scripts and webGUI will work as it
already does to handle the package reinstall and notifying the user that
a package reinstall is about to be done or in progress...

### DIFF
--- a/src/etc/rc.bootup
+++ b/src/etc/rc.bootup
@@ -132,6 +132,12 @@ if (file_exists("/root/firmware.tgz")) {
 	unlink("/root/firmware.tgz");
 }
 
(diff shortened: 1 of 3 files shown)


phil-davis:
Note: version for RELENG_2_2 is https://github.com/pfsense/pfsense/pull/2166

PR #2168, opened 2015-12-07
Status - System Logs - Manage Log

NOYB:
Build up th manage log section with options to override the "General Logging Options" settings on an individual log basis.
Remove over exuberant gettext's.
Set/adjust filter form field widths to be better fitting for the field types.
Open/Close filter form based on filtering state.

Doesn't restart syslog daemon for changes it is not required for.
 Forward/Reverse Display
 GUI Log Entires
 Formated/Raw Display
 Web Server Log (enable/disable)


src/usr/local/www/diag_logs.php
```diff
@@ -377,20 +486,141 @@ function getGETPOSTsettingvalue($settingname, $default) {
...
+
+$group->add(new Form_Checkbox(
+	'format',
+	null,
+	'General Logging Options Setting',
+	($pconfig['format'] == '') ? true : false,
+	''
+))->displayAsRadio();
+
+$group->setHelp('Show the log entries as formated or raw output as generated by the service. The raw output will reveal more detailed information, but it is more difficult to read.');
```
phil-davis: s/formated/formatted/

heper:
@NOYB just out of curiosity: why'd you remove the gettext ? doesn't this remove the possibility for translations?

sbeaver-netgate:
gettext and htmlspecialchars are applied where required within the Form_* framework so are no longer required when composing the page. See Section.class.php line 77 for an example.

Steve
--

heper:
thanks, good to know.

sbeaver-netgate:
Would it be worth making the table of formatted logs sortable? I can imagine that clicking a column header to sort by process or by PID might be useful/ It would only require changing the table declaration to:
````html
<table class="table table-striped table-hover table-condensed sortable-theme-bootstrap" data-sortable>
````

NOYB:
Don't know why but making the table sortable messes up the column header vertical alignment.
PID is like it's on it's own line above the others.
snap1

sbeaver-netgate:
I found that the sorting library is very picky about the table code. It must include the full thead/tbody format and the number of columns in the header must exactly match the number of columns in the body. You might check those things.

````html
<table class="table table-striped table-hover table-condensed sortable-theme-bootstrap" data-sortable>
    <thead>
        <<th>One</th>
        <<th>Two</th>
    </thead>
    <tbody>
        <td>One</td>
        <td>Two</td>
    </tbody>
</table>
````

PR #2182, opened 2015-12-08
Write Config Only Option

NOYB:
Add write_config function option to only write the config.  Sometimes syncing firewall is not necessary or desirable.  ex: changing log display options.


rbgarga:
Do you have any situation where you need this?

NOYB:
It would be nice in some situations such as just making a temporary change to status log display (formatted/raw, forward/reverse, quantity) not to initiate a sync or restart any services, etc.

PR #2183, opened 2015-12-08
Add pfBlockerNG as a known special widget name Redmine #5611

phil-davis:
I guess pfBlockerNG is a well known enough package to have its special case coded here?
Or does there need to be a way that packages can specify the form of their widget name when it is not just the capitalised name of the widget file?

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -261,8 +261,8 @@ if ($config['widgets'] && $config['widgets']['sequence'] != "") {
 }
 
 ## Replace any known acronyms in widget names with suitable mixed-case forms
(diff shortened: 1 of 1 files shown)


phil-davis:
And what other packages have a dashboard widget? Are there others that need a special widget display name?

doktornotor:
>And what other packages have a dashboard widget?

HAproxy, Squid, Suricata, Snort + BIND, NUT (not available from 2.3 currently but it really needs to be done).

phil-davis:
This should no longer be necessary. The code is now in place to use the include files in /usr/local/www/widgets/include which already define vars to be the widget title text and title link.
So any package that provides an include file there will get its widget title and link from that mechanism.
https://github.com/pfsense/pfsense/pull/2198

I guess the code did all this in 2.2.* but somehow in the bootstrap conversion it got in a muddle.

PR #2185, opened 2015-12-08

phil-davis:
on factory default boot.
Alix - vr1 vr0
APU - re1 re2
serial cable and do interface assignment.


phil-davis:
This is a resubmit of https://github.com/pfsense/pfsense/pull/1819

phil-davis:
Whoo-hoo, this got merged - thanks.

rbgarga:
Thank you for all you have been doing for pfSense!!!

PR #2194, opened 2015-12-10
Internationalize more of config.console.inc

phil-davis:
This code is a whole mix of some palces that use gettext() and other chunks of output that are hardcoded text between "EOD" markers.
I have got rid of the "EOD" blocks where there is a question asked - those needed to have a space at the end of the question. To meet code style guide it is easy if questions are all in their own echo statements and the space " " can be visibly coded in.

### DIFF
--- a/src/etc/inc/config.console.inc
+++ b/src/etc/inc/config.console.inc
@@ -101,15 +101,13 @@ BEGIN MANUAL CONFIGURATION OR WE WILL PROCEED WITH AUTO CONFIGURATION.
 EOD;
 	}
 
-	echo <<<EOD
-
-Do you want to set up VLANs first? 
-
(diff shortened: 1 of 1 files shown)


phil-davis:
Do we want to internationalize all of the "EOD" text blocks in these console text-based scripts?

I guess when translating this sort of stuff, the translators have to be aware that this is running in a text-based window. If translating to a language with a totally different script, then maybe it will not render properly (depends on both the FreeBSD console device and the end-device that is connected to the actual console - e.g. if someone has a real VT100 terminal connect to a real serial port, then the VT100 is not going to display Unicode Japanese Katakana characters).

doktornotor:
@phil-davis - Additionally, anyone translating these things should be aware of the state of UTF-8 "support" on console on FreeBSD... https://wiki.freebsd.org/Newcons

PR #2197, opened 2015-12-10
Dashboard widget links use underlying file name

phil-davis:
when constructing the var name for _title_link.
Prior to this, the link for Dynamic DNS Status was not happening. That was because the text $widgetconfig['name'] is "Dynamic DNS Status". But the key of this array ($widgetname inside this loop) is the underlying file name "dyn_dns_status".
It works for all the widgets - their underlying file names and the var names $carp_status_title_link $dyn_dns_status_title_link and so on all match up.


phil-davis:
This was also included in my code for https://github.com/pfsense/pfsense/pull/2198 and is committed.
Closing this one.

PR #2203, opened 2015-12-11
Only ask Proceed with upgrade question once

phil-davis:
On nanoBSD there is a "Proceed with upgrade?" question warning about the duplicate slice. After answering "y" to that, the system does the duplicate slice, which takes some minutes. Then it asks again "Proceed with upgrade?" after displaying all the packages it will install.

That is a bit annoying at the console - I answer "y" and go off to make a cup of tea, only to come back and find that it is waiting asking again.

On nanoBSD a reboot is known to be required anyway, because even if it is some little package that gets upgraded, and not the core OS, the partition is always switched. So we can say that in the first warning and then skip asking "Proceed with upgrade?" a second time.

What do you think? Should it ask a 2nd time after displaying the packages to be installed? Or is it OK to skip that confirmation prompt?

Also, I moved the call to setup_nanobsd_env inside the "if nanobsd" - it worked like it was because setup_nanobsd_env returns without doing anything if the system is not nanobsd, but it just looked odd and there seems no point calling it when not nanobsd.

### DIFF
--- a/src/usr/local/sbin/pfSense-upgrade
+++ b/src/usr/local/sbin/pfSense-upgrade
@@ -319,6 +319,8 @@ pkg_upgrade() {
 				_echo "Before starting the upgrade process, the currently mounted nanobsd partition"
 				_echo "needs to be cloned to the secondary partition, where the update will happen"
 				_echo ""
+				_echo "After installation a reboot will be required to switch partition."
+				_echo ""
 				if [ -z "${yes}" ]; then
 					_echo -n "Proceed with upgrade? (y/N) "
 					read answer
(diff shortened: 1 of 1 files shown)


src/usr/local/sbin/pfSense-upgrade
```diff
@@ -860,3 +864,4 @@ case "${action}" in
 esac
 
 _exit 0
+\
```
rbgarga: Looks like it shouldn't be here :)

phil-davis: Oops - I leaned on the keyboard. I will make a new PR, so there is not a silly little commit that reverses this.

phil-davis:
Closing this in favor of https://github.com/pfsense/pfsense/pull/2205 which does not have the accidental rubbish.

PR #2211, opened 2015-12-12
Status - System Logs - Firewall - Manage Log

NOYB:
Use the same methods of obtaining log file etc. and mimic much of the code that is common with status_logs.php.
(future: factor out common code into shared functions to remove duplication.)

### DIFF
--- a/src/usr/local/www/status_logs_filter.php
+++ b/src/usr/local/www/status_logs_filter.php
@@ -290,8 +290,8 @@ if ($save_settings) {
 				($oldnologbogons !== isset($config['syslog']['nologbogons'])) ||
 				($oldnologprivatenets !== isset($config['syslog']['nologprivatenets']))) {
 
-				$retval |= filter_configure();
 				require_once("filter.inc");
+				$retval |= filter_configure();
 				filter_pflog_start(true);
 			}
 		}
--- a/src/usr/local/www/status_logs_filter.php
+++ b/src/usr/local/www/status_logs_filter.php
@@ -2,7 +2,6 @@
 /*
 	status_logs_filter.php
 */
-
 /* ====================================================================
  *	Copyright (c)  2004-2015  Electric Sheep Fencing, LLC. All rights reserved.
  *
@@ -72,6 +71,7 @@ require("guiconfig.inc");
 require_once("ipsec.inc");
 require_once("filter_log.inc");
 
+
 # --- AJAX RESOLVE ---
 if (isset($_POST['resolve'])) {
 	$ip = strtolower($_POST['resolve']);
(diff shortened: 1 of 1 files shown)


src/usr/local/www/status_logs_filter.php
```diff
@@ -134,9 +162,156 @@ function getGETPOSTsettingvalue($settingname, $default) {
...
+$pconfig['nentries'] = $config['syslog'][$specific_log]['nentries'];
+$pconfig['logfilesize'] = $config['syslog'][$specific_log]['logfilesize'];
+$pconfig['format'] = $config['syslog'][$specific_log]['format'];
+
+# System General (main) Specific
+$pconfig['loglighttpd'] = !isset($config['syslog']['nologlighttpd']);
+
+# Firewall Specific
+$pconfig['logdefaultblock'] = !isset($config['syslog']['nologdefaultblock']);
+$pconfig['logdefaultpass'] = isset($config['syslog']['nologdefaultpass']);
```
phil-davis: Should this be "not isset", or did I miss something?

src/usr/local/www/status_logs_filter.php
```diff
@@ -134,9 +162,156 @@ function getGETPOSTsettingvalue($settingname, $default) {
...
+
+	# Firewall Specific
+		if ($logfile == 'filter') {
+			$oldnologdefaultblock = isset($config['syslog']['nologdefaultblock']);
+			$oldnologdefaultpass = isset($config['syslog']['nologdefaultpass']);
+			$oldnologbogons = isset($config['syslog']['nologbogons']);
+			$oldnologprivatenets = isset($config['syslog']['nologprivatenets']);
+
+			$config['syslog']['nologdefaultblock'] = $logdefaultblock ? false : true;
+			$config['syslog']['nologdefaultpass'] = $logdefaultpass ? true : false;
```
phil-davis: Now I really do not understand - the logic here is also the reverse of what I expected.

NOYB:
Didn't change the logic.  It's been that way (reverse logic) for a long time.  Checked 2.2.5 and it's the same.  The code was copied directly from status_logs_settings.php.  As long as the code making use of the setting knows how to interpret the logic correctly I think changing it should be a separate commit.

phil-davis:
OK - I just mentioned it because it looked odd.

src/usr/local/www/status_logs_filter.php
```diff
@@ -200,14 +391,18 @@ function build_if_list() {
 $tab_array[] = array(gettext("Summary View"), false, "/status_logs_filter_summary.php");
 display_top_tabs($tab_array, false, 'nav nav-tabs');
 
-$Include_Act = explode(",", str_replace(" ", ",", $filterfieldsarray['act']));
-if ($filterfieldsarray['interface'] == "All")
-	$interface = "";
+define(SEC_OPEN, 0x00);
+define(SEC_CLOSED, 0x04);
```
sbeaver-netgate: Since these are defined in classes/Form/Section.class.php I don't think they are required here.

NOYB:
You know Steve I thought the same thing.  But without them the filter section defaults to open.  But the manage log section continues to work as expected.  Perhaps those defines are not getting defined until form section has been called once.

sbeaver-netgate:
Could be. Perhaps it would make sense to include them in pfSenseHelpers.js, which is included by foot.inc ?

PR #2218, opened 2015-12-13
HTML Compliance - Dashboard

NOYB:
HTML compliance (aside from the active widgets).

Rename duplicate id's.
Fix available widgets collapsible section href with id.
Where obsolete replace name attribute with id.


sbeaver-netgate:
This change has been reverted since it broke the "click on the wrnech to configure" widget actions.

PR #2239, opened 2015-12-15
system_advanced_firewall display State Timeouts section

phil-davis:
A few little code style bits
Make the State Timeouts section actually appear (add the section to the form)
The State Timeouts section was trying to do groups of fields, which spread themselves horizontally. That did not fit across a reasonable screen nicely, specially for the 6 different TCP timeouts. So the groups thing is removed and each field is its own input in the section, all vertically below each other.
Fixed various cut-and-paste bugs in the variable naming of the timeouts code.
Added logic to be able to generate the key name (like tcpfinwaittimeout) from the screen displayed text (like "FIN Wait"). Previously the "FIN Wait" case would not work with the simple strtolower($name)

This works now. If someone wants to mess with grouping the TCP, UDP, ICMP and Other in some way, then great.

### DIFF
--- a/src/usr/local/www/system_advanced_firewall.php
+++ b/src/usr/local/www/system_advanced_firewall.php
@@ -388,7 +388,7 @@ if ($_POST) {
 
 		// Kill filterdns when value changes, filter_configure() will restart it
 		if (($old_aliasesresolveinterval != $config['system']['aliasesresolveinterval']) &&
-			isvalidpid("{$g['varrun_path']}/filterdns.pid")) {
+		    isvalidpid("{$g['varrun_path']}/filterdns.pid")) {
 			killbypid("{$g['varrun_path']}/filterdns.pid");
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
The state timeouts seem fine in rows of their own, although the help text is completely pointless. If the section heading said: “State timeouts in seconds. (Leave blank for default)” it could all be removed.

Steve
--

phil-davis:
Yes, but I guess at least it has replicated the functionality and text from 2.2.* - if someone wants to reduce the mindless repetitive help text, then great.
I will keep going with code style, because while scanning through for code style I am finding all sorts of odd bits of stuff to fix.

sbeaver-netgate:
Thank you Phil :)

phil-davis:
I see you already removed the help text duplication in a later commit - good stuff.

PR #2242, opened 2015-12-16
util.inc syntax error and code style

phil-davis:
Line 578 of the previous commit had a bonus ")".
Code style also fixed up.


phil-davis:
@cbuechler It would be good to fix the syntax error ASAP - the previous commit breaks a lot of stuff because most of the universe uses util.inc

PR #2265, opened 2015-12-17
Use different var name looping through columns

phil-davis:
There is some problem with the var $i getting reset somehow somewhere
related to the include of each widget PHP code. I found that this was
doing $i = 1, $i = 2 and then $i = 2 again and I got fatal errors
because each file of widget code was being included twice.

Crash report begins.  Anonymous machine information:

amd64
10.2-STABLE
FreeBSD 10.2-STABLE #245 1b7c029(devel): Mon Dec 14 04:17:47 CST 2015

root@pfs23-amd64-builder:/usr/home/pfsense/pfsense/tmp/obj/usr/home/pfsense/pfsense/tmp/FreeBSD-src/sys/pfSense

Crash report details:

PHP Errors:
[17-Dec-2015 10:10:50 Asia/Kathmandu] PHP Fatal error:  Cannot redeclare
getServerDateItems() (previously declared in
/usr/local/www/widgets/widgets/ntp_status.widget.php:64) in
/usr/local/www/widgets/widgets/ntp_status.widget.php on line 64
[17-Dec-2015 10:10:50 Asia/Kathmandu] PHP Stack trace:
[17-Dec-2015 10:10:50 Asia/Kathmandu] PHP   1. {main}()
/usr/local/www/index.php:0
[17-Dec-2015 10:10:53 Asia/Kathmandu] PHP Fatal error:  Cannot redeclare
getServerDateItems() (previously declared in
/usr/local/www/widgets/widgets/ntp_status.widget.php:64) in
/usr/local/www/widgets/widgets/ntp_status.widget.php on line 64
[17-Dec-2015 10:10:53 Asia/Kathmandu] PHP Stack trace:
[17-Dec-2015 10:10:53 Asia/Kathmandu] PHP   1. {main}()
/usr/local/www/index.php:0
[17-Dec-2015 10:11:27 Asia/Kathmandu] PHP Fatal error:  Cannot redeclare
getServerDateItems() (previously declared in
/usr/local/www/widgets/widgets/ntp_status.widget.php:64) in
/usr/local/www/widgets/widgets/ntp_status.widget.php on line 64
[17-Dec-2015 10:11:27 Asia/Kathmandu] PHP Stack trace:
[17-Dec-2015 10:11:27 Asia/Kathmandu] PHP   1. {main}()
/usr/local/www/index.php:0

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -381,13 +381,12 @@ foreach ($widgets as $widgetname => $widgetconfig) {
 <?php
 	$columnWidth = 12 / $numColumns;
 
-	for ($i=1; $i<=$numColumns; $i++) {
-		echo '<div class="col-md-' . $columnWidth . '" id="widgets-col' . $i . '">';
+	for ($currentColumnNumber = 1; $currentColumnNumber <= $numColumns; $currentColumnNumber++) {
+		echo '<div class="col-md-' . $columnWidth . '" id="widgets-col' . $currentColumnNumber . '">';
 
-		//if col$i exists
-		if (isset($widgetColumns['col'.$i])) {
-
-			$columnWidgets = $widgetColumns['col'.$i];
+		//if col$currentColumnNumber exists
+		if (isset($widgetColumns['col'.$currentColumnNumber])) {
+			$columnWidgets = $widgetColumns['col'.$currentColumnNumber];
 
 			foreach ($columnWidgets as $widgetname => $widgetconfig) {
 				// Compose the widget title and include the title link if available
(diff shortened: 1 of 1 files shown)


phil-davis:
I think there is some underlying issue here that var names used by top-level code in widgets must not conflict with var names used by the top-level dashboard index.php code. I guess they are all in the same name space.
This change fixes the nasty problem that appeared when $i was used here. So this would be good to commit and stop the dashboard from crashing.
In the longer term there probably needs to be some way to resolve potential name space conflicts? Maybe for now we just make sure that every widget at the top level uses var names that start with some unique string based on the widget name?

phil-davis:
@cbuechler @sbeaver-netgate @jdillard Is anybody awake now? Because this issue crashed my dashboard and menus, so probably it will do the same to other people.

phil-davis:
The offending widget is dyn_dns_status - it also uses a var $i that keeps a counter. So this problem is only evident if you have the dyn_dns_status widget enabled.
For practical reasons for now it would be good to avoid that also, so I did that in https://github.com/pfsense/pfsense/pull/2267

I suspect that this var naming thing is an existing potential issue in 2.2.* also.

PR #2267, opened 2015-12-17
dyn_dns_status widget avoid var naming issue

phil-davis:
This use of var $i was conflicting and interfering with a use of it in
index.php
Use a much more unique name for now to avoid future risk.
Maybe a proper final solution for var naming across dashboard and widget
is needed - but that is a design issue to be discussed.


phil-davis:
See also https://github.com/pfsense/pfsense/pull/2265

jdillard:
@phil-davis I agree with the var naming issue. I intended to go back and rename the $i variable (I hate coming up with names) before committing, but unfortunately forgot. Thanks for catching it.

phil-davis:
It was only a problem because some widget happened to use it and modify it in code that happened inside the outer loop. It took me a while to work out what in the heck was going wrong - I stared at your code and it looked fine.

jim-p:
We've been down this road before with widget variable re-use over the years... eb7454c4491c38c183f4c37df0a0987fe9114e24 6bc1e79a7dd5d7498adec42045382e7c5804e530

PR #2269, opened 2015-12-17
Form - Input - Number Attributes

NOYB:
One is the loneliest number but zero is a number too, and is valid value for some settings.  e.g. DHCP protocol timing select timeout.

### DIFF
--- a/src/usr/local/www/classes/Form/Input.class.php
+++ b/src/usr/local/www/classes/Form/Input.class.php
@@ -56,7 +56,7 @@ class Form_Input extends Form_Element
 		switch ($type)
 		{
 			case 'number':
-				$attributes += array('min' => 1, 'step' => 1);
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
It seems that some number fields allow zero, some don't. Where the required behavior does not correspond to the default, we add an attribute like this:
````php
$control = new Form_Input(
    'hatsize',
    'Hat Size',
    'number',
    $pconfig['hatsize'],
    ['min' => 4, ''max' => 20]
);
````

I suspect that if we change the default to allow zero, we will need to add a lot of "min"s where a zero is not permitted

NOYB:
Yeah probably right.  Few places where zero is valid than not.
I had already added min zero to the DHCP advanced protocol timing select timeout.
Will close this PR.

PR #2277, opened 2015-12-17
system_crlmanager dodgy else statement

phil-davis:
The "else" that was at line 550 should enclose the whole block of code below it. The "if" handles the case "No Certificates Found for this CA." and all the code below the else should only happen when there are certificates found.


phil-davis:
nocertforthisca
Screen shot of error.

PR #2280, opened 2015-12-17
services_rfc2136_edit remove unused $input

phil-davis:
These bits if code that put new Form_Checkbox() into var $input do not seem to do anything useful. They just made me wonder why they were there.
If someone knows how they do anything useful, then please say so. Otherwise we can get rid of $input here.


sbeaver-netgate:
Agreed they serve no purpose

PR #2281, opened 2015-12-17
Tidy up "status_wireless.php"

ExolonDX:
Remove closing FONT tags


sbeaver-netgate:
Looks like someone removed half of an unneeded font tag set :(

Thanks

PR #2293, opened 2015-12-18
Bump for Nuke fifolog support

phil-davis:
Should the default config version always be bumped to match the version in globals.inc when new upgrade code is committed?


rbgarga:
It depends.
If config blocks changed by that upgrade config function changes part of default config, you should not just bump the version but also apply the changes, otherwise default config will contain wrong parts and will never be converted by upgrade script

PR #2296, opened 2015-12-19
HTML Compliance - Form Button - HRef

NOYB:
The name attribute is obsolete. Consider putting an id attribute on the nearest container instead.
The <a> name attribute is not supported in HTML5. Use the id attribute instead.


sbeaver-netgate:
Needs extensive review of jQuery selectors that may reference names. Will merge when that has been done.

PR #2301, opened 2015-12-20
Staus - System - Logs - Write Config Only

NOYB:
If no logging changes have been made, only write the config file.

### DIFF
--- a/src/usr/local/www/status_logs.php
+++ b/src/usr/local/www/status_logs.php
@@ -214,17 +214,27 @@ if ($save_settings) {
 		if ($logfile == 'system') {
 			$oldnologlighttpd = isset($config['syslog']['nologlighttpd']);
 			$config['syslog']['nologlighttpd'] = $loglighttpd ? false : true;
+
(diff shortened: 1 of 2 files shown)


src/usr/local/www/status_logs.php
```diff
@@ -214,17 +214,27 @@ function getGETPOSTsettingvalue($settingname, $default) {
...
+				$logging_changed = $lighttpd_logging_changed = true;
+			}
 		}
 
 
-		write_config($desc = "Log Display Settings Saved: " . gettext($allowed_logs[$logfile]["name"]));
+		// If any of the logging settings were changed then backup and sync (standard write_config).  Otherwise only write config (don't backup, don't sync).
+		if ($logging_changed) {
+			write_config($desc = "Log Display Settings Saved: " . gettext($allowed_logs[$logfile]["name"]), $backup = true, $write_config_only = false);
+			$retval = 0;
```
sbeaver-netgate: Don't think this $retval = 0; is needed here.

PR #2304, opened 2015-12-20
adding active tab to breadcrumb

heper:
closely review L186 - L188 (commented out some code, that is possibly used in some packages)

### DIFF
--- a/src/usr/local/www/pkg.php
+++ b/src/usr/local/www/pkg.php
@@ -183,9 +183,9 @@ if ($pkg['custom_php_command_before_form'] != "") {
 
 // Breadcrumb
 if ($pkg['title'] != "") {
-	if (!$only_edit) {
- 		$pkg['title'] = $pkg['title'] . '/Edit';
-	}
+	/*if (!$only_edit) {						// Is any package still making use of this?? Is this something that is still wanted, considering the breadcrumb policy https://redmine.pfsense.org/issues/5527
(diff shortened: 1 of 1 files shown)


src/usr/local/www/pkg.php
```diff
@@ -183,9 +183,9 @@ function domTT_title($title_msg) {
 
 // Breadcrumb
 if ($pkg['title'] != "") {
-	if (!$only_edit) {
- 		$pkg['title'] = $pkg['title'] . '/Edit';
-	}
+	/*if (!$only_edit) {						// Is any package still making use of this?? Is this something that is still wanted, considering the breadcrumb policy https://redmine.pfsense.org/issues/5527
```
sbeaver-netgate: Yes. There are packages that still use edit_only.

heper: @sbeaver-netgate $only_edit is not set anywhere in pkg.php.
this means that, this code will always add '/Edit' to the title. Is that the way it is supposed to be then ?

sbeaver-netgate: Perhaps I was thinking on pkg_edit.php

I'll merge this and we can look out for any issues then. Thanks.

heper:
@sbeaver-netgate 
i think some of the code from pkg_edit.php got copied to pkg.php a couple of days ago:
see: https://github.com/pfsense/pfsense/commit/8cfdeb2ef63e283fad931f220d1f1930651097d4

sbeaver-netgate:
Ah yes. That explains it. Thanks!

Steve
--

PR #2303, opened 2015-12-20
Redmine #5613 avoid Reset to Factory Defaults PHP warnings

phil-davis:
On Reset to Factory Defaults there are some warnings given when unlink_if_exists() tries to unlink() directories in /conf - /conf/backup and /conf/sshd.
unlink does not work on directories.
This PR avoids the PHP warnings by not trying to remove the directories inside /conf.

As has been the case before this PR, the following things are thus left on the system after Reset to Factory Defaults:
/conf/backup - old config backups, which contain configs of the previous system.
/conf/sshd - old ssh keys.

For testing, this is handy. But for real "reset" should these be deleted?
In addition to this PR, should there be code to delete everything in /conf/backup and/or /conf/sshd?

PR #2307, opened 2015-12-20
Suppress warnings for directories

phil-davis:
If you call unlink_if_exists with some wildcard that happens to include some directory(ies) then it will give warnings about "operation not permitted" when the array_map() unlink is attempted on the directories.
Putting "@unlink" directly in the array_map() does not work.
So I have made unlink_silent() as a wrapper for @unlink().
This will fix the warnings reported in Redmine #5613 and for anything else calling unlink_if_exists() in similar circumstances.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1547,15 +1547,20 @@ function mwexec_bg($command, $clearsigmask = false) {
 	return mwexec($command, false, $clearsigmask, true);
 }
 
+function unlink_silent($fn) {
+	@unlink($fn);
+}
+
 /* unlink a file, or pattern-match of a file, if it exists
    if the file/path contains glob() compatible wildcards, all matching files will be unlinked
-   if no matches, no error occurs */
+   if no matches, no error occurs
+   attempts to unlink directories will fail silently */
 function unlink_if_exists($fn) {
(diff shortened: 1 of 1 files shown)


phil-davis:
Maybe this is the better more general solution than https://github.com/pfsense/pfsense/pull/2303
Although both together also works fine.

src/etc/inc/util.inc
```diff
@@ -1547,15 +1547,20 @@ function mwexec_bg($command, $clearsigmask = false) {
...
+
 /* unlink a file, or pattern-match of a file, if it exists
    if the file/path contains glob() compatible wildcards, all matching files will be unlinked
-   if no matches, no error occurs */
+   if no matches, no error occurs
+   attempts to unlink directories will fail silently */
 function unlink_if_exists($fn) {
 	$to_do = glob($fn);
 	if (is_array($to_do) && count($to_do) > 0) {
-		array_map("unlink", $to_do);
```
rbgarga: What do you think about add @array_map here instead of creating this new function?

phil-davis: I didn't think of that. If it works then do it that way and close this PR.

rbgarga:
Done, thanks!

PR #2309, opened 2015-12-21
Avoid triggering initial wizard before reboot

phil-davis:
If you use the webGUI to reset to factory defaults, then while the
existing system is shutting down you navigate off to the dashboard, the
initial setup wizard will start. The trigger_initial_wizard flag file is
deleted, and so after the reboot the initial setup wizard will not be
triggered.
This change fixes that sequence of events by using a different flag file
across the reboot, and switching it back to the usual flag file name
during the subsequent boot up.


phil-davis:
Fixes redmine https://redmine.pfsense.org/issues/5666

PR #2310, opened 2015-12-21
HTML Compliance - Form Select - Empty

NOYB:
Misses some corner cases such as boolean false and all white space.


phil-davis:
This improvement was included in #2316 so I believe this PR can be closed.

PR #2312, opened 2015-12-21
Bug #5527 wizard breadcrumb Concept

heper:
This would require some minor adjustments to the wizards xml , for it to make sense everywhere.

### DIFF
--- a/src/usr/local/www/wizard.php
+++ b/src/usr/local/www/wizard.php
@@ -214,8 +214,8 @@ do {
 	}
 } while ($oldstepid != $stepid);
 
-$pgtitle = array($title);
-$notitle = true;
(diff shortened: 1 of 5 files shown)


sbeaver-netgate:
Perhaps a good compromise would be to note here what those "minor adjustments" should be. THat would allow others to more easily take on the work?

heper:
@sbeaver-netgate found some time: i've adjusted the xml's but they might require some polishing.
the <description> tags could use some more information, but i'm not all that good with words (&  english isn't my native language)

sbeaver-netgate:
Thank you!

sbeaver-netgate:
Thanks again. Though it may not be your first language, your English is better than some people for whom it is their only language :)

PR #2313, opened 2015-12-22
Status - System Logs - Consolidation

NOYB:
Factor common code section into to functions and place in include file.

 getGETPOSTsettingvalue
 miscellaneous_common_code
 tab_array_logs_common

 log_filter_form_system_submit
 filter_form_system

 log_filter_form_firewall_submit
 filter_form_firewall
 build_if_list

 manage_log_code
 manage_log_section

Add advanced filtering to firewall dynamic view.
Add manage log to dynamic view.

Update dynamic view and summary view log file selection method to same method as firewall and system logs.

### DIFF
--- a/src/usr/local/www/status_logs_common.inc
+++ b/src/usr/local/www/status_logs_common.inc
@@ -1,0 +1,913 @@
+<?php
+/*
+	status_logs_common.inc
+*/
+/* ====================================================================
+ *	Copyright (c)  2004-2015  Electric Sheep Fencing, LLC. All rights reserved.
+ *
+ *  Some or all of this file is based on the m0n0wall project which is
+ *  Copyright (c)  2004 Manuel Kasper (BSD 2 clause)
+ *
+ *	Redistribution and use in source and binary forms, with or without modification,
+ *	are permitted provided that the following conditions are met:
+ *
+ *	1. Redistributions of source code must retain the above copyright notice,
+ *		this list of conditions and the following disclaimer.
+ *
+ *	2. Redistributions in binary form must reproduce the above copyright
+ *		notice, this list of conditions and the following disclaimer in
+ *		the documentation and/or other materials provided with the
+ *		distribution.
+ *
+ *	3. All advertising materials mentioning features or use of this software
+ *		must display the following acknowledgment:
+ *		"This product includes software developed by the pfSense Project
+ *		 for use in the pfSense software distribution. (http://www.pfsense.org/).
+ *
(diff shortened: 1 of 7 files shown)


src/usr/local/www/status_logs_common.inc
```diff
@@ -0,0 +1,913 @@
...
+	}
+	if ($_POST[$settingname]) {
+		$settingvalue = $_POST[$settingname];
+	}
+	return $settingvalue;
+}
+
+
+// Miscellaneous Common - Code
+function miscellaneous_common_code() {
```
phil-davis: Maybe this could be called status_logs_common_code?
That would avoid future function naming conflict or confusion about it.

src/usr/local/www/status_logs_common.inc
```diff
@@ -0,0 +1,913 @@
...
+ *	====================================================================
+ *
+ */
+
+require_once("guiconfig.inc");
+require_once("filter_log.inc");
+
+
+// Retrieve GET or POST Value
+function getGETPOSTsettingvalue($settingname, $default) {
```
phil-davis: This seems a really generic thing - perhaps it should even go in a more general include file so other things can use it if they want. @rbgarga might have an opinion about that.

src/usr/local/www/status_logs_common.inc
```diff
@@ -0,0 +1,913 @@
...
+
+	$group->setHelp('<a target="_blank" href="http://www.php.net/manual/en/book.pcre.php">' . gettext('Regular expression reference') . '</a> ' . gettext('Precede with exclamation (!) to exclude match.'));
+	$section->add($group);
+	$form->add($section);
+	print($form);
+}
+?>
+
+<?php
+function build_if_list() {
```
phil-davis: I see that this function had this name already in the old code. I am a bit surprised that a function like this does not already exist, but that is another thing that could be looked into for the future (rationalizing all these sorts of places where this stuff is done).
For now I am just thinking that it could have a more specific name, as build_if_list() is likely to be chosen as a function name one day in interfaces.inc .or util.inc or...
Maybe status_logs_build_if_list?

sbeaver-netgate:
Merged.

Thanks for all of the work on this, and thinks Phil for the suggested revisions. Another nice improvement to the project :)

NOYB:
Thanks Steve.  You had a significant hand in it as well.  Like the collapsible sections, tutelage, and I'm sure a lot of other stuff beyond my awareness.

PR #2314, opened 2015-12-22
Allow disabling DiffServ Code Point (DSCP) matching

simonair:
Allow disabling the addition of a DiffServ Code Point (DSCP) match to new or edited rules.

It is currently not possible to create a new rule without a DSCP matching attribute, which defaults to "af11". As a result, rules never match for no apparent reason.

This issue has previously been reported and fixed [1] but the fix no longer works, probably due to Bootstrap ignoring empty values in selects.

This changes adds a label "any" to the empty value in the select.

[1] https://redmine.pfsense.org/issues/4951#change-19876

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -1406,7 +1406,7 @@ $section->addInput(new Form_Select(
 	'dscp',
 	'Diffserv Code Point',
 	$pconfig['dscp'],
-	["" => ''] + array_combine($firewall_rules_dscp_types, $firewall_rules_dscp_types)
+	["" => 'any'] + array_combine($firewall_rules_dscp_types, $firewall_rules_dscp_types)
 ));
 
 $section->addInput(new Form_Checkbox(


pfSense-git-updates:

phil-davis:
Actually this was broken by https://github.com/pfsense/pfsense/commit/bc3b831d82b58a0bc4822e14d30e4643536b603e
If I put back the version of /usr/local/www/classes/Form/Select.class.php from before that commit, then the blank array entry appears as an option in the dropdown list and is the default when a new rule is created.

sbeaver-netgate:
Thanks Phil. I am leaning towards reverting the PR you cite here. That construct is used in many places in the code base. Fixing it as @simonair suggests here is technically superior, but more risky since so much code elsewhere would need to be reviewed/fixed.

simonair:
I'm fine with any fix that makes firewall_rules_edit.php work again.

sbeaver-netgate:
Reverted. Thanks again for pointing out hte issue.

simonair:
Thanks a lot for the fast response and fix.

phil-davis:
I am looking at some places that talk about the HTML5 validation problem for blank options in a list. I will submit a PR with a proposed solution for Select.class.php that might work all-round...

PR #2315, opened 2015-12-22
HTML Compliance - Form Select another try

phil-davis:
Element option without attribute label must not be empty.
This is a revised attempt at what was done in commit https://github.com/pfsense/pfsense/commit/bc3b831d82b58a0bc4822e14d30e4643536b603e
A few random searches gave the suggestion that when you want a blank as the option text, you can get past the validation by making it a non-breaking space. An ordinary space seems to be consider empty for the purposes of this HTML5 validation rule.
Here are some random search results for your reading pleasure:
http://w3schools.invisionzone.com/index.php?showtopic=50403
http://stackoverflow.com/questions/22237658/html5-validation-error-with-select-required-attribute
https://github.com/wet-boew/wet-boew/issues/2721

### DIFF
--- a/src/usr/local/www/classes/Form/Select.class.php
+++ b/src/usr/local/www/classes/Form/Select.class.php
@@ -66,7 +66,14 @@ class Form_Select extends Form_Input
 				$selected = ($sval == $value);
 			}
 
-			$options .= '<option value="'. htmlspecialchars($value) .'"'.($selected ? ' selected' : '').'>'. htmlspecialchars(gettext($name)) .'</option>';
+			if (!empty($name) || $name == '0') {
+				$name_str = htmlspecialchars(gettext($name));
+			} else {
+				// Fixes HTML5 validation: Element option without attribute label must not be empty
+				$name_str = "&nbsp;";
+			}
+
+			$options .= '<option value="'. htmlspecialchars($value) .'"'.($selected ? ' selected' : '').'>'. $name_str .'</option>';
 		}
 
 		return <<<EOT


sbeaver-netgate:
Thanks Phil. This looks promising. I will take some time to test this before merging.

phil-davis:
For reference the previous PR that dealt with this was https://github.com/pfsense/pfsense/pull/2294

phil-davis:
Yes, don't rush into anything here! It worked for me on the DSCP Code Point stuff. But it seems like a dodgy thing to have to do all-round when you really do want a blank entry in a list of options. It seems to be just a work-around for a [faulty|limited] HTML5 spec to me.

NOYB:
'(!empty($name) || $name == '0')'
This misses some corner cases such as Boolean false and all white space.

''(!empty(trim($name)) || is_numeric($name)'
May be better.

NOYB:
You just want to keep coddling the broken code don't you.  That's why it never gets fixed.
Accommodating it is hack.

phil-davis:
I am no expert at all.
If the UI designer says that a drop-down list needs to have an empty option, then how is that supposed to be implemented in HTML5?
Once I (someone) knows the answer to that then it can be coded.

jim-p:
If nothing else, just use a few dashes "----" to indicate there is no choice. Using a word ("any", "none", etc) is context-specific and likely to be wrong for some number of choices.

NOYB:
"If the UI designer says that a drop-down list needs to have an empty option, then how is that supposed to be implemented in HTML5?"

It's not supposed to be implemented in HTML5.  The designer is supposed to work within the spec.

sbeaver-netgate:
Why don't we just test and go with the work-around Phil suggests today, as well as creating a new "ToDo" item in Redmine to review, correct and test all selectors that employ blank options ?

That way it will not be too broken while the work is scheduled/performed.

There are certainly "What people are used to"/support implications to such a change as well.

phil-davis:
This could become a p-ing match that actually I did not intend to start. I guess it is too late to have the HTML5 spec fixed (I should have got involved in the review a few years ago!). But I would have thought that when producing a spec for a hyper-text markup language that you would try hard to minimize restrictions that might end up putting unnecessary constraints on what can be done. Anyway, nothing we can do about that.

The (lucky?) hole that seems to be left in the spec is that there are some forms of effective white-space that are not considered empty and thus render as empty-looking space and also pass the validation.

phil-davis:
Combined version of this and #2310 is #2316

PR #2316, opened 2015-12-22
HTML Compliance - Form Select 3rd try

phil-davis:
This is a combined version of:
https://github.com/pfsense/pfsense/pull/2310
https://github.com/pfsense/pfsense/pull/2315


sbeaver-netgate:
I have not been able to find any negative side effects, so let's merge and see if anyone else does :)

PR #2317, opened 2015-12-22
data sanitising: ip2long32, ip2ulong, long2ip32 (FIXED RESUBMIT of #2152)

stilez:
Self explanatory. If these functions find themselves trying to convert non-int data (or an x64 int with non-zeros in any bits >32) to dotted IPv4, or non-dotted IPv4 to integer IPv4 values, something's wrong and they shouldn't return a value that looks like they succeeded.

The original PR caused issues with VPN. This was because, to check the presence of any bits beyond #32 were zero (if INT was 64 bits or larger), the operator >>32 was used. Unfortunately this was undefined on x32 platforms. (See https://forum.pfsense.org/index.php?topic=104175 ). 

The fix below was tested on x32, on the same thread.

TEST PR'ed IN #2152 (FAILS ON x32): 
return ((is_int($ip) && ($ip >> 32) == 0) ? long2ip($ip & 0xFFFFFFFF) : '');

TEST NOW USED (SEEMS RELIABLE ON ALL SYSTEM INT SIZES):
return ((is_int($ip) && ($ip & ~0xFFFFFFFF) == 0) ? long2ip($ip & 0xFFFFFFFF) : ''); 

Other than this line and a comment, this code is identical to PR #2152

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -422,21 +422,28 @@ function gen_subnet_mask($bits) {
 }
 
 /* Convert long int to IPv4 address
-   Returns '' if not valid IPv4 (including if any bits >32 are non-zero) */
+   Returns '' if not valid IPv4 (including if any bits >32 are non-zero) 
+   NOTE: The test needs care, as several methods to test whether any high bits >= 33 are set, fails on x32 platforms.
+   The test uses ~0xFFFFFFFF which set all high bits >= 33 (if any) whatever the system INT size
+   Alternative test if ever needed that's known to work, is: (PHP_INT_SIZE <= 4 || ($ip >> 32) == 0)) 
+   since if INT <4 bytes it's always ok and if not >>32 should be reliable.
+*/
 function long2ip32($ip) {
-	return long2ip($ip & 0xFFFFFFFF);
-}
+      return ((is_int($ip) && ($ip & ~0xFFFFFFFF) == 0) ? long2ip($ip & 0xFFFFFFFF) : '');
+ }
 
 /* Convert IPv4 address to long int, truncated to 32-bits to avoid sign extension on 64-bit platforms.
    Returns '' if not valid IPv4. */
 function ip2long32($ip) {
-	return (ip2long($ip) & 0xFFFFFFFF);
+	$a = ip2long($ip);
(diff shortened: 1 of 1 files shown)


cbuechler:
I reverted this because it broke things again. First thing noticed was services_dhcp.php range checking started failing (on 64 bit at least, didn't check 32 bit). Maybe other things, didn't dig much into it beyond that. Please make sure every area that uses these functions still works on 32 and 64 bit before re-submitting.

stilez:
just wow. Okay, will do.....

What is the best (as in, most reliable) way to test if x32 will work, if the hardware available is x64 or x64 + VM on x64?

stilez:
@cbuechler - the code is correct. 

What it's doing is precisely what it * should * be doing - namely, **identifying invalid data and bad calculations in the originating/calling code**.

In this case the issue is a bug in services_dhcp.php.  The $input_errors code tries to calculate the end point of the range by using (start_IP) || (~NETMASK).

This method isn't, and never has been, valid code on x64, because on x64 inverting an IPv4 netmask would produces a 64 bit integer with 0xFFFFFFFF in the high bits, which is automatically bad data for any kind of IPv4 conversion function.  So naturally now we're checking for bad data, it gets-correctly-picked up. This code had only worked in the past because - by chance - the high bits are silently ignored or ignored by some other oversight, either in PHP's function long2ip() or in the old version of long2ip32(). That's not guaranteed behaviour and isn't correct behaviour, as any integer >=2^32 **should** give an error when converting to an IPv4 IP, and any netmask that is logical NOTted on x64 will give such a value.

It should use **gen_subnetv4_max()** or **gen_subnetv4_max** to get this value - which **does** correctly use (~$sn & 0xFFFFFFFF) rather than ~$sn   .....

Not sure how pervasive this logic bug is. Please re-add this code and in case of future issues check for the same problem in calling code.....

jim-p:
Re-adding this code without first (or simultaneously) fixing all of the calls it breaks would be irresponsible. It doesn't matter if it's technically correct, if it introduces breakage it won't be imported so we can trip over the land mines it plants later.

stilez:
hehe - I was reporting the issue for better ideas as you responded. See redmine https://redmine.pfsense.org/issues/5702

PR #2320, opened 2015-12-23
REBASE of #1786 and #1788, tightening three IP functions

stilez:
Resubmit of two PRs that couldn't be merged due to basecode conflicts

is_linklocal() - tightened, fixed validating arbitrary text, & made correctly IPv4/v6 agnostic per RFCs
is_literalipaddrv6()  - simplified 
is_hostnamewithport() - simplified
 
 
IS_LINKLOCAL()

is_linklocal has a few issues, including validating as linklocal, addresses that aren't linklocal according to RFC 4291, validating as a linklocal address input that could contain arbitrary text/no validation of reasonableness on any %(scope/interface) present, and appearing from its function name to be suitable for all linklocal addresses but actually not IPv4/v6 agnostic.

1) IPv4/6 agnostic: while IPv4 linklocal testing isn't much needed, not it should probably be recognised because some code handling linklocal may reasonably expect is_linklocal() to be IPv4/IPv6 agnostic.

2) For IPv6, it tests at least, that the purported scope/interface is [0-9a-z]+ otherwise user input or other text such as "fe80::%\n;ARBIRARYTEXT;" would be validated as a linklocal address and inserted into pf and perhaps other places without further detection, leading to possible vulnerabilities. Also tests scope/interface for a reasonable length of <= 64 chars "just in case". But it doesn't test more than this (and probably should test for valid scope/interface if present).

Note:  Net_IPv6::_Ip2Bin() can return shorter binary strings for IPv4 or "junk" input. So this code tests that it returned a 128 bit length, which ensure it was meaningful IPv6.
 
 
IS_HOSTNAMEWITHPORT()
simplified - we don't need to pop() or assign a new variable just to test 2nd member of the array
 
 
IS_LITERALIPADDRV6()
simplified - we don't need an expensive preg_match() to test if it's a valid IPv6 wrapped in "[" ... "]"

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -698,9 +698,31 @@ function is_ipaddrv4($ipaddr) {
 	return true;
 }
 
-/* returns true if $ipaddr is a valid IPv6 linklocal address */
+/* returns 4 or 6 respectively (== TRUE) if $ipaddr is a valid IPv4 or IPv6 linklocal address
+   returns '' if not a valid linklocal address
+   TODO: does not attempt to validate any IPv6 %scope (if present) other than to check [0-9a-z_] and non-pathological max length 64 */
 function is_linklocal($ipaddr) {
-	return (strtolower(substr($ipaddr, 0, 5)) == "fe80:");
+	if (is_string($ipaddr)) {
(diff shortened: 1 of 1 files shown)


rbgarga:
I merged but after that I simplified is_linklocal(). Thanks

PR #2324, opened 2015-12-25
HTML Compliance - System / Advanced / Networking

NOYB:
Close div tag
Close php tag


sbeaver-netgate:
Fixed by removing unneeded div, rather than adding a closing /div

PR #2337, opened 2015-12-30
Update pkg_mgr.php

BBcan177:
* **Note - There are two previously commented out Lines! (L127-128)**

* Mods to if/else/elseif/endif statements for consistency.
* Change package dependency href link icon from "paper-clip" to "globe".
* Put package dependency links into a < ul > for a cleaner output when dependencies span over two lines.
* Add missing gettext()

### DIFF
--- a/src/usr/local/www/pkg_mgr.php
+++ b/src/usr/local/www/pkg_mgr.php
@@ -67,7 +67,7 @@ require_once("globals.inc");
 require_once("guiconfig.inc");
 require_once("pkg-utils.inc");
 
-/* if upgrade in progress, alert user */
+// if upgrade in progress, alert user
 if (is_subsystem_dirty('packagelock')) {
 	$pgtitle = array(gettext("System"), gettext("Package Manager"));
 	include("head.inc");
@@ -76,10 +76,7 @@ if (is_subsystem_dirty('packagelock')) {
 	exit;
 }
 
-$pkg_info = get_pkg_info();
-
 $pgtitle = array(gettext("System"), gettext("Package Manager"), gettext("Available Packages"));
-
 include("head.inc");
 
 $tab_array = array();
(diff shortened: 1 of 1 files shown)


src/usr/local/www/pkg_mgr.php
```diff
@@ -179,72 +177,58 @@
...
-		if ($menu_category != "All" && $index['categories'][0] != $menu_category && !($menu_category == "Other" && !in_array($index['categories'][0], $visible_categories))) {
-			continue;
-		}
+<?php foreach ($pkg_info as $index):
+	if (isset($index['installed'])) {
+		continue;
+	}
 
+	if ($menu_category != "All" && $index['categories'][0] != $menu_category &&
+	  !($menu_category == "Other" && !in_array($index['categories'][0], $visible_categories))) {
```
sbeaver-netgate: Style guide requires 4 space indent after a compound "if"

BBcan177: np.. Just look ugly that way :)

src/usr/local/www/pkg_mgr.php
```diff
@@ -130,12 +128,12 @@
 //		display_top_tabs($tab_array);
 }
 
-if (!$pkg_info || !is_array($pkg_info)):
-?>
+if (!$pkg_info || !is_array($pkg_info)):?>
```
sbeaver-netgate: I have yet to find a text editor that understands PHP "alternate" control format :)

BBcan177: What would you like that line to look like then? Drop the ?> to the following line?

sbeaver-netgate: NO. Its fine the way you have it. I was just whining :)

PR #2340, opened 2015-12-31
redmine 5702 - change to use high level IPv4 functions where low level ip2long32() etc are used

stilez:
Redmine 5702 identified functions ip2long32(), long2ip32() and ip2ulong() as being used quite often in a way that might break on x64 (eg "~netmask | ip" without " ... & 0xFFFFFFFF"). A review shows that these functions are almost entirely being used to do IPv4 range comparisons and manipulation for which robust high level functions already exist (or for which robustness commits are ready but can't be PR'ed until the calling code is cleaned up!)

Phil Davis suggested that code shouldn't call these functions directly, but should use the high level functions for IP manipulation instead if possible. So this is the first go at doing so. Don't know if I got them all, but got almost all if not :) Also same needs doing in packages repo.

Github says this can't be merged but "can still be created". No idea what that means.

I've reported 4 minor possible issues and queries that I noticed in passing, at Redmine # 5720, which need very quick review. I've changed one of them in this commit *(src/usr/local/www/services_dhcp.php line 447-448: surely <= and >=, not < and > to test for overlap?)* because it looks wrong, but not the other 3. Please check if my "fix" is correct.

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1243,7 +1243,7 @@ function openvpn_resync_csc(& $settings) {
 							$csc_conf .= "ifconfig-push {$ip} " . gen_subnet_mask($mask) . "\n";
 						} else {
 							/* Because this is being pushed, the order from the client's point of view. */
-							$baselong = ip2long32($ip) & gen_subnet_mask_long($mask);
+							$baselong = ip2ulong(gen_subnetv4($ip, $mask));
 							$serverip = long2ip32($baselong + 1);
 							$clientip = long2ip32($baselong + 2);
 							$csc_conf .= "ifconfig-push {$clientip} {$serverip}\n";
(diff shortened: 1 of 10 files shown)


pfSense-git-updates:

stilez:
@sbeaver-netgate - any idea why CLA missing has appeared here?

gonzopancho:
no idea, but once beaver (or someone else) tells me you're really good-to-go, I'll hand force things.

stilez:
@gonzopancho  See PR #2325 - CLA added routinely there, as with all previous PRs. Nothing's changed as far as I know, still good. Glitch of some kind?

stilez:
Superseded by 9 individual PRs (#2406 - #2414) to make merging easier and avoid conflicts. One file, utyil.inc, not updated. Hence closed this.

PR #2349, opened 2016-01-04
Feature #5729 Add Description to DHCP Leases list RELENG_2_2

phil-davis:
The start and end columns are n/a for static leases, and it is static leases for which we (might) have a description. So we could reuse the Start and End columns to show the description. That avoids any issue of using more screen width and how to fit all the columns.
I tried it at home on my 2.2.6 system and actually it is useful for me - the descriptions can be a lot more descriptive than the host name that I have put in the static mappings, so it is easier to know what is what on the status DHCP leases display with the description shown.
This is just an example based on RELENG_2_2 - but it might be the way to go?

### DIFF
--- a/usr/local/www/status_dhcp_leases.php
+++ b/usr/local/www/status_dhcp_leases.php
@@ -294,6 +294,7 @@ foreach($config['interfaces'] as $ifname => $ifarr) {
 			$slease['start'] = "";
 			$slease['end'] = "";
 			$slease['hostname'] = htmlentities($static['hostname']);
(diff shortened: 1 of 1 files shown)


phil-davis:
@sbeaver-netgate has done this in master for 2.3 and it works.
So either:
a) This could be done in 2.2.* also - so merge this
or
b) This sort of enhancement should not be back-ported to 2.2.*, in which case just close this PR.

cbuechler:
Thanks Phil. We'll leave 2_2 as is.

PR #2350, opened 2016-01-04
Redmine #5731 Error when setting various system log settings from fac…

phil-davis:
…tory default config

Forum https://forum.pfsense.org/index.php?topic=104762.0


NOYB:
Thanks Phil.  Appreciate the fix.

PR #2353, opened 2016-01-05
HTML Compliance - Firewall / NAT / Outbound / Edit

NOYB:
Bad value [0-9, a-z, A-Z and . for attribute pattern on element input: Unterminated character class
<input class="form-control" name="source" id="source" type="text" pattern="[0-9, a-z, A-Z and .">
<input class="form-control" name="destination" id="destination" type="text" value="" pattern="[0-9, a-z, A-Z and .">

Element ul not allowed as child of element span in this context. (Suppressing further errors from this subtree.)
Contexts in which element ul may be used:
 Where flow content is expected.
Content model for element span:
 Phrasing content.

Element br not allowed as child of element ul in this context. (Suppressing further errors from this subtree.)
Contexts in which element br may be used:
 Where phrasing content is expected.
Content model for element ul:
 Zero or more li and script-supporting elements.

### DIFF
--- a/src/usr/local/www/firewall_nat_out_edit.php
+++ b/src/usr/local/www/firewall_nat_out_edit.php
@@ -517,7 +517,7 @@ $group->add(new Form_IpAddress(
 	'source',
 	null,
 	$pconfig['source']
-))->addMask('source_subnet', $pconfig['source_subnet'])->setHelp('Source network for the outbound NAT mapping.')->setPattern('[0-9, a-z, A-Z and .');
+))->addMask('source_subnet', $pconfig['source_subnet'])->setHelp('Source network for the outbound NAT mapping.')->setPattern('[a-zA-Z0-9\_\.\:]+');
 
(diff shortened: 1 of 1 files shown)


src/usr/local/www/firewall_nat_out_edit.php
```diff
@@ -541,7 +541,7 @@ function build_target_list() {
 	'destination',
 	null,
 	$pconfig['destination'] == "any" ? "":$pconfig['destination']
-))->addMask('destination_subnet', $pconfig['destination_subnet'])->setHelp('Destination network for the outbound NAT mapping.')->setPattern('[0-9, a-z, A-Z and .');
+))->addMask('destination_subnet', $pconfig['destination_subnet'])->setHelp('Destination network for the outbound NAT mapping.')->setPattern('[[a-zA-Z0-9\_\.\:]+');
```
sbeaver-netgate: Too many '[' here ?

PR #2355, opened 2016-01-06
Remove preload token from HSTS header

jlduran:
Please see the documentation on how to include your domain in the preload lists:

https://hstspreload.appspot.com


cbuechler:
Merged, thanks @jlduran. I adapted that bit of the nginx conf from another use and didn't catch removing that part as well.

jlduran:
:+1: no problem. And thank you for switching to nginx!

PR #2361, opened 2016-01-07
firewall_shaper_vinterface URL reference

phil-davis:
As far as I can see this URL should be self-referring - to firewall_shaper_vinterface.php
But I am having trouble finding how to test it, because whenever $pipe is set, so is $queue so I never get down the "else".
Anyway, this seems an obvious fix, and there might be other stuff related to when creating a limiter as distinct from a queue inside that limiter.


sbeaver-netgate:
I think you are right. There no references at all to firewall_shaper.php in the 2.2.x version of this file.

PR #2364, opened 2016-01-07
Config history info - make it an info box

phil-davis:
Currently this display as an orange box. It seems to me that it is ordinary info and should be an info box (displaying in blue).
Note: with the current behavior the text is always displayed, after changing it to an info box the "i" icon has to be clicked to show the text.
Do we sometimes(like in this example) want the info text to be automagically displayed on initial page load?

### DIFF
--- a/src/usr/local/www/diag_confbak.php
+++ b/src/usr/local/www/diag_confbak.php
@@ -206,9 +206,18 @@ $form->add($section);
 print($form);
 
 if (is_array($confvers)) {
-	print_info_box(gettext('To view the differences between an older configuration and a newer configuration, ' .
-						   'select the older configuration using the left column of radio options and select the newer configuration in the right column, ' .
-						   'then press the "Diff" button.'));
+?>
+<div>
+	<div id="infoblock">
+		<?=print_info_box(
+			gettext(
+				'To view the differences between an older configuration and a newer configuration, ' .
+				'select the older configuration using the left column of radio options and select the newer configuration in the right column, ' .
+				'then press the "Diff" button.'),
+			'info')?>
+	</div>
+</div>
+<?php
 }
 ?>
 


src/usr/local/www/diag_confbak.php
```diff
@@ -206,9 +206,18 @@
 print($form);
 
 if (is_array($confvers)) {
-	print_info_box(gettext('To view the differences between an older configuration and a newer configuration, ' .
-						   'select the older configuration using the left column of radio options and select the newer configuration in the right column, ' .
-						   'then press the "Diff" button.'));
+?>
+<div>
+	<div id="infoblock">
```
sbeaver-netgate: It is the id="infobock" here that causes the block to be hidden and the "i" to appear.

To just change the color of the block (and not hide it) just add the bootstrap style as in:

print_Info_box("Some text for the box", "info"); or print_Info_box("Some text for the box", "success");

phil-davis: If I remove the infoblock stuff then sure, it becomes just and ordinary info box.
But then if I "x" it, it goes away completely and I can't open it again (of course if I refresh the whole page it comes back).
1) Do we want info boxes like that? That can be dismissed and they are gone.
2) Or do we always want ones that have the little "i" icon and can be toggled on and off?
If (2), then (maybe) sometimes we would like them to be open on initial page load.

sbeaver-netgate: Maybe id=infoblock_open" to specify it is open on page load? I'll look at that.

sbeaver-netgate: Just pushed that. You may now use
````html5
<div id="infoblock_open" . . 
````
To have the infoblock open on initial page load

PR #2365, opened 2016-01-07
status_logs_filter_summary make the info an info box

phil-davis:
The existing behavior in 2.3-BETA is that if you close this box with the "x" it goes away completely and you cannot get it back.
Do we want to make it a "proper" info box that can be open and closed as desired? (like I did here)
Note: with this change the info "i" is displayed in the closed state. Maybe it should be open by default?


sbeaver-netgate:
Setting the div id to "infoblock_open" will display the infoblock on page load. Clicking hte 'i' will toggle the display as before

phil-davis:
Yep, I am just sorting out the conflict with the 'info' quoting change.

phil-davis:
Good - you sorted out the conflict yourself.

PR #2369, opened 2016-01-07
Make bogons update message a success

phil-davis:
When this actually works (after https://redmine.pfsense.org/issues/5742 is addressed) then the message here should be a success one.


sbeaver-netgate:
No longer required due to page re-write just completed

phil-davis:
Yep, looks good.

PR #2371, opened 2016-01-07
Some improvements in PHP code and configurations

NewEraCracker:
I've noticed some minor inconsistencies in how print_info_box was called with info parameter. This brings all the code to the same standard.

Additionally, since Suhosin has a configured limit of 5000 maximum input variables, for consistency, PHP should be configured with a similar setting (the default is just 1000). This will allow BIG forms to save without any problem in the future.

Thanks.

### DIFF
--- a/src/etc/rc.php_ini_setup
+++ b/src/etc/rc.php_ini_setup
@@ -181,6 +181,7 @@ implicit_flush = true
 magic_quotes_gpc = Off
 max_execution_time = 900
 max_input_time = 1800
+max_input_vars = 5000
 register_argc_argv = On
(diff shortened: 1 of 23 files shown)


phil-davis:
I happen to have got myself side-tracked into looking at a bunch of print_info_box() stuff today. Most of my things are about putting the 'success' parameter in needed places, so they have not conflicted with this change. But it would be nice to get this PR committed. Then I won't see all those unquoted info parameters any more.

sbeaver-netgate:
Agreed Phil.

Merged.

Thanks @NewEraCracker

PR #2377, opened 2016-01-07
diag_tables give different message when last updated is unknown

phil-davis:
and put the message in an infoblock_open so the user can close/open it as they wish.


sbeaver-netgate:
FYI, the more I look at infoblock, the more I think it should be triggered by <div class="infoblock" rather than id="infoblock". Just seems to make more sense that way.

I may change that globally later today. I'll let you know if I do.

phil-davis:
Yep, that's fine. It's getting late here so I won't do anything much more tonight - I will see what you have done in the morning.

PR #2378, opened 2016-01-07
Convert short open tags to full tags

NewEraCracker:
Short open tag is discouraged since it is only available if enabled:
See https://secure.php.net/manual/en/language.basic-syntax.phptags.php

This PR may break other PRs that interact with this same files, so please consider if this is possible to integrate or not. I don't mind if this is rejected for that reason.

The ```<?php``` syntax is already used far more than the short one ```<?```.

Also the the tag ```<?=``` is always available regardless of the ```short_open_tag``` ini setting since PHP 5.4. The chances of that ever being removed from PHP are by far low, so I didn't touch that.


sbeaver-netgate:
Looks good. Thank you. I want to test this fairly carefully so it may be later in the day before you see it merged :)

NewEraCracker:
No problem. Feedback is welcome :)
I've adapted the PR to the latest changes that were committed.

sbeaver-netgate:
Found no issues. Thanks for the work :)

PR #2381, opened 2016-01-08
Fix webgui-redirect in system_advanced_admin

phil-davis:
This one also seems to be wrong.
Why do people use a different name for the $POST var?

### DIFF
--- a/src/usr/local/www/system_advanced_admin.php
+++ b/src/usr/local/www/system_advanced_admin.php
@@ -393,7 +393,7 @@ $section->addInput(new Form_Input(
 	'users/browsers to access the GUI concurrently.');
 
 $section->addInput(new Form_Checkbox(
-	'disablehttpredirect',
+	'webgui-redirect',
 	'WebGUI redirect',
 	'Disable webConfigurator redirect rule',
 	$pconfig['disablehttpredirect']


src/usr/local/www/system_advanced_admin.php
```diff
@@ -597,4 +597,4 @@
 	log_error(gettext("webConfigurator configuration has changed. Restarting webConfigurator."));
 	send_event("service restart webgui");
 }
-?>
\ No newline at end of file
+?>
```
NewEraCracker: Seems like your editor has added a new line after the closing tag. This is a bad idea as per:
https://secure.php.net/manual/en/language.basic-syntax.phptags.php
> If a file is pure PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.```

phil-davis: I tried a 2nd time - the GitHub online editor is doing that to me. But if I get the version before and after the change I can't see any difference.

NewEraCracker: Alright. This shouldn't cause any trouble either. Because that file is not included anywhere, but rather a directly accessed page.

PR #2382, opened 2016-01-08
Remove whitespace and closing-tags of .inc files before EOF

NewEraCracker:
https://secure.php.net/manual/en/language.basic-syntax.phptags.php
> If a file is pure PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

There are 128 files with this problem. 71 inc and 57 php.

This PR deals with the .inc files as those are more important because, afaik, they're always included before any intended output. I did not touch the PHP files yet, but can do so if requested.

Additionally note this may break automatic merging of other PRs produced prior to this change which edit this files.

Thanks.


phil-davis:
It only effects the last line of each file, so Git should be easily able to automagically merge it with other pending changes that do not touch the last line.

NewEraCracker:
Thanks for the feedback.

This is how I did this fairly quickly:
https://github.com/NewEraCracker/php-work/blob/master/php_whitspace_removal.php

rbgarga:
The text says "it is preferable to omit the PHP closing tag at the end of the file" what means completely remove the line containing ?>. This patch is only removing the EOL char

phil-davis:
Now there are just a few random files where you have not removed the PHP closing tag. Is there a reason for that?

NewEraCracker:
Yes. Because those files are not pure PHP code and contain output outside PHP tags (HTML) or multiple tags, where removing the last one is not a safe operation.

Also note, if this files are only used in things like 'include' or 'require' type calls, removing closing tags where possible is safe, but if files are combined by other means then removing whitespace is the only truly safe operation.

Example:
1. Merging test1.inc (with ending tag removed) and test2.inc to generate a new test3.inc would result in an invalid PHP file because the ending tag of test1.inc has been removed, so code wouldn't work when combined.
2. Merging test4.inc (with only whitespace removed) and test5.inc to generate a new test6.inc would result in a valid PHP file because the ending tag of test4.inc has been kept, so code would not suffer any parsing problems.

rbgarga:
In which cases this kind of merge can occour? I didn't remember to see any other method being used on pfSense, only require(), require_once() and include()

NewEraCracker:
If done by the command line / shell scripting to combine multiple files in one. I don't think that's being used here. But is good to keep that in mind.

NewEraCracker:
Did further inspections and found some short open tags that should be converted to full open tags. See PR #2378 for where that work has begun.

gonzopancho:
I think you should close this PR and open a new one with the problem fixed.

rbgarga:
Please submit a clean PR to this change, or rebase your fork to make it apply automatically again

NewEraCracker:
Done.

rbgarga:
I applied it on a system to test and i completely broke GUI

Fatal error: Cannot redeclare getThermalSensorsData() (previously declared in /usr/local/www/widgets/include/thermal_sensors.inc:20) in /usr/local/www/widgets/include/thermal_sensors.inc.orig on line 20 Call Stack: 0.0000 227008 1. {main}() /usr/local/www/index.php:0 PHP ERROR: Type: 1, File: /usr/local/www/widgets/include/thermal_sensors.inc.orig, Line: 20, Message: Cannot redeclare getThermalSensorsData() (previously declared in /usr/local/www/widgets/include/thermal_sensors.inc:20)

I'd say it would be better to do not touch all these files at this point. All .inc files have the ?> in the end for years and it's been working as is, this can introduce undesired damage at this point (2.3 BETA).

If you want to try it again after 2.3 is branched. please do in small chunks and test all pages that use (directly or indirectly) that .inc file to be absolutely sure it won't break anything

NewEraCracker:
Problem is that the patch util creates the .orig files by default and in FreeBSD there is no way I know of to change that behavior.

Only way is to kill those orig files by running:
```find . -type f -name "*.orig" -exec rm '{}' \; ```

Things like widget processing will include files that have '.inc' anywhere in their name and cause this double include nastiness.

PR #2385, opened 2016-01-08
firewall_shaper_vinterface display savemsg according to severity

phil-davis:


sbeaver-netgate:
Would setting $input_errors not work here?

sbeaver-netgate:
Then again no, because this isn't an input error. Never mind :(

PR #2393, opened 2016-01-09
Add possibility to configure radvd valid lifetime

schinken:


pfSense-git-updates:

src/etc/inc/services.inc
```diff
@@ -181,6 +181,17 @@ function services_radvd_configure($blacklist = array()) {
 				$radvdconf .= "\t\tAdvRouterAddr on;\n";
 				break;
 		}
+
+		$validlifetime = $dhcpv6ifconf['ravalidlifetime'];
+		if(is_numeric($validlifetime)) {
+		  $radvdconf .= "\t\tAdvValidLifetime {$validlifetime};\n";
```
rbgarga: IMO you don't need to set $validlifetime here, you can check using array value direct and using is_numericint() to make sure it's a positive number
```
if (is_numericint($dhcpv6ifconf['ravalidlifetime'])) {
```

src/etc/inc/services.inc
```diff
@@ -181,6 +181,17 @@ function services_radvd_configure($blacklist = array()) {
...
+		}
+
+		$preferredlifetime = $dhcpv6ifconf['rapreferredlifetime'];
+		if(is_numeric($preferredlifetime)) {
+		  $radvdconf .= "\t\tAdvPreferredLifetime {$preferredlifetime};\n";
```
rbgarga: Same suggestions here

schinken:
Changed the pull request according to your feedback

davidjwood:
Point taken on the defaults, but the possibility exists that the default in radvd will change silently or does not conform with the documentation, at which point the help text is misleading. If the default is configured explicitly in an else {} block, there is no ambiguity as the default you claim will be 'baked in' to the generated radvd configuration.

schinken:
Yeah okay, you're right. I'll do another pull request soon

PR #2400, opened 2016-01-11
services_dnsmasq put info in infoblock

phil-davis:
multiinfoblock
There are 3 sets of info. But when I put each in an infoblock then:
1) Only the first "i" icon works to open and close, and it opens and closes all 3 blocks of text.
2) The first 2 text boxes still have an "x" and they can be dismissed.

Do we want to enhance the supporting code for infoblock so that it can work with multiple infoblocks on the same page?

Or do we have a policy that there is only to be 1 infoblock on a page, and other 'info' text just has to not be in an infoblock?

### DIFF
--- a/src/usr/local/www/services_dnsmasq.php
+++ b/src/usr/local/www/services_dnsmasq.php
@@ -341,7 +341,9 @@ $section->addInput(new Form_Textarea(
 
 $form->add($section);
 print($form);
-
+?>
+<div class="infoblock_open">
+<?php
 print_info_box(sprintf("If the DNS forwarder is enabled, the DHCP".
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
There can only be one infoblock per page. It would be good to revise the system to allow for multiple infoblocks, but to do that I would want to revise print_info_block() and friends, particularly the way they make buttons and controls based on strings in the text ("apply" etc). Obviously that will have significant ramifications and take a lot of testing, so one infoblock per page for now.

I have a couple of ideas that may help so I will hold off on this PR until I have had time to check them.

sbeaver-netgate:
Please see: https://forum.pfsense.org/index.php?topic=101865.msg586139#msg586139

PR #2401, opened 2016-01-11
Redmine #5759 print_info_box style

phil-davis:
Define the 'success' 'info' etc style of calls to print_info_box()

and I changed just a var name in head.inc so it would be obvious to future maintainers that the text strings there are all warnings (not info).


sbeaver-netgate:
Thanks Phil.

Committed with some revisions since I was working on an overlapping issue.

It is now possible to have multiple infoblock divs on one page and in doing this I changed print_info_box("My message", 'info', false); to take that last argument. false = don't print a 'close' button. Default = print it.

PR #2418, opened 2016-01-12
APU description correction

geijt:
Just a small cosmetic change. I've corrected the description for the 'APU' embedded system. Netgate is not the manufacturer of the APU but just one of the many distributors. PC Engines is the manufacturer (and owner of it), same as for ALIX and WRAP.

Also see the output of /bin/kenv on my APU
smbios.chassis.maker="PC Engines"
smbios.planar.maker="PC Engines"
smbios.planar.product="APU"
...
smbios.system.maker="PC Engines"
smbios.system.product="APU"


cbuechler:
The purpose of that was/is for Netgate-sold systems. Not changing.

PR #2419, opened 2016-01-13
Continued: Fix port aliases (Seems you can't autosuggest an element of type "number")

NOYB:
Also
apply to src too.
remove is_numeric check
remove min/max (not valid on text fields; does not work)
add pattern in place of min/max

### DIFF
--- a/src/usr/local/www/firewall_nat_edit.php
+++ b/src/usr/local/www/firewall_nat_edit.php
@@ -771,10 +771,9 @@ $group->add(new Form_Select(
 $group->add(new Form_Input(
 	'srcbeginport_cust',
 	null,
-	'number',
-	is_numeric($pconfig['srcbeginport']) ? $pconfig['srcbeginport'] : null,
-	['min' => '1', 'max' => '65536']
-))->setHelp('Custom');
+	'text',
+	$pconfig['srcbeginport']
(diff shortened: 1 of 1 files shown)


cbuechler:
The is_numeric may be a part of XSS protection. Should probably check is_portoralias instead to be safe. Will leave for @sbeaver-netgate to review

NOYB:
The is_numeric was added as part of my original commit ( PR #2347 ).  It's existence will prohibit the text field from being populated with a non numeric existing value such as an alias name.  Don't think that is the desired behavior.  But sure no problem letting Steve check it to be sure.

rbgarga:
IMO it would be nice if we have a special type of form_input for "port or alias" fields, but @sbeaver-netgate is the best person to say

sbeaver-netgate:
The PR is acceptable as submitted since XSS protection is provided in the FORM_* framework (pfForm?) , but I will check the possibility of a special input class. We would need to do the same for IP addresses (and accommodate v4/v6 addresses) so it would be a rather significant change.

Another solution would be to provide an allowAlias() method that could be applied to Form_Input and Form_IpAddress classes.

Care has to be taken because there are three levels of validation: What can you type into an element, what will the browser allow you to submit, what do you do with the submitted data. Browsers have slight differences in how the first two are handled.

PR #2420, opened 2016-01-13
Gateways Widget Refactor init-ajax

NOYB:
Move table content creation into a function so initial table load and reload can be done normally without resize flicker.
Subsequent content updates remain via ajax.


sbeaver-netgate:
Looks good. Thanks. I'll merge as-is, and if you feel like updating as below we can add that later.

In other parts of the codebase we do something similar, but make the function returns a string that we then print:

if ($_REQUEST && $_REQUEST['ajax']) {
    print( compose_table_body_contents() );
    exit;
}

compose_table_body_contents() {
    $rtnstr = '<tr>';
    $rtnstr .=    '<td>' . "\n");

    blah, blah, blah

    $rtnstr .=    '</tr>' . "\n");

    return($rtnstr);
}

PR #2421, opened 2016-01-13
Fix logic for subnet overlap functions + canonical version + IPv6 agnostic as usual

stilez:
The subnet overlap functions came up as a concern while fixing redmine 5702. 

Specifically - 
*   The "canonical" function check_subnets_overlap() doesn't handle IPv6 (util.inc has standardised on v4/v6/agnostic versions, but this doesn't fit). Fixed by adding transparent detection of v4/v6 and a specific IPv4-only version
*   The IPv6 version is wrong (if sub1 *contains* sub2 then neither of sub1's endpoints will be detected as "inrange" of sub2 and result will be incorrect: this logic error has been fixed recently in other code too)
*  Bad data isn't detected - this still isn't detected for compatibility and is tagged "FIXME" instead to look at in future. Reason - not to break anything, at present always returns "overlap = true/false", not "true/false/invalid input".
*  Because CIDR overlap implies containment, the IPv4 version uses a very efficient logic, namely calculates largest size subnet and checks this is same for both. Adopting this for both, and simplifying, makes these functions far "neater"
*   The old v4 version allowed for non-numeric $bits which doesn't make sense and I've omitted. Cannot think of a single situation where we would provide empty or bad data when we actually mean a /32 single IP. (If /32 was hard coded and intended we would test IP in subnet anyhow, not subnets overlapping)

Solution in this commit - a canonical "overlap" test (IPv4/IPv6 agnostic), IPv4/v6 "overlap" versions that actually do the work, in each case using the same logic as the old v4 (identify largest bit size and test if subnets created are valid and identical), and tag lack of "bad data" detection as fixme for now, returning FALSE instead to avoid breaking anything until fixed.  Should be transparent from outside.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -862,7 +862,7 @@ function check_subnets_overlap($subnet1, $bits1, $subnet2, $bits2) {
 function check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2) {
 	$largest_sn = max($bits1, $bits2);
 	$subnetv4_start1 = gen_subnetv4($subnet1, $largest_sn);
-	$subnetv4_start2 = gen_subnetv4($subnet1, $largest_sn);
+	$subnetv4_start2 = gen_subnetv4($subnet2, $largest_sn);
 	
 	if($subnetv4_start1 == '' || $subnetv4_start2 == '') {
 		// One or both args is not a valid IPv4 subnet
@@ -877,7 +877,7 @@ function check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2) {
 function check_subnetsv6_overlap($subnet1, $bits1, $subnet2, $bits2) {
 	$largest_sn = max($bits1, $bits2);
 	$subnetv6_start1 = gen_subnetv6($subnet1, $largest_sn);
-	$subnetv6_start2 = gen_subnetv6($subnet1, $largest_sn);
+	$subnetv6_start2 = gen_subnetv6($subnet2, $largest_sn);
 	
 	if($subnetv6_start1 == '' || $subnetv6_start2 == '') {
 		// One or both args is not a valid IPv6 subnet
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -850,7 +850,7 @@ function subnetv4_expand($subnet) {
 /* find out whether two IPv4/IPv6 CIDR subnets overlap.
    Note: CIDR overlap implies one is identical or included so largest sn will be the same */
 function check_subnets_overlap($subnet1, $bits1, $subnet2, $bits2) {
-	if (is_ipaddrv4($ipaddr)) {
+	if (is_ipaddrv4($subnet1)) {
 		return check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2);
 	} else {
 		return check_subnetsv6_overlap($subnet1, $bits1, $subnet2, $bits2);
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -860,7 +860,7 @@ function check_subnets_overlap($subnet1, $bits1, $subnet2, $bits2) {
 /* find out whether two IPv4 CIDR subnets overlap.
    Note: CIDR overlap means sn1/sn2 are identical or one is included in other. So sn using largest $bits will be the same  */
 function check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2) {
-	$largest_sn = max($bits1, $bits2);
+	$largest_sn = min($bits1, $bits2);
 	$subnetv4_start1 = gen_subnetv4($subnet1, $largest_sn);
 	$subnetv4_start2 = gen_subnetv4($subnet2, $largest_sn);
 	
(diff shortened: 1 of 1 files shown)


src/etc/inc/util.inc
```diff
@@ -847,36 +847,44 @@ function subnetv4_expand($subnet) {
...
+		//FIXME: needs to return "bad data" not true/false if bad. For now return false, best we can do until fixed
+		return false;
+	}
+	return ($subnetv4_start1 == $subnetv4_start2);
+}
 
-	return (is_inrange_v6($sub1_min, $sub2_min, $sub2_max) || is_inrange_v6($sub1_max, $sub2_min, $sub2_max) || is_inrange_v6($sub2_min, $sub1_min, $sub1_max));
+/* find out whether two IPv6 CIDR subnets overlap.
+   Note: CIDR overlap means sn1/sn2 are identical or one is included in other. So sn using largest $bits will be the same  */
+function check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2) {
```
rbgarga: I guess it meant to be check_subnetsv6_overlap

src/etc/inc/util.inc
```diff
@@ -847,36 +847,44 @@ function subnetv4_expand($subnet) {
...
+	$largest_sn = max($bits1, $bits2);
+	$subnetv4_start1 = gen_subnetv6($subnet1, $largest_sn);
```
rbgarga: And also here

stilez:
Thanks - you're right

src/etc/inc/util.inc
```diff
@@ -847,36 +847,44 @@ function subnetv4_expand($subnet) {
...
 function check_subnetsv6_overlap($subnet1, $bits1, $subnet2, $bits2) {
-	$sub1_min = gen_subnetv6($subnet1, $bits1);
-	$sub1_max = gen_subnetv6_max($subnet1, $bits1);
-	$sub2_min = gen_subnetv6($subnet2, $bits2);
-	$sub2_max = gen_subnetv6_max($subnet2, $bits2);
-
-	return (is_inrange_v6($sub1_min, $sub2_min, $sub2_max) || is_inrange_v6($sub1_max, $sub2_min, $sub2_max) || is_inrange_v6($sub2_min, $sub1_min, $sub1_max));
+	$largest_sn = max($bits1, $bits2);
+	$subnetv6_start1 = gen_subnetv6($subnet1, $largest_sn);
+	$subnetv6_start2 = gen_subnetv6($subnet1, $largest_sn);
```
rbgarga: It should be $subnet2 here

src/etc/inc/util.inc
```diff
@@ -847,36 +847,44 @@ function subnetv4_expand($subnet) {
...
-	$sn1 = gen_subnet_mask_long($relbits) & ip2long($subnet1);
-	$sn2 = gen_subnet_mask_long($relbits) & ip2long($subnet2);
-
-	return ($sn1 == $sn2);
+/* find out whether two IPv4 CIDR subnets overlap.
+   Note: CIDR overlap means sn1/sn2 are identical or one is included in other. So sn using largest $bits will be the same  */
+function check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2) {
+	$largest_sn = max($bits1, $bits2);
+	$subnetv4_start1 = gen_subnetv4($subnet1, $largest_sn);
+	$subnetv4_start2 = gen_subnetv4($subnet1, $largest_sn);
```
rbgarga: It should be $subnet2 here

src/etc/inc/util.inc
```diff
@@ -847,36 +847,44 @@ function subnetv4_expand($subnet) {
...
+	$largest_sn = max($bits1, $bits2);
+	$subnetv6_start1 = gen_subnetv6($subnet1, $largest_sn);
+	$subnetv6_start2 = gen_subnetv6($subnet1, $largest_sn);
+	
+	if($subnetv6_start1 == '' || $subnetv6_start2 == '') {
+		// One or both args is not a valid IPv6 subnet
+		//FIXME: needs to return "bad data" not true/false if bad. For now return false, best we can do until fixed
+		return false;
+	}
+	return ($subnetv6_start1 == $subnetv6_start2);
```
rbgarga: This if state can be removed and code simplified as:
```
return ($subnetv6_start1 != '' && $subnet_start1 == $subnet_start2)
```

phil-davis: The same could be done in check_subnetsv4_overlap.
And there are a couple of missing v6 in the var names of @rbgarga example.

stilez:
@rbgarba - *headdesk* 4 or 5 typos is 4 or 5 too many. Thanks for being watchful where I slipped.

@rbgarba, @phil-davis - I deliberately have not combined the logic into one line as rbgarba suggests. That's because the current function needs a FIXME - it simply doesn't test or handle invalid input (eg "do 1.1.1/23.7 and ::ff00/arbitrarytext overlap"? Answer currently is true/false which can't be best response). Learning from the recent past, I'm not fixing this aspect until I've checked first that anything it would affect is checked + ok/fixed which I can't do yet. However eventually this will return 3 values (true/false/bad input) and at that point the 2 separate tests won't so neatly compact into 1. So I've separated them and essentially forced it to return false for bad data for compatibility, but eventually it won't do so. So no point simplifying the code, as that code oonly appears sensible becaue the FIXME isn't yet fixed (sort of thing!). Best leave as-is so it's easier to FIX the FIXME :)

src/etc/inc/util.inc
```diff
@@ -847,36 +847,44 @@ function subnetv4_expand($subnet) {
...
+}
 
-	$sn1 = gen_subnet_mask_long($relbits) & ip2long($subnet1);
-	$sn2 = gen_subnet_mask_long($relbits) & ip2long($subnet2);
-
-	return ($sn1 == $sn2);
+/* find out whether two IPv4 CIDR subnets overlap.
+   Note: CIDR overlap means sn1/sn2 are identical or one is included in other. So sn using largest $bits will be the same  */
+function check_subnetsv4_overlap($subnet1, $bits1, $subnet2, $bits2) {
+	$largest_sn = max($bits1, $bits2);
```
rbgarga: I believe it's reversed here and you should use smallest value. Here is a test case where it fails
It's returning false to:
```
check_subnets_overlap("192.168.1.0", "25", "192.168.1.130", "24")
```

stilez: Indeed. In an earlier version of this, I used 32-$bits and 128-$bits to do them as one function, before realising this could itself be simplified.  When I removed that piece, I should have replaced max() by min()

PR #2432, opened 2016-01-14
Status Logs breadcrumbs and tab array highlighting

phil-davis:
1) When on Status Logs->VPN the VPN part of the breadcrumb list was
missing.
2) When selecting Status Logs-> System->Gateways... the underlining of
the "System" level tab went AWOL.
3) When selecting Status Logs->VPN-> PPPoE Service or L2TP Service the
underlining of the "VPN" level tab went AWOL.


NOYB:
Thanks Phil
Haven't tried it but the code looks like it should work.

PR #2433, opened 2016-01-14
Gateways Widget - Return String and Print

NOYB:
Return table body content string from function and print.


sbeaver-netgate:
Thanks!

PR #2435, opened 2016-01-14
subnet_size() and underlying function

stilez:
Transparent tightening/enhancement. Added function subnet_size_from_netmask and modified existing function to use it.

1. Added a version of this that uses IP type and bit size rather than a string containing a subnet. More useful where $bits is known but base IP is irrelevant, which is the case in some DHCP/CARP/VPN code where we just want a `for (i=0; i<=$count-1; i++)` loop.
2. Rewrote existing function to use this (avoids duplication)
3. Added transparent optional argument $exact=false that allows calling code to specify an exact INT is required. This may be needed for loop control and some IP calculations, because for large subnets a FLOAT may be returned. (It's hard to think where exactness will matter on such a huge subnet that INT overflows, but this guarantees for calling code that the result will be an int, and may be useful in future)
4. Use ** operator (PHP 5.6) rather than exp(log). Tested and appears this returns INT where possible, up to 2^63 on x64, and a FLOAT if larger. (pow() seems identical. 1<< $bits wraps round to negative at 31/63 of course)

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -811,11 +811,9 @@ function is_subnetoralias($subnet) {
 	}
 }
 
-/* Get number of addresses in an IPv4/IPv6 subnet.
-   optional arg $exact=true forces an error to be returned if it can't be represented exactly
-   
-   
-   
+/* Get number of addresses in an IPv4/IPv6 subnet (represented as a string)
+   optional $exact=true forces error (0) to be returned if it can't be represented exactly
+   Exact result not possible above PHP_MAX_INT which is about 2^31 addresses on x32 or 2^63 on x64
    Returns 0 for bad data or if cannot represent size as an INT when $exact is set. */
 function subnet_size($subnet, $exact=false) {
 	$parts = explode("/", $subnet);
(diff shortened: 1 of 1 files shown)


src/etc/inc/util.inc
```diff
@@ -811,16 +811,48 @@ function is_subnetoralias($subnet) {
...
-		list ($ip, $bits) = explode("/", $subnet);
-		return round(exp(log(2) * (128 - $bits)));
+/* Get number of addresses in an IPv4/IPv6 subnet (represented as a string)
+   optional $exact=true forces error (0) to be returned if it can't be represented exactly
+   Exact result not possible above PHP_MAX_INT which is about 2^31 addresses on x32 or 2^63 on x64
+   Returns 0 for bad data or if cannot represent size as an INT when $exact is set. */
+function subnet_size($subnet, $exact=false) {
+	$parts = explode("/", $subnet);
+	if (count($parts) == 2) {
+		if (is_ipaddrv4($ip)) {
```
phil-davis: Where does $ip and $bits get set in this function?

stilez: Fixed as you were typing. I changed the explode from exploding to a list to an array, in case of malformed arg.

rbgarga:
Calls to subnet_size() must be adjusted to deal with return 0, for example:
https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/firewall_aliases_edit.php#L368
will always consider 0 <= max allowed items

stilez:
The original code used return 0 as error (not an IP), so I retained that. Presumably it should be okay because it shouldn't be possible for any valid input data to return 0, because no valid CIDR covers 0 IPs. 

If calling code does needs this issue addressing as you identify (which makes sense) then I'll switch it to use empty string for "bad data" at the same time, to ensure consistency with many of the other IP functions which use the same return for bad data. I'll look at this in a bit and sort it out somehow :)

PR #2436, opened 2016-01-14
Replace abbreviations with complete word

NewEraCracker:
I've found some abbreviations that could be replaced with complete words.
This should ease the understanding of every table column.


NOYB:
Not too fond of the action, interface and protocol abbreviations on the firewall logs table.  The action, interface and protocol fields rarely if ever have content even close to the length of the full word column header.  Thus take up valuable horizontal space from fields that make use of the space.

In the context of the firewall log table, and given the contents of the fields, the abbreviations are clear enough for anyone knowledgeable enough to run a pfSense box.  Much prefer the abbreviation in such cases.

PR #2440, opened 2016-01-14
Consistent panel-heading h2

phil-davis:
Make all the panel-heading uses also have consistent <h2
class="panel-title"> so they all display the same size.
Forum https://forum.pfsense.org/index.php?topic=105275.0

### DIFF
--- a/src/usr/local/www/crash_reporter.php
+++ b/src/usr/local/www/crash_reporter.php
@@ -151,7 +151,7 @@ exec("/usr/bin/grep -vi warning /tmp/PHP_errors.log", $php_errors);
 		}
 ?>
 	<div class="panel panel-default">
-		<div class="panel-heading"><?=gettext("Unfortunately we have detected a programming bug.")?></div>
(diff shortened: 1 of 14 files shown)


sbeaver-netgate:
I'm not sure about applying the change to index.php. Might make it look ugly but let's give it a go and see.

phil-davis:
Yes, I wondered about index.php - you guys can discuss that and I guess there might be feedback in the forum also. But if that size/style is being used everywhere else, then it will seem odd for it to be different when switching between dashboard and every other form.

sbeaver-netgate:
I put:

.panel-title {
    font-size: 16px;
}

In my pfSense-Beta.css file and switched to that. I much prefer the look, but we'll have to see what other think :)

phil-davis:
I tried that. IMHO it looks good and is big enough for fingers to hit the +/- button in places where there is one in the panel-title. Pass it by a few others to get opinion.

PR #2450, opened 2016-01-17
Add bcrypt password hashing as per #4120

hexaclock:
Made bcrypt the default password hash function. Upon the next password change for a user account, the old descrypt and MD5 hashes are removed and a new bcrypt hash is stored.

I tested authentication via the web UI and SSH - both seem to work before and after a password change.

If anyone has any suggestions for improvement, please let me know!


src/etc/inc/auth.inc
```diff
@@ -593,8 +610,18 @@ function local_user_del($user) {
 
 function local_user_set_password(&$user, $password) {
 
-	$user['password'] = crypt($password);
-	$user['md5-hash'] = md5($password);
+	$user['password'] = "REMOVED";
+	$user['md5-hash'] = "REMOVED";
```
rbgarga: Why keep these instead of unset()?

src/etc/inc/auth.inc
```diff
@@ -528,7 +540,12 @@ function local_user_set(& $user) {
 		log_error(sprintf(gettext("Running: %s"), $cmd));
 	}
 	$fd = popen($cmd, "w");
-	fwrite($fd, $user['password']);
+	if (empty($user['bcrypt-hash'])) {
+		fwrite($fd, $user['password']);
+	}
+	else {
```
rbgarga: style: Move else to same line of }

src/etc/inc/auth.inc
```diff
@@ -498,7 +505,12 @@ function local_user_set(& $user) {
 			log_error(sprintf(gettext("Running: %s"), $cmd));
 		}
 		$fd = popen($cmd, "w");
-		fwrite($fd, $user['password']);
+		if (empty($user['bcrypt-hash'])) {
+			fwrite($fd, $user['password']);
+		}
+		else {
```
rbgarga: style: move else to same line of }

src/etc/inc/auth.inc
```diff
@@ -360,6 +360,13 @@ function local_backed($username, $passwd) {
 		return false;
 	}
 
+	if ($user['bcrypt-hash']) {
+		if (password_verify($passwd, $user['bcrypt-hash'])) {
```
rbgarga: Please combine both if statements in once with &&

src/etc/inc/auth.inc
```diff
@@ -593,8 +610,18 @@ function local_user_del($user) {
 
 function local_user_set_password(&$user, $password) {
 
-	$user['password'] = crypt($password);
-	$user['md5-hash'] = md5($password);
+	$user['password'] = "REMOVED";
+	$user['md5-hash'] = "REMOVED";
+	$user['bcrypt-hash'] = password_hash($password,PASSWORD_BCRYPT);
```
rbgarga: style: add space after ,

rbgarga:
It passed all my tests, thanks!
After style fixes I'm going to get it merged

src/etc/inc/auth.inc
```diff
@@ -593,8 +610,18 @@ function local_user_del($user) {
...
+
+	// Maintain compatibility with FreeBSD
+	// https://reviews.freebsd.org/D2742
+	// change $2y$ prefix to $2b$
+
+	if ($user['bcrypt-hash'][2] == "y") {
+		$user['bcrypt-hash'][2] = "b";
+	}
```
rbgarga: I was wondering about it, if FreeBSD already deal with it we don't need to care about it.

hexaclock: I wasn't sure what version of FreeBSD would be used for pfSense 2.3. It definitely breaks local authentication (e.g. when trying to SSH in) on pfsense 2.2.6.

src/etc/inc/auth.inc
hexaclock: I was matching the style in the statements below this one:

```
	//for backwards compatibility
	if ($user['password']) {
		if (crypt($passwd, $user['password']) == $user['password']) {
			return true;
		}
	}

	if ($user['md5-hash']) {
		if (md5($passwd) == $user['md5-hash']) {
			return true;
		}
	}
```

src/etc/inc/auth.inc
hexaclock: I left the fields in just in case something was relying upon them for something, but if that's not the case I'll unset them.

src/etc/inc/auth.inc
rbgarga: Yeah, but we are not implementing in 2.2.x, but anyway it needs to stay for now since the commit was never MFC'd

Please add a XXX: line in comment saying it can be removed as soon as r284483 is MFC'd

src/etc/inc/auth.inc
rbgarga: OK, just leave it as is for now and we can revisit and change all of them at once

hexaclock:
I think it's all set now.

PR #2452, opened 2016-01-17
fix Bug #5776

heper:


jlduran:
Should work fine, although I would prefer to change all _username_ inputs from `type=text` to `type=email`, you get the added bonus of a keyboard switch. i.e.:

```html
<input type="email" class="form-control" name="usernamefld" id="usernamefld" placeholder="Enter your username">
```

sbeaver-netgate:
We recently discovered some issues with type="email" when the field is left blank. Even though that may not be an issue here I would prefer that we stick with "text" for now.

PR #2454, opened 2016-01-17
Rationalize message display on diag_testport

phil-davis:
1) The logic of the various tests that determined which messages were displayed was all a bit odd, and there were ways that things could happen and no message was displayed. Now it displays the "warning" information message the first time, and 1 of the 4 possible result messages depending on the result status and if showtest was on.
2) Internationalize - put the gettext() into calls with full sentences and sprintf() wrappers as necessary.

Now I can get each of the 4 possible result messages, and always get a result message.

### DIFF
--- a/src/usr/local/www/diag_testport.php
+++ b/src/usr/local/www/diag_testport.php
@@ -221,9 +221,9 @@ include("head.inc");
 // Handle the display of all messages here where the user can readily see them
 if ($input_errors) {
 	print_input_errors($input_errors);
-} elseif ($do_testport)	{
+} elseif ($do_testport) {
 	// User asked for a port test
-	if ($retval == 0)	{
+	if ($retval == 0) {
 		// Good host & port
 		$alert_text = '<div class="alert alert-success" role="alert">' . sprintf(gettext('Port test to host: %1$s Port: %2$s successful'), $host, $port);
 		if ($showtext) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_testport.php
```diff
@@ -222,28 +222,28 @@
...
-		} else {
-			print('<div class="alert alert-success" role="alert">'.gettext("Port test to host: " . $host . " Port: " . $port . " successful") . '. Any text received from the host will be shown below the form.</div>');
-		}
-	}
-
-	// netcat exit value != 0
-	if ($retval != 0 && !empty($result)) {
-		if ($showtext) {
-			print('<div class="alert alert-danger" role="alert">'.gettext('No output received, or connection failed. Try with "Show Remote Text" unchecked first.').'</div>');
+	if ($do_testport)	{
```
rbgarga: This test could be done in above 'else' statement using elseif

src/usr/local/www/diag_testport.php
```diff
@@ -222,28 +222,28 @@
...
+		// User asked for a port test
+		if ($retval == 0)	{
+			// Good host & port
+			if (!$showtext) {
+				print('<div class="alert alert-success" role="alert">' . sprintf(gettext('Port test to host: %1$s Port: %2$s successful'), $host, $port) . '</div>');
+			} else {
+				print('<div class="alert alert-success" role="alert">' . sprintf(gettext('Port test to host: %1$s Port: %2$s successful.'), $host, $port) . ' ' . gettext('Any text received from the host will be shown below the form.') . '</div>');
```
rbgarga: First sentence is the same in both cases, it could print it and then print 2nd sentence only when $showtext is true

phil-davis:
Please review again.

PR #2460, opened 2016-01-18
Firewall / Schedules / Edit - Add Minutes 45

NOYB:


phil-davis:
Whoa - well spotted. How did that get lost for so long! I have been through all the code (at speed-reading pace) looking for various things but missed that. How many other dumb little errors are still waiting to be found in this bootstrap thing?

NOYB:
Found it because I was manually entering a couple of schedules that use it.

PR #2471, opened 2016-01-18
classes/Form/Textarea.class.php, setNoWrap, make it work properly on FireFox and Chromium

PiBa-NL:
classes/Form/Textarea.class.php, setNoWrap, make it work properly on FireFox and Chromium

### DIFF
--- a/src/usr/local/www/classes/Form/Textarea.class.php
+++ b/src/usr/local/www/classes/Form/Textarea.class.php
@@ -51,7 +51,7 @@ class Form_Textarea extends Form_Input
 
 	public function setNoWrap()
 	{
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
I don't think this method is actually used anywhere (yet) . All of the textareas that require nowrap seem to be written in HTML.

Merging this for future use.

PiBa-NL:
The HAProxy-devel package uses it on the settings tab for the "Global Advanced pass thru" field.
Anyway thanks for merging. 
(p.s. Chrome does currently wrap with this applied, but i believe that as a bug that will be fixed in the near future as Chromium does show it properly.) With previous code Firefox for example would not show a second line, even when pressing enter. Imho the current code is the best option. Unless we want to code some Chrome version specific case..

PR #2473, opened 2016-01-19
Standardize NAT NPt

phil-davis:
Give NAT NPt GUI the same sort of interface as NAT 1:1
1) Checkbox to select multiple mappings and then delete selected
mappings
2) Tick/cross to enable/disable a mapping
3) Ability to "Add a new mapping based on this one"
4) Double-click to edit an entry
5) Ability to reorder mappings and save the new order - probably nobody
cares about that but it is easy to provide it the same as on other
forms, and so users can order the entries as they like.
Note: It used to try to automagically sort the entries based on ipcmp()
of the value of 'external' - that does not work in 2.2.* because
'external' is not a key at all for NPt. So I took that out - seemed more
consistent to provide the user drag-drop entry ordering feature.

Other bits of text in firewall NAT, 1:1 and Out standardized to make all
of them match.

### DIFF
--- a/src/usr/local/www/firewall_nat_out.php
+++ b/src/usr/local/www/firewall_nat_out.php
@@ -244,7 +244,8 @@ if ($savemsg) {
 }
 
 if (is_subsystem_dirty('natconf')) {
-	print_info_box_np(gettext("The NAT configuration has been changed.")."<br />".gettext("You must apply the changes in order for them to take effect."));
+	print_info_box_np(gettext('The NAT configuration has been changed.') . '<br />' .
+					  gettext('You must apply the changes in order for them to take effect.'));
 }
 
 $tab_array = array();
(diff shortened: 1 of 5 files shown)


src/usr/local/www/firewall_nat_out.php
```diff
@@ -244,7 +244,8 @@
 }
 
 if (is_subsystem_dirty('natconf')) {
-	print_info_box_np(gettext("The NAT configuration has been changed.")."<br />".gettext("You must apply the changes in order for them to take effect."));
+	print_info_box_np(gettext('The NAT configuration has been changed.') . '<br />' .
+					  gettext('You must apply the changes in order for them to take effect.'));
```
rbgarga: Looks like a lot of extra blank space here

phil-davis: That is a cut-paste from firewall_nat.php - it has a bunch of tabs that, in an editor with tabs displayed at 4-spaces intervals, will make the 2 gettext() calls line up vertically.
I have never really understood why people have done that, because the resulting vertical line-up depends on the editor display settings used. Personally I would put just 1 tab extra indent from the outer calling function.
This is the same in all 4 files that it is in - I can easily modify them all, but exactly how should it be (is there a definitive style-guide for this?)?

rbgarga: Unfortunately part of style was lost during the big bootstrap merge. the style guide is still the same, TAB with size of 8 and 4 spaces for multiple line statements

PR #2474, opened 2016-01-19
A proposal for 5431 https://redmine.pfsense.org/issues/5431

NOYB:
A proposal for 5431 https://redmine.pfsense.org/issues/5431

A simple button next to the help icon (?) at right end of breadcrumb bar to show/hide the available widgets panel outer div.

The available widgets panel itself remains unchanged.
css btn-breadcrumbbar to blend button in with the breadcrumb bar to not be distracting.
Bootstrap framework remains intact.

### DIFF
--- a/src/usr/local/www/head.inc
+++ b/src/usr/local/www/head.inc
@@ -486,7 +486,7 @@ echo '<li>'. get_shortcut_status_link($shortcut_section, false). '</li>';
 echo '<li>'. get_shortcut_log_link($shortcut_section, false). '</li>';
 
 ?>
-	<?php if ($pagetitle === 'Status: Dashboard'): ?>
+	<?php if ($dashboard_available_widgets_hidden): ?>
 		<li>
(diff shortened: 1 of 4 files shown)


src/usr/local/www/head.inc
```diff
@@ -486,6 +486,14 @@ echo '<li>'. get_shortcut_status_link($shortcut_section, false). '</li>';
 echo '<li>'. get_shortcut_log_link($shortcut_section, false). '</li>';
 
 ?>
+	<?php if ($pagetitle === 'Status: Dashboard'): ?>
```
rbgarga: If system is using a different language it will fail

sbeaver-netgate:
I would prefer to wait before merging this PR since I am doing some other layout work on the dashboard at the moment. It may also be better to enable this functionality (once fixed) via a control on the General Setup page, since hiding the already collapsed panel is a matter of preference, and opinions are divided.

PR #2486, opened 2016-01-20
Special Panels - Show/Hide

NOYB:
Add configuration option to show or hide special panels.
When a panel is hidden a button is displayed on the page title bar to toggle the panel display.

### DIFF
--- a/src/usr/local/www/bootstrap/css/pfSense-dark.css
+++ b/src/usr/local/www/bootstrap/css/pfSense-dark.css
@@ -157,6 +157,11 @@ ul.context-links li a {
     border-color: #212121;
 }
 
+.btn-breadcrumbbar {
+	color: #808080;
+    background-color: #404040;
+}
+
 .btn-default.active, .btn-default.focus, .btn-default:active, .btn-default:focus, .btn-default:hover, .open>.dropdown-toggle.btn-default {
     color: #ffffff;
     background-color: #424242;
--- a/src/usr/local/www/bootstrap/css/pfSense-dark.css
+++ b/src/usr/local/www/bootstrap/css/pfSense-dark.css
@@ -157,11 +157,6 @@ ul.context-links li a {
     border-color: #212121;
 }
 
-.btn-breadcrumbbar {
-	color: #808080;
-    background-color: #404040;
-}
-
 .btn-default.active, .btn-default.focus, .btn-default:active, .btn-default:focus, .btn-default:hover, .open>.dropdown-toggle.btn-default {
     color: #ffffff;
     background-color: #424242;
--- a/src/usr/local/www/bootstrap/css/pfSense.css
+++ b/src/usr/local/www/bootstrap/css/pfSense.css
@@ -509,6 +509,11 @@ footer a {
     background-color: #1976D2;
 }
 
+.btn-breadcrumbbar {
+	color: #808080;
+    background-color: #D0D0D0;
+}
+
 .text-warning {
     color: #F57F17;
 }
--- a/src/usr/local/www/bootstrap/css/pfSense.css
+++ b/src/usr/local/www/bootstrap/css/pfSense.css
@@ -509,11 +509,6 @@ footer a {
     background-color: #1976D2;
 }
 
-.btn-breadcrumbbar {
-	color: #808080;
-    background-color: #D0D0D0;
-}
-
 .text-warning {
     color: #F57F17;
 }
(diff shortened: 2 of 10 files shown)


src/usr/local/www/bootstrap/css/pfSense-dark.css
```diff
@@ -157,6 +157,11 @@ ul.context-links li a {
     border-color: #212121;
 }
 
+.btn-breadcrumbbar {
+	color: #808080;
+    background-color: #404040;
+}
```
sbeaver-netgate: I'm not sure it is necessary to change the background color here, or if it is it should be the same color as the breadcrumb background so that there does not appear to be a rectangular box around the text.

Also consider adding a cursor statement here to that the cursor changes to a pointer when hovering over the control.

src/usr/local/www/bootstrap/css/pfSense.css
```diff
@@ -509,6 +509,11 @@ footer a {
     background-color: #1976D2;
 }
 
+.btn-breadcrumbbar {
+	color: #808080;
+    background-color: #D0D0D0;
+}
+
```
sbeaver-netgate: Same as above

src/usr/local/www/head.inc
```diff
@@ -461,8 +461,32 @@ if (($pagename === "index.php") && ($numColumns > 2)) {
...
+				<i class="btn-sm btn-breadcrumbbar">Filter</i>
+			</a>
+		</li>
+	<?php endif ?>
+
+	<?php if ($system_logs_manage_log_form_hidden): ?>
+		<li>
+			<a onclick="$('#manage-log-form').toggle(360)" title="<?=gettext("Show/Hide manage log panel")?>">
+				<i class="btn-sm btn-breadcrumbbar">Manage Log</i>
+			</a>
```
sbeaver-netgate: Does "Manage log" need a gettext()? Maybe not?

src/usr/local/www/system.php
```diff
@@ -506,6 +518,34 @@ function is_timezone($elt) {
...
+$group->add(new Form_Checkbox(
+	'systemlogsmanagelogpanel',
+	null,
+	'Manage Logs',
+	$pconfig['systemlogsmanagelogpanel']
+))->setHelp('Show the Manage Log panel in System Logs.');
+
+$group->setHelp('These options set which special panels to show or hide.  When a panel is hidden a button to show/hide the panel is displayed in the title bar.
+<br /><span class="badge" title="This feature is in BETA">BETA</span>');
+
```
sbeaver-netgate: 'These options allow certain panels to be automatically hidden on page load. A control is provided in the title bar to un-hide the panel.'

src/usr/local/www/head.inc
```diff
@@ -461,8 +461,32 @@ if (($pagename === "index.php") && ($numColumns > 2)) {
 	}
 ?>
 		<ul class="context-links">
-<?php
 
+	<?php if ($dashboard_available_widgets_hidden): ?>
+		<li>
+			<a onclick="$('#widget-available').toggle(360);" title="<?=gettext("Show/Hide available widgets panel")?>">
+				<i class="btn-sm btn-breadcrumbbar">Widgets</i>
+			</a>
```
sbeaver-netgate: Perhaps this doesn't need to be a button, but just (bold) text. That would eliminate the special styling because it would inherit the title bar styling I think ??

sbeaver-netgate:
Overall a very nicely done change. Thanks!

src/usr/local/www/head.inc
phil-davis: As a general rule there should be gettext() around anything that uses an "ordinary" word/s. That way the whole UI retains the potential to be translated into languages with completely different character sets. In "Latin character-based languages" we have some "IT words that are as good as international" and it feels silly to translate those, but other different script-bases need to be considered.
I think the exceptions to translation are some completely technical acronyms that have become so well-used in the IT industry that nobody would expand the acronym, translate it to another language, then make a new acronym out of it - "TCP "IP" "DHCP" "DNS" come to mind.
I am just having a brain-dump on the principle! I am currently doing a pass through the code to gettext() everything, so I am happy to do these after this PR is committed.

sbeaver-netgate: Agreed. There are a few places where it could be questionable, like "squid". Portuguese users are going to have a hard time trying to install "lula" :) I wondered if "Filter" falls into the same category.

sbeaver-netgate:
Thanks for the changes :)

PR #2487, opened 2016-01-20
Solved RADIUS Session-Timeout with wrong AcctStopTime

dariomas:
When Idle-Timeout  set 
$stop_time = $lastact
at line 694, we need to reset 
$stop_time = $pruning_time
for the new iteration.


rbgarga:
I've pushed the change since PR has indentation issues and is just one line change. Thank you for submission!

PR #2496, opened 2016-01-21
Use a single call to dpinger for a coherent view of latency/stddev/loss.

dennypage:


src/etc/inc/rrd.inc
```diff
@@ -984,14 +984,14 @@ for sock in {$g['varrun_path']}/dpinger_*.sock; do
 		continue
 	fi
 
-	gw=\$(/usr/bin/nc -U \$sock | awk '{ print \$1 }')
-
-	if [ -z "\$gw" ]; then
+    t=\$(/usr/bin/nc -U \$sock)
```
rbgarga: Good catch. Can you just fix indent in this line please?

dennypage:
Sorry, I had turned tabs completely off to make it easier for you and I to work on dpinger. Fixed now.

PR #2498, opened 2016-01-21
Fix misplaced quote (resolves syntax error)

NewEraCracker:
My automatic syntax checker found this error:
```Parse error: syntax error, unexpected '=>' (T_DOUBLE_ARROW) in ./pfsense/src/usr/local/www/firewall_nat_out_edit.php on line 513```

I've corrected it.

**edit:**
Added fix for
```Parse error: syntax error, unexpected ';' in ./pfsense/src/usr/local/www/interfaces_wireless_edit.php on line 228```


sbeaver-netgate:
I just knew I would make a couple of those editing 91 files :)  Thanks.

PR #2499, opened 2016-01-21
allow packages in a www/subfolder/ (like haproxy) to use the service start/stop/restart links

PiBa-NL:
allow packages in a www/subfolder/ (like haproxy) to use the service start/stop/restart links

### DIFF
--- a/src/etc/inc/service-utils.inc
+++ b/src/etc/inc/service-utils.inc
@@ -518,7 +518,7 @@ function get_service_control_GET_links($service, $addname = false) {
 			$link = '<a title="%s" href="status_services.php?mode=%s&amp;service=' . $service['name'] . '&amp;zone=' . $service['zone'] . '">';
 		break;
 		default:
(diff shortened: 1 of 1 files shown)


cbuechler:
We've tried to stay consistent with relative links not having a leading /. One reason for that was accessing the web interface through a reverse proxy, though that's a rare circumstance, some complain when things like this are done. That's possibly not reason enough to do that. 

I was thinking there was some other more important reason to not have that leading / though. @jim-p you recall anything there?

jim-p:
IIRC I had the same question years ago and a proxy was the only reason we could come up with, we've had absolute links in 2.2.x in some places though, I thought we'd given up on limiting that after 1.2.x but my memory is hazy.

jim-p:
Looks OK to merge to me, all of the links in head.inc (formerly in fbegin.inc) are absolute, not relative, so this would actually make it consistent with what's being done now.

PR #2501, opened 2016-01-22
Crash Reporter compare choice with language-specific string #5792

phil-davis:
I don't really understand why this has to be - the HTML has value="Yes" and value="No" - I thought those would be the strings sent along with "submit".
The translation of Yes and No should have been just for display purposes.
But after this change, I can submit or delete crash reports when the UI isin other languages.

### DIFF
--- a/src/usr/local/www/crash_reporter.php
+++ b/src/usr/local/www/crash_reporter.php
@@ -99,7 +99,7 @@ $crash_report_header .= "\nCrash report details:\n";
 exec("/usr/bin/grep -vi warning /tmp/PHP_errors.log", $php_errors);
 ?>
 <?php
-	if (gettext($_POST['Submit']) == "Yes") {
+	if (gettext($_POST['Submit']) == gettext("Yes")) {
 		echo gettext("Processing...");
 		if (!is_dir("/var/crash")) {
 			mkdir("/var/crash", 0750, true);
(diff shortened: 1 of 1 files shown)


NOYB:
In looking at the html form buttons have value="Yes" and value="No".  Those are what is posted and is not gettext'ed.
So would think the following is what was intended.  Test the posted submit value without gettexting it.
`if ($_POST['Submit'] == "Yes") {`
`if ($_POST['Submit'] == "No") {`

phil-davis:
That's what I thought, but that did not work???

NOYB:
Hmm, that is strange.  It works like that here for me.

This is cut and paste directly from crash_reporter.php.

`if ($_POST['Submit'] == "Yes") {`
. . .
`} else if ($_POST['Submit'] == "No") {`

`<button class="btn btn-primary" name="Submit" type="submit" value="Yes"> . . .`
`<button class="btn btn-default" name="Submit" type="submit" value="No"> . . .`

phil-davis:
Ohhh! I need to actually read the code.
Now I see that someone already had gettext($_POST['Submit']) - which was the thing that I should have removed. It was dumb to gettext() the posted value and compare it to the literal string "Yes". It is nearly as dumb to also gettext('Yes') and then compare :)
I will make another PR to clean it up.

rbgarga:
Yes, please. I was reading the code right now and would comment about it

phil-davis:
https://github.com/pfsense/pfsense/pull/2505

PR #2503, opened 2016-01-22
Form Group - Empty Title GetText

NOYB:
Don't call gettext with an empty title.

There are undoubtedly other places that a check should be done before passing a variable only to a gettext call.

Fixes use case of form group class without a title from passing empty var to gettext.
Gettext returns header info. when an empty string/var is passed.

Perhaps a function should be created for this for consistency.
http://php.net/manual/en/function.gettext.php#108594

For groups without a title gettext returns header info.

To reproduce the issue select a non English language and open any of the following:
Services Load Balancer - edit
Diagnostics - Backup/Restore
Diagnostics - Status SMART
Status - System Logs - log filter and manage log panels
(not a comprehensive list)

### DIFF
--- a/src/usr/local/www/classes/Form/Group.class.php
+++ b/src/usr/local/www/classes/Form/Group.class.php
@@ -148,8 +148,9 @@ EOT;
 		$label = new Form_Element('label', false, ['for' => $target]);
 		$label->addClass('col-sm-'.Form::LABEL_WIDTH, 'control-label');
 
-		if (!empty(trim($this->_title)) || is_numeric($this->_title))
+		if (!empty(trim($this->_title)) || is_numeric($this->_title)) {
 			$title = htmlspecialchars(gettext($this->_title));
+		}
 
 		return <<<EOT
 	{$element}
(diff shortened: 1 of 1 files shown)


src/usr/local/www/classes/Form/Group.class.php
```diff
@@ -148,7 +148,8 @@ public function __toString()
 		$label = new Form_Element('label', false, ['for' => $target]);
 		$label->addClass('col-sm-'.Form::LABEL_WIDTH, 'control-label');
 
-		$title = htmlspecialchars(gettext($this->_title));
+		if (!empty(trim($this->_title)) || is_numeric($this->_title))
+			$title = htmlspecialchars(gettext($this->_title));
```
rbgarga: good catch, thanks. Just add { } please

NOYB: So not using {} for one liners seems to be the norm in all the classes code.  So was just following what seemed to be the norm there.  Given that.  Should the {} be added or not?  I'm okay either way.  Just let me know the official code style policy for the classes and I'll do that.  Thanks.

rbgarga: According developer guide it should be added. @phil-davis spent a long time in the past adding it all around, with all other style fixes. Unfortunately classes code was merged and diverge in some rules, we will need to fix it all in the future, but the idea is to avoid introducing new cases to be fixed later

NOYB: Okay thanks for the clarification.  I'll fix it right up.

PR #2506, opened 2016-01-22
Experimental fix to correct 2.2 to 2.3 upgrade (RELENG_2_2)

NewEraCracker:
This applies to pfSense platform (the one installed from LiveCD to HDD). It prevents a bunch of php errors related to extension loading failures. Other platforms have not been tested and thereby this procedure was not implemented.

**This has been tested in the following way:**
1. Installed 2.2.6 amd64 from LiveCD to HDD
2. Enabled SSH (and thereby SFTP capabilities)
3. Uploaded the modified file
4. Did a firmware upgrade procedure to 2.3 via Web GUI using latest full upgrade snapshot

Previously, doing a firmware upgrade procedure to 2.3 would yell a lot of php errors during post_upgrade_command tasks. This commit fixes (or workarounds it) for pfSense platform.

Thank you! :)

### DIFF
--- a/etc/rc.firmware
+++ b/etc/rc.firmware
@@ -477,6 +477,12 @@ pfSenseupgrade)
 	# Remove saved commit ID for gitsync
 	rm -f /etc/version.gitsync
 
+	PFSENSETYPE=`cat /etc/platform`
+
+	if [ "${PFSENSETYPE}" = "pfSense" ]; then
+		/etc/rc.php_ini_setup
+	fi
+
 	# If /tmp/post_upgrade_command exists after update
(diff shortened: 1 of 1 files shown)


rbgarga:
See comments on #2507

NewEraCracker:
This PR should be relatively be safe because it will only apply when 2.2.x is used to upgrade to a newer distribution (2.3+). I have tested it and found no problems.

Still I believe it needs further testing just to be sure it works in all scenarios.

rbgarga:
Exactly! It's not something we should do at this point of 2.2

PR #2507, opened 2016-01-22
Experimental fix to correct 2.2 to 2.3 upgrade (master)

NewEraCracker:
See PR #2506

The method of this PR was **not tested** as it would require modifying the snapshot tarball, something I don't know how to do.

This places the same fix on the *post_upgrade_command* file.
Please double-check if this fix alone (in a system without #2506 applied) works.

Thanks. :)

### DIFF
--- a/src/usr/local/share/pfSense/post_upgrade_command
+++ b/src/usr/local/share/pfSense/post_upgrade_command
@@ -4,6 +4,10 @@
 
 PFSENSETYPE=`cat /etc/platform`
 
+if [ "${PFSENSETYPE}" = "pfSense" ]; then
+	/etc/rc.php_ini_setup
+fi
+
 if [ "${PFSENSETYPE}" = "pfSense" -o "${PFSENSETYPE}" = "nanobsd" ]; then
 	touch /conf/needs_package_sync_after_reboot
 fi


rbgarga:
I didn't test but I strong believe this can cause a big damage in some cases. Have you tried to upgrade from 2.0 or 2.1 to 2.3 using this change?

2.0 / 2.1 are based on FreeBSD 8.x, after extract 2.3 binaries system ends up with 8.x kernel and 10.x binaries, and I believe /etc/rc.php_ini_setup will not run properly at this point.

NewEraCracker:
Good point. I have not tested this change, in the current form, at all because I'd need a custom generated snapshot with this change implemented.

For testing purposes, I've instead edited the firmware updater itself of 2.2 on #2506.

The problem I am having with upgrades (untouched 2.2.6 -> 2.3 snapshots) are errors running post_upgrade_command.php invoked from post_upgrade_command script (fails to load all extension modules + throws an error when calling textdomain as gettext extension is not loaded).

The php file has a shebang which invokes the php executable which was updated during tarball extraction into the filesystem and thus requires config regeneration (php.ini) to be able to correctly execute.

As you said, the consequences of doing this on 2.1 could cause a big damage. But from my point of view - after taking your considerations in account - the process, as is, is already flawed.

This needs a closer look by someone that is deeply familiar with the internals of pfsense before being considered a final solution.

jim-p:
The upgrade itself is not failing. It's failing to redraw the banner when the upgrade finishes and the user gets logged out just as it reboots. It's ugly but harmless and there isn't a good way around it because at that point the kernel and world are mismatched. It won't be a problem after 2.3 due to the way the upgrade process has been changed already. I'm not sure this is worth messing with.

NewEraCracker:
Thanks for your input. The errors on post_upgrade_command.php are not fatal to the entire upgrade procedure but would be nice if they get fixed.

If fixing them, would bring more problems, well... then it is better to leave things as is, and hope for the best.

jim-p:
The error is not coming from post_upgrade_command.php, the entire upgrade process is finished before the error is printed, see here: https://gist.github.com/jim-p/60584be7afe41beca7cf
It has started to reboot before the error displays, so it's done with this script. What it's failing to do is redraw the banner on the menu (see /etc/rc.banner in the printed stack trace). It's failing to run the 2.3 version of the banner with 2.3's php on the system as it reboots, which is no surprise. It's already rebooting at the time after the upgrade succeeded.

NewEraCracker:
My mistake. Closing this.

PR #2508, opened 2016-01-22
add call 'gettext' function in the status menu files

brunostein:
Added call to 'gettext' function for translation.


src/usr/local/www/status_filter_reload.php
```diff
@@ -135,16 +135,16 @@ function update_data(obj) {
...
 		jQuery('#status').effect('highlight');
-		jQuery('#status').html('Done.  The filter rules have been reloaded.');
+		jQuery('#status').html('<?=gettext("Done.  The filter rules have been reloaded.");?>');
 		jQuery('#reloadinfo').css("visibility", "hidden");
 		jQuery('#doneurl').css("visibility", "visible");
-		jQuery('#doneurl').html("<p><a href='status_queues.php'>Queue Status<\/a><\/p>");
+		jQuery('#doneurl').html("<p><a href='status_queues.php'><?=gettext("Queue Status");?><\/a><\/p>");
 	}
 	window.setTimeout('update_status_thread()', 2500);
 }
```
sbeaver-netgate: Perhaps we should replace the 'jQuery's with '$' while you are editing?

src/usr/local/www/status_captiveportal.php
```diff
@@ -259,7 +259,7 @@ function clientcmp($a, $b) {
 	<div class="panel-heading"><h2 class="panel-title"><?=gettext("Captive Portal Status")?></h2></div>
 	<div class="panel-body"><br />
 <?php
-	print_info_box(gettext("No captive portal zones have been configured. You may add new zones here: ") . '<a href="services_captiveportal_zones.php">' . 'Services->Captive portal' . '</a>');
+	print_info_box(gettext("No captive portal zones have been configured. You may add new zones here: ") . '<a href="services_captiveportal_zones.php">' . gettext('Services->Captive portal') . '</a>');
 ?>
 	</div>
 </div>
```
sbeaver-netgate: The menus use gettext, so adding it here makes sense. Thanks.

sbeaver-netgate:
Nice set of changes. Thanks!

PR #2515, opened 2016-01-22
Fix syntax error. Missing semicolon.

NewEraCracker:


NewEraCracker:
This kind of errors can be busted pretty easily using a script like this
```sh
#!/bin/sh

PHP_EXECUTABLE=`which php`

if [ -x "${PHP_EXECUTABLE}" ]; then
	find . -type f -name "*.inc" -exec "${PHP_EXECUTABLE}" -n -l {} \; | grep -v "No syntax errors detected"
	find . -type f -name "*.php" -exec "${PHP_EXECUTABLE}" -n -l {} \; | grep -v "No syntax errors detected"
else
	echo "PHP executable not found!"
fi
```
Feel free to add something like this to a test-suite or something ;)

jim-p:
Already fixed. I have a git hook that usually flags such things but it must have broken since it didn't warn me.

PR #2517, opened 2016-01-22
Add script to lint all files with PHP code

NewEraCracker:
Since there are sometimes syntax errors, having a script to perform some automated checking is a good idea. This is the first shell script I've coded, so I believe there is plenty of room for improvement.

It will output currently installed PHP version, then scan current directory and run php lint on all *.inc and *.php files and output any errors if found.

Tested & Working while running on pfSense shell.

### DIFF
--- a/tools/scripts/php_lint.sh
+++ b/tools/scripts/php_lint.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+#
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
Thanks. Most of the devs use pretty much this script on their workstations. Might be good for you to post it on the forum for others to learn from.comment on.

PR #2522, opened 2016-01-23
Remove print_info_box_np

phil-davis:
Put the real code into print_info_box, adding just a $btntext parameter
to the existing 3 params of print_info_box.
Change the remaining print_info_box_np() calls to use print_info_box()

Note: The previous code also had a "bug" - the "apply" button said
"apply" literally on the UI, when it should have said "Apply changes".
That is fixed so that now the "Apply changes" or other specified button
text is actually displayed to the user. (In the old code it was
literally displaying $name when it should have been $value)

### DIFF
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -307,57 +307,35 @@ function verify_gzip_file($fname) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: https://github.com/pfsense/pfsense/pull/2521 needs to also be committed before (or at the same time as) this.

sbeaver-netgate:
Thanks Phil. I like this version better :)

PR #2527, opened 2016-01-25
Save selected wireless interface

phil-davis:
This looks like a bug. I don't have a system with the necessary hardware running 2.3 right now to really test, but the code that processes the submit/apply looks for $_POST['if']
Please double-check on suitable hardware!


jim-p:
Looks good to me, especially comparing it to the 2.2.x version of the page.

PR #2529, opened 2016-01-25
Enhance error output when package file not found

phil-davis:
I had a system that had upgraded but the Notes package had not managed to (re)install. In that situation the Notes item is on the Status menu but there is no notes.xml or other notes code on the the system.
When I go to Status->Notes I get:
Warning: substr_compare(): The start position cannot exceed initial string length in /usr/local/www/pkg.php ...

The warning is because realpath() returns false in this case and cannot be used is a parameter to substr_compare().

Handle this case, and make the error message more informative.


phil-davis:
See https://github.com/pfsense/pfsense/pull/2528 for the equivalent code for RELENG_2_2.

PR #2533, opened 2016-01-25
Add leaving page message to NAT pages

phil-davis:
These pages use the draggable/sortable rows stuff, so could use the same
warning message if the user tries to change page before saving


sbeaver-netgate:
I was planning to add that when I add the separator row things to those pages later this week so you have saved me the trouble :)

Unfortunately the custom text is ignored by Firefox now. "Security", they claim. You will just see a standard browser message.

phil-davis:
I am using Chrome at the moment, so I get to see the custom message :)

PR #2534, opened 2016-01-25
Internationalization of pkg*.php

phil-davis:
Various sentences were being built up from gettext() of individual words
or phrases concatenated together. That will not work for translation to
languages with different word order, so I have restructured that.
Note: TODO: The countdown timer text with seconds countdown is all in JS
so I will think about how that can be translated nicely.

### DIFF
--- a/src/usr/local/www/pkg_mgr_install.php
+++ b/src/usr/local/www/pkg_mgr_install.php
@@ -292,13 +292,14 @@ if ($input_errors) {
 
 	switch ($pkgmode) {
 		case 'reinstallpkg':
-			$pkgtxt = 'reinstalled';
+			$pkgtxt = sprintf(gettext('Package <b>%s</b> will be reinstalled'), $pkgname);
 			break;
 		case 'delete':
-			$pkgtxt = 'removed';
+			$pkgtxt = sprintf(gettext('Package <b>%s</b> will be removed'), $pkgname);
 			break;
+		case 'installed':
 		default:
-			$pkgtxt = $pkgmode;
+			$pkgtxt = sprintf(gettext('Package <b>%s</b> will be installed'), $pkgname);
 			break;
 	}
 ?>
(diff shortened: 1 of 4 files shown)


sbeaver-netgate:
You can generally uses gettext there too. e.g.:

$('#some-element').val( "<?=gettext('This is the text to display')?>");

src/usr/local/www/pkg_mgr_install.php
```diff
@@ -389,12 +390,25 @@ function waitfor_string_in_file($filename, $string, $timeout) {
...
+	$pkg_wait_txt = sprintf(gettext('Please wait while the removal of <b>%1$s</b> completes.'), $pkgid);
 } else if (($_POST['mode'] == 'reinstallpkg') || ($_POST['mode'] == 'reinstallall')) {
-	$modetxt = gettext("reinstallation");
+	$panel_heading_txt = gettext("Package reinstallation");
+	$pkg_success_txt = sprintf(gettext('<b>%1$s</b> reinstallation successfully completed'), $pkgid);
+	$pkg_fail_txt = sprintf(gettext('<b>%1$s</b> reinstallation failed!'), $pkgid);
+	$pkg_wait_txt = sprintf(gettext('Please wait while the reinstallation of <b>%1$s</b> completes.'), $pkgid);
 } else {
-	$modetxt = gettext("installation");
+	$panel_heading_txt = gettext("Package installation");
```
sbeaver-netgate: The same code is used when updating the base system. Is the word "Package" still appropriate then?

phil-davis: There is a separate if () lower down that has the special text for updating the base system
gettext("Updating system")
So $panel_heading_txt is never used in that case.

sbeaver-netgate: OK thanks

phil-davis:
The tricky part is that there is a JS var for the number of seconds remaining that has to be substituted in the middle of the sentence, and that var keeps changing in real-time. So the fully-translated sentence has to be adjusted for the countdown timer.
Anyway, it is late here - if you are happy with what I have done here then commit it. I (we...) can sort out how to do internationalization of variable JS later.

sbeaver-netgate:
Look at the same code in diag_reboot.php Looks like it is appropriate there, but that I copied it to the pkg system before doing that.

PR #2535, opened 2016-01-26
Use template string for JS countdown in diag_reboot

phil-davis:
Tested on Windows10 with:
Chrome 47.0.2526.111 m
Firefox 38.5.2
Edge (current with all Windows 10 updates)

This will allow translators to translate a whole string:
'Page will automatically reload in ${time} seconds.'
into some different language word order, retaining the text of the
${time} reference, like:
'Page ${time} seconds in automatically will reload'
and that should get JS substitution of the time var as it counts down in
real-time.
This provides similar functionality to PHP
sprintf(gettext('Page will automatically reload in %s seconds.'), $time);

Please test, and if this seems like the way to go for this sort of
thing, then we can do similar in other places that have JS vars that are
variable at run-time and need to be substituted into sentences at
run-time.

### DIFF
--- a/src/usr/local/www/diag_reboot.php
+++ b/src/usr/local/www/diag_reboot.php
@@ -76,7 +76,6 @@ $guiretry = 20;		// Seconds to try again if $guitimeout was not long enough
 $pgtitle = array(gettext("Diagnostics"), gettext("Reboot System"));
 include("head.inc");
 
-
 if ($_SERVER['REQUEST_METHOD'] == 'POST') {
 	if (DEBUG) {
 	   print_info_box(gettext("Not actually rebooting (DEBUG is set true)"), 'success');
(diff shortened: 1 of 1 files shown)


phil-davis:
Here is a reference to Template Strings in JS
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings

sbeaver-netgate:
I wonder if this might be a more straightforward solution:

sprintf(gettext("Rebooting.%s The page will automatically reboot in %s seconds."), "<br />", "<span id='seconds'></span>"));

The the JS will update '#seconds'

sbeaver-netgate:
I'd like to try an alternative solution before merging this

sbeaver-netgate:
Phil,
Please take a look at diag_reboot.php where I have tried something a little different. Let me know which solution you prefer and we can make both instances that way.

phil-davis:
@sbeaver-netgate Yes, your way is good - I hadn;t thought about the <span> tag for a bit of run-time replaceable text like that.

PR #2560, opened 2016-01-27
status_ntpd whitespace

phil-davis:
Various lines were not on tab stops, some PHP end tags were on the end of lines and others on new lines... I found it a bit hard to read when tracking through a previous missing PHP start tag. Maybe this is better? or maybe I have gone too crazy re-formatting it?


sbeaver-netgate:
I much prefer that (your) style of <?php tagging :)

PR #2571, opened 2016-01-29
Panels Overflow - Horizontal Scroll Bar

NOYB:
Been looking into ways to eliminate the horizontal scroll bar in the available widgets, log filter, and manage log panels.  Not sure which would be preferred.  Using the specific panel ID or the class hierarchy for applying more globally.


sbeaver-netgate:
Thanks for looking at this. It had been driving me crazy :)

PR #2573, opened 2016-01-29
Remove executable bit of all static content in www

NewEraCracker:
There are multiple static files that have their executable bit set, while they should not be executable at all. This PR changes mode of all of them from 0755 to 0644.

Thanks.


cbuechler:
merged, thanks!

PR #2574, opened 2016-01-30
Fix issues #3900 and #5051 to make No-ip round robin DNS works.

tiagobar:
The paramater "&hostname" should be changed to "&h[]" (issue #5051).
I think this affect only NO-IP DNS that uses round robin. Anyway the no-ip official client uses the "&h[]" parameter.

There's also a change to Dynamic DNS hostname allow '@' character (issue #3900).

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -395,7 +395,7 @@
 					} else {
 						$iptoset = $this->_dnsIP;
 					}
-					curl_setopt($ch, CURLOPT_URL, $server . $port . '?username=' . urlencode($this->_dnsUser) . '&pass=' . urlencode($this->_dnsPass) . '&hostname=' . $this->_dnsHost.'&ip=' . $iptoset);
+					curl_setopt($ch, CURLOPT_URL, $server . $port . '?username=' . urlencode($this->_dnsUser) . '&pass=' . urlencode($this->_dnsPass) . '&h[]=' . $this->_dnsHost.'&ip=' . $iptoset);
 					break;
 				case 'easydns':
 					$needsIP = TRUE;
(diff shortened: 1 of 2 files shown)


NewEraCracker:
I had an idea after reading countless posts about this. Code could be relaxed even further to allow more customization.
```
		/* Namecheap can have a @. in hostname */
		if ($pconfig['type'] == "namecheap" && substr($_POST['host'], 0, 2) == '@.') {
			$host_to_check = substr($_POST['host'], 2);
		} else {
			$host_to_check = $_POST['host'];
			/* No-ip (and maybe others) can have a @ in hostname */
			$last_to_check = strrpos($host_to_check, '@');
			if ($last_to_check !== false) {
				$host_to_check = substr_replace($host_to_check, '.', $last_to_check, 1);
			}
			unset($last_to_check);
		}
```
To:
```
// Namecheap can have a @. or a *. in the start of hostname
// No-ip (and maybe others) can have a @ in hostname
$host_to_check = preg_replace(['/^@\\./', '/^\\*\\./', '/@/'], ['', '', '.'], $_POST['host']);
```
This way, far more inputs would pass the validation.
See: https://3v4l.org/A0UES

(maybe I am exaggerating too much, so I'd appreciate all feedback)

tiagobar:
I like your idea @NewEraCracker and I have some considerations:

1) With no-ip domains only the last "@" must be accepted. That's the reason I replace it with a ".", to produce a valid domain and invalidate it if there's others "@".
The regex below should be ok:
 
`
$host_to_check` = preg_replace(['/^@\\./', '/^\\*\\./', '/(@(?!.*@.*))/'], ['', '', '.'], `$_POST['host']);
`

2) One difference from previous code is that it doesn't check for specific hosts rules and turn on all validations for all host types. 
For example, all hosts will accept the leading "@." used by "namecheap".
I think this shouldn't be a problem. The validation's goal is provide basic sintax/typos validation and specific sintax should be user responsability.

Anyway I guess some senior dev could state what rule best fit to pfSense policy.

PR #2576, opened 2016-01-30
HTML Compliance - VPN / IPsec / Tunnel - Edit Phase II

NOYB:
Duplicate ID ealgos[].
Duplicate ID halgos[].

### DIFF
--- a/src/usr/local/www/vpn_ipsec_phase2.php
+++ b/src/usr/local/www/vpn_ipsec_phase2.php
@@ -639,14 +639,14 @@ foreach ($p2_ealgos as $algo => $algodata) {
 	$group = new Form_Group($i == 0 ? 'Encryption Algorithms':'');
 	$group->addClass('encalg');
 
-	// Note: ID attribute of each element created is to be unique.
+	// Note: ID attribute of each element created is to be unique.  Not being used, suppressing it.
 	$group->add(new Form_Checkbox(
 		'ealgos[]',
 		null,
 		$algodata['name'],
 		(is_array($pconfig['ealgos']) && in_array($algo, $pconfig['ealgos'])),
 		$algo
-	))->addClass('multi')->setAttribute('id', 'ealgos[]' . '_' . $algodata['name']);
+	))->addClass('multi')->setAttribute('id');
 
 	if (is_array($algodata['keysel'])) {
 		$list = array();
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
I'd like to run some tests on this proposal before merging.

NOYB:
I think that's a good idea.  I didn't see anyplace where the ID's are being used and they are automatically created by the class.  But that doesn't that certainly doesn't make it's a sure thing.  Thanks.

sbeaver-netgate:
I think the existing code is correct. When you define multiple checkboxes with the same ID, they are treated as an array of elements each with a different value. That is how the page validation/save is configured to work. The 'multi' class is what triggers the use of an array.

NOYB:
The page validation is using the checkbox name attribute though is it not?  ID's are always to be unique.  Don't see where the page is using the ID's.  The ID's in this case appear to just be a side effect of the class creating ID's for everything with the same value as the name.  It's basically the same problem that existed with the display as radio buttons.  Except here it is caused by the for loop.

sbeaver-netgate:
If you print_r the $POST data you will see that it contains:

    [ealgos] => Array
        (
            [0] => aes
            [1] => aes192gcm
            [2] => blowfish
            [3] => cast128
        )

That is exactly the same as 2.2.x - You can also see that the information is being correctly checked and saved in the config file as:

<encryption-algorithm-option>
	<name>aes</name>
	<keylen>auto</keylen>
</encryption-algorithm-option>
<encryption-algorithm-option>
	<name>aes192gcm</name>
	<keylen>auto</keylen>
</encryption-algorithm-option>
<encryption-algorithm-option>
	<name>blowfish</name>
	<keylen>auto</keylen>
</encryption-algorithm-option>
<encryption-algorithm-option>
	<name>cast128</name>
</encryption-algorithm-option>

Which again is the correct format for pfSense.

Everything seems to work as expected no?

NOYB:
Yes, I understand that.  But that is all using the name attribute is it not? GET and POST methods use the name attribute do they not?  Not the ID attribute.  The ID attribute would be used for client side scripts to access the individual elements.  Look at the HTML produced.  All those elements have the same ID.  How would client side scripts access those individual elements when they all have the same ID?  They are not an array in the HTML.

Also it does not appear to me that their is anything using the ID's.  If so then they could just as well be removed.  The only reason I didn't remove them is because they are created by the class.  Which is fine but the are required to be unique.  So when put in a for loop that needs to be accounted for.

sbeaver-netgate:
True.
As long as the functionality is not broken, I have no objection to removing the name (with a suitable comment explaining why)

NOYB:
Not the name.  The ID.  But the class creates the ID attribute from the name.  So the ID can't really be removed unless the class is changed to not create ID attributes from the name attribute.  In which case everywhere a class created ID attribute is used it would need to be added to the list of arguments of the class and added throughout the code base.  Starting from scratch I'd say that's the way it should be designed.  But probably more work than worth at this point.  So the best option I think is to handle it in the for loop.

I don't know how to tell the class to suppress creation of the ID attribute though.  If there is, I'm all ears to that.  So appending the setting to the name to make the ID's unique.

PR #2583, opened 2016-02-04
Status / DHCP leases - Root Cause Sort Issue

NOYB:
Sort routine apparently does not honor colspan attribute.

Here are three possible solutions:
1) Add a description column instead of spanning the start and end columns.  This is the method presented in this commit.

2) Include an empty second column with the description instead of spanning.  This would not be very display/sort friendly though because it would be sharing the start column with the description.

3) Display two empty cells when lease type is static, instead of the description.  This is the way status DHCPv6 leases currently works.

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -379,11 +379,12 @@ if (count($pools) > 0) {
 					<th><?=gettext("IP address")?></th>
 					<th><?=gettext("MAC address")?></th>
 					<th><?=gettext("Hostname")?></th>
+					<th><?=gettext("Description")?></th>
 					<th><?=gettext("Start")?></th>
 					<th><?=gettext("End")?></th>
-					<th data-sortable="false"><?=gettext("Online")?></th>
-					<th data-sortable="false"><?=gettext("Lease Type")?></th>
-					<th data-sortable="false"><?=gettext("Actions")?></th>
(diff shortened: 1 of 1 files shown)


src/usr/local/www/status_dhcp_leases.php
```diff
@@ -379,11 +379,12 @@ function remove_duplicate($array, $field) {
 					<th><?=gettext("IP address")?></th>
 					<th><?=gettext("MAC address")?></th>
 					<th><?=gettext("Hostname")?></th>
+					<th><?=gettext("Description")?></th>
 					<th><?=gettext("Start")?></th>
 					<th><?=gettext("End")?></th>
-					<th data-sortable="false"><?=gettext("Online")?></th>
-					<th data-sortable="false"><?=gettext("Lease Type")?></th>
-					<th data-sortable="false"><?=gettext("Actions")?></th>
```
sbeaver-netgate: Perhaps the actions column should remain un-sortable?

sbeaver-netgate:
Thanks for looking into this.

sbeaver-netgate:
Merged but with unsortable actions column

src/usr/local/www/status_dhcp_leases.php
NOYB: Don't think it really matters one way or the other.  Not much point in sorting on the actions column, but not a problem for it to be sortable either.  So either way is fine by me.  Thanks.

PR #2587, opened 2016-02-06
Don't kill an sshd parent process on exit from /etc/rc.initial

davidjwood:
This stops exiting a shell running in an SSH session from terminating the parent sshd process, which has the unwanted side effect of killing all other sessions running under that sshd process (including cloned shell sessions).

Cloned sessions are commonly encountered with Vandyke's SecureCRT client. It can be a huge problem to CTRL-D out of one shell only to find all other shells running in cloned sessions have been killed!

### DIFF
--- a/src/etc/rc.initial
+++ b/src/etc/rc.initial
@@ -171,7 +171,12 @@ case ${opmode} in
 	fi
 	;;
 "")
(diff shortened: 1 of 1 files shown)


cbuechler:
Merged, thanks!

PR #2590, opened 2016-02-06
Firewall / Rules - Empty Horizontal Scrollbar

NOYB:
Fixes #5559
Also fixes display logic of the no rues messages.

If there are any rules defined, but none to display for the selected interface (all rules are hidden) IE will produce a horizontal scrollbar.
To prevent this add an empty unhidden row when there are no rules to display for the selected interface.

Also use rules displayed count for the no rules messages logic.  Otherwise it will only be displayed if none of the interfaces has any rules defined.  Just a single rule defined on any interface will prevent the message from being displayed on all interfaces.

Hide tbodies, messages, and buttons until loaded for cleaner page rendering appearance.

### DIFF
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -398,7 +398,7 @@ $columns_in_table = 13;
 
 <?php if ($showblockbogons || $showantilockout || $showprivate) :
 ?>
-				<tbody>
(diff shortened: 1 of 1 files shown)


NOYB:
Closed in favor of Firewall / Rules - Page Efficiency Upgrade #2600

PR #2598, opened 2016-02-08
Limit interface MTU to 1500 if no explicit MTU configured

davidjwood:
This fixes a regression introduced by 5815d13659250c8319424268a1c8836928cd8cbf . Following that commit, VLANs on the same parent interface as a PPPoE parent interface finish up with an MTU of (PPPoE MTU + 8) if no MTU is configured explicitly. A LAN MTU of 1508 without explicit configuration was rather unexpected!

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3131,7 +3131,7 @@ function interface_vlan_adapt_mtu($vlanifs, $mtu) {
 		    !empty($config['interfaces'][$assignedport]['mtu'])) {
 			$if_mtu = $config['interfaces'][$assignedport]['mtu'];
 		} else {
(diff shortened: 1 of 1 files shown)


cbuechler:
Good catch, David. Though that fix isn't workable because it'll change behavior of jumbo frames on VLANs. 

There is likely a better general fix. It's doing what the comment there deems necessary - "All vlans need to use the same mtu value as their parent" - but that isn't actually true (at least on FreeBSD 10.x and newer, may have been diff at the time that was done, which was on an 8.x base). The parent interface MTU just must be >= the VLAN MTUs. 

Needs a bit more digging into to determine the best course of action to fix. Opened bug. https://redmine.pfsense.org/issues/5857

PR #2600, opened 2016-02-08
Firewall / Rules - Page Efficiency Upgrade

NOYB:
Take it for a test drive and see what you think.  May need to reboot after gitsync for the config upgrade to take hold to re-index the separators.  Don't forget to clear browser cache.

1) Only load the rules of the selected interface instead of loading all the rules of all interfaces and hiding them.  Very inefficient and can result in sluggishness.  Especially with large number of rules and/or large alias tables.

1.1) Rule separators indexed to their position within their interface so they work with only loading the selected interface rules.

2) Supports consecutive separators.

3) Fixes #5559 - horizontal scrollbar when no rules to display (all hidden).  No longer hiding rules.

4) Corrects case where a single rule defined on any interface would prevent the alert messages from being displayed on all interfaces.


sbeaver-netgate:
Thanks for this work. Obviously it will take a while to thoroughly test, but I will try to get to it today or tomorrow.

NOYB:
One caveat I know of is that in floating rules adding a rule via copy/dup doesn't move the separators.  Checked in previous build (built on Sat Feb 06 14:22:10 CST 2016) and behavior was the same.  Appears to be using a different code path.

NOYB:
Figured out the cause for copy/dup in floating rules not working.  The $if var is set to 'wan'.
Not sure where to implement a fix that only applies to copy/dup floating rules though.

sbeaver-netgate:
I have completed half a day of testing and as far as I can see this update works nicely. Thanks a lot for all the work.

Merged.

PR #2604, opened 2016-02-08
Verify types of input username and password

NewEraCracker:
This should prevent the possibility of illegal offsets. If you poke pfSense WebGUI you'll find some funny stuff. Username and Password should never be arrays at all.

Fussing version 2.2.6 with vulnerability auditing tools throws things like:
```Illegal offset type in isset or empty in /etc/inc/auth.inc on line 229```
```crypt() expects parameter 1 to be string, array given in /etc/inc/auth.inc on line 305```
```md5() expects parameter 1 to be string, array given in /etc/inc/auth.inc on line 311```

Thanks.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1530,6 +1530,10 @@ function getUserGroups($username, $authcfg, &$attributes = array()) {
 
 function authenticate_user($username, $password, $authcfg = NULL, &$attributes = array()) {
 
+	if (is_array($username) || is_array($password)) {
+		return false;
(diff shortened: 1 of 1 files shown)


zBart:
Wouldn't [is_string()](https://secure.php.net/is_string) make more sense here?

NewEraCracker:
That could also be an option. For now the objective was just to prevent any illegal uses where an array should not be input at all, but any other type should not yell any errors.

This implementation is open for changes in the future by someone more experienced.

PR #2607, opened 2016-02-09
Fix #3209 editing unassigned VLAN tag can change an assigned interfac…

phil-davis:
…e - RELENG_2_2

Might as well fix this bug in 2.2.* as well, since it is easy and was obviously an error.


cbuechler:
Yeah might as well, merged, thanks @phil-davis

PR #2615, opened 2016-02-10
Firewall / Rules - Page Efficiency Upgrade

NOYB:
Add is_array test and change !empty test to is_array test.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -4243,21 +4243,23 @@ function ifridx($if, $ridx) {
 	}
 	
 	$i = $ifridx = 0;
-	foreach ($config['filter']['rule'] as $rulen => $filterent) {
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
I already pushed a change that is almost the same as this and the DEV team have tested it. Would you take a look at that please and let me know if this PR adds anything?

cbuechler:
@sbeaver-netgate wasn't the change you're referencing was just in upgrade_config.inc? 

This looks legit. The checking whether $config['filter']['rule'] is an array should always be unnecessary since I don't think short of manually editing the config it's possible for that to not be an array, but it's a good check to have regardless. The foreach on $separators likely hasn't triggered an error because it's never been passed anything other than an array, but is_array is better than !empty there and may prevent PHP from throwing errors in the future.

sbeaver-netgate:
Yes. You are right. The code is so similar I didn't notice this was another instance. This is fine.

NOYB:
Yeah I think you are referring to the upgrade script function.
That triggered me to take a look at these two functions in filter.inc.

PR #2621, opened 2016-02-11
Logic correction

NewEraCracker:
From my understanding ppp['mtu'] and ppp['ports'] are directly related with each other and they correlate information about mtu and ports. This specific commit doesn't introduce anything new other than simplifying the code slightly.

All feedback is appreciated as I dig into the internals of pfSense.


loos-br:
Committed, thanks!

PR #2627, opened 2016-02-12
Remove German from English message text

phil-davis:


sbeaver-netgate:
Vielen Dank Herr Davis.

PR #2635, opened 2016-02-14
Do not echo printf return value

NewEraCracker:
The current implementation would echo both the contents and printf return value which is incorrect. This commit fixes the issue.


sbeaver-netgate:
Perhaps <?=sprintf(... )?> would accomplish the same thing, but be more in line with the style of project?

NewEraCracker:
Yes. I can change to that if you wish.

sbeaver-netgate:
Thanks :)

NewEraCracker:
Done.

PR #2654, opened 2016-02-17
Status / System Logs - Associated Panels - Remove Icon Text Buffer Space

NOYB:
The space that was being inserted as a buffer between button icon and text is no longer necessary. Recently addressed in css.


sbeaver-netgate:
Congratulations! You are the (un)official winner of the "Longest patch file name ever submitted to pfSense" competion :)

NOYB:
You should have seen it before I word smithed it to be concise. ;)

PR #2657, opened 2016-02-17
status_logs_vpn.php - Remove icon text buffer space

NewEraCracker:
Fix another occurrence similar to PR #2654


NOYB:
Thanks.  I missed those.

PR #2659, opened 2016-02-17
Fix printf(printf())

phil-davis:
This looked odd having nested calls to printf()
Should it actually log_error()?


sbeaver-netgate:
I think so.

This may have crept in via a recent PR to correct some <?=printf(... that should have been sprintf()

PR #2661, opened 2016-02-18
Fix hardcoded pfSense on login page

phil-davis:
I guess this hard-coded mention of pfSense should use the value from $g ?


cbuechler:
Yep, thanks @phil-davis

PR #2665, opened 2016-02-18
Minor tweaks to simplify code

NewEraCracker:
Calling ```print(sprintf(...))``` can be simplified to ```printf(...)```.
```<?php print``` can be simplified to ```<?=```.

**edit**
added a commit that deals with define() calls with unquoted 1st parameter.


sbeaver-netgate:
Bug fixes only from no on please :)

PR #2669, opened 2016-02-18
Diagnostics / Tables - Large Tables Perf Band-Aid

NOYB:
Here's my band-aid proposal for large tables performance issue.
I would prefer for large tables be un-formated and usable than formated and unusable.


sbeaver-netgate:
Thank you.

PR #2671, opened 2016-02-19
Internationalize etc inc uvx

phil-davis:


phil-davis:
This is the end of my general internationalization effort. Those doing translations can get on with their work from now, with a stable base of text strings to work on.

sbeaver-netgate:
Thank you so much Phil.

Steve

PR #2673, opened 2016-02-19
Allow user to dismiss diag_edit messages

phil-davis:
In particular, I like to dismiss the "File saved successfully." message. Then when I save again I can see that the message re-appears.
Also, when you put a wrong filename and you get a "nosuch file" or similar message, then use "Browse", that does not refresh the page. So you end up successfully browsed to somewhere but the danger message is still sitting there. So it is nice to be able to choose to dismiss a danger message also.

### DIFF
--- a/src/usr/local/www/diag_edit.php
+++ b/src/usr/local/www/diag_edit.php
@@ -1,6 +1,6 @@
 <?php
 /*
-	diag_edit.php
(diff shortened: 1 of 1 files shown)


phil-davis:
I crossed with the other change to diag_edit - resubmit coming

PR #2678, opened 2016-02-21
Fix regression on system page preventing save with blank DNS servers

TheUbuntuGuy:
Introduced in f700dc9 for ticket #5915.
Fixes ticket #5916.

Blank DNS slots will always cause the count check to return false so you can't save the page with less than 3 DNS servers listed.

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -201,7 +201,7 @@ if ($_POST) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

TheUbuntuGuy:
I apologise for not having the ICLA signed beforehand. I have rectified this.

sbeaver-netgate:
Thanks

PR #2682, opened 2016-02-22
Initialize variables correctly

NewEraCracker:
I have spotted and corrected the following:

1) gmirror.inc: there are two calls to exec(), those calls expect the 2nd parameter to be an array, this commit initializes the variables correctly.

2) system.php: variable dnslist is used as an array without prior type initialization. This is one of the causes for apparent PR #2680 problems (which led to its reversal).

Note: After this commit is merged, PR #2680 may be safely reapplied for more tough code if desired.


sbeaver-netgate:
Thanks. 2680 is no longer required since the validation is performed elsewhere now. (Later PR)

PR #2683, opened 2016-02-22
Fix lack of Interface Selection on DHCP Advanced page.

WilliamHiles:
The advanced DHCP (client) settings is lacking the entry for interface.  Adding this (and corresponding logic) fixes issues related to DHCP PD.

Developing the same for 2.3.


cbuechler:
What "issues related to DHCP PD" does this fix? The track6 configuration on the interface that receives the PD is what populates this field. Don't see any reason this would be desirable.

PR #2686, opened 2016-02-25
Collapsible Panels JavaScript Error - Fixes 5926

NOYB:
Not the most elegant solution but should be functional.
Would prefer each widget to have an id attribute with something unique to match that identifies it as a widget panel.  Or test if 'updateWidgets' is defined as a function since it should only be so on the index page (dashboard).  But that seems sort of a hackish means.

Changing the widgets available panel href to "#widgets-available_panel-body" (note the added 's' to widget) may be the next best thing so that only the one if match test would be needed.

Is a call to updateWidgets even need here?  It's not readily obvious to me what it's accomplishing here when collapsing/expanding a widget panel.  My guess is that the intention is to save the panel collapsible state.  But I'm not experiencing that.

### DIFF
--- a/src/usr/local/www/bootstrap/css/pfSense.css
+++ b/src/usr/local/www/bootstrap/css/pfSense.css
@@ -790,7 +790,7 @@ a[href]:after {
 }
 
 /** Eliminate overflow in available widgets, log filter, and manage log panels. (cause of scroll bar) */
-#widget-available_panel-body>.content>.row,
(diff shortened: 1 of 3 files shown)


sbeaver-netgate:
Thanks for the fix, but he updateWidgets call in pfSenseHelpers.js is no longer required since we explicitly save the dashboard layout via the 'save' icon in the breadcrumbs bar. I removed them.

PR #2690, opened 2016-02-27
ICMPv6 Firewall Log Display - Fixes 5933

NOYB:
ICMPv6 logging was not being displayed in firewall log.


cbuechler:
thanks @NOYB! I was hoping to get a chance to fix that today, been tied up on other things.

PR #2693, opened 2016-02-29
Diagnostics / Tables - URL Table Aliases

NOYB:
Handle URL table aliases similar to bogons tables.  Suppress record remove buttons.
Update (download) only the selected URL table alias instead of all of them.
Retain full line comments in downloaded URL table files so we can view and use them.
When URL table file full line comments are present, provide view option in info box.
Increase ability to extract last update date from wider variety of URL table file comments.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -1919,7 +1919,7 @@ function update_alias_names_upon_change($section, $field, $new_alias_name, $orig
 
 }
 
-function parse_aliases_file($filename, $type = "url", $max_items = -1) {
+function parse_aliases_file($filename, $type = "url", $max_items = -1, $kflc = false) {
 	/*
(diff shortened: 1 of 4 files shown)


sbeaver-netgate:
There is a small glitch in that "Hide table comments." seems to get broken into two lines?

break

NOYB:
Hmmm.  I've never seen that.  What is your browser width there?

sbeaver-netgate:
Very wide indeed. The issue shows up at any width.

Steve
--

NOYB:
Only way I seem to be able to reproduce that (Chrome & IE; Win 81.Pro) is to go very narrow.  Any idea's?  Do you have a system that it doesn't occur on?  Maybe the differences between them could provide a clue.

sbeaver-netgate:
Sorry for delay. It seems it is only an issue with fief. We could move the <br> to the beginning of the line (in front of the second “i”) or just live with it.

Steve
--

NOYB:
Pull request submitted.
https://github.com/pfsense/pfsense/pull/2694

PR #2700, opened 2016-03-02
Fixes issues raised in #5944

al-right:
This pull request fixes various issues raised in https://redmine.pfsense.org/issues/5944 concerning prefixes.php and status_dhcpv6_leases.php


cbuechler:
Merged, thanks @al-right

PR #2701, opened 2016-03-03
Only add fake DHCPDv6 when not already enabled

phil-davis:
The current code is adding the "fake" entry when track6-interface is defined, and the user has enabled DHCPv6 on the interface - that is causing the user-defined settings to be overridden by the hard-coded settings in this "if" block.
This change should make this code be executed only for track6-interface where DHCPv6 is not enabled by the user, as the comments above intended:
We add a fake entry for interfaces that are set to track6 another WAN unless DHCP has been configured.

Ref forum: https://forum.pfsense.org/index.php?topic=107537.msg600111#msg600111

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1249,7 +1249,7 @@ function services_dhcpdv6_configure($blacklist = array()) {
 		if (isset($blacklist[$ifname])) {
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


cbuechler:
That code block is completely redundant now, since the changes I made to allow disabling of DHCPv6 server and radvd with track interfaces. If it's not enabled, it won't be started. What I missed is that it still was configured and started later. I just removed that code block entirely, so this isn't applicable. Thanks though @phil-davis

phil-davis:
Happy to see dodgy code gone :)

PR #2704, opened 2016-03-04
Revise explanation and validation of dpinger parameters.

dennypage:
Note that tools/scripts/update_pot.sh still needs to be run to update pfSense.pot.  I'm unable to successfully run update_pot.sh doesn't run on my system.


phil-davis:
Looks good.

PR #2712, opened 2016-03-08
Add action column heading to group manager

phil-davis:
Similar fixup to the user manager code in https://github.com/pfsense/pfsense/pull/2711


phil-davis:
@NOYB provided even better code committed in https://github.com/pfsense/pfsense/commit/bf2602a9761c4b296ffa473cb673e4238a688f7b
This is no longer needed.

PR #2718, opened 2016-03-09
JavaScript Polyfill - String Includes Method

NOYB:
A place for JavaScript polyfills.

JavaScript string includes method polyfill.  A new method that has not yet found it's way into many browser implementations.
Documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

Note: An array & typedarray includes method, that could also be useful, is currently in draft status.
Documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes
Documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/includes


sbeaver-netgate:
Did you come across an instance of string.includes() in the pfSense codebase? I was not able to find one so I wonder why this is needed.

NewEraCracker:
The **status_monitoring.php** page from pfSense-Status_Monitoring package needs it.

sbeaver-netgate:
OK thanks. That's why grep didn't find it :)

PR #2719, opened 2016-03-09
Suricata 3.0 binary build options update

bmeeks8:
This update adds the NETMAP build option to the Suricata binary for supporting the new inline IPS mode.


rbgarga:
I did a manual commit to avoid all these merges to go to the tree just to change a single line. Thanks!

PR #2723, opened 2016-03-11
Translate submit button name

phil-davis:
I noticed that the submit button 'Save' did not appear as 'Salvar' in pt_BR, even though that translation is there for pt_BR.
This change here makes the buttons have translated text.
But maybe there is a deeper place that the gettext() should be done - inside Form_Button() itself?
And maybe there are other places in the whole bootstrap/classes/... where strings still need to get translated?

Note: This change causes breakage to services_dhcpv6.php because that has code which explicitly compares to the untranslated string "Save":

```
if ($_POST['apply'] == "Apply Changes") {
	$savemsg = dhcpv6_apply_changes(false);
} elseif ($_POST['save'] == "Save") {
```
So whatever solution is done to make button text translate, code like the above will also need to be fixed at the same time.

### DIFF
--- a/src/usr/local/www/classes/Form.class.php
+++ b/src/usr/local/www/classes/Form.class.php
@@ -47,7 +47,7 @@ class Form extends Form_Element
 		if (gettype($submit) == 'string') {
 			$submit = new Form_Button(
 				'save',
-				$submit,
+				gettext($submit),
 				null,
 				'fa-save'
 			);


sbeaver-netgate:
Phil,

The translatable strings are harvested from the source code files, so while:

`$btn = gettext("Save"); print($btn); `

would print the correct translation, the human translator would be asked to translate the string "$btn" rather than "Save". A better solution is probablly this:
```
	if (!isset($submit))
			$submit = gettext('Save');

		if (gettype($submit) == 'string') {
			$submit = new Form_Button(
				'save',
				$submit,
				null,
				'fa-save'
			);
			$submit->addClass('btn-primary');
		}
```
Where custom "SAve" button text is required, that will be (is?) explicitly translated.

phil-davis:
Yes, true. Making a gettext() around the string 'Save' will guarantee that it appears as a string to be translated.
If the caller provides a string, then do we want to translate it for them? Or do we expect the caller to have translated it before providing it?

phil-davis:
Because I think that a lot of the bootstrap stuff already translates the strings itself - is that correct?
In code fragments like:
```
$section->addInput(new Form_Input(
	'domain',
	'Domain Name',
	'text',
	$pconfig['domain']
))->setHelp('The default is to use the domain name of this system as the default domain name provided by DHCP. You may specify an alternate domain name here. ');
```
The text in setHelp() is already being translated internally, and so we do not also bother to enclose it in gettext(). And somehow the utility that parses the code to pull out strings for translation is looking for setHelp() and other constructions that it knows have text string parameters.
So sometimes we do not have to enclose a string in gettext()

sbeaver-netgate:
Yes. That is correct. The harvester extracts strings from setHelp() (and a number of other functions) as well as gettext().

phil-davis:
Good, now I understand properly what is going on. Closing this in favor of https://github.com/pfsense/pfsense/pull/2726

rbgarga:
@phil-davis You can check update_pot.sh [1] source to see what functions and what parameters of each function are being considered to translations

[1] https://github.com/pfsense/pfsense/blob/master/tools/scripts/update_pot.sh

PR #2725, opened 2016-03-11
System certmanager fixes

NewEraCracker:
Hello,

From my point of view and prompted by inspecting the changes at **system_certmanager.php** on 141d8913e20d814f9dc2897a2eab7c44e3ea1c13, I have found some (unrelated) problem with how the code flows.

This set of commits should correct the code and retain the correct behavior.
Needs testing to ensure everything is working as expected and, as usual, feedback is appreciated.

Still the ```$_POST['save'] == gettext(``` need review (and might need editing) to make sure those checks work as expected without any internationalization issues.

Thanks.


sbeaver-netgate:
Thank you. Reviewing these patches now.

PR #2728, opened 2016-03-11
Internationalize Form_Button text

phil-davis:
The text of a Form_Button is not translated internally. Some Form_Button
calls already had the button text enclosed in gettext(), this does it
for the remaining ones.
So now all buttons have the chance to have a translation.

### DIFF
--- a/src/usr/local/www/diag_authentication.php
+++ b/src/usr/local/www/diag_authentication.php
@@ -147,7 +147,7 @@ $form->add($section);
 
 $form->addGlobal(new Form_Button(
 	'Submit',
-	'Test',
+	gettext('Test'),
 	null,
 	'fa-wrench'
 ))->addClass('btn-primary');
--- a/src/usr/local/www/diag_backup.php
+++ b/src/usr/local/www/diag_backup.php
@@ -664,7 +664,7 @@ $group = new Form_Group('');
 // Note: ID attribute of each element created is to be unique.  Not being used, suppressing it.
 $group->add(new Form_Button(
 	'Submit',
-	'Download configuration as XML',
+	gettext('Download configuration as XML'),
 	null,
 	'fa-download'
 ))->setAttribute('id')->addClass('btn-primary');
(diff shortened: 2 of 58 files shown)


jim-p:
@phil-davis Looks like somehow this already has conflicts, can you check it and rebase?

phil-davis:
Yes, I had started in a branch from a couple of hours ago. Should be good now.

rbgarga:
If Form_Button 2nd parameter must be considered on gettext, it would be easier to add this to tools/scripts/update_pot.sh as it was done for other functions. Then we don't need to remember to add explicit calls on this specific function

phil-davis:
It would need both:
1) A gettext() to actually be done somewhere inside Form_Button so that the underlying code actually translates the parameter.
2) Enhance update_pot.sh to gather the strings in the parameter as @rbgarga mentions.

Is that the way you want to go?

phil-davis:
In update_pot.sh there is already:
```
-kForm_Button:2
```
So the strings should be being collected for translation.
I guess the missing thing is inside Form_Button - it does not actually attempt translation anywhere in the code.

phil-davis:
Maybe it is this easy: https://github.com/pfsense/pfsense/pull/2736

phil-davis:
This won't have done harm in normal operation, but actually should have been unnecessary since #2736 does the translation inside Form_Button.
It does result in gettext() being called twice on the string. That is a little bit of wasted performance, and for unusual cases can mean the translation ends up wrong if the first translated word happens to translate to a word that is also a translatable word in the original language.
e.g. Let's imagine that in some language en_US "Filter'" -> zz_XY "Fritter"
Then we send "Fritter" through gettext(). If en_US "Fritter" appeared somewhere else in the UI, then it will have a translation, and so now the translation of en_US "Fritter" will get displayed when we really want just the translation of en_US "Filter".
Anyway, in reality that rarely (or never) happens.

rbgarga:
I've reverted this

PR #2732, opened 2016-03-12
Tidy up "system_gateways_edit"

ExolonDX:
Tidy up missing semi-colon in non-breaking space


NOYB:
Why reducing from 4 spaces to only 2 spaces?
Shouldn't it be this?
`&nbsp; &nbsp; `

ExolonDX:
Checking both Chrome and Mozilla, it's difficult to see if there are two or four spaces, I looked at lines 869, 872 and 875, which are a hyphen then a space, 2 characters, therefore used two spaces.

NOYB:
Given the layout and the code it would seem to me that 4 spaces was the intention.

PR #2736, opened 2016-03-14
Translate button text

phil-davis:


phil-davis:
This works. Is it the right place in the various nested classes to put the gettext()?

PR #2738, opened 2016-03-15
Redmine #5994 Make DHCP advanced buttons work consistently

phil-davis:
This is the first UI page to be changed so that the Advanced buttons work as specified in Redmine #5994
If this seems like a reasonable way to do it, then I will make similar code for the other UI pages with Advanced buttons.


phil-davis:
Closed because of conflicts. New PR with conflicts resolved is https://github.com/pfsense/pfsense/pull/2739

PR #2739, opened 2016-03-15
Redmine #5994 Make DHCP advanced buttons work consistently

phil-davis:
This is the first UI page to be changed so that the Advanced buttons work as specified in Redmine #5994
If this seems like a reasonable way to do it, then I will make similar code for the other UI pages with Advanced buttons.


sbeaver-netgate:
I'm not sure about the gear wheel, but otherwise I think this is a good foundation for the rest of the pages. Thank you. We can always change the icon later.

Now that I think about it, it would have been nice to centralize the icon selection by using class="icon-savebtn" on hte pages, then setting that to the icon we actually want to see in CSS.

Need to save something for 2.3.1 though :)

PR #2743, opened 2016-03-15
Add field to specify source-hash key

fredronnv:
The source-hash pool option uses a hash of the source address to
determine the translation address. This hashing algorithm is also fed a
key, which unless specified defaults to a random value. This random
value is then generated each time pf is reloaded.

This commit adds the ability to specify the key in order to provide
consistent hashing, even when pf is reloaded.

I welcome feedback as I am primarily a systems/network engineer, not a programmer. We have however been developing a larger NAT-solution based on pfsense when we ran into issues with inconsistent hashing, especially when trying to configure high-availability.

### DIFF
--- a/src/usr/local/www/firewall_nat_out_edit.php
+++ b/src/usr/local/www/firewall_nat_out_edit.php
@@ -126,7 +126,13 @@ if (isset($id) && $a_out[$id]) {
 	$pconfig['target'] = $a_out[$id]['target'];
 	$pconfig['targetip'] = $a_out[$id]['targetip'];
 	$pconfig['targetip_subnet'] = $a_out[$id]['targetip_subnet'];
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

cbuechler:
Thanks for the submission. Several issues there that would need to be addressed to merge. 

1) the key should be its own config field stored separately rather than appended to poolopts, then filter.inc would include the key if a valid one is configured. 
2) no input validation of key field, lots of possibilities for creating a broken ruleset
3) numerous style issues. "if(" should be "if (", "}else{" as "} else {", "){" at end of if lines should be ") {"
4) also need a CLA on file, as in the link above. 

If you can correct that and re-submit, we'll get it merged.

fredronnv:
I've escalated the CLA to my manager, in the meantime I'll make changes as suggested, thanks.

PR #2750, opened 2016-03-16
Status / Dashboard - Conceal save icon when not dirty.

NOYB:


sbeaver-netgate:
Icon should be shown when a widget is opened or closed so that the new state can be saved.

NOYB:
I have no problem with that.  But why is that condition not marked dirty?

sbeaver-netgate:
Because a widget is frequently opened to view its information, and then closed again. We don't want to for the user to deal with "Stay on page/Leave page" in that case.

NOYB:
When the user want's to save widget collapse state changes they could expose the save icon by moving a widget to dirty the page.  ;)

But seriously that is a bit of a conundrum.  The technical means are there.  But nothing seems to present a smooth usage experience.

sbeaver-netgate:
Agreed. I'm going to temporarily remove the class="invisible" for now. Let's think about how we can detect opening/collapsing widgets and enabling the icon.
I would have to try it, but something like $('.panel-body').bind('collapse', function() {enable_icon() }); might do it.

NOYB:
Think the sticking point is with the smoothness of usage.  For instance...
 
1. Don't want the save icon popping in and out every time a widget is expanded/collapsed.  Annoying.
2. Don't want to present warning when leaving the dashboard with a widget that has been left in a different collapse state.
 
Detecting the users intention is the real sticking point.  e.g. Did they just expand a widget to view something and now is going to that widgets page.  Or are they wanting to change the widgets collapse state setting.

sbeaver-netgate:
PR has been cherry picked and merged, but the action temporarily disabled per the above discussion

PR #2754, opened 2016-03-16
Redmine #5994 Firewall Rules advanced buttons

phil-davis:
Standardize the source port advanced button:
1) Put some common code fragments into show_source_port_range() and use code that actually successfully changes the button text Display/Hide. Around line 1808.
2) Make the button itself be hidden when a protocol is selected that does not use source port range (as well as the source port fields being hidden) - line 1908-1909
3) When the user re-selects a protocol that can use source port range, unhide the button and set the button text and hide/unhide of source port fields appropriately depending on the run-time values in the source port fields - line 1900-1905

This standardizes the behavior of the source port Advanced button.

I will do the other general advanced button separately.


phil-davis:
2nd commit added to standardize the Firewall Rules Advanced button.
This is all the code for firewall_rule_edit.php - it cam be reviewed and merged.

PR #2757, opened 2016-03-17
Allow IPs in althostnames

NewEraCracker:
This change will allow implementation of workarounds for bugs involving 'An HTTP_REFERER was detected other than what is defined in System' by adding the IP address to the alternate hostnames list, without having to disable protection completely.

This is useful to workaround things like Redmine #5939 and/or enabling access to WebGUI over IPv6 linklocal (fe80::1:1) by adding IPs manually.

Thanks.


sbeaver-netgate:
Thanks. If local IPs are not being added to the list, that should be reported as a bug for further action.

PR #2764, opened 2016-03-18
Fix Redmine #6011

NewEraCracker:
This will add linklocal fallback where no IPv6 is otherwise configured to allow login using this IPs.
See: https://redmine.pfsense.org/issues/6011


NewEraCracker:
Also it may be worth mentioning that my WAN isn't yet IPv6 ready. At the moment, I only use IPv6 in LAN for testing purposes.

cbuechler:
Thanks @NewEraCracker, merged

PR #2765, opened 2016-03-19
Services / DNS Resolver / General Settings

NOYB:
Static should be static.
Elements that don't need to be dynamic should be static.
More robust than depending on client-side processing to set them.

### DIFF
--- a/src/usr/local/www/services_unbound.php
+++ b/src/usr/local/www/services_unbound.php
@@ -378,7 +378,7 @@ $btnadv = new Form_Button(
 	'fa-cog'
 );
 
-$btnadv->addClass('btn-info btn-sm');
+$btnadv->setAttribute('type','button')->addClass('btn-info btn-sm');
 
 $section->addInput(new Form_StaticText(
 	'Display Custom Options',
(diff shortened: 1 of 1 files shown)


phil-davis:
There are about a dozen of these that were created this last week, because I was using a "template" of how the Advanced button code was done in services_dhcp
So I guess they all should be changed.
After this one gets committed (so we know that is the agreed way to go), I am happy to find the all and do the update.

NOYB:
I for one definitely think they should be changed and welcome your offer.
In my opinion client-side processing on static items, adds unnecessary complexity and compatibility risk.

For instance in this particular case had this been static setting the JavaScript failure, fixed yesterday, would not have resulted in an unsuspected submit.  Instead it would have just failed to toggle its target.

sbeaver-netgate:
I have no objection to this and agree that it is somewhat more robust. The change is not strictly necessary but since the matter was introduced (mostly) in a recent change, it is OK to fix it. There appear to be 16 instances of this altogether so not too bad a task (easy for me to say)

phil-davis:
Commit this one whenever you like, and I will make a PR for the rest of them that will be waiting for you on Monday morning...

sbeaver-netgate:
Thank you Phil.

PR #2766, opened 2016-03-19
Cleanup nginx configuration file

jlduran:
- Fix indentations
- Use the `ssl` parameter of the `listen` directive[¹]
- Change the rewrite rule to use the recommended syntax[²]

---

¹ <http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl>
² <http://nginx.org/en/docs/http/converting_rewrite_rules.html>

[¹]: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl
[²]: http://nginx.org/en/docs/http/converting_rewrite_rules.html

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1278,11 +1278,11 @@ worker_processes  {$max_procs};
 
 EOD;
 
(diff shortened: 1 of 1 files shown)


cbuechler:
merged, thanks!

PR #2770, opened 2016-03-20
Allow namecheap dynamic dns host to start with *.

ryanshow:
Namecheap allows wildcard subdomains in their dynamic dns update url by providing the host as *.example.com. This commit allows hosts that start with *. to pass validation if the provider is namecheap.

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -146,8 +146,8 @@ if ($_POST) {
 	}
 
 	if (isset($_POST['host']) && in_array("host", $reqdfields)) {
-		/* Namecheap can have a @. in hostname */
-		if ($pconfig['type'] == "namecheap" && substr($_POST['host'], 0, 2) == '@.') {
+		/* Namecheap can have @. and *. in hostname */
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

NewEraCracker:
This PR needs to be tweaked to re-apply on the current master.

2.3.1 and the master now have:
```
if ($pconfig['type'] == "namecheap" && ($_POST['host'] == '@.' || $_POST['host'] == '@')) {
```
which could, perhaps, be replaced with:
```
if ($pconfig['type'] == "namecheap" && ($_POST['host'] == '*.' || $_POST['host'] == '*' || $_POST['host'] == '@.' || $_POST['host'] == '@')) {
```

See: https://github.com/pfsense/pfsense/blob/RELENG_2_3_1/src/usr/local/www/services_dyndns_edit.php#L150

cbuechler:
I merged something similar to what @NewEraCracker noted. 

@ryanshow could you try the next 2.3.2 snapshot to verify this change? Thanks!

lukehamburg:
Tested with pfSense-CE-2.3.2-DEVELOPMENT-amd64-20160702-0342 + Namecheap DDNS using `*` wildcard. Works!  Testing was a little rough because the GUI (or PHP?) kept hanging on the status page, but I did check and confirm that the DNS was properly updated.

PR #2773, opened 2016-03-21
Echo boolean value instead of string

NewEraCracker:
Prior to this change, code would echo an empty string if false and 1 if true. While this shouldn't cause any harm, echoing boolean representation is cleaner as done in all other files with similar code.

This PR puts code in line, while simplifying it with a ternary expression.

In other places it happens as:
```<?php if (is_aoadv_used($pconfig)) {echo 'true';} else {echo 'false';} ?>``` and
```<?php if ($showadv) {echo 'true';} else {echo 'false';} ?>```
Which could be optimized, respectively, to:
```<?=(is_aoadv_used($pconfig) ? 'true' : 'false');?>``` and
```<?=($showadv ? 'true' : 'false');?>``` if desired.

Thanks.

### DIFF
--- a/src/usr/local/www/services_unbound.php
+++ b/src/usr/local/www/services_unbound.php
@@ -406,7 +406,7 @@ events.push(function() {
 		var text;
 		// On page load decide the initial state based on the data.
 		if (ispageload) {
-			showadvcustom = "<?=$customoptions?>";
+			showadvcustom = <?=($customoptions ? 'true' : 'false');?>;
 		} else {
 			// It was a click, swap the state.
 			showadvcustom = !showadvcustom;


phil-davis:
Agree, what you have done here is much better. If $customoptions has multiple lines of stuff then all sorts of weird stuff can happen from echoing the PHP variable value back into the Java script.
Anywhere else that does this echoing of a user-entered string should also be fixed up, as there is potential for error depending what the user filled in.
The other changes to ternary expressions can be done if somebody cares enough, as both are logically equivalent.

NewEraCracker:
I thought about making things better when printed to javascript.

At the moment ```$customoptions``` only has boolean values as it was taken care [here](https://github.com/pfsense/pfsense/pull/2763/files). They are converted to string implicitly (either empty or 1) and then printed.

Thereby I present better implementation which prints 'true' or 'false' depending on the situation.

phil-davis:
OK, now I see that @NOYB fixed the big problem and you are just tidying it up a bit more.
I checked through the other code that evaluates whether to display or hide the advanced options, all the other uses are doing proper boolean tests in PHP and sending a boolean back into the JavaScript. So I don't think there is any actual bug remaining in that stuff.
As you say, it is just a matter of code-style as to what looks clean and readable.

sbeaver-netgate:
Thanks NEC and Phil. Empty strings in PHP always risky :)

PR #2777, opened 2016-03-22
Allow mail.php to be used in shell with subject

deajan:
Permits scripts to use /usr/local/bin/mail.php with subject like:
```
echo "This mail has a subject" | /usr/local/bin/mail.php subject="My Subject"
```


jim-p:
What advantage does this have over using -s="Subject" which does the exact same thing?

deajan:
Oh, sorry, I tried /usr/local/bin/mail.php -s "subject" without the equal sign. Sorry :)

PR #2778, opened 2016-03-22
DHCPD: Allow DDNS Hostname option

Tommi2Day:
For entries into a remote  bind9 server i would like to provide the dhcp ddnshostname option for static dhcp entries.


rbgarga:
Looks like there is something wrong with this PR, please re-submit.

Tommi2Day:
from tag RELENG_2_2_6 these are only 2 small changes, doenst know, why it would merge 1000 commits. But from this tag i cant submit a PR

rbgarga:
If you want to submit a patch to 2.2.x you should use the branch RELENG_2_2 and not the tag RELENG_2_2_6

PR #2781, opened 2016-03-23
Fix OpenVPN client additional options typo

jlduran:
Should read client instead of server.  Also remove the `push` example.


NOYB:
Why remove the push example?  Example seem to be appropriate.
Also not so sure that this should say client instead of server.  The server configuration is what is being specified here.

jlduran:
This page (OpenVPN client) supposed to configure the `/var/etc/openvpn/client<number>.conf` file, not the `server.conf`file. I guess the push example is more appropriate for the server config.

NOYB:
Ah.  I missed that this was on *client.php

PR #2782, opened 2016-03-23
Add field to specify source-hash key

fredronnv:
Resubmission of https://github.com/pfsense/pfsense/pull/2743 with corrections. CLA has been submitted and should be on file.

The source-hash pool option uses a hash of the source address to
determine the translation address. This hashing algorithm is also fed a
key, which unless specified defaults to a random value. This random
value is then generated each time pf is reloaded.

This PR adds the ability to specify the key in order to provide
consistent hashing, even when pf is reloaded.

Validation in firewall_nat_out_edit.php, in filter.inc the value is assumed valid if set. Allowed values are 0x followed by 32 hexadecimal digits. If a random string is entered it will be converted to the allowed value using md5, which generates 32 hexadecimal digits.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2067,7 +2067,7 @@ function filter_nat_rules_generate() {
 				$poolopts = (is_subnet($obtarget) || is_alias($obtarget)) ? $obent['poolopts'] : "";
 
 				/* pool option source-hash allows specification of an optional source-hash key */
-				if ($poolopts == "source-hash" && isset($obent['source_hash_key'])) {
+				if ($poolopts == "source-hash" && !empty($obent['source_hash_key'])) {
 					$poolopts = "source-hash ".$obent['source_hash_key'];
 				}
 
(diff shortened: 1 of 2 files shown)


pfSense-git-updates:

rbgarga:
@jdillard could you please check this CLA state?

fredronnv:
Hey @rbgarga, I checked my mail and I sent a CLA to licensing@pfsense.org on 21st March, perhaps it got filtered? Should I re-send the CLA?

gonzopancho:
I just checked, found fredronnv's CLA in my email, on 21 March.  Wasn't filtered, but difficult to find.  I've forwarded to @rbgarga and @jdillard  please resubmit your patch.  thank you.

rbgarga:
Thanks @gonzopancho

fredronnv:
Thanks all, I have a few other patches that we're using internally that I wanted to get in a ready state for a few other PRs. It helps when the CLA is in and registered properly :)

src/etc/inc/filter.inc
```diff
@@ -2066,6 +2066,11 @@ function filter_nat_rules_generate() {
 				$obtarget = ($obent['target'] == "other-subnet") ? $obent['targetip'] . '/' . $obent['targetip_subnet']: $obent['target'];
 				$poolopts = (is_subnet($obtarget) || is_alias($obtarget)) ? $obent['poolopts'] : "";
 
+				/* pool option source-hash allows specification of an optional source-hash key */
+				if ($poolopts == "source-hash" && isset($obent['source_hash_key'])) {
```
rbgarga: It's better to use !empty() here instead of isset()

PR #2783, opened 2016-03-23
Harden sshd_config

jlduran:
The suggested changes are better explained in the following article:

https://stribika.github.io/2015/01/04/secure-secure-shell.html

While I have tried to keep as much backward compatibility as possible, there might be some edge cases.  The main goal should be to harden the sshd (for now, client is next) configuration. 

> **NOTE:**
> You may receive a warning stating that the host identification has changed.  This is because previously an ECDSA key may have been used. Now, only ED25519 and RSA keys are used.
> To solve, remove the line from your `known_hosts` file


cbuechler:
That should be fine. Not changing this close to a release though, we'll merge post-2.3.

PR #2785, opened 2016-03-23
This is my attempt to fix redmine #5721

NewEraCracker:
Toughen all checks and fix logic errors.

I may have added back some "dummy" checking, but better be safe than sorry. Also made the assumption that ```$interface``` is a string. I don't believe anyone would set interface names as numbers or any other type (array, object, resource). Please correct me it that is not the case.

It is my belief that the code is improved by this commit. It needs extensive testing to ensure everything works as it should.

Thanks.


loos-br:
A different fix was committed, there were other bugs from previous commit that I had to address.

Thank you for the report!

PR #2786, opened 2016-03-23
Remove calls which rely in uninitizalized vars

NewEraCracker:
After analyzing implementations in RELENG_2_1 and RELENG_2_2, this code seems to be a left over after code rewrite.

Also remove a trailing space from comment.

Thanks.


NewEraCracker:
This was introduced in 43517fcc1b616b7443b26247dc59dbd65bde2819

cbuechler:
This seems OK, but it doesn't fix any problems and would require a variety of re-testing, we'll merge post-2.3. Thanks

PR #2789, opened 2016-03-25
Interfaces - Local Identification Networks - Filter Option

NOYB:
Separate local identification networks filtering from the bogons filter so the bogons filter can be used on interfaces that serve DHCP clients.


cbuechler:
We're not going to add all that and a new checkbox just to omit 0.0.0.0. There could be a billion checkboxes there if you start trying to cover every edge case. 

The rules that allow access to the DHCP server and relay could be moved above the block bogons to accommodate this.

NOYB:
No more an edge case than private networks.  Nor is it every edge case.  Just those created by a few networks that really shouldn't be in the bogons list and handled separately.  Greatly impairs bogons list use flexibility.
Reordering the rules makes it undesirable in the other direction.
Regardless it has been part of my config for a long time and will continue to be.

PR #2791, opened 2016-03-26
Interfaces - Section Title - Reserved Networks

NOYB:
More accurate section title.


cbuechler:
I don't know that reserved is the perfect title either, but I couldn't think of anything better and it's more apt than private. Merged, thanks.

PR #2792, opened 2016-03-26
Firewall / Aliases / Edit - Exclude Reserved Networks

NOYB:
Add options to exclude reserved networks from downloaded lists for compatibility with interfaces utilizing those networks.
(private address networks and local identification networks (DHCP))

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -2173,7 +2173,7 @@ function update_alias_names_upon_change($section, $field, $new_alias_name, $orig
 
 }
 
-function parse_aliases_file($filename, $type = "url", $max_items = -1, $kflc = false) {
+function parse_aliases_file($filename, $type = "url", $max_items = -1, $kflc = false, $exclude_private_networks = false, $exclude_local_identification_networks = false) {
 	/*
 	 * $filename = file to process for example blocklist like DROP:  http://www.spamhaus.org/drop/drop.txt
 	 * $type = if set to 'url' then subnets and ips will be returned,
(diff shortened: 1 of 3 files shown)


cbuechler:
I'm leaning towards not merging this in the interest of not adding technical debt that's unnecessary for the vast majority (don't fetch aliases with contents you don't want). But there is code that excludes RFC 1918 from bogons separately, where this function could be used instead. Will reconsider post-2.3.

NOYB:
The "don't fetch aliases with contents you don't want" argument is kind of like throwing out the baby with the bath water.  Just because a list contains a few undesirables doesn't make the rest/remainder/vast majority of the list undesirable and useless.

BBcan177:
pfBlockerNG has an option to remove RFC1918 addresses in Feeds. There are some corner cases where these feeds add loopback and other RFC1918 addresses...  Seen this with IBlock Feeds on occasion.

NOYB:
BBcan177 brings up another pretty good point here.  In regard to the "don't fetch aliases with contents you don't want" argument, just because a particular list is "clean" today doesn't mean it will be next month.  Then suddenly the network is broken.  Think it is a good idea to be able to exclude critical networks like private and id from the downloads.

rbgarga:
I would consider more useful if user could insert a list of exceptions instead of having these 2 fixed options. But I would like to hear from @jim-p what he thinks

jim-p:
Having a configurable list of networks to exclude would be better. It could default to RFC1918, RFC4193, maybe even RFC5737 & RFC3849

rbgarga:
It won't be accepted the way it is, if you are still interested please re-implement as a list of networks to be excluded. Thanks!

PR #2793, opened 2016-03-27
Interfaces - Adv DHCP Options Str to Array

NOYB:
Add function to convert DHCP options strings to array that maintains comma's within the double quoted enclosed declaration portions.


NOYB:
Don't know if there is any possibility for this to be advanced to 2.3 inclusion.  But there are some folks in Switzerland that would probably appreciated it.  https://forum.pfsense.org/index.php?topic=87570

What this does is process the send and modifier options strings of the DHCP Advanced settings such that commas (the delimiter) are protected within the double quote declaration portion of the options.

Thanks for the extra consideration.

rbgarga:
Closing in favor of #3020, thanks!

PR #2796, opened 2016-03-28
System / Advanced / Networking - Remove Personalizations

NOYB:
Remove a bunch of "you" personalizations.


src/usr/local/www/system_advanced_network.php
```diff
@@ -251,7 +251,7 @@
 ))->setHelp('Checking this option will disable hardware checksum offloading.<br/>'.
 	'Checksum offloading is broken in some hardware, particularly some Realtek cards. '.
 	'Rarely, drivers may have problems with checksum offloading and some specific '.
-	'NICs.This will take effect after you reboot the machine or re-configure each '.
+	'NICs. This will take effect after a machine reboot or re-configure each '.
```
phil-davis: "or re-configure each interface" also needs to become passive here, "or re-configure of each interface".
3 places to change.

PR #2800, opened 2016-03-29
Add brackets for clarity

phil-davis:
Add some extra brackets for clarity, rather than relying on the operator
precedence rules. IMHO this makes it more readable, and no need for the
reader to wonder if anything might go wrong with the operator
precedence.


cbuechler:
Merged, thanks Phil!

PR #2832, opened 2016-04-04
Fixes to linguistic issues introduced by recent commits

NewEraCracker:
Fixes:
1) encryption vs decryption typos;
2) unnecessary sprintf.

Thanks.

### DIFF
--- a/src/usr/local/www/diag_backup.php
+++ b/src/usr/local/www/diag_backup.php
@@ -226,7 +226,7 @@ if ($_POST) {
 		if ($mode == "download") {
 			if ($_POST['encrypt']) {
 				if (!$_POST['encrypt_password']) {
-					$input_errors[] = gettext("A password for decryption must be supplied and confirmed.");
+					$input_errors[] = gettext("A password for encryption must be supplied and confirmed.");
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


NewEraCracker:
By the way. Regeneration of .pot may be a good idea.

phil-davis:
Well spotted- I must have missed those while reviewing - :+1: 
And yes, those doing real translation will need a new .pot and will have a bit of work to do if they want all those (now passive) text strings to have translations in the real release of 2.3

NOYB:
Well you caught the encrypt/decrypt, but think I messed up the correction.  Didn't realize there were two of them; one each way for save/restore.

Nice finds and corrections NEC.  Thanks.

phil-davis:
@NOYB have you finished reviewing everything for "remove personalisations"?
If so, then I suggest someone regenerates.pot

NOYB:
Not quite yet.  Working on some stragglers.  Will be submitting them tomorrow.

Then there is some high hanging fruit that I avoided because public school did not ed-u-ma-cated me sufficiently enough to come up with correct rephrasing.  Stuff like and similar to: "Are you sure you want to ...(reboot/factory reset/etc.).  So those probably need to be tackled by someone else.

I'll post a list in the forum of what's remaining.
The wizards is the majority of what remains, and I think there are lot of tough ones in there too.

PR #2833, opened 2016-04-04
Prevent foreach crash if no IPsec

phil-davis:
On a system with no IPsec defined, this would crash with:
[04-Apr-2016 21:33:24 Etc/UTC] PHP Stack trace:
[04-Apr-2016 21:33:24 Etc/UTC] PHP   1. {main}() /usr/local/www/status_ipsec.php:0
[04-Apr-2016 21:33:24 Etc/UTC] PHP   2. print_ipsec_body() /usr/local/www/status_ipsec.php:74

I did belts-and-braces here - moved the "if" up at line 72 and also added an if(is_array()) at line 394

### DIFF
--- a/src/usr/local/www/status_ipsec.php
+++ b/src/usr/local/www/status_ipsec.php
@@ -381,7 +381,7 @@ function print_ipsec_body() {
 				print("</tbody>\n");
 				print("	</table>\n");
 				print("</td>\n");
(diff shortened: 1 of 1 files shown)


phil-davis:
Will submit again with conflict resolved.

phil-davis:
See #2835

PR #2838, opened 2016-04-05
Remove "we" personalizations from config.console.inc

phil-davis:
I changed "we" to "the system" - if someone thinks so, it could be changed to "pfSense" or $g["product_name"]\
If you want to do something different to what I did here, then feel free to close this and make the required changes.


phil-davis:
Note: This script has a lot of "you" in it also. But that seems fine to me - it really is a setup script that talks to "you" the sysadmin at the console who is doing the initial setup.

PR #2844, opened 2016-04-05
system.inc - improve package info to syslog.conf

marcelloc:
packages like postfix needs some custom settings to syslog.conf. As it is dangerous to be done by file hacking.(https://github.com/pfsense/FreeBSD-ports/pull/23) I'm sending this code change to improve system.inc tests on package logging config.


spec1re:
:+1:

rbgarga:
Since postfix is not coming back, this change doesn't make sense.

marcelloc:
Will be there any place for "non official" packages? This is something that I need on my pfsense boxes. the pull request for the package is still there.

rbgarga:
No, there are no plans for non-official package repository

PR #2850, opened 2016-04-07
Traffic Shaper Wizard - Remove Personalizations

NOYB:
Remove "you" personalizations.


phil-davis:
In 2 places:
This wizard will provide guidance through setting up the pfSense traffic shaper.
Maybe could be:
This wizard will provide guidance **in** setting up the pfSense traffic shaper.

Otherwise looks good :+1:

NOYB:
Thanks for the suggestion.  Also considered ".. for setting up ...".  Prefer the original "... through setting up ..." though, as in talking a person through the process.
Would like to stay with the original "trough".  Hope that's okay.

phil-davis:
That's OK

PR #2854, opened 2016-04-10
Fix add delete of rows with buttons

phil-davis:
Redmine #6101
Forum https://forum.pfsense.org/index.php?topic=109554.0


sbeaver-netgate:
Changed this slightly to only increment the deleterow button. There is usually only one button per row, but it is possible to have more so this seemed safer.

PR #2866, opened 2016-04-14
Firewall / Rules - Underscore Line Wrap

NOYB:
Replace underscore with space in display text of certain fields to allow them to line wrap.  ex: long alias names


sbeaver-netgate:
On other pages I have inserted <wbr> tags to facilitate wrapping. Perhaps if you replace "_" with "_<wrb>"  it would break without losing the underscores?

NOYB:
I did experiment breaking with zero width space and wrb.  But found it to be cleaner and easier to read without the underscore.

PR #2868, opened 2016-04-14
Fix #6146 saving of source tracking timeout

phil-davis:
1) The $POST array index key source-tracking-timeout was incorrect - fixes the direct issue reported.
2) If "use sticky" was unchecked and the user made a change to source tracking timeout, then that change was not being saved (the old value would be "stuck" in the confgi). There seemed no god reason for that. If the user wants to, they should be able to change the source tracking timeout value whenever they wish, then come along later and enable "use sticky".

### DIFF
--- a/src/usr/local/www/system_advanced_misc.php
+++ b/src/usr/local/www/system_advanced_misc.php
@@ -193,13 +193,16 @@ if ($_POST) {
 				$config['system']['lb_use_sticky'] = true;
 				$need_relayd_restart = true;
 			}
-			if ($config['system']['srctrack'] != $_POST['srctrack']) {
-				$config['system']['srctrack'] = $_POST['srctrack'];
-				$need_relayd_restart = true;
-			}
 		} else {
 			if (isset($config['system']['lb_use_sticky'])) {
 				unset($config['system']['lb_use_sticky']);
+				$need_relayd_restart = true;
+			}
+		}
+
+		if ($config['system']['srctrack'] != $_POST['srctrack']) {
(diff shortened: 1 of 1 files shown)


phil-davis:
@sbeaver-netgate already did a fix.

sbeaver-netgate:
Phil. - Sorry, I was working on this issue at the same time as you and already pushed the fix :(
I added some JS to retain the behavior of 2.2.x

phil-davis:
Yes, I saw that the same minute. There is a question about general user functionality for anything where there is an enable/disable checkbox and then a bunch of parameters (in this case only 1) that tune that functionality.
If edit of the tuning parameters is disabled when the checkbox is disabled, then there is no chance for the user to "fixup" or adjust the tuning parameters without enabling the function. What if the user wants to turn the function on sometimes? But today wants it off and just wants to set up the parameters ready for tomorrow, or notices that the parameters are set to crud/old values and wants to fix that up in case someone in future turns it on.
I reckon parameters like that should be editable at any time, and just become run-time functional when the checkbox is enabled.

sbeaver-netgate:
Yep. On the other hand, I suppose it is reasonable to have some visual indication that the thing you are adjusting here isn't actually going to do anything because the parent function is disabled.

If you make a change and save, should we still validate the input if the parent is disabled?

phil-davis:
Yes, the input would have to be validated, otherwise it is misleading to users to be able to put in a rubbish combination of parameters.
And yes, it would be nice to have some visual indication that the parameter will not be run-time implemented (because the check box is unchecked). I wonder if there is a UI "standard" for that?

PR #2869, opened 2016-04-14
Fix #6142 validate and adjust dpinger params on upgrade

phil-davis:
1) The previous "down" value was being converted to msec and put into
"loss_interval". It should go into "time_period".
2) loss_interval must always be at least latencyhigh - make it so if
latencyhigh is big.
3) alert_interval must always be at least probe interval - make it so if
the probe interval is high.
4) The time_period for averaging the results must be at least 2 probe
intervals plus loss_interval (ensuring there should always be 2 probe
results by the time_period expiry) - make it so.

These various conditions taken from the validation code in
system_gateways_edit.php

Note: I have explicitly put the necessary default dpinger parameter
values here, rather than calling return_dpinger_defaults() because at
this point in any future conversion we want to use these particular
numbers, not what the defaults happen to be in pfSense 2.4 or 3.0 or...

### DIFF
--- a/src/etc/inc/upgrade_config.inc
+++ b/src/etc/inc/upgrade_config.inc
@@ -4147,6 +4147,12 @@ function upgrade_129_to_130() {
 function upgrade_130_to_131() {
 	global $config;
 
+	// Default dpinger parameters at time of this upgrade (2.3)
+	$default_interval = 500;
+	$default_alert_interval = 1000;
+	$default_loss_interval = 2000;
+	$default_time_period = 60000;
+
 	if (isset($config['syslog']['apinger'])) {
 		$config['syslog']['dpinger'] = true;
 		unset($config['syslog']['apinger']);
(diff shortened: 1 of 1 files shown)


phil-davis:
If this code ends up in 2.3.1, then future upgrades from 2.2.6 or earlier to 2.3.1 will work fine.
On systems that have already upgraded to 2.3 using the 2.3 version of upgrade_130_to_131() step, they might have broken the dpinger settings, but presumably the admin will have had to fix it manually by editing the gateway advanced parameters. In case they have not, we could add a new upgrade step for 2.3 to 2.3.1 that just checks dpinger (new) parameter validity and corrects/adjusts a few things of they are still not valid (i.e. if the user has not already corrected manually).

cbuechler:
merged to all 3 branches, thanks Phil!

phil-davis:
I will submit another PR to help handle 2.3 to 2.3.1 upgrade later tonight.

PR #2872, opened 2016-04-15
Add checks in case there are dpinger param problems

phil-davis:
in 2.3 installations that were upgraded from 2.2.* when the apinger to
dpinger params conversion code in upgrade_130_to_131() was not so good.


phil-davis:
This is the last piece of redmine #6142

PR #2874, opened 2016-04-16
Fix sshd can't load key log spam

jlduran:
I have not been able to replicate [6143](https://redmine.pfsense.org/issues/6143), but from what I understood on the comments, this should do it.


NOYB:
Yup.  Looks like that fixes it.
Apply the code and no more error in the log.
Remove the code and error returns to the log.
Thanks.

PR #2876, opened 2016-04-16
Removed redundant if statement part

Fusl:


pfSense-git-updates:

cbuechler:
That's not redundant. It prevents use of a gateway with synproxy state rules because it isn't functional in that circumstance (or at least didn't used to be, have you seen otherwise?). This is wrong regardless though as it would make it impossible to save any firewall rule that specified a gateway, which breaks tons of things.

Fusl:
Of course this is redundant, please take a look here:

https://github.com/pfsense/pfsense/pull/2876/commits/d19ad3c23a4461cde5dc74ddc9e4dcf20b010655#diff-34254bd85b68c71eddb8fc0eed510e66L455

You have the exact same if statement in line 455.

cbuechler:
I didn't notice the check above it, looked like you just made a change without having any idea of why. I'll remove it. Thanks

cbuechler:
Fixed in 34515250ca726f096071b139fcebd5d18601162a, thanks @Fusl

PR #2878, opened 2016-04-17
Alias Tables RAM Disk Store

NOYB:
Fixes 6189
If ramdisk is enabled keep a copy of the alias tables to restore at boot time.  Otherwise unpredictable behavior may occur due to some aliases not being  available when the firewall rules load.


sbeaver-netgate:
Nice work. Thanks you!

NOYB:
Thanks.  And you're welcome.

PR #2880, opened 2016-04-18
Added back GET parameters support in status_graph.php (as in pre-2.3)

LucaTNT:
In version 2.3 the Traffic Graph page switched from GET to POST parameters to control the graph.
This patch restores the old functionality, which was useful to bookmark a given graph configuration.

For instance, I had a bookmark that shows the bandwidth usage on the LAN interface, using hostnames:
`/status_graph.php?if=lan&sort=&filter=&hostipformat=hostname`

### DIFF
--- a/src/usr/local/www/status_graph.php
+++ b/src/usr/local/www/status_graph.php
@@ -97,6 +97,21 @@ foreach (array('server', 'client') as $mode) {
 	}
 }
 
+// Compatiblity to restore GET parameters used pre-2.3
+// Useful to save a URL for a given graph configuration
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

phil-davis:
IMHO this sort of GET is fine because it only controls display. In other places GET was removed because people could send values in GET that were applied to the config, making real change on a system.

sbeaver-netgate:
I agree. We need to examine each case carefully, but this one seems reasonable. I will merge the shortly. PR

PR #2883, opened 2016-04-19
Periodic RRD & DHCP Leases RAM Disk Backup

NOYB:
Remove these cron jobs on full install if not using ramdisk.


cbuechler:
merged, thanks @NOYB

PR #2886, opened 2016-04-19
Handle single quotes in user names for CP

phil-davis:
Forum: https://forum.pfsense.org/index.php?topic=110243.0
This is a trial - make sure it really works before committing.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -2086,7 +2086,7 @@ function portal_allow($clientip, $clientmac, $username, $password = null, $attri
 
 	/* read in client database */
 	$query = "WHERE ip = '{$clientip}'";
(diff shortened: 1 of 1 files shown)


heper:
@phil-davis i can confirm your fix works for my AD testuser: test'abcd
thanks a lot

phil-davis:
I noticed that there is code that does stuff like:
			$safe_username = SQLite3::escapeString($username);
Maybe this escapeString() call is the more correct generic way to make sure that odd characters in strings are appropriately escaped for use in SQL.

phil-davis:
See https://github.com/pfsense/pfsense/pull/2887 for alternate solution that should be better, assuming it owrks.

phil-davis:
Closing in favor of #2887

PR #2887, opened 2016-04-19
Escape username before use in CP SQL

phil-davis:
A maybe "better" version to try.


phil-davis:
@heper Can you check if this version works?
The call to escapeString should be a more generic thing that will work for all weird strings for the rest of eternity.

heper:
@phil-davis seems to work fine. thanks :)

phil-davis:
Thanks for the quick testing! Then this is the preferred fix IMHO.

phil-davis:
Redmine issue https://redmine.pfsense.org/issues/6203 created to document this.

PR #2888, opened 2016-04-19
Feature 4923: update to rfc2307 filter and group name length

jobewan:
- The rfc2307 filter currently doesn't account for the situation where a member is added only via their DN.  This filter extension simply adds the capability to search via DN "or" user_attr.  I would think a wildcard search would be acceptable in some situations as well, but less secure depending on existing hierarchy of the tree.
- The other commit extends the group length requirement to allow 32 characters opposed to 16.  LDAP group names can get quite long in length and BSD isn't impacted by this, so I'd think that 32 would be acceptable.


pfSense-git-updates:

jobewan:
CLA has been signed... Sorry, I didn't see that.

rbgarga:
@jobewan could you please rebase your fork?

sbeaver-netgate:
No response to rebase request in six months

PR #2891, opened 2016-04-20
Feature #6210 create user privilege for only DNS Host and Domain Overrides in DNS Resolver

gottaloveit:
Currently, the only way to allow privilege to add/edit DNS overrides is to allow access to the entire DNS Resolver Options General Settings page. I have a use case currently where admins want to extend override capability to a set of users for a network dedicated to testing devices and DNS overrides happen all the time, but admins are nervous of allowing these overrides to somehow get out of this network. So if the testers only had access to the overrides without the general settings everyone would be happy.  These changes did not add or remove any functionality, just split it between pages to be able to assign privileges.


cbuechler:
This was reverted for the reasons noted in the redmine ticket.

PR #2899, opened 2016-04-22
Don't foreach unless it's an array. Ticket #6142

phil-davis:
Similar thing to https://github.com/pfsense/pfsense/commit/13dab3538fe8b6c1688142c89a017520ad0aaac6 should be done here in upgrade_130_to_131()


cbuechler:
Good catch. Merged, thanks @phil-davis

PR #2902, opened 2016-04-23
RAM Disk Management

NOYB:
1) Treat the RAM disk more like a permanent storage device with content managed/restored by the system and made available at boot up, before needed by any services.
 a) Handle saving and restoring RAM disk content at reboot/shutdown/boot centrally in more of a system manged fashion.
 b) If it's in the RAM disk store it gets restored early in the pfSense startup so it's available for whatever needs to use it.
 c) Services utilizing RAM disk don't need to be aware that their content is on a RAM disk, and handling content restore individually.

2) Has the benefit of eliminating some issues with the previous code as well.  Such as...
 a) Restoring RRD multiple times during boot, potentially at least 3 times, by rc.newwanip, rc.newwanipv6, and rc.boot.  Some even overlapping.
 b) Not removing the backups if/when not being utilized.  Such as on a full install with the RAM disk option not enabled.
 c) Eliminate some duplicate code.

3) Looking forward.
 a) The more centrally system managed approach may crack the door open to making it easier to include some of the logs in the RAM disk store.  As well as anything else that may be useful/desirable to retain in the RAM disk across reboots.


gonzopancho:
I'd like @rbgarga to look this over.

stilez:
@NOYB - how easy would it be to add a config option _"don't save ramdisk content on shutdown [Y/N]"_?

Use case (just one, but it's very useful) - protecting data such as scanned, cached, and proxied http/ftp traffic that is stored briefly during processing or until completely received/sent, where it's undesirable for it to be recoverable from disk at a later time.  In this case an option to _retain_ volatility is desirable; a ramdisk based cache keeps possibly private data from being written to disk in the first place.

NOYB:
Don't know.

The appropriate security measures for this would be for such features that are handling sensitive data to either not write it to disk (RAM disk or otherwise) in the first place, or to manage it; i.e. clean up after itself.  Dependency on RAM disk volatility as a security management feature is really super bad design.

Even with such an option to retain RAM disk volatility across reboots, such data that is written to the RAM disk would still be exposed there for as long as the system remained up.  Until either some process cleaned it up or the system is rebooted.  Doesn't seem like a very effective, nor efficient (reboot required), nor appropriate security measure.

I hope no one is actually using system reboots to purge sensitive data from RAM disk as a security measure.  If they are, I wonder what their reboot schedule has to be to meet their security requirements.

Lets say a weekly reboot is used to purge the sensitive data from RAM disk.  The data has been exposed there for up to a week already.  If it is sensitive enough to warrant a reboot to flush it then the design of the feature that wrote it is flawed and should be addressed to correct the handling/management of the data.

stilez:
Good points. My thinking was around things like office printer/scanners which can be made to reprint sensitive documents from their disk cache. Suppose someone wanted to use squid as a quick and builtin way to achieve antivirus scanning, squid is ideal and a ramdisk cache will be faster too. But AFAIK squid doesn't have a secure wipe-files-immediately-when-no-longer-needed setting. I could be wrong on that, it's an example though. It should be fixed in squid but if it isn't, then at least putting the cache on volatile store will add some element of additional risk mitigation that they can't easily get from the package.  A physical disk can be very large (hundreds of GB or more) and might contain data at rest for some time. Ram sizes are smaller so stale pages are quickly overwritten or zeroed for reuse by the OS, whether or not the system is rebooted. Again my apologies if I'm technically incorrect on this!

Andrew17856:
Another option might be to add an "exclude paths" option, so that particular parts of the RAM disk are excluded from backup.  I would use that to exclude directories that contained large volumes of data that you wouldn't necessarily want to backup/restore on each reboot, such as ntopng or RRD data.  This isn't for security reasons; rather that it would take a while to backup/restore, I wouldn't need it backed up and restored, and it wears out your SSD.

That said, I'm really keen to get the backup/restore functionality added, because a number of plugins/other features of pfSense stop working properly on reboot at the moment if you're using a RAM disk.  I would be very happy for it to be rolled out in its current form, with these sorts of options coming as a subsequent update if they are needed.

Thanks for all your work on this.

stilez:
@Andrew17856 - that would work fine and answer my point. Also a good point on its own merits, and keen to see this done, as you say, because of the huge wider benefits.

Andrew17856:
Thanks. Thinking about it, it might be easier to implement as a list of paths to include (with a recursive copy) rather than a list to exclude.  By default it could be set up to just list the root of var and tmp (ie copy everything), but then a user could specify what to copy by amending the list.

I think the negative list would be better though (but more difficult to implement).

Very happy just to have it as currently proposed though, with any refinements to follow on later.

NOYB:
Took a quick cursory look the code and it seems that only the alias tables, dhcp leases, and rrd get backup up and thus restored.

Compared to the original code, alias tables should be the only additional content that is being backed up.  If that's the case then the rest of the RAM disk should still be volatile across reboots.

But it's been quite awhile since submitting this so it's not fresh in mind.  So I could be totally missing it.

/etc/rc.shutdown.sh calls these three backup routines.

	/etc/rc.backup_aliastables.sh
	/etc/rc.backup_rrd.sh
	/etc/rc.backup_dhcpleases.sh

Andrew17856:
Thanks. To take us further than we are already, it would ideally be backing up the entire drive (or be customisable). 

At the moment, various addons assume that the var and tmp directories persist.  For example, ntopng needs a reinstall after a  reboot if you're running off a RAM disk.   pfBlockerNG used to be the same, but has now had some backup functionality added to deal with RAM disks (but still needs an force update after rebooting).

NOYB:
Understand.  This method of handling RAM disk backup/restore in more centrally managed system fashion should make it pretty easy to incorporate persistent content needed by other features/packages etc.

Basically just add the content to the RAM disk store such as by adding an rc.xxxx.sh and calling it at shutdown.  Every thing placed in the RAM disk store should be restored early in the boot up so that it is available prior to anything needing it.  As though it was on a persistent disk.

So if/when this PR gets merged, then those could be incorporated.

NOYB:
It's been about 6 months since this PR was submitted and has been marked as "Pending review/testing" for quite some time.  Could we please get a review/testing status update?  Thanks

Please note.  Added a commit this week to move the RAM disk restore out of PHP (/etc/rc.bootup) and into /etc/pfSense-rc so that the backed-up contents get restored as soon as the MFS is created.
This way everything that is backed-up in the RAM Disk Store as a tar *.tgz will be restore before pfSense even begins to load.
This makes it possible to include content such as logs that if restored latter would overwrite any new entires that were created earlier in the boot up process.
In short what is in the RAM Disk Store at shutdown/reboot will be restored to the RAM disk before pfSense even starts to load.  So features/packages etc. do not even need to be aware that their content is on a volatile RAM disk.  It is just there for them as if on a persistent disk.

Centralized management of the RAM disk content backup/restore would make it much easier to include content needed by other features/packages etc.
Just by placing a tar *.tgz in the RAM Disk Store it will be restore at system boot up as soon as the MFS is created.  Eliminating the need for feature/package specific code to manage and restore it's content.
Just create an /etc/rc.backup_*.sh and include a call in /etc/rc.reboot and /etc/pfSense-rc.shutdown.  See other similar rc files for examples.

Periodic backup cronjob config can also easily be added to the RAM disk configuration in /usr/local/www/system_advanced_misc.php.  Or by other means if desired.  Such as in a package's configuration.

If a feature/package already has it's own schedule for updating it's content the /etc/rc.backup_*.sh can be called from there rather than using a periodic backup that may be end up backing up unchanged content.  Such as is done for URL alias tables (see /etc/inc/pfsense-utils.inc for example).

This streamlines and simplifies the RAM disk management, and eliminates several issues mentioned earlier in the previous commits.

I realize that the "nanobsd" platform is being eliminated and with that one of the uses of MFS.  But there are still significant usage benefits of MFS for "full" installs with CF cards and USB flash drives, etc.  Please let's not also eliminate the MFS/RAM disk.

Would be nice if this could be included.

Thanks

src/etc/inc/pfsense-utils.inc
```diff
@@ -2448,13 +2448,12 @@ function process_alias_urltable($name, $type, $url, $freq, $forceupdate=false, $
...
-			/* If this backup is still there on a full install, but we aren't going to use ram disks, remove the archive since this is a transition. */
-			if (!isset($config['system']['use_mfs_tmpvar'])) {
-				unlink_if_exists("{$g['cf_conf_path']}/RAM_Disk_Store{$urltable_filename}.tgz");
-			} else {
-				/* Update the RAM disk store with the new/updated table file. */
-				mwexec("cd / && /usr/bin/tar -czf \"{$g['cf_conf_path']}/RAM_Disk_Store{$urltable_filename}.tgz\" -C / \"{$urltable_filename}\"");
+			/* Remove existing archive and create an up to date archive if RAM disk is enabled. */
+			unlink_if_exists("{$g['cf_conf_path']}/RAM_Disk_Store/{$name}.txt.tgz");
+			if (isset($config['system']['use_mfs_tmpvar'])) {
+				mwexec("cd / && /usr/bin/tar -czf \"{$g['cf_conf_path']}/RAM_Disk_Store/{$name}.txt.tgz\" -C / \"{$urltable_filename}\"");
```
rbgarga: cd / is redundant here since tar already contains -C /, please remove it

src/etc/rc.backup_aliastables.sh
```diff
@@ -3,18 +3,23 @@
...
 		mkdir -p "${RAM_Disk_Store}"
 	fi
 
 	for aliastablefile in "${DBPATH}"/* ; do
 		filename="$(basename ${aliastablefile})"
-		if [ ! -f "${RAM_Disk_Store}/${filename}.tgz" ]; then
-			cd / && /usr/bin/tar -czf "${RAM_Disk_Store}/${filename}.tgz" -C / "${DBPATH}/${filename}"
+		if [ ! -f "${RAM_Disk_Store}/${filename}.tgz" -o "${RAM_Disk_Store}/${filename}.tgz" -ot "${DBPATH#/}/${filename}" ]; then
+			[ -f "${RAM_Disk_Store}/${filename}.tgz" ] && /bin/rm -f "${RAM_Disk_Store}/${filename}.tgz"
+			cd / && /usr/bin/tar -czf "${RAM_Disk_Store}/${filename}.tgz" -C / "${DBPATH#/}/${filename}"
```
rbgarga: cd / is redundant, please remove

src/etc/rc.backup_dhcpleases.sh
```diff
@@ -1,6 +1,21 @@
...
+if [ -d "${DBPATH}" ]; then
+	echo -n "Saving DHCP Leases to RAM disk store...";
+
+	[ -f "${RAM_Disk_Store}/dhcpleases.tgz" ] && /bin/rm -f "${RAM_Disk_Store}/dhcpleases.tgz"
+
+	if [ ! -d "${RAM_Disk_Store}" ]; then
+		mkdir -p "${RAM_Disk_Store}"
+	fi
+
+	cd / && /usr/bin/tar -czf "${RAM_Disk_Store}/dhcpleases.tgz" -C / "${DBPATH#/}/"
```
rbgarga: cd / is redundant, please remove

src/etc/rc.backup_logs.sh
```diff
@@ -0,0 +1,21 @@
...
+if [ -d "${DBPATH}" ]; then
+	echo -n "Saving Logs to RAM disk store...";
+
+	[ -f "${RAM_Disk_Store}/logs.tgz" ] && /bin/rm -f "${RAM_Disk_Store}/logs.tgz"
+
+	if [ ! -d "${RAM_Disk_Store}" ]; then
+		mkdir -p "${RAM_Disk_Store}"
+	fi
+
+	cd / && /usr/bin/tar -czf "${RAM_Disk_Store}/logs.tgz" -C / "${DBPATH#/}/"
```
rbgarga: cd / is redundant, please remove

src/etc/rc.backup_rrd.sh
```diff
@@ -1,25 +1,21 @@
...
-		cd / && /usr/bin/tar -czf "${CF_CONF_PATH}/rrd.tgz" ${tgzlist}
-		/bin/rm -f "${RRDDBPATH}"/*.tgz
+	[ -f "${RAM_Disk_Store}/rrd.tgz" ] && /bin/rm -f "${RAM_Disk_Store}/rrd.tgz"
+
+	if [ ! -d "${RAM_Disk_Store}" ]; then
+		mkdir -p "${RAM_Disk_Store}"
 	fi
-fi
 
+	cd / && /usr/bin/tar -czf "${RAM_Disk_Store}/rrd.tgz" -C / "${DBPATH#/}/"
```
rbgarga: cd / is redundant, please remove

src/etc/rc.backup_logs.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
rbgarga: It misses license text and copyright notice

src/etc/rc.restore_ramdisk_store
```diff
@@ -0,0 +1,113 @@
+#!/bin/sh
+#
+# rc.restore_ramdisk_store
```
rbgarga: Please move license text to Apache 2.0

src/etc/rc.restore_ramdisk_store
```diff
@@ -0,0 +1,113 @@
...
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Wildcard file existence check function
+wildcard_file_exists() {
+	for file in "$1"; do
+		if [ -f "$file" ]; then
+			return 0
+			break
```
rbgarga: break will never be reached here after a return, please remove it

src/etc/rc.restore_ramdisk_store
```diff
@@ -0,0 +1,113 @@
...
+
+	mesg="Restoring contents of RAM disk store..."
+	echo -n "$mesg"
+
+	USE_MFS_TMPVAR=$(/usr/local/sbin/read_xml_tag.sh boolean system/use_mfs_tmpvar)
+
+	# Restore the ram disk
+	for backup_file in "${RAM_Disk_Store}/"*".tgz"; do
+		if [ -f "$backup_file" ]; then
+			cd / && /usr/bin/tar -xzf "${backup_file}" -C / 2>&1
```
rbgarga: cd / is redundant here, please remove

NOYB:
Applied requested changes.  i.e. redundant "cd / ", "break" after "return", and license text.

NOYB:
> One question, is this code respecting all backups done before the change? I mean, when someone update to new version containing this change, all old backups will be restored, right?

Have an update script laying around somewhere.   To avoid having to rebase every time there is a new update script added, didn't want to include it until this was on the verge of being accepted and merged.

src/etc/inc/pfsense-utils.inc
```diff
@@ -2395,13 +2395,12 @@ function process_alias_urltable($name, $type, $url, $freq, $forceupdate=false, $
...
-			/* If this backup is still there on a full install, but we aren't going to use ram disks, remove the archive since this is a transition. */
-			if (!isset($config['system']['use_mfs_tmpvar'])) {
-				unlink_if_exists("{$g['cf_conf_path']}/RAM_Disk_Store{$urltable_filename}.tgz");
-			} else {
-				/* Update the RAM disk store with the new/updated table file. */
-				mwexec("cd / && /usr/bin/tar -czf \"{$g['cf_conf_path']}/RAM_Disk_Store{$urltable_filename}.tgz\" -C / \"{$urltable_filename}\"");
+			/* Remove existing archive and create an up to date archive if RAM disk is enabled. */
+			unlink_if_exists("{$g['cf_conf_path']}/RAM_Disk_Store/{$name}.txt.tgz");
+			if (isset($config['system']['use_mfs_tmpvar'])) {
+				mwexec("/usr/bin/tar -czf \"{$g['cf_conf_path']}/RAM_Disk_Store/{$name}.txt.tgz\" -C / \"{$urltable_filename}\"");
```
jim-p: This must be changed, the double quoting is not good enough, it should be:
`escapeshellarg("{$g['cf_conf_path']}/RAM_Disk_Store/{$name}.txt.tgz")`
And then
`escapeshellarg($urltable_filename)`

So the line should become:
`mwexec("/usr/bin/tar -czf " . escapeshellarg("{$g['cf_conf_path']}/RAM_Disk_Store/{$name}.txt.tgz") . " -C / " . escapeshellarg($urltable_filename));`

src/usr/local/www/system_advanced_misc.php
```diff
@@ -541,6 +550,16 @@
...
 	'frequent the backup, the more writes will happen to the media.');
 
+$section->addInput(new Form_Input(
+	'logsbackup',
+	'Periodic Logs Backup',
+	'number',
+	$config['system']['logsbackup'],
+	['min' => 0, 'max' => 24, 'placeholder' => 'Period between 1 and 24 hours']
+))->setHelp('This will periodically backup the logs so '.
+	'it can be restored automatically on the next boot. Keep in mind that the more '.
```
jim-p: Wording isn't quite right, either s/the logs/the log directory/ or s/it can/they can/

NOYB:
> I haven't tested it yet, but it seems OK from a quick read-through, aside from the items I have noted.

Made the requested changes.

NOYB:
Unknown to me is at what point in the upgrade process these new rc.* scripts become active and thus whether or not an upgrade script/function is needed to restore the old backups.  For example.  Will the new rc.* scripts be active at shutdown/restart during the upgrade process so the RAM disk store gets created with the new scripts.  Or will the old scripts still be in use and thus require an upgrade script to restore the old backups for the first time during the upgrade reboot.

This is something that needs to be tested to verify.  I'm not setup to confidently do that in an actual upgrade process.  I have an upgrade script that I've simulated with and can provide/add that in the PR if it is needed.

jim-p:
If the timing is uncertain then the safest way would be to detect an old-style backup and restore it, then remove the old backup.

During an upgrade from 2.3.x to 2.3.x or 2.4.x, the old script will be in place at shutdown and the new script will be in place before the bootup code is processed. During an upgrade from an older version (<= 2.2.x) the new code would be in place at shutdown and the next boot, but other changes might prevent it from running properly on shutdown (e.g. PHP syntax changes).

NOYB:
Thanks for the explanation Jim.  I will add the upgrade script to the PR.

NOYB:
Upgrade script added to PR.

jim-p:
@NOYB can you update your fork, there are conflicting changes

NOYB:
Conflicts resolved.

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
 		}
 	}
 }
+
+/* RAM Disk Management */
+function upgrade_160_to_161() {
+	global $g, $config;
+
+	if (isset($config['system']['use_mfs_tmpvar'])) {
```
rbgarga: please change it to if (!isset(...)) and return, then reduce one indent level

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
...
+
+		// Move existing RRD backup to the RAM Disk Store if it don't already exist there.
+		// Restore existing RRD XML dump backup.
+		if (file_exists("{$g['cf_conf_path']}/rrd.tgz") && !file_exists("{$g['cf_conf_path']}/RAM_Disk_Store/rrd.tgz")) {
+			$rrddbpath = "{$g['vardb_path']}/rrd/";
+			$rrdtool = "/usr/bin/nice -n20 /usr/local/bin/rrdtool";
+
+			$rrdrestore = "";
+			$rrdreturn = "";
+			foreach (glob("{$rrddbpath}/*.xml") as $xml_file) {
```
rbgarga: unlink_if_exists("{$rrddbpath}/*.xml"); can be used here

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
...
+				@unlink($xml_file);
+			}
+			unset($rrdrestore);
+			$_gb = exec("cd /;LANG=C /usr/bin/tar -tf {$g['cf_conf_path']}/rrd.tgz", $rrdrestore, $rrdreturn);
```
rbgarga: cd / can be removed, tar -tf is a read-only operation that will only print filenames

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
...
+			if ($rrdreturn != 0) {
+				log_error(sprintf(gettext('RRD restore failed exited with %1$s, the error is: %2$s'), $rrdreturn, $rrdrestore));
+//				return;
```
rbgarga: Remove commented out code

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
...
+			} else {
+				foreach ($rrdrestore as $xml_file) {
+					$rrd_file = '/' . substr($xml_file, 0, -4) . '.rrd';
+					if (file_exists("{$rrd_file}")) {
+						@unlink($rrd_file);
```
rbgarga: Use unlink_if_exists() and remove if (file_exists())

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
...
+					}
+					file_put_contents("{$g['tmp_path']}/rrd_restore", $xml_file);
+					$_gb = exec("cd /;LANG=C /usr/bin/tar -xf {$g['cf_conf_path']}/rrd.tgz -T {$g['tmp_path']}/rrd_restore");
```
rbgarga: Replace "cd /;" by tar parameter "-C /"

src/etc/inc/upgrade_config.inc
```diff
@@ -5063,4 +5063,88 @@ function upgrade_159_to_160() {
...
+			rename("{$g['cf_conf_path']}/dhcpleases.tgz", "{$g['cf_conf_path']}/RAM_Disk_Store/dhcpleases.tgz");
+			$ramds_updated = true;
+		}
+
+		// Move existing alias table backups to the RAM Disk Store if they don't already exist there.
+		$dbpath = "{$g['vardb_path']}/aliastables/";
+		$files = glob("{$g['cf_conf_path']}/RAM_Disk_Store{$dbpath}*.tgz");
+		if (count($files)) {
+			foreach ($files as $file) {
+				if (file_exists($file) && ! file_exists("{$g['cf_conf_path']}/RAM_Disk_Store/".basename($file))) {
```
rbgarga: file_exists($file) is redundant check since list was obtained with glob()

src/etc/rc.backup_logs.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+# rc.backup_logs.sh
+#
+# part of pfSense (https://www.pfsense.org)
+# Copyright (c) 2004-2016 Rubicon Communications, LLC (Netgate)
```
rbgarga: It's a new file, copyright should be only 2016

NOYB:
Requested changes made.

jim-p:
One additional request: For tracking purposes, please open a feature request ticket on https://redmine.pfsense.org/projects/pfsense that briefly describes the change (like your PR summary) and links back to this PR. Set the target to 2.4.

NOYB:
Jim,  Not able to reproduce the issue.  Opened discussion thread in 2.4 Development Snapshots forum.
https://forum.pfsense.org/index.php?topic=123602.0

NOYB:
Feature request ticket added in redmine.
https://redmine.pfsense.org/issues/7098

PR #2905, opened 2016-04-24
Fix issue with QinQ on LAGG interfaces where MTU doesn't apply to parent

Chrisc-c-c:
Fix for an issue with MTU not being applied to parent interfaces when QinQ on top of LAGG.

https://redmine.pfsense.org/issues/6227


cbuechler:
merged, thanks @Chrisc-c-c !

PR #2779, opened 2016-03-22
DHCPD: Allow DDNS Hostname option

Tommi2Day:
For entries into a remote  bind9 server i would like to provide the dhcp ddnshostname option for static dhcp entries.

PR #2907, opened 2016-04-25
add ddns-hostname option to DHCP Server

Tommi2Day:
add ddns-hostname option field to dhcp server.

This is the same as PR #2779, but now based on pfsense RELENG_2_3


src/usr/local/www/services_dhcp_edit.php
```diff
@@ -468,6 +482,13 @@ function staticmaps_sort($ifgui) {
 	$pconfig['hostname']
 ))->setHelp('Name of the host, without domain part.');
 
+$section->addInput(new Form_Input(
+	'ddnshostname',
+	'DDNS Hostname',
+	'text',
+	$pconfig['ddnshostname']
+))->setHelp('Name of the host to be register in DNS, without domain part.');
```
NOYB: 'Name of the host to ~~be~~ register in DNS, without domain part.'
Thanks for contributing.

rbgarga:
This PR is clearly broken, probably due to a bad rebase. Could you please submit a fresh one?

PR #2922, opened 2016-04-30
Fix Redmine #6097

NewEraCracker:
Based on my suggestion here:
https://forum.pfsense.org/index.php?topic=110698.msg618503#msg618503


cbuechler:
merged, thanks!

PR #2925, opened 2016-05-01
Fix checking of config.cache validity

NewEraCracker:
Make sure $config contains something other than NULL or FALSE. Previously code would only check for NULL.

On previous implementation: If, for any reason, config.cache were corrupted, then unserialize would return FALSE and configuration would be considered parsed anyway.

### DIFF
--- a/src/etc/inc/config.lib.inc
+++ b/src/etc/inc/config.lib.inc
@@ -138,7 +138,7 @@ function parse_config($parse = false) {
 	if (!$parse) {
 		if (file_exists($g['tmp_path'] . '/config.cache')) {
 			$config = unserialize(file_get_contents($g['tmp_path'] . '/config.cache'));
-			if (is_null($config)) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/config.lib.inc
```diff
@@ -138,7 +138,7 @@ function parse_config($parse = false) {
 	if (!$parse) {
 		if (file_exists($g['tmp_path'] . '/config.cache')) {
 			$config = unserialize(file_get_contents($g['tmp_path'] . '/config.cache'));
-			if (is_null($config)) {
+			if (!$config) {
```
phil-davis: It could even do
if (!is_array($config))
because if $config is not an array then there is trouble.

cbuechler:
Thanks @NewEraCracker and @phil-davis. I merged the one line change manually for cleaner history.

PR #2932, opened 2016-05-04
Removed global variables used on the version 2.2

SilvioGiunge:
Removed global vars that is not more used on this version in the file rrd.inc


pfSense-git-updates:

PR #2938, opened 2016-05-06
 Display hostname on header

antun3s:
It includes the hostname in the header next to the logout button.
This will facilitate the recognition of pfsense is operating.
selecao_001

I had a similar feature in previous versions, so I brought it to the latest version.


pfSense-git-updates:

sbeaver-netgate:
Thanks for you work. This feature has been added in another way (that does not consume menu real-estate.

PR #2940, opened 2016-05-06
Fix versions alignment in System Update

thelinuxfr:
Proposal for minor corrections of the alignment of the elements "version" and "status"


thelinuxfr:
Ongoing discussion to solve the general problem https://redmine.pfsense.org/issues/6327

sbeaver-netgate:
Thanks for this work. Inspired by your changes here I implemented them in CSS so that all input elements benefit from the change. I will probably revise it again once we see how it works in the field (big OS/browser matrix)

PR #2944, opened 2016-05-08
Add missing recommended key lengths/digest to Cert system

stilez:
Add missing key lengths to the certificate management pages, for keys of size 3072 (for current use), 7680, 15360 (for long term resistance), 8192 and 16384 (common binary exponents), and the missing digest "whirlpool".

The key lengths are supported by OpenSSL and for certain uses are currently recommended (eg long term resistance to replay/decryption). See keylength.com for citations.

The "whirlpool" digest is the only non-SHA1/2/3 family digest to be recommended and is especially valuable for users wishing to avoid the SHA families (for whatever reason).

Files changed - user manager, cert manager, ca manager, certs.inc

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -56,7 +56,7 @@ define("OPEN_SSL_CONF_PATH", "/etc/ssl/openssl.cnf");
 require_once("functions.inc");
 
 global $openssl_digest_algs;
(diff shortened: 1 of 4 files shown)


cbuechler:
Merged to master only, for now at least. If it checks out in more widespread usage, might cherry-pick back for 2.3.2.

PR #2951, opened 2016-05-11
Hide CIDR for alias types that do not use it

phil-davis:
Suggested/discussed in forum https://forum.pfsense.org/index.php?topic=111593.0
Note: This solution hides the CIDR entry dropdown and changes the grey-shaded box pfIpMask with the "/" to be empty. The grey-shaded box has a "span" which makes the preceding text box (for the port number, IP address, FQDN...) stretch out across the available column space - which is good.
If I hide the whole pfIpMask then both the slash and the whole grey-shaded box go away, but it the span effect is also lost.
So this solution leaves a little grey-shaded empty box between Port and Description fields.
If someone has an even better way to enhance this, go for it. I didn't want to mess with the underlying form class code.

### DIFF
--- a/src/usr/local/www/firewall_aliases_edit.php
+++ b/src/usr/local/www/firewall_aliases_edit.php
@@ -774,7 +774,17 @@ events.push(function() {
 
 		disable_subnets = (tab == 'host') || (tab == 'port') || (tab == 'url') || (tab == 'url_ports');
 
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
Phil,

I added a new JS function hideMask() that might make this easier for you.

Try: `hideMask('address_subnet', true);
`

phil-davis:
New PR coming soon...

phil-davis:
See #2952

PR #2952, opened 2016-05-11
Hide CIDR for alias types that do not use it

phil-davis:
Suggested/discussed in forum https://forum.pfsense.org/index.php?topic=111593.0


cbuechler:
Looks nice, thanks @phil-davis and @sbeaver-netgate

PR #2955, opened 2016-05-11
Tweak upgrade - Remove unused code and document changes better

NewEraCracker:
This will ensure a smooth upgrade for those running config 15.4 to any next version.

See my comment at https://github.com/pfsense/pfsense/commit/6086293b6203769634fbff6a0fe2d2d1c49f1e12 as I believe this is important.

Just my two cents.
Thanks!


cbuechler:
Merged, thanks! I may still re-use that 154 if a <=2.2.x config upgrade is needed pre-2.3.1, but that's fine for now.

PR #2971, opened 2016-05-19
Bug #6364 Memory increase, Bug#6367 LDAP timeout

Sandeep1991:
PHP Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 32 bytes) in /usr/local/www/status_carp.php on line 261

So increasing the memory in the includes php file might solve the issue.


pfSense-git-updates:

Sandeep1991:
I have submitted the CLA

cbuechler:
Thanks @Sandeep1991. I merged mostly the same changes here into all 3 branches, didn't merge this PR since we don't want to bump memory on 32 bit, there is an extraneous merge commit there that would just add noise to the history, and wanted the commits to be associated with their tickets ("Ticket #x" in commit log).

PR #2975, opened 2016-05-21
Services Status

NOYB:
Get Service Status Function

Add option for thumbs up/down icon.
Add option to specify title contents.
Distinguish between Stopped and Disabled with icons and text classes.

Replace status indicator icon fa-play with animated fa-circle-notch.
Replace status indicator icon fa-times with fa-stop and fa-ban.

Re-factor code logic for greater output flexibility.  Can return many combinations of status icons and text based on the function call parameters.

Status / OpenVPN - Status Indicator

Add prefixes to Status and Actions.
Remove textual status.  Icon and title should be fine.

Status / Services - Status Indicator

Use status icon instead of text.  Icon and title should be fine.

### DIFF
--- a/src/etc/inc/service-utils.inc
+++ b/src/etc/inc/service-utils.inc
@@ -492,7 +492,7 @@ function get_service_status_icon($service, $withtext = true, $smallicon = false,
 	if (get_service_status($service)) {
 		$statustext = gettext("Running");
 		$text_class = "text-success";
-		$fa_class = "fa fa-circle-o-notch fa-spin";
+		$fa_class = "fa fa-play";
 		$fa_class_thumbs = "fa fa-thumbs-o-up";
 		$Thumbs_UpDown = "Thumbs up";
 	} else {
--- a/src/etc/inc/service-utils.inc
+++ b/src/etc/inc/service-utils.inc
@@ -486,27 +486,50 @@ function get_service_status($service) {
 	return $running;
 }
 
-function get_service_status_icon($service, $withtext = true, $smallicon = false) {
-	global $g;
+function get_service_status_icon($service, $withtext = true, $smallicon = false, $withthumbs = false, $title = "service_state") {
 	$output = "";
+
 	if (get_service_status($service)) {
 		$statustext = gettext("Running");
-		$output .= "<a title=\"" . sprintf(gettext("%s Service is"), $service["name"]) . " {$statustext}\" ><i class=\"";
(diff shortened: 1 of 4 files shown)


sbeaver-netgate:
Thanks for this work. It can be hard to find just the right icon for status indicators. Somehow we need to differentiate "stop" the action, and "stopped" the status. The double icon approach you have taken here is OK but I have some concerns.

In general we use the fa-spin animation attribute to indicate that we are waiting for something to happen (e.g.: retrieving the current software versions) so also using that to indicate a service status may be confusing. This is also the only time we use two icons to indicate one status.

I still prefer to display the words "stopped" and "running" (or their translations) since we have room to do so in these cases. There is less ambiguity with words. c.f. Schipol airport.

A more appropriate single icon would be OK too. Perhaps we need to design our own?

NOYB:
re: differentiating "stop" the action, and "stopped" the status.
In addition to using different icons, most everywhere has an "Actions" and "Status" column or section.

re: fa-spin dedicated to single concept.  It's not just about the spin but the icon.  The spinning cog is good for waiting on a local action such as starting and stopping services etc.  The well known spinner is good for downloading.  The spinning circle notch I think is very appropriate for a "running" status indicator.  It is analogous to a power meter.  When it spins, stuff is running.

re: confusing.  Nothing that hovering once to see the title can't cure.

"This is also the only time we use two icons to indicate one status."  I didn't completely get what that is referring too.

Many combinations of text and/or icon are possible based on the parameters passed to the get_service_status_icon function.  Text only, icon only, thumbs up/down only, or combinations of them.  All colored with the text-success, text-warning, and text-danger classes.  As well as an assortment of titles for hover (state only, service and state, description and state, description service and state).

re: Space for text.  Some places there is some, such as widgets, there is not.  Nice to have consistent status indicators.  Where there is space, such as on the services status page, maybe a compromise; status icon plus status text.

sbeaver-netgate:
Sorry I have left this PR for so long. Now that I have reviewed it again I find that I like the changes to the logic, but I am less happy with the choice of icons. I think that if we could compromise on those it would be acceptable though.

I suggest replacing:
fa-play with fa-check-circle
fa-stop with fa-times-circle

While retaining the same colors/sizes you have employed.

I would also like to see the table column "status" on status_services.php centered so that the icons appear under the middle of the word "status, not at its left edge. This should be done via CSS if possible.

Apologies again for forgetting about this one :)

sbeaver-netgate:
I  merged this and changed the icons per the suggestions above. Was working on something else and it was easier for me to do this.

Thanks for the change.

PR #2977, opened 2016-05-22
Diagnostics - DNS Lookup, also query for IPv6 addresses

PiBa-NL:


pfSense-git-updates:

phil-davis:
You might like to:
s/resolve_host_adresses/resolve_host_addresses/
Just for future maintenance ease for English people.

PiBa-NL:
@phil-davis, ok thanks typo fixed

As for the 'missing CLA'.. ive signed that over a year ago or something like that.. should i do again.?

rbgarga:
@PiBa-NL You probably singed it when it was set to expire in 1 year. Now when you sign it again it won't expire anymore. Can you please do it?

src/usr/local/www/diag_dns.php
```diff
@@ -84,12 +83,20 @@
 	$counter++;
 }
 
+function resolve_host_addresses($host) {	
+	$host_esc = escapeshellarg($host);
+	$resolved = array();
+	exec("/usr/bin/drill {$host_esc} A | /usr/bin/grep {$host_esc} | grep '\tA\t\|\tCNAME\t' | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
+	$resolved_v6 = array();
+	exec("/usr/bin/drill {$host_esc} AAAA | /usr/bin/grep {$host_esc} | grep '\tAAAA\t\|\tCNAME\t' | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved_v6);
```
rbgarga: While you are touching this code, what do you think about using dns_get_record() instead shell calls to /usr/bin/drill?

http://php.net/manual/en/function.dns-get-record.php

src/usr/local/www/diag_dns.php
```diff
@@ -84,12 +84,33 @@
...
-	$resolved_v6 = array();
-	exec("/usr/bin/drill {$host_esc} AAAA | /usr/bin/grep {$host_esc} | grep '\tAAAA\t\|\tCNAME\t' | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved_v6);
-	return array_unique(array_merge($resolved, $resolved_v6));
+
+	$errreporting = error_reporting();
+	error_reporting($errreporting & ~E_WARNING);// dns_get_record throws a warning if nothing is resolved..
+
+	$dnsresult_cname = dns_get_record($host, DNS_CNAME);
+	$dnsresult_v4 = dns_get_record($host, DNS_A);
+	$dnsresult_v6 = dns_get_record($host, DNS_AAAA);
```
rbgarga: You can run a single call here, using DNS_A | DNS_AAAA | DNS_CNAME

rbgarga: Nevermind, it will return false if one of these items are not available. I did some tests and looks like the best way of doing that is running a single call using DNS_ALL, it'll return all CNAME, A and AAAA records.

It's better to use a single call because of timeout in case of problems, it will be reduced by 2/3

PiBa-NL: If i try DNS_ALL on webdir.online.lync.com i get nothing in return. If i ask for DNS_A it does return a ip.

cbuechler: DNS_ANY seems to be a better solution. If DNS_ALL gets NXDOMAIN back for anything, it returns false. DNS_ANY returns an array of records that exist among A, AAAA, CNAME, NS, and MX.

cbuechler:
There are several unnecessary commits here which will clutter up the history and make it harder to cherry-pick so I'm closing this one. @PiBa-NL if you can re-submit with DNS_ANY, we'll get that merged. Thanks!

src/usr/local/www/diag_dns.php
PiBa-NL: DNS_ANY is also bad, its return values depend on what was previously cached.. Which can be missing some records that actually do exist.

cbuechler: @PiBa-NL you mean previously cached by the configured DNS server? PHP issues new DNS queries every time dns_get_record is run. So yeah you might end up with a reply from cache from the configured DNS server, but I don't see how that's any diff regardless of whether it's DNS_ANY or DNS_A or DNS_AAAA.

PiBa-NL: With code below first time the IPv6 record is 'missing'.
```
	$host = "www.google.com";
	echo "<br/>\nANY:";
	$tmp = dns_get_record($host, DNS_ANY);
	print_r($tmp);
	echo "<br/>\nAAAA:";
	$tmp = dns_get_record($host, DNS_AAAA);
	print_r($tmp);
	echo "<br/>\nANY:";
	$tmp = dns_get_record($host, DNS_ANY);
	print_r($tmp);

```
```
ANY:Array ( [0] => Array ( [host] => www.google.com [class] => IN [ttl] => 299 [type] => A [ip] => 172.217.2.132 ) ) 
AAAA:Array ( [0] => Array ( [host] => www.google.com [class] => IN [ttl] => 299 [type] => AAAA [ipv6] => 2607:f8b0:400b:80b::2004 ) ) 
ANY:Array ( [0] => Array ( [host] => www.google.com [class] => IN [ttl] => 298 [type] => A [ip] => 172.217.2.132 ) [1] => Array ( [host] => www.google.com [class] => IN [ttl] => 299 [type] => AAAA [ipv6] => 2607:f8b0:400b:80b::2004 ) ) 
```

PiBa-NL: Other discussion regarding DNS ANY queries from haproxy.. http://thread.gmane.org/gmane.comp.web.haproxy/24691/focus=24698

PR #2979, opened 2016-05-24
Fix #6395

phil-davis:
Port tables do not seem to cope with having comments in them when used in pf.
This change puts back the editing (sed) for ports tables that was in RELENG_2_2 and moves the different editing (leaving comments) that is done in RELENG_2_3 so that it is only applied for ordinary URL table aliases.
Hopefully it will fix the reported issue. Needs to be checked/tested.


NOYB:
Full line comments are allowed in url table files.
It is unnecessary to revert that code.  Simply set the kflc flag to false on line 2186.

The real fix though is to correctly treat URL Table (Ports) aliases as a table.

phil-davis:
@NOYB that looks a much cleaner fix.
Everybody, see discussion in the Redmine issue.
I happy for a fix other than this one here to be made, and then close this PR.

phil-davis:
#2980 looks a nicer way to fix this.

rbgarga:
#2980 was merged. Thanks!

PR #2981, opened 2016-05-24
Fix GUI timeout and csrf expiry handling when timeout=0 or unset

stilez:
The gui timeout setting is handled inconsistently. 

- _system_usermanager_settings.php_ states that unset = default 240 mins and "0" = no timeout. But _guiconfig.inc_ treats "0" as zero minutes rather than a setting used to disable csrf expiry timeout
- _csrf_magic.php_ automatically sets the 'expires' variable = 7200. If guiconfig.inc doesn't explicitly set a value then this value might be used. This can never be correct handling (valid values are only default=240*60, user specified in $config or "never expire")

Fix:
 - add a $g['default_gui_timeout_mins'] = 240 to avoid a hard-coded value in multiple places;
 - ensure guiconfig.inc always sets a value for expiry, even if configured never to expire (in which case use 50 years or something) - see below. 
 - also clarify comment in system_usermanager_settings.php, and ensure it handles any possible future change to the default value displayed to the user (unlikely but still)

(**NOTE** - it could be changed more efficiently in csrf_magic.php where the expires default is first set, or by simply not calling csrf_start() in guiconfig.inc when non-expiry is configured. But the php file looks like a direct import from the csrf project so it's left unchanged, and it's not clear whether totally avoiding csrf_start() will have other side effects on session security, so I've instead chosen to set a very long timeout explicitly. If these are not issues then the handling should be changed from this PR for improved handling of never-expires and defaults)

### DIFF
--- a/src/etc/inc/globals.inc
+++ b/src/etc/inc/globals.inc
@@ -112,7 +112,8 @@ $g = array(
 	"help_base_url" => "/help.php",
 	"pkg_prefix" => "pfSense-pkg-",
 	"default_timezone" => "Etc/UTC",
-	"language" => "en_US"
+	"language" => "en_US",
+	"default_gui_timeout" => 240
 );
 
 /* IP TOS flags */
--- a/src/etc/inc/globals.inc
+++ b/src/etc/inc/globals.inc
@@ -113,7 +113,7 @@ $g = array(
 	"pkg_prefix" => "pfSense-pkg-",
 	"default_timezone" => "Etc/UTC",
 	"language" => "en_US",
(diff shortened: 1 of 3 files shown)


cbuechler:
The timeout on system_usermanager_settings.php is the session timeout, which is completely unrelated to the CSRF timeout, and is handled correctly in auth.inc. The CSRF timeout should be independent of the session timeout, as those aren't directly related things. Even if the session never expires, you don't want the CSRF to never expire. You don't want to change the CSRF timeout value based on the session timeout. 

I don't see anything here that's desirable.

PR #2984, opened 2016-05-24
Relax Suhosin to allow a 512M memory limit

NewEraCracker:
Because b20c7ef13d47d058bdea845c3b28503529e153fd and PHP uses 1024 multiples.


cbuechler:
merged, thanks!

PR #2987, opened 2016-05-31
Add "delete entry" for ARP table

stilez:
Useful function in some circumstances - seems no reason to force people to use command prompt for it.

Uses IP rather than hostname since not all ARP entries have hostnames.

Probably should also have "delete all" but not done that.


GTAXL:
I agree, this would be nice, and funny as I was going to suggest this as well.

PR #2988, opened 2016-05-31
Simplify convert_seconds_to_hms() and show days for large numbers of hours

stilez:
If the time represented is large (eg uptime might be several months) then hours becomes unhelpful, Put simply, it's clearer and more intuitive/meaningful to show "4921:02:06" as "205d 01:02:06". (Leading "days" value isn't shown unless >=1 for simplicity)

Optional arg added forcing HH:MM:SS without days (old behaviour) even when it's a large number of hours, in case it's useful in some contexts.

Notes -

1) Function  simplified and all "if" statements removed
2) Input validity check before trying to convert format
3) Renamed to convert_seconds_to_dhms()

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -1144,27 +1144,17 @@ function is_pppoe_server_enabled() {
 	return $pppoeenable;
 }
 
-function convert_seconds_to_hms($sec) {
-	$min = $hrs = 0;
-	if ($sec != 0) {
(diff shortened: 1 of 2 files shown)


stilez:
typo in commit c57e936 
"function name hms -> dhms to match edit to util.inc"

should be "...to pfsense-utils.inc".

sbeaver-netgate:
It would be a lot easier to merge if everything was in a single commit :)

PR #2989, opened 2016-06-01
Feature #6388 custom GUI preference settings per user

phil-davis:
This allows the admin to create/edit a user and check the new
customprofile box. The various GUI preferences are then displayed. The
admin can choose the various GUI preferences for that user - CS file
(theme), top navigation, dashboard columns, and a host of checkboxes
that turn on/off the other preferences that are available for the GUI.
That user then gets the specified GUI preferences, rather than the
system settings.
When that user changes their dashboard widgets and widget layout, that
is saved for that individual user also.
Note: This feature is only for local users. If a user is authenticated
remotely then this code provides no way to specify custom options for
the user.


phil-davis:
I had been thinking about this for a while. With multiple "admins" on a system doing different things it is nice if they (or the fully-privileged admin) can customize their own dashboard and other GUI preferences. It was mentioned in https://redmine.pfsense.org/issues/6388 by Manuel M and I had the time to have a look today.

NOYB:
I've not tried it out but thank you for this.  User specific config and settings has been noticeably absent thus far.  Hope it gets added.  Nice work.

stilez:
I'm not sure the GUI for this is in the right place. One would assume that a user should be able to modify their own basic custom GUI settings (if they can't then the custom user settings themselves need permissions). But if the GUI is in "user manager", the only person who can change them is someone with permission to manage users on the router, a much more restrictive set of users and not necessary. It would force people to ask someone with access to that page, to change their theme for them, or their custom widgets.

I think the GUI for this should be moved to an included file, so it can be used in two places - (1) where it is, so whoever manages the router can modify it for any user, but also (2) shown as an entry "User GUI customisation" in the system menu, where it only allows the current user to change their own custom settings.

phil-davis:
Yes, I will think about the detail. At the moment the various GUI preference settings are in a few places where the admin can set them. So I added equivalent ones to the user settings. That at least lets the admin set up the preferences when creating the user.
The dashboard layout is done interactively via manipulating the dashboard, so I kept that the same and made it so that a user with "custom profile" for their account will have their dashboard layout saved separately to the default system dashboard layout.
Perhaps it just needs an extra page where a user can set their own preferences. Then an admin can give the user access to that page if they want the user to be able to choose their own preferences.
I will think a bit more...

phil-davis:
Requirement) It shall be possible to grant a user the right to change GUI preferences of other users, without also granting that user the ability to create/edit users in general (i.e. without giving them the right to change other user's privs, password...)
Question) Is the above requirement needed?
I reckon that allowing an admin with User Manager privs to also setup initial user GUI preferences, then providing a way for each user to change their own preferences, would be sufficient.
The answer makes a bit of difference to the code needed.

stilez:
That (your last comment above) sounds fine.  Let user manager be used to set up users and also the  _default_ user GUI prefs, if desired and if any defaults are to be non-standard.  From there on, let  individual users change their own but (unless they have user manager permission) not those of other users). Avoids over complicating. 

I think this might be the easiest way to work it:  Use a url like system_gui-customise.php?account=9999. Add to the menu structure, a fixed link to that with the users' own account #, and to user manager a link next to each user for the same page with that user's account #. On the page itself have a dropdown "Customise GUI for the following user: <MY_NAME, ALL USERS, Adam, Beverley, Chris, ...>". Hide this driopdown if the current user doesn't have user_manager permission.  if they have the permission, then reload the page with the given account #. Deny page access or attempted changes, if (($_GET['account'] != their own account #) && (they don't have user manager permission)). Done.

phil-davis:
In the "Add page to manage user profile" commit, I have added a "User Profile" entry in the System menu. It only shows if the user has "customprofile" set for them. The page allows the user to choose the various profile settings that they want.

phil-davis:
The workflow is:
1) Admin person adds a new user (or edits an existing user) and selects "Custom Profile" and chooses the initial profile settings.
2) Admin person gives the user privs (e.g. puts them in admins group, gives them access to all pages, or gives them specific page access). If giving a user only specific pages, then the admin would normally give them "WebCfg - System: User Password Manager" (so the user can change their password), and would likely also now give them "WebCfg - System: User Profile" (so the user can choose their own profile settings).
3) User logs in. When they add/remove/move widgets on the dashboard, the layout is saved just for them. They can chose other profile settings in System->User Profile.

Comments please...

stilez:
Sorry for the longish comments, Phil -

1. **Page title** - Can we change it to "<u>User settings</u>"? To me, "User profile" suggests something quite a lot different than this should be, while "User settings" is clear as to what it might cover and matches the brevity we use for other page titles. Going forward, it also allows us eventually to centralise other user-centric customisations and settings that a user might want to have customised (for example some people may want 2000 log lines per page and others 100 lines, some may want graphs and traffic in bits and others in bytes, and so on).  A single page "user_settings.php" where the user can access all their UX settings in one place is probably easier all round, in the long term. <br /><br />(Note - being able to change any of your own user settings is safe. If you don't have permission to use a _page that uses the setting_, it does nothing)

2.  **Permissions** - To me this page seems a bit like the "logout" button or other fundamental of the GUI that any user should be able to access for themselves without needing permission to do so. If there is a permission to edit your own settings, maybe the default should be that it is granted, but I don't actually think one is needed - after all, it doesn't allow them to modify anything affecting anyone else, or impeding anyone elses' use of the UI.  (At worst using a different account would stop one person's GUI settings affecting another person).<br/><br/>In any event it would need a permission specifically for "WebCfg - System: User Settings (other users)", so it's clear what aspect the permission controls. But if we really must prevent some users changing their own GUI settings, add a _second_ new permission "WebCfg - System: User Settings (self)".  The latter would also need a help tip that it is overridden by both "user manager" and "user settings (other users)" permissions, since someone with either of those two permissions can always change any user's GUI, or can change far more serious settings for any other user than their GUI. (Could some permissions be non-binary with multiple levels?)

3. **Workflow** - The suggested workflow seems to put a lot of work on the user-managing admin, which could probably be pushed onto the user.  I'd suggest doing the workflow differently:<br /><br />(a) **system_usermanager.php -** A dropdown + button on the "create new user" form that pre-fills the user settings of the form (or specific sections of them) from another existing user's settings.  Templating by copying the settings from "any existing user" (or creating "dummy" users as templates) makes it simple to manage the defaults.<br /><br />(b) **"user_settings.php" -** add two options, first to "copy from existing user" (like that described at (a) for creating new user), and second - only for users with "user manager" or "user settings (other users)" permissions - a dropdown box "edit another user's settings" to select the user whose settings are being modified. The "save button" text should read "Save settings for <USERNAME>" if not the logged in user.

4. **Console reset theme/css** - One knock-on effect, the console menu needs an item "Reset web configurator theme+css" for the logged in user, so that if you manage to screw it up badly (eg you somehow set a dud theme with 0.1 point or 1000 point font, white text on near-white background, or just broke the CSS during testing, making the GUI unreadable) you can reset the few GUI settings that you need working, in order to login and use the GUI as usual to fix the rest.

5. **Consistency with dashboard control of widgets** - Last, for clarity, how do you plan that changing/controlling/general UI for widgets will play out between user_custom_settings.php" and the dashboard?  I'm a bit unclear whether you plan for widgets to be changed/configured on both of these pages, or some widget settings managed on the dashboard and some on the new page, or all customised on the dashboard as at present. Does this PR ensure that if some settings exist that affect widgets and can be changed via _both_ the dashboard _and_ this new page, then those settings and their controlling permissions will remain consistent and consistently managed between the two interfaces which they can be changed in, and that the settings for widgets aren't fragmented in an annoying way :)

phil-davis:
I'll start with point 5.
The widget layout on the dashboard (which widgets appear, in which column, in which order) all happens as it does now - the user adds/removes widgets and drags them around to position them. Underneath, if the user has "customprofile" set, that stuff is saved individually for that user.
The "User Profile" (which you suggest calling User Settings"), contains all the other user preferences that are not widget layout. So the 2 things have no overlap.
Note: There are widgets that have settings in them (e.g. Gateways widget has a display type setting). There is generic backend code already that seeks to generically cope with settings that are internal to widgets - index.php has stuff that thinks that widgets will set stuff in a key called widgetname-config (e.g. gateways-config). If the widgets used this "standard" then the code here would automagically "just work" to save the settings system-wide or per-user as appropriate. But most (all?) widgets just "do their own thing" in how they store their settings. I have not attempted to address that at all yet - so if someone changes "display type" in the gateways widget, it changes for everyone. To fix that requires looking at each widget and sorting out what it should do - which can be done as a later task after the main features here are sorted.

phil-davis:
1. Page Title - yes, happy with "User Settings" - I will make that change in the morning (I am in Oz and bed time is coming already), unless someone else has a better idea or argument.

phil-davis:
2. Permissions
Yes, perhaps there is no use case for giving a user "customprofile" but then preventing that user from changing their user settings. I already let such a user modify their dashboard widget layout anyway! But maybe such a user might have no dashboard, just a few specific menu items and the preference settings are already set up for them.
One way to make life easier is to automatically add the "WebCfg - System: User Profile" priv when "Custom Profile" is checked. Then it is effectively "on by default" but the admin can take it away if they like.

phil-davis:
3 Workflow
a) Templating a new user from an existing user sounds like "a good thing", and would include putting the new user in the same groups, privs... as the existing user. Need to think for a minute about certificates - in what cases they should be copied and in what cases they should really be made new for every user... - in any case, that sort of function would be "a good thing" in the existing system, regardless of this enhancement. So I will politely suggest that I (or someone) implements that as a separate thing.
b) A user's settings can already be modified in System->User Manager, by admin types. I am guessing that it would be only those type of users that would be changing/presetting the settings of other users anyway, so I am not sure that system_user_settings.php needs to be capable of editing the settings of other users?

phil-davis:
4 Console theme reset
I guess that is potentially needed now anyway, if someone sets a bad theme for the system-wide default. Perhaps it should set the system-wide default back, and, if the admin user has a custom profile, also set the admin user custom settings back to the default theme. That way the admin user webGUI can be made functional, and from there the admin can edit settings of other user(s) as required. Again, this could be done in a separate change, since it would be handy regardless of this PR.

stilez:
1 - yep
2 - yep as well - on by default works, granularity might not be used much but doesn't hurt
3 - (a) I didn't think of templating as copying the whole user, or a way to create new users (aslthough it can be headed that way and would be useful). But that kind of scope would raise a lot of issues to be thought through (as you mention). I simply thought of it as a way to copy "less problematic" user/GUI settings over, such as themes, css, and settings controlling some UI screens such as logs (length? sort order?) and graphs (bits? bytes?). Beyond that I'd suggest the KISS principle - get the page working with simple stuff, and anything else can be dealt with later "ad hoc" in future (like your answer to point 5 on widgets). That way it's available and helping users, it can be enhanced for other stuff later if the need arises.
3(b) - in that case the page would need to appear twice in the UI - once as a menu item in its own right ("change your own settings") and again as a subpage of user manager ("change someone elses)". That seems over loaded and possibly confusing, as well as unnecessary/clutter. I figured it's easier to have it in one place, a hidden-if-not-permitted button to select the user to apply to, and maybe a link to that page from user manager (much like other pages link to other relevant pages elsewhere). What do you think?
4 - I'm not _au fait_ with the shell side of the codebase, could you do this one?
5 - yep, leave widget fine tuning issues till another time, as long as they aren't broken and their permissions are still meaningful it's all good.

stilez:
In the code, all the variables use the term "profile" - not used elsewhere in the diffs for these files in this PR, so can probably be changed easily to be consistent ("profile" -> "settings"), which will avoid future confusion.

phil-davis:
Yeh, I just changed the front-facing text and minimal PHP filename... I will do a global search/replace on var names etc also...

phil-davis:
The other possible pieces of this are:
a) Ability to reset vital webGUI settings to defaults at the console (for if someone really gets the webGUI broken somehow) - https://github.com/pfsense/pfsense/pull/2997 - if that PR is merged, then I can add some code to that basic version to also reset the customsettings flag for admin.
b) Some way to easily give a new user the "User Settings" page permission, which you usually want to do if they are given "Custom Settings". This is kind of like the issue of giving them the page that lets them change their password - it sounds like something "everyone" can have, but then maybe not. An Admin can easily create a group that has these 2 page permissions and then put the users in that group, which is all quite easy to do. So I am not sure if a hard-wired button is really useful for this.

I will leave this code as-is for now, until the final version of https://github.com/pfsense/pfsense/pull/2993 is decided. Then  will integrate appropriately and make a fresh PR with just a single commit for the (hopefully) final version of this.

phil-davis:
I will check this after resolving conflicts and merging the latest changes in master and submit a new PR with single clean commit in a few hours.

phil-davis:
I need to sort out the new webguihostnamemenu stuff.

phil-davis:
Closing in favor of https://github.com/pfsense/pfsense/pull/3018

PR #2992, opened 2016-06-04
Add basic PHP info dump

stilez:
Although a "hidden" file, status.php is useful for debugging and development. 

With pfSense using a lot of PHP functionality, probably should include a section that dumps the key parts of phpinfo() for debugging/checking. Add at the end.


cbuechler:
Appreciate the suggestion. That's intended for support purposes, where I can't recall phpinfo ever being useful. That page is getting a bit unwieldy as-is (it's long enough to hang up some people's browsers for at least several seconds), I don't think it's a good idea to make it several pages longer by including information that's virtually never useful for its primary intended function.

PR #2993, opened 2016-06-05
Do not allow deleting your own user name

phil-davis:
Currently if you delete your own user name, then the config ends up with a blank user tag in it. Rather than fix that up, it seems dangerous to be able to delete yourself anyway, because if you are the last user with admin privs for which you know the password (i.e. if you have not recorded the password for "admin" somewhere), then you can lock yourself out. That would require console access to fix, which for some people is a pain.
It seems reasonable to me to make the person login as some other user with admin privs to delete "themselves".
Bit of boots and braces done here:
1) Don't show the trash bin icon for "yourself", and also disable the delete_check checkbox. So you can't opt to delete yourself from the ordinary front-end GUI.
2) Enhance the back-end validation to prevent deleting yourself, just in case someone mucks about in the front-end code.
3) Put error messages to tell people when something is not deleted, and why.
4) In the success message for multi-deletion, tell which user names have actually been deleted.

### DIFF
--- a/src/usr/local/www/system_usermanager.php
+++ b/src/usr/local/www/system_usermanager.php
@@ -110,13 +110,17 @@ if ($_GET['act'] == "deluser") {
 		exit;
 	}
 
-	conf_mount_rw();
-	local_user_del($a_user[$id]);
-	conf_mount_ro();
-	$userdeleted = $a_user[$id]['name'];
-	unset($a_user[$id]);
-	write_config();
-	$savemsg = sprintf(gettext("User %s successfully deleted."), $userdeleted);
+	if ($_GET['username'] == $_SESSION['Username']) {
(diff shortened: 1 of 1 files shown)


phil-davis:
@cbuechler I guess you are relaxing at BSD Canada. Can someone just comment on if this is the desired way to go. I have another PR developing for having custom GUI preference settings per user, and that will have conflicts with this. So it would be nice to sort out this PR, then I can easily ingrate the other with whatever code is committed here.

sbeaver-netgate:
I think this is a very sensible way to go so merging now.

Interesting regional variations on the idiom:
English: Belt and braces
American: Belt and suspenders
Australian: Boots and braces

I wonder what South Africans say.

PR #2994, opened 2016-06-06
Get certificate keysize

stilez:
Useful utility function when it's necessary to verify that existing keys meets current practices


cbuechler:
merged to master and 2_3, thanks @stilez

PR #2995, opened 2016-06-06
Add array of submenu interfaces

marianamorim:
Add array to show menu in pages edit of assign's interfaces submenu


cbuechler:
The edit screens, as a general standard, don't have tabs of their parent screens. To retain consistency across them, we'll leave that as-is.

rbgarga:
Looks like interfaces_groups_edit.php still have it, I'll remove to be consistent with others.

PR #2996, opened 2016-06-07
Fix #6460 Interface mismatch apply changes button

phil-davis:


cbuechler:
merged, thanks! 

not sure why the cleanup script didn't pick that up

PR #2997, opened 2016-06-07
Add resetwebgui to developer shell

phil-davis:
This might be helpful to people if they have set the theme to something
that they are having trouble displaying, reading... or enabled some
widget that is not good or...
It allows them to get back to a known-working dashboard state, from
where they can try again with customizations.
Having something like this was suggested by stilez in amongst discussion
of features for https://github.com/pfsense/pfsense/pull/2989

### DIFF
--- a/src/etc/phpshellsessions/resetwebgui
+++ b/src/etc/phpshellsessions/resetwebgui
@@ -1,0 +1,26 @@
+global $config;
+
+$config = parse_config(true);
+$default_theme = "pfSense.css";
+$default_columns = 2;
+$default_widgets = "system_information:col1:show,interfaces:col2:show";
+
+echo "Resetting webGUI:\n";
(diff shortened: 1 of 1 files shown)


stilez:
It also needs a link in the menu in rc.initial, doesn't it?

phil-davis:
This uses option 12, developer shell. There are a bunch of scripts there that do things to the config to "get out of a hole", so I put it there. At the prompt you:
playback resetwebgui

I thought the main devs might be averse to having extra things on the main menu - I'm sure they will give their opinion.

cbuechler:
That's obscure enough of a need that it doesn't belong in the main menu. Looks fine at a glance, will review closer and get merged later this week. Thanks!

phil-davis:
:+1:

PR #3000, opened 2016-06-08
Rename item #12 in console menu

stilez:
The console menu item #12 (_"pfSense/$product developer shell"_) has a pointless title that doesn't help users much.  It also actively obscures from router admins the useful tools it contains which is a negative side-effect.

* Describing it as a "developer" shell is redundant.  pfSsh.php isn't any more or less a "developer shell" than any other shell provided. Both shells can be used by developers or by end-user router admins, but its title as a "developer" [only] shell suggests it contains nothing of relevance to an end-user admin, which is very incorrect - it contains very useful packaged tools and versatility, to get the router recovered from a bad borking or if the web UI can't be used for any reason.

* It also isn't just a "pfSense/$product" shell. It's a full-fledged PHP shell, and obscuring this also hides its usefulness.

It's unlikely that non-dev users will "mess things up more with it" or "need protecting from themselves" -  any user prone to do so would already have used option (8) and full shell access for that purpose.

Renaming the menu item should help end-user admins, as the discoverability and usefulness that pfSsh.php provides will be more readily visible to users who might benefit from it.


cbuechler:
I was hesitant because it invalidates a lot of references all over the Internet, but this is a better name. Merged to master and 2_3. Thanks @stilez

PR #3003, opened 2016-06-14
added missing div.panel-body wrapper when reloading filters

hailthemelody:
Without this, content on status_filter_reload.php is missing the reference to the 10px CSS border.


hailthemelody:
error - my apologies will correct and resubmit

PR #3006, opened 2016-06-17
Remove subnet_expand()

stilez:
Function isn't used in main or packages repo, and in any case would need a complete rewrite to handle IPv6.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -877,30 +877,6 @@ function subnet_size_by_netmask($iptype, $bits, $exact=false) {
 	}
 }
 
-
-function subnet_expand($subnet) {
-	if (is_subnetv4($subnet)) {
-		return subnetv4_expand($subnet);
-	} else if (is_subnetv6($subnet)) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/util.inc
```diff
@@ -877,30 +877,6 @@ function subnet_size_by_netmask($iptype, $bits, $exact=false) {
 	}
 }
 
-
-function subnet_expand($subnet) {
-	if (is_subnetv4($subnet)) {
-		return subnetv4_expand($subnet);
-	} else if (is_subnetv6($subnet)) {
-		return subnetv6_expand($subnet);
```
phil-davis: That's weird that subnetv6_expand does not occur anywhere. This got commented on last year https://forum.pfsense.org/index.php?topic=90966.0
And for some reason it does not generate an error - I guess the PHP parser only looks for and complains about missing functions if the code in question is actually executed.

stilez: Yeah. I guess the function could be dynamically added, or something. But irrelevant really :)

PR #3008, opened 2016-06-17
Check IP Services

NOYB:
Configurable check ip services.
Used by Dynamic DNS services to check ip addresses.
Configure as many as check ip services as desired.  The first (highest in the list) that is enabled will be used as the default.
A future possibility could be to add an option to the Dynamic DNS entires to use a specific check ip service.
Addresses bug #6493


src/etc/inc/services.inc
```diff
@@ -1972,14 +1972,29 @@ function dyndnsCheckIP($int) {
...
 		}
-		$hosttocheck = "http://checkip.dyndns.org";
+
+		if (is_array($config['checkipservices']['checkipservice'])) {
+			foreach ($config['checkipservices']['checkipservice'] as $i => $checkipservice) {
+				if (isset($checkipservice['enable'])) {
+					$url = $checkipservice['url'];
+					$username = $checkipservice['username'];
+					$password = $checkipservice['password'];
+					$verifysslpeer = isset($checkipservice['verifysslpear']) ? true : false;
```
phil-davis: s/verifysslpear/verifysslpeer

src/etc/inc/services.inc
```diff
@@ -1972,14 +1972,29 @@ function dyndnsCheckIP($int) {
...
 		$ip_ch = curl_init($hosttocheck);
 		curl_setopt($ip_ch, CURLOPT_RETURNTRANSFER, 1);
-		curl_setopt($ip_ch, CURLOPT_SSL_VERIFYPEER, FALSE);
+		curl_setopt($ip_ch, CURLOPT_SSL_VERIFYPEER, $verifysslpeer);
 		curl_setopt($ip_ch, CURLOPT_INTERFACE, 'host!' . $ip_address);
 		curl_setopt($ip_ch, CURLOPT_CONNECTTIMEOUT, '30');
 		curl_setopt($ip_ch, CURLOPT_TIMEOUT, 120);
 		curl_setopt($ip_ch, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
+		curl_setopt($ip_ch, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
+		curl_setopt($ip_ch, CURLOPT_USERPWD, "{$username}:{$password}");
```
phil-davis: What happens here when username and/or password are blank? Is it happy to get CURLOPT_USERPWD ":"?

src/usr/local/www/services_checkip.php
```diff
@@ -0,0 +1,175 @@
...
+ *	STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *	ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ *	OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *	====================================================================
+ *
+ */
+
+##|+PRIV
+##|*IDENT=page-services-checkipserviceclients
```
phil-davis: Should this just be:
IDENT=page-services-checkipservices

src/etc/inc/services.inc
```diff
@@ -1972,14 +1972,29 @@ function dyndnsCheckIP($int) {
...
+					$url = $checkipservice['url'];
+					$username = $checkipservice['username'];
+					$password = $checkipservice['password'];
+					$verifysslpeer = isset($checkipservice['verifysslpear']) ? true : false;
+					break;
+				}
+			}
+		}
+
+		$hosttocheck = $url;
```
phil-davis: If the user deletes all the "Check IP Services" then $url is going to be empty here.
Perhaps there could be a function that returns the array of the "default" DynDNS check IP settings. That could be used here in an else of the above "if", and also in services_checkip_edit.php around line 90, where those "default" settings are currently done.

src/etc/inc/services.inc
NOYB: It doesn't seem to mind far as I've seen.

src/etc/inc/services.inc
NOYB: If user deletes them all then the user is saying don't check.  Not second guessing or mind reading the users intentions.  An empty URL is no different than an invalid or none existent URL.  The check will fail.

src/usr/local/www/services_checkip.php
NOYB: Corrected.  Thanks.

src/etc/inc/services.inc
NOYB: Corrected.  Thanks.

src/etc/inc/services.inc
cbuechler: I think the default needs to be hard-coded (probably in globals.inc), and show as such on the checkip page. We might point that to one of our own servers in the near future. Disabling is something that should be done on the individual config entries (though that's only on the RFC 2136 types currently, it should be added to the others as well), not by deleting the services available.

cbuechler:
@NOYB: could you please address the comment I left, and re-submit so the history's clean?

PR #3012, opened 2016-06-19
Add samba in the available packages

LeaderSoul:
Hello
I don't know if I'm writing this in the right place if not then I apologise.
My request is that add in the available packages samba 4 or which is newest version.
I think many people will be glad if this package can install and configurate via webgui.


NewEraCracker:
This is not the right way to elaborate a PR. Pushing from different version branches is a BAD idea.

AFAIK, you may request packages here: https://forum.pfsense.org/index.php?topic=6.540

phil-davis:
You need to start again with master, make a branch from master, put your edits into that branch, then make a pull request from that branch to master.
After review, the ESF guys will sort out cherry-picking (back-porting) whatever is appropriate to other maintenance branches (like RELENG_2_3 amnd/or RELENG_2_3_1).

jim-p:
This PR is broken, and samba isn't likely to ever be included.

PR #3013, opened 2016-06-20
Add include of functions.inc for declaration of gettext

dennypage:
Include functions.inc for gettext().

### DIFF
--- a/src/etc/inc/notices.inc
+++ b/src/etc/inc/notices.inc
@@ -53,6 +53,7 @@
  */
 
 require_once("globals.inc");
+require_once("functions.inc");
 require_once("led.inc");
 
 $notice_path = $g['tmp_path'] . '/notices';


cbuechler:
The only reason that gettext is in functions.inc is dating back to years ago when gettext didn't exist in the PHP version used in stable releases at the time, so people could gitsync without breaking things. 

I'm guessing the root cause you ran into is the same as this bug. https://redmine.pfsense.org/issues/6422

where somehow some people end up with broken/missing gettext in PHP. The only time I've been able to replicate that is in Azure, where it fails to write out extensions.ini during boot because it ends up getting "access denied". One other person somehow ended up with a system with screwed up libraries that broke gettext. In other cases, I don't know the root cause. 

I'm guessing you ended up in a similar circumstance as described in redmine 6422? That was probably breaking things in a lot more than just notices.inc, so not sure it's worth merging this.

dennypage:
The error results from invocation of /usr/local/bin/mail.php. Straight install of 2.3.1 on an SG unit. There should be a few crash reports submitted for this. Not sure why the builtin isn't being picked up, but the require of functions.inc corrects the issue. The require could have been added to mail.php, but several other files in /etc/inc have the require of functions.inc so it seemed to make sense. If there is a better solution, I'm good with that.

Discussion is here: https://forum.pfsense.org/index.php?topic=113728.msg633146

NewEraCracker:
It depends whether PHP is called with or without loading php.ini.
Bad things will happen if PHP is called with ```-n``` option.
This PR should be merged as band aid if that situation ever occurs.

cbuechler:
PHP isn't called with -n outside of where it's intentionally done in early bootup. Not sure why php-cgi isn't picking it up in that case, but this is fine. Merged to all 3 branches. Thanks!

PR #3014, opened 2016-06-21
fix rowhelper select_source empty combo

marcelloc:
while using $config['installedpackage']{['...'] as source


cbuechler:
This was cherry-picked from the other PR

PR #3016, opened 2016-06-22
added checking in the Form class

brunostein:
Added checking in the Form class to avoid empty strings when the gettext function is called.


sbeaver-netgate:
Duplicates #6934

PR #3017, opened 2016-06-23
Load actual value of webguihostnamemenu

phil-davis:
This is a string from a list of valid values (empty, "hostonly", "fqdn"). So it is not correct to just gather a true/false value here.


phil-davis:
Test case:
1) System->General Setup, for "Hostname in Menu" select "Hostname only" and save.
2) Go to some other page, then come back to System->General Setup.
"Hostname in Menu" displays "Fully Qualified Domain Name", which is not what is in config.xml.

after this fix:
"Hostname in Menu" displays "Hostname only", which is correct.

PR #3018, opened 2016-06-23
Feature #6388 custom GUI preference settings per user

phil-davis:
This allows the admin to create/edit a user and check the new
customprofile checkbox. The various GUI preferences are then displayed. The
admin can choose the various GUI preferences for that user (or can leave them as-is and let the user chose):
- CSS file (theme)
- top navigation
- hostname in menu
- dashboard columns
- Show the Available Widgets panel on the Dashboard (on/off)
- Show the Log Filter panel in System Logs (on/off)
- Show the Manage Log panel in System Logs (on/off)
- Show the Settings panel in Status Monitoring (on/off)
- Left Column Labels (on/off)
- Display page name first in browser tab (on/off)

That user then gets the specified GUI preferences, rather than the
system settings.
When that user changes their dashboard widgets and widget layout, that
is saved for that individual user also.
If the user has been granted the "WebCfg - System: User Settings" privilege/page then they can go to System->User Settings and choose/change the user settings for themselves.

So user preference settings can be managed in 3 general places:
a) Where they are now for system-wide settings (mostly in System->General Setup, webConfigurator section)
b) System->User Manager - for an admin to set user preference settings for particular user/s.
c) System->User Settings - for the user themselves to choose their preferred settings.

In addition, the dashboard widgets and their layout are saved using the same UI as at present. If the user has customsettings enabled, then those settings are saved just for them, otherwise they are saved as system-wide settings like now.

Note: This feature is only for local users. If a user is authenticated
remotely then this code provides no way to specify custom options for
the user.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1513,6 +1513,66 @@ function is_account_disabled($username) {
 	return false;
 }
 
+function get_user_settings($username) {
+	global $config;
+	$settings = array();
+	$settings['widgets'] = $config['widgets'];
+	$settings['webgui']['dashboardcolumns'] = $config['system']['webgui']['dashboardcolumns'];
+	$settings['webgui']['webguihostnamemenu'] = $config['system']['webgui']['webguihostnamemenu'];
+	$settings['webgui']['webguicss'] = $config['system']['webgui']['webguicss'];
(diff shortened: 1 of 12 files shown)


rbgarga:
It's also related with pfsense/freebsd-ports#142

phil-davis:
@rbgarga thanks for reminding me about that other PR, I updated it to use the latest $user_settings array name.

PR #3020, opened 2016-06-23
Allow passing comma in string quoted DHCP client options

shieldwed:
For some ISPs it is necessary to send DHCP client options, which contain commas in a quoted string. E.g:
```
dhcp-class-identifier "100008,0001,,pfSense dhclient 2.1"
```
Up until now, inserting such an option into `Interfaces > Lease Requirements and Requests >  Send options` and related options creates invalid dhclient config files. This patch allows setting commas in quoted strings.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4364,12 +4364,8 @@ function DHCP_Config_File_Advanced($interface, $wancfg, $wanif) {
 }
 
 function DHCP_Config_Option_Split($option_string) {
-	$options = [];
-	preg_match_all('/[^",]*(?:"[^"]*"[^",]*)+(?:"[^",]*)?|[^,]+/m', $option_string, $result, PREG_PATTERN_ORDER);
-	for ($i = 0; $i < count($result[0]); $i++) {
-		$options[] = $result[0][$i];
-	}
-	return $options;
+	preg_match_all('/[^",]*(?:"[^"]*"[^",]*)+(?:"[^",]*)?|[^,]+/m', $option_string, $matches, PREG_PATTERN_ORDER);
+	return $matches ? $matches[0] : [];
 }
 
 function DHCP_Config_File_Override($wancfg, $wanif) {
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4036,7 +4036,7 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif) {
 
 	$send_options = "";
 	if ($wancfg['adv_dhcp6_interface_statement_send_options'] != '') {
-		$options = explode(',', $wancfg['adv_dhcp6_interface_statement_send_options']);
+		$options = DHCP_Config_Option_Split($wancfg['adv_dhcp6_interface_statement_send_options']);
 		foreach ($options as $option) {
 			$send_options .= "\tsend " . trim($option) . ";\n";
 		}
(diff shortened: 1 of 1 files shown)


shieldwed:
Hmm, this seems to duplicate #2793. This might be an alternative implementation ;-).

NOYB:
Consider using foreach loop.  Cleaner than a for count loop.
Consider making the function generic.  ex: csv_str_to_array, $str, $matches, $array.
What is the use case for the second regex substring?  Given the first substring is it possible for the second substring to begin with a double quote?

Don't know which method is more desirable, explicit regex or the preg match call back.  I'm on the fence myself.  Like the delimiter and enclosure flexibility of the preg match call back way, but like the "simplicity" of the explicit regex way.

shieldwed:
Thank you for your review. I simplified my version quite a bit. I also tried to make it more flexible, but the regex became a monster as soon as i tried to interpolate variables. Therefore if we need a very flexible solution, then yours it is.

The first part of the regex `[^",]*(?:"[^"]*"[^",]*)+(?:"[^",]*)?` only matches tokes with quotes. The second part `[^,]+` only matches tokens without quotes. The thing is, that if you replace the `+` of the first part with `*` it matches empty strings as well and would return them.

Anyway, I don't care which solution is going to be merged as long as one of them is.

NOYB:
What about something like this?

Regex parts broken out for example clarity.  Personally that's probably how I'd write it too.  I like breaking up long or complex regex patterns to be more manageable and comprehendible.  But others may not like it like that.

```
function csv_str_to_array($str, $enclosure = '"', $delimiter = ',') {
/*
	$part1 = '[^'.$enclosure.$delimiter.']*';
	$part2 = '(?:"[^'.$enclosure.']*'.$enclosure.'[^'.$enclosure.$delimiter.']*)+';
	$part3 = '(?:'.$enclosure.'[^'.$enclosure.$delimiter.']*)?';
	$part4 = '[^'.$delimiter.']+';

	$csv_pattern = $part1.$part2.$part3.'|'.$part4;
/**/
	$csv_pattern = '[^'.$enclosure.$delimiter.']*(?:"[^'.$enclosure.']*'.$enclosure.'[^'.$enclosure.$delimiter.']*)+(?:'.$enclosure.'[^'.$enclosure.$delimiter.']*)?|[^'.$delimiter.']+';
	preg_match_all('/'.$csv_pattern.'/m', $str, $matches, PREG_PATTERN_ORDER);

	return $matches ? $matches[0] : [];
}
```
I was referring to the second sub-pattern.  The second set of parenthesis.  Not "or" part.
 
Anyway, nice work.  Yes I'd like to see one or the other get merged too.  Don't know why it hasn't been merged.  Would be nice if someone would at least make some kind of feedback comment on it.  Like "no way not going to happen because xyz", or "like it but probably hold off until abc".  etc.  The crickets MO (mode of operation) leaves people in the dark.  And in my opinion is disrespectful.

shieldwed:
I added some comments to the regex, hopefully they might help:
```
[^",]*        # match text in front of protected string
(?:
  ".*?"       # match protected string (with separators included)
  [^",]*      # match text after protected string
)+            # match one! or more protected strings in one token
(?:"[^",]*)?  # match string until next separator in case of unbalanced string quotes
|[^,]+        # match tokens without protected strings
```
The following two images should display the difference in result:

with the second sub-pattern:
with the 2nd sub-pattern

without the second sub-pattern:
without the 2nd sub-pattern

So instead of totally failing in case of unbalanced quotes, this part is trying to save, what can be saved. Probably this is worth a discussion.
- Should the code try to interpret, what it can and avoid issues?
- Should it just stubbornly do what it was being told?
- Should it raise an exception and therefore fail hard?

shieldwed:
I was actually writing this code with interpolated variables as well. But what a mess, this is is quite far from readable code. Compared to your solution with regex callbacks, yours feels more like reading a poem 😄 .

PR #3021, opened 2016-06-26
Fix Fatal Error - Cannot redeclare interfaces_bring_up()

NewEraCracker:
https://forum.pfsense.org/index.php?topic=114212.0

### DIFF
--- a/src/usr/local/bin/dhcpd_gather_stats.php
+++ b/src/usr/local/bin/dhcpd_gather_stats.php
@@ -25,7 +25,7 @@
 */
 
 require_once("config.inc");
-require("interfaces.inc");
+require_once("interfaces.inc");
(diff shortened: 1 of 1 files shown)


NewEraCracker:
This was caused by PR #3013 which lead to the following inclusion chain on this file
```config.inc > notices.inc > functions.inc > interfaces.inc```
This PR makes sure ```interfaces.inc``` is only included once.
I've done quick auditing on other files and no other issues have been found.

NewEraCracker:
On certain situations functions.inc may be included twice or more times, but code is already protected to handle that situation.

cbuechler:
Thanks @NewEraCracker. Closing this one in favor of PR 3022 which includes this change

PR #3022, opened 2016-06-27
Always use require_once

phil-davis:
The usage of require() and require_once() throughout the system is inconsistent, and "bugs" come up now and then when the order of "requires" is a bit different and some require() happens after the include file is already included/required.
It seems to me that there is no harm at all in always using require_once().


cbuechler:
Thanks @phil-davis. Merged this to all 3 branches. 

Where we use include I guess it's just a diff in that it won't fail if the included file is not found. Every use of it seems to be where it wouldn't matter much if the included file were missing. include_once should be fine. I don't see any uses of that where require_once would be preferable.

PR #3023, opened 2016-06-27
Add syslog support for tinc VPN daemon log files

dmegyesi:
Tinc VPN daemon is logging to /var/log/tinc.log. The Status->Tinc VPN page uses the "clog /var/log/tinc.log" command to check the status of the connected VPN clients. In order to fix this status page, syslog has to handle the tinc.log file.

(Further details: tinc can only output the connected client list to stdout (=> therefore, to tinc.log) by issuing a SIGUSR1 or SIGUSR2 signal to the daemon. Unfortunately, this is the only solution to make its status page work.)

If the tinc package is not installed on the given system, this change shouldn't cause any issue or change for those installations.

Related PR: https://github.com/pfsense/FreeBSD-ports/pull/149

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -860,7 +860,7 @@ function clear_all_log_files($restart = false) {
 	global $g;
 	exec("/usr/bin/killall syslogd");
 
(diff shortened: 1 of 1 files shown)


rbgarga:
This change is not necessary, it can be done direct via package XML file using following tags:
```
<logging><facilityname>
<logging><logfilename>
```

When it's defined on package XML file, code will create a separate logfile for specific facility using clog. You can check code here:

https://github.com/pfsense/pfsense/blob/master/src/etc/inc/system.inc#L879

PR #3024, opened 2016-06-27
Update include() to include_once()

phil-davis:
For safety, use include_once() when including various "side" files.
There are a couple of instances of include("guiconfig,inc") that I
changes to use require_once() to be consistent with everywhere else.
The remaining cases of include() are just (hundreds of) head.inc and
footer.inc - those are only at the start/end of the ordinary top-level
interactive PHP files. There is no nested interaction of those to think
about, so that should be fine.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -2289,7 +2289,7 @@ function portal_allow($clientip, $clientmac, $username, $password = null, $attri
 			$message = 0;
 		}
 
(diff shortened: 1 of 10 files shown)


cbuechler:
merged to master and 2_3, intentionally excluding 2_3_1 though I doubt it'll hurt anything, doesn't fix anything either (that we're aware of). 

Thanks @phil-davis !

phil-davis:
:+1: 
Correct - there is nothing here with a demonstrable bug, just changes for consistency and "safety".

PR #3029, opened 2016-06-30
Fix #6552 Handle more invalid IPv6 formats

phil-davis:
Redmine https://redmine.pfsense.org/issues/6552

All the following formats can currently be entered and accepted as "valid" IPv6 addresses, when actually they should be rejected.

1:2:3:4:5:6:7:8:1.2.3.4 - a full 8 pieces of IPv6 followed by an IPv4 on the end. (a valid address in this form has 6 colon-separated IPv6 parts followed by the IPv4 format on the end).
- fixed by line 930

::1:2:3:4:5:6:7:8 - the front "::" was getting uncompressed into "", resulting in "1:2:3:4:5:6:7:8" which looked good, but actually the input is bad.
- fixed by line 618

1:2:3:4:5:6:7:8:: - the end "::" was getting uncompressed into "", resulting in "1:2:3:4:5:6:7:8" which looked good, but actually the input is bad.
- fixed by line 623

1:2:3:4:5:6:::8 - the triple-colon ":::" happened to uncompress into "1:2:3:4:5:6:0:8" which looked good, but actually the input is bad.
- fixed by line 628

::::a - start with 4 colons
- fixed by lines 560 and 609-625

:::: - just input 4 colons
- fixed with a bit of enhancement at lines 610-612. We don't want to give back "0:0:0:0:0:0:0:0:" from uncompress(), so add a 9th 0 on the end to make it look nicer but invalid.

That's all I can find for now.

### DIFF
--- a/src/etc/inc/IPv6.inc
+++ b/src/etc/inc/IPv6.inc
@@ -557,7 +557,7 @@ class Net_IPv6
 
         if (false !== strpos($uip, '::') ) {
 
-            list($ip1, $ip2) = explode('::', $uip);
+            list($ip1, $ip2, $ip3) = explode('::', $uip);
 
             if ("" == $ip1) {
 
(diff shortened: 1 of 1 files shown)


cbuechler:
Merged, thanks @phil-davis. Now if only Pear would accept changes so all these issues could be fixed there, saw your PR there from a while back is still outstanding.

phil-davis:
I found out that they are still using SVN as the master for PHP PEAR stuff, so I have started by raising a first bug report to get all the whitespace, spelling and grammar stuff updated at https://pear.php.net/bugs/bug.php?id=21094
That will reduce the diffs between the PEAR repo and the pfSense version.
Then I will look at what we have done in pfSense and submit real bug reports with test cases.
I am manually keeping a Github repo https://github.com/phil-davis/Net_IPv6_svn that has the "official" PEAR version of the code and then commits for what I am reporting upstream. That way they can take versions of the code in response to each of my bug reports. (I have no idea how they want people to do the equivalent of a pull request, and am happy not to bother learning!)

PR #3030, opened 2016-06-30
Rationalize System Update GUI messages

phil-davis:
At present, when doing a System Update, there is a message box that says:
"Please wait while the installation of completes.
This may take several minutes."

Between "of" and "completes" is the package name, which is blank in the case of a System Update.

This should fix that issue by defining $pkg_wait_txt appropriately for the $firmwareupdate case.

Note: It was also easy to refactor out a couple of special "if firmwareupdate" tests, as the code "fell out" nicely by defining the other *_txt variables to suitable strings for the $firmwareupdate case.

### DIFF
--- a/src/usr/local/www/pkg_mgr_install.php
+++ b/src/usr/local/www/pkg_mgr_install.php
@@ -401,7 +401,12 @@ if ($_POST) {
 	}
 }
 
-if ($pkgmode == 'delete') {
+if ($firmwareupdate) {
+	$panel_heading_txt = gettext("Updating System");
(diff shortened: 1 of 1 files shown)


phil-davis:
I noticed this while updating a test VM to 2.3.2 development. I need to test that it really works when the next snapshot comes out and I can run this code while doing the update.

cbuechler:
Looks good, thanks @phil-davis. Merged to master and 2_3. If it checks out in 2.3.2 snapshots, we can cherry-pick back to 2_3_1 as well.

PR #3031, opened 2016-06-30
GitSync Diff Only Option

NOYB:
Diff Option
--diff
SSD Friendlier

Verbose Option
--verbose
Display constructed command.
In combination with the --diff option, display the array of different and missing files.

Dry Run Option
--dry-run
Run without syncing files.

Fixes

Files to copy variable state is used to determine whether or not to run the command.  Tar options being passed in the files to copy variable causes the command to always run.  Even if there are no files to be copied.  Separate out tar options from files to copy variable so command only runs if there are files to be copied.

Accommodate providing tar options such as '-C ./src' with the --minimal option.

Resolve function redefinition that prevents running gitsync again in same dev shell session (command recall) (function_exists).


cbuechler:
Merged to master and 2_3, thanks @NOYB

PR #3032, opened 2016-06-30
Fix firewall rules for PPPoE server

NewEraCracker:
1) PPPoE Clients are located within 'Remote Address Range' (sa) and not Server Address (ip), see lines 1194 and 1195.

2) Interfaces for floating rules were not being correctly added due to a bug on interface detection caused by PPPoE differences.

This commit fixes those two issues.


cbuechler:
Thanks @NewEraCracker 

also opened bug ticket here for tracking: 
https://redmine.pfsense.org/issues/6597

PR #3033, opened 2016-06-30
Diag Tables Last Update

NOYB:
Use correct Thursday abbreviation (s/Thr/Thu/).


cbuechler:
merged to all 3 branches, thanks @NOYB

PR #3034, opened 2016-07-01
startsWith Polyfill

NOYB:
Fixes hidding of custom individual settings when not in use on user manager page.

https://forum.pfsense.org/index.php?topic=114467.0


cbuechler:
Merged, thanks @NOYB

PR #3035, opened 2016-07-01
Fix some PHP notice issues

BBcan177:
Use of undefined constant min - assumed 'min'
Use of undefined constant max - assumed 'max'


cbuechler:
merged to all 3 branches, thanks @BBcan177 !

PR #3037, opened 2016-07-02
Check IP Services

NOYB:
Configurable check IP services.
Used by Dynamic DNS services to check IP addresses.
Configure as many check IP services as desired.  The first (highest in the list) that is enabled will be used as the default.
A future possibility could be to add an option to the Dynamic DNS entires to use a specific check IP service.
Addresses bug #6493 by allowing additional check IP service to be added and selected as the default.

### DIFF
--- a/src/etc/inc/globals.inc
+++ b/src/etc/inc/globals.inc
@@ -238,4 +238,15 @@ if (file_exists("{$g['cf_conf_path']}/enableserial_force")) {
 
 $config_parsed = false;
 
+/* Factory default check IP service. */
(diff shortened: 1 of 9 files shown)


cbuechler:
Merged for 2.4, thanks @NOYB

cbuechler:
added feature ticket https://redmine.pfsense.org/issues/6591

NOYB:
Disappointed that it won't be available until 2.4.  But at least I have it.  :)

phil-davis:
See #3071 for a backport PR of this to RELENG_2_3

PR #3042, opened 2016-07-05
Protect get_dir from causing PHP errors

NewEraCracker:
See: https://forum.pfsense.org/index.php?topic=114570.0

Thanks!


cbuechler:
Merged to all 3 branches, thanks @NewEraCracker

PR #3044, opened 2016-07-07
Automagically determine if imported alias is host or network type

phil-davis:
At the moment, if all the IP addresses in an imported list are single IPs, then the alias type is still being set to "network". It does not cause any real issue, since a "network" alias with all /32 entries is just like a "host" alias. But it is easy to determine, so why not.
This is the first easy step in seeing if we can auto-detect more alias types on input, as suggested by https://redmine.pfsense.org/issues/6582


phil-davis:
Note: #3046 includes the code here, so if you like that PR, then take it and close this one.

rbgarga:
#3046 was merged, thanks!

PR #3046, opened 2016-07-07
Fix #6582 Allow import of ports

phil-davis:
This change does:
1) Allow bulk import of a port alias. That happens when the Import
button is pressed from the Ports tab of the firewall_aliases screen.
2) Allow bulk import of an IP alias, automatically determine if the
imported data fits a host or network alias type. That happens when the
Import button is pressed from the IP tab of the firewall_aliases screen.
3) When Import is pressed from the "All" tab of the firewall_aliases
screen, make an IP alias.
4) Do not display the Import button when on the URL tab of the
firewall_aliases screen.

### DIFF
--- a/src/usr/local/www/firewall_aliases.php
+++ b/src/usr/local/www/firewall_aliases.php
@@ -306,10 +306,16 @@ display_top_tabs($tab_array);
 		<i class="fa fa-plus icon-embed-btn"></i>
 		<?=gettext("Add");?>
 	</a>
-	<a href="firewall_aliases_import.php" role="button" class="btn btn-primary btn-sm">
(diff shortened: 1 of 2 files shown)


rbgarga:
There is a case it's failing, if you add a port item when importing IP aliases it is not detecting it

phil-davis:
Numbers are valid names also - you can have an FQDN like:
1234.example.com
So if you are in the domain example.com, you can use just "1234" as the name, which can go in an IP alias and can be successfully resolved as 1234.example.com
The spec for host/domain names makes it impossible to tell the difference between a host name that is a string of digits and a port number.

PR #3048, opened 2016-07-08
Added option for DHCPv6 before RA

marjohn56:
Some ISP BNG's require that a DHCP6 solicit is sent before RA.

Interfaces.php has additions to add new flag 
'dhcp6withoutra' to configuration.

Interfaces.inc uses this flag to decide whether to launch dhcp6c on its own or as presently done via rtsold.

Added an extra function call 'kill_dhcp6client_process($interface) and removed multiple posix_kill calls, this is to make it operate in similar way to the  kill_dhclient_process(***)


cbuechler:
Thanks @marjohn56. This will need to be resubmitted without the merge commits to not clutter up the history unnecessarily and ease cherry-picking. I'm guessing your local copy became outdated and you didn't do a 'git fetch && git rebase origin' before the git push. 

If you can re-submit as a single commit PR in the next 3 days or so, we can get this in for 2.3.2 release. Otherwise it can make 2.4.

PR #3049, opened 2016-07-08
Create /var/run/dmesg.boot symlink for vm-bhyve (Feature #6573)

doktornotor:
See https://redmine.pfsense.org/issues/6573


cbuechler:
He's back! Thanks @doktornotor

PR #3050, opened 2016-07-09
Fix #6590 NTPd Leapfile Upload

phil-davis:


cbuechler:
Good catch, thanks @phil-davis !

PR #3051, opened 2016-07-09
Fix #6050 DHCP - provide Network Booting display/hide advanced button

phil-davis:
This keeps the section header, but instead of a +/- on the right of the section header bar, it puts a Display/Hide Advanced button, like other parts of this GUI page.
The section header looks difficult to remove, because the "Additional BOOTP/DHCP Options" above has its own section, which has to end. So I am leaving the "Network Booting" section header.

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -1237,12 +1237,23 @@ $section->addInput(new Form_Button(
 
 $form->add($section);
 
-if ($pconfig['netboot']) {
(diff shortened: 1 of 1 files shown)


phil-davis:
Also look at #3052 where I have put the "Network Booting" fields above "Additional BOOTP/DHCP Options", which (IMHO) makes the UI more consistent.
Pick whichever PR you like and close the other.

rbgarga:
Closing in favor of #3052

PR #3054, opened 2016-07-10
Advanced DHCP config breaks option to ignore DHCP lease from specific IP

nkaminski:
When the advanced DHCP configuration option checkbox is checked for an interface setup as a DHCP client, such as the WAN interface, the config option to ignore leases from a certain DHCP server is no longer included in the actual DHCP server config file and therefore no longer has any effect. This change modifies the DHCP_Config_File_Advanced function to include the "reject x.x.x.x;" line if applicable in the advanced configuration as well.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4355,6 +4355,9 @@ function DHCP_Config_File_Advanced($interface, $wancfg, $wanif) {
 	$dhclientconf .= "{$required_options}";
 	$dhclientconf .= "{$option_modifiers}";
 	$dhclientconf .= "\n";
+	if(is_ipaddrv4($wancfg['dhcprejectfrom'])){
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

nkaminski:
CLA submitted

src/etc/inc/interfaces.inc
```diff
@@ -4355,6 +4355,9 @@ function DHCP_Config_File_Advanced($interface, $wancfg, $wanif) {
 	$dhclientconf .= "{$required_options}";
 	$dhclientconf .= "{$option_modifiers}";
 	$dhclientconf .= "\n";
+	if(is_ipaddrv4($wancfg['dhcprejectfrom'])){
```
phil-davis: Just style - we put spaces between "if" and "(", and also between ")" and "{".

cbuechler:
Merged, thanks @nkaminski

PR #3056, opened 2016-07-13
Some tweaks to improve alignment in table with checkbox

NewEraCracker:
1) If a checkbox does not have a description (even if it is empty), layout will be broken as checkbox won't be aligned correctly

2) UPnP checkbox looks better with description instead of help

This commit fixes the two issues


sbeaver-netgate:
Thanks for the improvement.

PR #3057, opened 2016-07-13
The ability for the Radius auth server to detect openVPN and distinguished different server configurations

wokis:
I've added the ability for the Radius auth server to detect openVPN and distinguished different server configurations (by port). I've also corrected some of the Radius attributes to be more in line with what strongSwan uses.

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -851,7 +851,7 @@ function openvpn_reconfigure($mode, $settings) {
 					if ($settings['strictusercn']) {
 						$strictusercn = "true";
 					}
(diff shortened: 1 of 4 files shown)


src/etc/inc/openvpn.inc
```diff
@@ -103,8 +103,7 @@ $openvpn_verbosity_level = array(
 
 global $openvpn_dh_lengths;
 $openvpn_dh_lengths = array(
-	1024, 2048, 3072, 4096, 7680, 8192, 15360, 16384
-);
+	1024, 2048, 4096);
```
cbuechler: this overwrites a recent change that it shouldn't

wokis:
I've must have pushed changes that shouldn't have been there. I force pushed and overwrote my repo, this should have fixed the pull request.

PR #3059, opened 2016-07-15
Fix xml syntax after license update

NewEraCracker:
This looks like an error to me so making a quick fix.


rbgarga:
good catch, thanks!

PR #3060, opened 2016-07-16
Fix "http://pf.sense/UNKNOWN" links in Pkg Manager

lukehamburg:
I noticed that some packages do not have the 'www' field in the database filled, or have it set to UNKNOWN.  The way the table is built on the 'Available Packages' page causes bad links to be generated pointing to e.g. http://pf.sense/UNKNOWN.  This patch fixes it, causing only packages with actual links to get the `<a>` tag.


cbuechler:
Merged, thanks @lukehamburg 

ticket opened for tracking: https://redmine.pfsense.org/issues/6617

PR #3061, opened 2016-07-17
dyndns: add IPv6 support for CloudFlare

thoughtpolice:
This adds support for managing IPv6/AAAA records to CloudFlare. It uses the extant IPv6 code paths, with only a minor tweak needed to the primary code path to distinguish which record type to update.

I developed and tested this patch against the stable `RELENG_2_3_1` branch (as this is what my router was running), and forward-ported it to the `master` branch. The patch against 2.3.1 which I'm currently using in anger can be found here: https://github.com/thoughtpolice/pfsense/tree/cloudflarev6_2_3_1

Tested on a pfSense 2.3.1_p5 CE virtual machine with a public IPv6 address, as well as pfSense 2.3.1_p5 official Netgate build, on an SG-4860 (my home router with IPv6 provided by my ISP).

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -47,6 +47,7 @@
 	 *    - Custom DDNS (any URL)
 	 *    - Custom DDNS IPv6 (any URL)
 	 *    - CloudFlare (www.cloudflare.com)
+	 *    - CloudFlare IPv6 (www.cloudflare.com)
 	 *    - Eurodns (eurodns.com)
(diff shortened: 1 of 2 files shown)


cbuechler:
Merged, thanks @thoughtpolice 

ticket for tracking: https://redmine.pfsense.org/issues/6623

thoughtpolice:
Thanks for the quick response! And thank you for pfSense. It's a great product and I actually love my router now :heart:

PR #3055, opened 2016-07-10
DHCP6 init before RA

marjohn56:
Some ISP BNG's  require that a dhcp6c init is sent before RA. The
additions to interfaces.php adds a new option flag 'dhcp6withoutra',
this flag is set or unset in the WAN DHCP6 section 'Do not wait for a
RA'.

The interfaces.inc file handles this flag and either launches dhcp6c on
interface_dhcpv6_configure or if the flag is not set rtsold will
launch dhcp6c on receipt of RA.

An extra addition is the new function kill_dhcp6client_process($wanif).
This is added to try and create a more robust termination of a dhcp6c
clients that may be running on that interface.

PR #3063, opened 2016-07-20
Backport DHCP6 init before RA

phil-davis:
Original PR to master was #3055
Now that 2.3.2 has been branched, this is an example of stuff that could be back-ported to RELENG_2_3 to make it available to 32-bit systems that will never be able to upgrade to 2.4

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4048,7 +4048,7 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 	if (isset($wancfg['dhcp6withoutra'])) {
 		kill_dhcp6client_process($wanif);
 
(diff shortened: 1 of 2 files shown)


marjohn56:
There is a fix needed to fix Phil. My only tester has found that an extra '-x' flag exists on the command line for dhcp6c when launched if dhcpwithoutra is true, This extra flag is used by a version of dhcp6c I am working on and will cause an exit without release. It needs to be removed from the existing files before and backports are made.

phil-davis:
@marjohn56 Can you make a pull request to master for the fix needed in master. Then I will add a back-port commit here to do the same thing, which I guess will be to remove the "-x" from line 4051 of interfaces.inc

marjohn56:
Cheers Phil, have done so.

PR #3071, opened 2016-07-20
Backport Check IP Services

phil-davis:
Original pull request to master was #3037


phil-davis:
These pull requests are all the code committed to master since 1 July 2016 that are for stuff that will work fine in the RELENG_2_3 branch on pfense 2.3.* and FreeBSD 10.3
#3062 #3063 #3064 #3065 #3066 #3067 #3068 #3069 #3070 #3071 
I believe that these would all be useful to people running 2.3.* and particularly those on 32 bit systems like Alix who will not be able to upgrade to 2.4

PR #3072, opened 2016-07-20
Fix DHCP Release and some Status > Interfaces changes

GTAXL:
Ticket #4209: Send DHCP Release packet
When you hit 'Release' on the status_interfaces.php the dhclient did not
send the proper release packet to the DHCP server. This is sometimes
required by some ISP's and the freebsd dhclient doesn't do it, however I
found a quick fix. A copy of ISC's dhclient is sitting on pfSense 2.3 so
we will utilize it for this function.

A few minor touchups on the wording along with showing the configuration
type after the IP address so you know where the IP was obtained from,
similar to DD-WRT/Tomato, etc. It also shows the CIDR notation after the
subnet masks since pfSense refers to the CIDR notation everywhere else.
It also shows the name of the NIC cuz I thought it was interesting to
show there.
https://forum.pfsense.org/index.php?topic=113088.msg629905#msg629905

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1273,6 +1273,7 @@ function interface_bring_down($interface = "wan", $destroy = false, $ifacecfg = 
 			}
 			break;
 		case "dhcp":
+			mwexec("/usr/local/sbin/dhclient -v -r -lf /var/db/dhclient.leases.{$realif} -i {$realif}");
 			kill_dhclient_process($realif);
 			unlink_if_exists("{$g['varetc_path']}/dhclient_{$interface}.conf");
 			if (does_interface_exist("$realif")) {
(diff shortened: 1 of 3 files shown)


cbuechler:
Thanks for the submission. That's not the dhclient that's actually used, can't do a release like that. The getinterfacename and netmask2cidr functions duplicate other functions that already exist and need to be used instead. status_interfaces.php already shows DHCP, PPPoE, etc. for non-static IP type interfaces, so adding that again is redundant clutter IMO. 

I think the only desirable change here would be to do a release from dhclient.

GTAXL:
I know that's not the dhclient that you use, but you ship it along with pfSense and the one you use doesn't release properly like this one. It does work and I've tested it. Along with the functions, which ones do I use? I'm new to PHP but know the basics and did quite a bit of Googling just to figure this out. I'm upset you find my changes clutter as I aim to improve pfSense, but if it's not appreciated this is a deterrent for me to keep contributing as I have many tweaks/improvments all across pfSense's web UI. :( I'm also wanting to pull the DHCP client's lease information and display that, if that's not clutter, and I find this will be a very hard task as there is a lot of parsing involved and converting the format of the date, etc. Maybe you know best..

PR #3073, opened 2016-07-21
Backport Add missing recommended key lengths/digest to Cert system

phil-davis:
Original pull request to master was #2944


phil-davis:
This is the last thing that I noticed that seems reasonable to backport to RELENG_2_3
The other diffs in master are mostly license text and build-related script/file changes.

PR #3074, opened 2016-07-21
Backport openvpn_add_dhcpopts already sets redirect-gateway

phil-davis:
Ticket 6633
Original commit to master was
https://github.com/pfsense/pfsense/commit/f8038899f250c656b1ef03fe351fb9cfdadeaf0c
Adding this PR for completeness so that this is visible as something
that can be back-ported to RELENG_2_3 at some time in the future after 2.3.2 is done and dusted.

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1233,10 +1233,6 @@ function openvpn_resync_csc(& $settings) {
 
 	openvpn_add_dhcpopts($settings, $conf);
 
-	if ($settings['gwredir']) {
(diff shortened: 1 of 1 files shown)


phil-davis:
General note on all these "back-port" pull requests:
I have put them here so that they are visible, making it easy for anyone to see what functionality is in master (heading for 2.4) but not (yet) in the 2.3.* series.
In practice, it might prove to be easier, and make the history easier to follow for the future, if these things are back-ported by actually cherry-picking the relevant commits from master back into RELENG_2_3. In that case, feel free to do the cherry-picks and then close the related pull request without merging it.

PR #3078, opened 2016-07-25
Correction to dhcp6 before RA

marjohn56:
Remove '-x' flag from dhcpwithoutra launch of dhcp6c.
In a previous commit ( accepted ) that was made to fix an issue with certain ISP's requiring that a dhcp6 init be sent before RA there is an extra '-x' flag on the command line for dhcp6c. This flag is going to be used for an updated dhcp6c client to enable a 'no release' option. In the current release of dhcp6c it causes the client to error and exit and must therefore be removed.


phil-davis:
The equivalent fix for RELENG_2_3 is in pull request #3063 commit https://github.com/pfsense/pfsense/pull/3063/commits/efd07e5e116a60745706e5d306c1f912eaa0ed78

rbgarga:
Merged manually due to lack of original repo. Thanks!

PR #3080, opened 2016-07-26
Display local DHCP lease times in 24-hour clock

phil-davis:
It seems odd to me that when the times are displayed in UTC they have
24-hour clock, but when displayed in local time they are formatted with
12-hour clock and AM/PM.
24-hour format takes less screen space, and I would have thought that
network admins would prefer working consistently in 24-hour clock
format.


sbeaver-netgate:
Agreed. Thank you Phil

PR #3081, opened 2016-07-29
Fix a redundant HTTP "User-Agent" string.

philipsoeberg:
CURLOPT_USERAGENT expect the value to the user-agent string, not the entire key-value pair.
Before this fix, HTTP header "User-Agent: User-Agent: phpDynDNS/0.7" was sent for DDNS updates.
NGINX configuration at GratisDNS will not accept a user-agent formatted in the above way.
This commit fixes GratisDNS Dynamic DNS service.


philipsoeberg:
incorrect e-mail address used in commit. Affects CLA.. will resubmit.,

PR #3083, opened 2016-07-31
Allow URLs for TFTP Server (Bug #6634)

doktornotor:
The setPattern() thing ain't usable for this and just causes regressions; also is_URL() from util.inc is way too limited for this purpose.


pfSense-git-updates:

doktornotor:
Guys, you've got some issues with your CLA script (already done), plus I do not think this damned paperwork is useful any more after relicensing the whole thing to Apache license.

jim-p:
The original CLA signing was set to expire in a year, new one won't expire, you do need to sign it again though, but just one last time.

ravenblue12:
I tested it and it works!

PR #3084, opened 2016-07-31
Experiment using the OS' native font stack

jlduran:
Following the steps of bfbfa4dd254c110db4122925d0a3716a34d4c664, experiment using the OS' native font stack. Obtained from the Bootstrap project (twbs/bootstrap#19098, twbs/bootstrap@0a31dacf3e0c43a73b2e61e266155b9dcce35339)

### DIFF
--- a/src/usr/local/www/css/pfSense-BETA.css
+++ b/src/usr/local/www/css/pfSense-BETA.css
@@ -23,7 +23,8 @@
 /*** Experimental Changes Go Here ***/
 
 body {
-	font-size: 12px;
+    font-size: 12px;
+    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
 }
 
 .navbar-nav {
@@ -31,7 +32,8 @@ body {
 }
 
 table[data-sortable].sortable-theme-bootstrap {
-	font-size: 12px;
+    font-size: 12px;
+    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
 }
 
 .table>tbody>tr>td {
--- a/src/usr/local/www/css/pfSense-dark-BETA.css
+++ b/src/usr/local/www/css/pfSense-dark-BETA.css
@@ -23,7 +23,8 @@
 /*** Experimental Changes Go Here ***/
 
 body {
-	font-size: 12px;
+    font-size: 12px;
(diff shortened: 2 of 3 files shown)


gonzopancho:
@sbeaver-netgate should look this over

gonzopancho:
@jdillard too

jdillard:
You would just need to leave sans-serif as the font-family for this to happen. Although I'm not sure why we would want to provide an inconsistent 'look and feel' to all users, when a user can prevent their browser from allowing websites to choose their own fonts and achieve that same result if that is their preference.

jlduran:
> You would just need to leave sans-serif as the font-family for this to happen.

It's the last resource:

```css
font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
```

> I'm not sure why we would want to provide an inconsistent 'look and feel'

Inconsistency is _very_ subjective, I for one think about consistency in terms of the OS: i.e. `Roboto` will look perfect on an Android phone, but not on an iPhone.

It's supposed to be an experiment. Try it with different OSs, the difference is subtle enough not to "break" the current styling.

If you are watching this message on GitHub, you'll see what I mean.

jdillard:
Ah so that's why Github's font changed recently, never took the time to look into it. Wouldn't the better approach be to update bootstrap to the version that includes that fix instead of putting it in the css override file?

jdillard:
Also, experimental CSS changes are best places in pfSense-BETA.css. If you make the change there and leave the Roboto files, I could see merging it with a lot less chance of conflict. That gives people the opportunity to try it without changing their current experience and to get feed back in the forums before putting it in the "default" CSS sheet.

jlduran:
> Wouldn't the better approach be to update bootstrap to the version that includes that fix

Absolutely, although that would be the v4 of Bootstrap, which is still alpha. Also, [this line](https://github.com/pfsense/pfsense/pull/3084/files#diff-96e5c95f451ec946f95e5c95ac150835R756) overrides the default styling for sortable tables, which is inconsistent with the current styling.

> experimental CSS changes are best placed in `pfSense-BETA.css`

I'll amend the commit then 👍

jlduran:
@jdillard done.

The first commit fixes the sortable tables (currently not using `Roboto`). In case you want to cherry-pick and back port it.

The second commit tells the BETA stylesheets to use the operating system's native font.

jdillard:
It is merged now, thanks for the PR, jlduran! Also, thanks for taking the time to explain and amend!

PR #3086, opened 2016-07-31
services_dhcp: Ignore BOOTP queries

jlduran:
BOOTP leases do not have a maximum lease time by default, this could potentially lead to a DHCP address pool exhaustion.

This commit adds an option to ignore BOOTP queries.

Redmine [#4351](https://redmine.pfsense.org/issues/4351)


jlduran:
This can be easily reproduced:

step-1

profit

PR #3087, opened 2016-08-01
Corrections and additions to dhcp6 before RA

marjohn56:
Changed kill dhcp6c process to use kill -9 $pid, it seems more reliable. 

Modified rtsold script creation to prevent double launch of dhcp6c.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1301,9 +1301,9 @@ function interface_bring_down($interface = "wan", $destroy = false, $ifacecfg = 
 		case "slaac":
 		case "dhcp6":
 			$pidv6 = find_dhcp6c_process($realif);
-			if ($pidv6) {
-				posix_kill($pidv6, SIGTERM);
-			}
+			mwexec("/usr/bin/logger -t mwtag 'killing dhcp6 on {$realif}'");  // Debug info only
+			kill_dhcp6client_process($realif);
+			
 			sleep(3);
 			unlink_if_exists("{$g['varetc_path']}/dhcp6c_{$interface}.conf");
 			unlink_if_exists("{$g['varetc_path']}/dhcp6c_{$interface}_script.sh");
@@ -3035,11 +3035,21 @@ function kill_dhcp6client_process($interface) {
  	}
 
  	$i = 0;
+ 	
+ 	$pid = find_dhcp6c_process($interface);
+	
+	if($pid > 0){
+		mwexec("/usr/bin/logger -t mwtag 'found dhcp6 process {$pid}'"); // debug info only
+	}
+	else {
+		mwexec("/usr/bin/logger -t mwtag 'no dhcp6c process to kill'"); // debug info only
+		unset ($i);
+		return;
+	}
+	
  	while ((($pid = find_dhcp6c_process($interface)) != 0) && ($i < 3)) {
- 		/* 3rd time make it die for sure */
- 		$sig = ($i == 2 ? SIGKILL : SIGTERM);
- 		posix_kill($pid, $sig);
- 		sleep(1);
+ 		mwexec("kill -9 $pid");
+ 		sleep(2);
  		$i++;
  	}
  	unset($i);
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3035,11 +3035,21 @@ function kill_dhcp6client_process($interface) {
  	}
 
  	$i = 0;
+ 	
+ 	$pid = find_dhcp6c_process($interface);
+	
+	if($pid > 0){
+		mwexec("/usr/bin/logger -t mwtag 'found dhcp6 process {$pid}'"); // debug info only
+	}
```
rbgarga: Please fix style, else should be on the same line of  }

src/etc/inc/interfaces.inc
```diff
@@ -3035,11 +3035,21 @@ function kill_dhcp6client_process($interface) {
...
+		unset ($i);
+		return;
+	}
+	
  	while ((($pid = find_dhcp6c_process($interface)) != 0) && ($i < 3)) {
- 		/* 3rd time make it die for sure */
- 		$sig = ($i == 2 ? SIGKILL : SIGTERM);
- 		posix_kill($pid, $sig);
- 		sleep(1);
+ 		mwexec("kill -9 $pid");
```
rbgarga: What is the reason to kill it with -9 in the first attempt?

src/etc/inc/interfaces.inc
```diff
@@ -3991,10 +3996,10 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$rtsoldscript .= "\t/bin/sleep 1\n";
 	}
 	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
-	if (!isset($wancfg['dhcp6withoutra'])) {
+	if (!isset($wancfg['dhcp6withoutra'])){
 		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption} -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
 		$rtsoldscript .= "/usr/bin/logger -t rtsold \"Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
-	}
+        }
```
rbgarga: Please do not break style, it should use tab here, as it was before

src/etc/inc/interfaces.inc
```diff
@@ -4003,28 +4008,24 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
...
 	/* fire up rtsold for IPv6 RAs first, this backgrounds immediately. It will call dhcp6c */
 	if (isvalidpid("{$g['varrun_path']}/rtsold_{$wanif}.pid")) {
 		killbypid("{$g['varrun_path']}/rtsold_{$wanif}.pid");
 		sleep(2);
 	}
 	if (isset($wancfg['dhcp6withoutra'])) {
-		kill_dhcp6client_process($wanif);
-
+		kill_dhcp6client_process($wanif);	
+
```
rbgarga: Check style, it's adding extra spaces

src/etc/inc/interfaces.inc
```diff
@@ -4003,28 +4008,24 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
...
 	if (isset($wancfg['dhcp6withoutra'])) {
-		kill_dhcp6client_process($wanif);
-
+		kill_dhcp6client_process($wanif);	
+		
 		mwexec("/usr/local/sbin/dhcp6c {$debugOption} -c {$g['varetc_path']}/dhcp6c_wan.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}");
- 		mwexec("/usr/bin/logger -t mwtag 'Starting dhcp6 client for interface wan({$wanif} in IPoE mode)'");
-	}
+ 		mwexec("/usr/bin/logger -t mwtag 'Starting dhcp6 client for interface wan({$wanif} in IPoE mode)'"); 
+    }
```
rbgarga: style

src/etc/inc/interfaces.inc
marjohn56: This is a really strange problem. For whatever reason the posix_kill... function did not always kill the dhcp6c process. I put logging in there to prove that the function was being called and it was. Changing posix_kill to mwexec("kill -9 $pid"); worked every time.

rbgarga: Even when $sig == SIGKILL it was not working?

marjohn56: Not 100% of the time no, I've added debug code in there and watched the function called and have a shell open so I can do a before and after ps -auxw | grep dhcpc6 and the dhcp6c process is still there. Also have someone else testing the patches and he sees the same thing. kill -9 works every time though.

I did say it was weird! :)

marjohn56: Just to add, the process ID was the same, so it was not another process being called. Although thinking on that, maybe a second dhcp6c was launched by rtsold ( that's what the other bits of this patch fix ) and that was killed instead of the first process. I'll hammer it again this evening and see if I can definitely prove it one way or the other,

marjohn56: OK, tried this several times now and this does not kill the dhcp6c process when a release is done on the WAN interface.

 while ((($pid = find_dhcp6c_process($interface)) != 0) && ($i < 3)) {
 		/* 3rd time make it die for sure */
 		$sig = ($i == 2 ? SIGKILL : SIGTERM);
 		posix_kill($pid, $sig);
 		sleep(1);
	}

This does, every time.
mwexec("kill -9 $pid");
sleep(1);

Anyone else like to try it?

PR #3088, opened 2016-08-02
Save widget settings per user

phil-davis:
For users that have "custom settings" enabled, save the "tool" settings
of their widgets on a per-user basis.
User that do not have "custom settings" enabled will continue to use and
save widget settings to/from the system-wide settings.


phil-davis:
Note: Redmine feature 6674 is for this.

sbeaver-netgate:
Testing this now

PR #3089, opened 2016-08-03
Update firewall rules separators when NAT associated rule is deleted.

NOYB:
Bug: https://redmine.pfsense.org/issues/6676


sbeaver-netgate:
Merged. Thank you

PR #3090, opened 2016-08-07
Style corrected DHCP6c before RA additions and corections

marjohn56:
Correct possible extra dhcp6c launch by rtsold script when dhcpwithoutra is set.

Replaced posix_kill call in kill_dhcp6client_process() with mwexec("kill -9 $pid"); as the posix_kill function is randomly failing. The mwexec call  kills the client every time.

HOPEFULLY,,, style is correct now too!


marjohn56:
Sorry, 2nd commit was a typo, called the wrong kill_* process.

PR #3091, opened 2016-08-07
DHCP6C without RA - Corrections and additions

marjohn56:
Correction to the RTSOLD script creation as the existing code was allowing dhcp6c to be launched on receipt of RA when it was in dhcpwithoutra mode. This results in multiple dhcp6c clients,

kill_dhcp6client_process() now uses mwexec("kill -9 $pid"); rather than posix_kill() as the later is not reliable, the former kills the process every time.

In interfaces_bring_down(*) the pid find and posix_kill is replaced with the call to kill_dhcp6client_process()

Style is now - hopefiully  - correct!

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1300,11 +1300,7 @@ function interface_bring_down($interface = "wan", $destroy = false, $ifacecfg = 
 	switch ($ifcfg['ipaddrv6']) {
 		case "slaac":
 		case "dhcp6":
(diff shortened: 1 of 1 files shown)


NewEraCracker:
The ```$i``` variable on ```kill_dhcp6client_process($interface)``` function is no longer used and can be removed along the ```unset($i)``` calls.

Regarding style, you added certain lines with only tabs on them, I'd suggest a find ```\t\n``` and replace that with ```\n``` only.

PR #3092, opened 2016-08-07
DHCP6 Before RA. Additions and ammendments

marjohn56:
Replaced posix_kill() in kill_dhcp6client_process() with mwexec("kill -9 $pid"), this is because the posix_kill call was not reliably killing the dhcp6c process, kill -9 works every time.

Changes to the rtsold script creation. The script lines starting dhcp6c should not have be written to the script when dhcpwithoutra is true.

Style corrections, I spend more time doing these than changing code!


NewEraCracker:
Needs conflicts to be resolved. ```Recieved``` has been replaced by ```Recieved``` by PR #3094

rbgarga:
Closing in favor of #3102

PR #3093, opened 2016-08-08
Trigger for external script when gateway status changes

deajan:
Adds a way for users to execute trivial code when a gateway is added or removed from a LB group.
Needed to make squid use failover & loadbalancing toghether, see https://forum.pfsense.org/index.php?topic=109000
I've asked if such a trigger exists before here: https://forum.pfsense.org/index.php?topic=112619.0

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -923,6 +923,7 @@ function return_gateway_groups_array() {
 						$gwdown = true;
 					}
 					if ($gwdown == true) {
+						if (file_exists('/usr/local/bin/gateway_status.sh')) { mwexec("/usr/local/bin/gateway_status.sh $gwname down"); }
                                         	if (!file_exists("/tmp/.down.$gwname")) {
                                                 	$msg .= "\n".implode("|", $status);
 							touch("/tmp/.down.$gwname");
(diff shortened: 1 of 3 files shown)


rbgarga:
If users manually insert code inside gateway_status.sh, what in my opinion is a bad idea, changes will be lost during the first upgrade it runs, since file will be overwritten.

deajan:
Renaming this file to a template solves this issue I think.

rbgarga:
It would still have issues since custom scripts will not be part of backup/restore.

Other packages also could benefit of such mechanism that let it know when a gateway goes up/down, so we need a more generic implementation. I'll take a look at the code and try to think about a different approach

deajan:
Any news about this ?

loos-br:
It need to be redone, there is a plugin method used by packages that need to get CARP state events.

That is the way to go.

deajan:
@loos-br Hmm, I'm not used to pfSense source.
Care to tell me where to dig so I can redo the PR ? Thanks

loos-br:
Committed in:
https://github.com/pfsense/pfsense/commit/ca8ab3c958ace8fb42eb2f00cea7db74ae0bf707

Here is a plugin implementation example:
https://github.com/pfsense/FreeBSD-ports/blob/devel/net/pfSense-pkg-OpenBGPD/files/usr/local/pkg/openbgpd.inc#L409
https://github.com/pfsense/FreeBSD-ports/blob/devel/net/pfSense-pkg-OpenBGPD/files/usr/local/pkg/openbgpd.xml#L65

PR #3096, opened 2016-08-11
Fixed to properly save configs in PPPs interfaces

ch1c4um:
-Fix country, provider and providerplan configs were not being saved;
-Fix country, provider and providerplan configs were not being show in ppp edit page;
-Fix error in load edit page when try hideInterfaces() function when ppp type was selected;
-Fix conflict in provider input when pppoe type was selected;


jlduran:
Should we do the same in `interfaces.php`?

PR #3100, opened 2016-08-12
Fix links for IP WHOIS and IP Info

NewEraCracker:
Logic behind resolving IPs has changed and code wasn't entirely updated. This commit will pick the first A record as the IP to perform actions on.

If accepted, please backport to RELENG_2_3 as well.
Thanks!


rbgarga:
I've pushed a different fix. Thanks!

PR #3101, opened 2016-08-12
Uniformize memory limits and remove old code

NewEraCracker:
1) Increase memory_limit to 192M on i386 (edit: this change has now been reverted)
2) Allow setting a memory_limit up to 768M (Suhosin)
3) Remove old workarounds. Memory limits on config.inc will be new defaults

This will help in working around this issue and improve the code:
https://forum.pfsense.org/index.php?topic=116307.0

Developer of the package can add calls such as ```ini_set('memory_limit', '650M');``` on the affected pages, the limit will be accepted by Suhosin and if system has enough memory everything will play nice (people use this in servers with 8GB+ RAM so no problem).

Thanks!
Backport to RELENG_2_3 if accepted :)

### DIFF
--- a/src/etc/rc.php_ini_setup
+++ b/src/etc/rc.php_ini_setup
@@ -266,7 +266,7 @@ suhosin.post.max_value_length = 500000
 suhosin.request.max_array_index_length = 256
 suhosin.request.max_vars = 5000
 suhosin.request.max_value_length = 500000
-suhosin.memory_limit = 536870912
+suhosin.memory_limit = 805306368
 
 EOF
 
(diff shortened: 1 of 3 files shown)


phil-davis:
For the change to config.inc
1) I am surprised that there is still and if-then for amd64 vs i386. Because in master there is no i386 build. So in master the test could be taken out (or substitute some other case statement for all the upcoming architectures like ARM, if there is a desire to have different limits used for different architectures.
2) If this is back-ported to RELENG_2_3 then people with (almost end of life) Alix boards may be effected. Those have only 256MB total hard-soldered memory. So they are stuck with some need for smaller memory limits.

NewEraCracker:
Thanks for your input phil. I have removed the config.inc change from this PR.

PR #3102, opened 2016-08-13
DHCP6 Before RA. Additions and ammendments (another version)

NewEraCracker:
An alternative implementation of PR #3092

> Replaced posix_kill() in kill_dhcp6client_process() with mwexec("kill -9 $pid"), this is because the posix_kill call was not reliably killing the dhcp6c process, kill -9 works every time.
> 
> Changes to the rtsold script creation. The script lines starting dhcp6c should not have be written to the script when dhcpwithoutra is true.
> 
> Style corrections, I spend more time doing these than changing code!

I have done some changes to style, fixed the conflict (the received typo, see #3094), and changed ```kill_dhcp6client_process``` to look more like it's v4 counterpart.

Credits are kept as @marjohn56 had the original idea.

Thanks!
Please backport to RELENG_2_3 if accepted.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3974,7 +3974,8 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$rtsoldscript .= "\t/bin/sleep 1\n";
 		$rtsoldscript .= "fi\n";
 	} else {
-		$rtsoldscript .= "\t/bin/sleep 1\n";
+		$rtsoldscript .= "{$g['varetc_path']}/dhcp6c_{$interface}_script.sh\n";
+		$rtsoldscript .= "/bin/sleep 1\n";
 	}
 	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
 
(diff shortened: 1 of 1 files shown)


NewEraCracker:
Tested this on a small IPv6 network based on virtual machines on VMware 12.1.1 with IPv6 enabled and works as expected.

NewEraCracker:
Closing this PR

After a lot discussion, this surely needs much more research and testing before being considered a good solution for the problem at hand.

Research will continue. There might be an easier solution lying around that hasn't been spotted yet.

NewEraCracker:
Reopening this. I am open to suggestions and improvements before considering this good to go.

Good news are, while researching on this I've found #3103 and this will surely fix the concurrency issues with quick link down/up transitions and the behavior of the dhcpd counterparts. That PR is safe and I am sure will help a lot of people affected by process double launch issues.

Now off to have some sleep! :)

marjohn56:
I'll try and run this up over the next twenty four, Nice work.

marjohn56:
Applied this along with #3103 to the latest snapshot.... looking good.

PR #3103, opened 2016-08-14
Ensure only one instance of services_dhcpd_configure runs concurrently

NewEraCracker:
Sometimes when interfaces are disconnected and reconnected too fast, you may get multiple ```dhcpd -6``` and/or ```dhcpleases6``` running because the whole kill/respawn reaches concurrency conditions.

By securing ```services_dhcpd_configure()``` it ensures only one instance runs concurrently, thereby controlling the flow of the scripts calling it.

Thanks!
Please backport to RELENG_2_3 if accepted.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -395,6 +395,8 @@ function services_radvd_configure($blacklist = array()) {
 function services_dhcpd_configure($family = "all", $blacklist = array()) {
 	global $config, $g;
 
+	$dhcpdconfigurelck = lock("dhcpdconfigure", LOCK_EX);
+
 	/* configure DHCPD chroot once */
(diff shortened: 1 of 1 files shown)


NewEraCracker:
Further details:

This issue is also reproducible in both 2.3.3-dev (most likely other lower versions as well) and 2.4.0-dev without any other changes than running the latest development snapshot (except that in 2.4.0-dev I had to [fix PEAR Net/IPv6.php](https://forum.pfsense.org/index.php?topic=117601.0))

For it to happen, both IPv4 and IPv6 need to be enabled on WAN (via DHCP clients) and LAN (static + DHCP servers). If only IPv4 configured and used, this issue does not happen.

Regards,
Jorge M. Oliveira

PR #3104, opened 2016-08-14
dnsresolver, make interface boxes resizable, to allow for easier picking/checking of multiple selected interfaces

PiBa-NL:
dnsresolver, make interface boxes resizable, to allow for easier picking/checking of multiple selected interfaces

### DIFF
--- a/src/usr/local/www/css/pfSense.css
+++ b/src/usr/local/www/css/pfSense.css
@@ -231,6 +231,12 @@ tr.disabled th {
     padding: inherit;
 }
 
+/* location of bottom right drag icon on resizable elements */
+.ui-resizable-se {
+	bottom: 2px;
+	right: 46px;
+}
+
 #login .form-control {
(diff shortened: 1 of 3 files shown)


jdillard:
@PiBa-NL, you don't need to edit `src/usr/local/www/head.inc`, you can add:

```
$(".resizable").each(function() {
	$(this).css('height', 80).resizable({minHeight: 80, minWidth: 200}).parent().css('padding-bottom', 0);
	$(this).css('height', 78);
});
```
to `src/usr/local/www/js/pfSense.js` underneath the following code (~L#234) instead:

```
// Focus first input
//$(':input:enabled:visible:first').focus();
```
otherwise, I don't see any issues.

PiBa-NL:
Ok thanks, moved it to pfSense.js .

jdillard:
Thanks. @rbgarga this looks good now.

PR #3112, opened 2016-08-16
Fix #6720 DHCP options by pool

phil-davis:
It is a little bit tricky having to generate the unique "option custom-if-n-m code ..." lines at first where n = pool index and m = item index in the items of the pool. Then make sure to reference that later, getting the same pool index into the array of pools. The $all_pools array as the "overall" or "base" pool first (at index 0), followed by the user-specified pools at index 1, 2, 3,... - which are actually at indexes 0, 1, 2,... in the ordinary array of pools in the config. So the -1 at line 910 has to happen.
But it works for me.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -911,9 +911,9 @@ EOPP;
 				foreach ($poolconf['numberoptions']['item'] as $itemidx => $item) {
 					$item_value = base64_decode($item['value']);
 					if (empty($item['type']) || $item['type'] == "text") {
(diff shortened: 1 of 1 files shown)


phil-davis:
In Redmine bug I have asked the OP to try this and report back. In my testing it produces dhcpd.conf files that look valid to me, but it would be nice to know that the clients then do actually get the specified options by pool for real.

phil-davis:
See bug report. OP has reported it is working for him. I added an extra indent to the "option custom..." line to make it line up nicely.

PR #3119, opened 2016-08-22
css: Fix jQuery UI widgets' font

jlduran:
Use the main font with jQuery UI widgets (e.g. autocomplete forms)


jdillard:
This looks good @rbgarga

PR #3123, opened 2016-08-25
Fix invalid regular expression error on Chrome Beta channel

NewEraCracker:
https://forum.pfsense.org/index.php?topic=117007.msg650206

### DIFF
--- a/src/usr/local/www/firewall_nat_out_edit.php
+++ b/src/usr/local/www/firewall_nat_out_edit.php
@@ -484,7 +484,7 @@ $group->add(new Form_IpAddress(
 	'source',
 	null,
 	$pconfig['source']
-))->addMask('source_subnet', $pconfig['source_subnet'])->setHelp('Source network for the outbound NAT mapping.')->setPattern('[a-zA-Z0-9\_\.\:]+');
+))->addMask('source_subnet', $pconfig['source_subnet'])->setHelp('Source network for the outbound NAT mapping.')->setPattern('[a-zA-Z0-9_.:]+');
 
 $group->add(new Form_Input(
 	'sourceport',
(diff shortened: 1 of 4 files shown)


NOYB:
Why does Chrome think escaping characters in a regular expression is invalid?  It should be perfectly legit.  If Chrome does not accept that then I think it's Chrome that is broken.

rbgarga:
This change is wrong. If you don't escape '.' on a regular expression it means 'any character' so it will never fail

NOYB:
That was one of my thoughts too but didn’t point it out because not sure if ‘.’ is ‘any character’ when part of a range, and too lazy to look it up.  If I recall some special meanings are a little different inside of a range.  But rather than committing them all to memory I like to just escape them all and be done with it.

NewEraCracker:
> This change is wrong. If you don't escape '.' on a regular expression it means 'any character' so it will never fail

> That was one of my thoughts too but didn’t point it out because not sure if ‘.’ is ‘any character’ when part of a range, and too lazy to look it up.

Checking the expression online gives:
```
    [a-zA-Z0-9_.:]+ match a single character present in the list below
        Quantifier: + Between one and unlimited times, as many times as possible, giving back as needed [greedy]
        a-z a single character in the range between a and z (case sensitive)
        A-Z a single character in the range between A and Z (case sensitive)
        0-9 a single character in the range between 0 and 9
        _.: a single character in the list _.: literally
```
Source: https://regex101.com/

NOYB:
The existing escaped expression is valid.  Anything that doesn’t deal with it is what is broken and should be fixed.  Including Chrome.  Especially if it is beta version.

`
[a-zA-Z0-9\_\.\:]+ match a single character present in the list below
Quantifier: + Between one and unlimited times, as many times as possible, giving back as needed [greedy]
    a-z a single character in the range between a and z (case sensitive)
    A-Z a single character in the range between A and Z (case sensitive)
    0-9 a single character in the range between 0 and 9
    \_ matches the character _ literally
    \. matches the character . literally
    \: matches the character : literally
`

NewEraCracker:
This bug has made its way into Google Chrome 53 Stable (53.0.2785.89)
This workaround is the only way to make pages work.

NOYB:
"This workaround is the only way to make pages work."
There are other ways to make the pages work.  Use a browser that adheres to the regex expression spec.

NewEraCracker:
Closing this in favor of #3127 which contains further fixes for same issue in other areas.

PR #3124, opened 2016-08-27
Support for special cases of TLD for specific countrys, like Brazil.

ancientz:
Small code change to support TLDs from Brazil which have two levels, like .com.br, and in the current code does not work. It should work for countries with even more levels, but haven't tried it.

Didn't have time to make some better working code, but It works and suit me immediate need, if anyone want to make a better code, be my guest :)

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -704,7 +704,22 @@
 					$dnsServer ='api.cloudflare.com';
 					$dnsHost = str_replace(' ', '', $this->_dnsHost);
 					$host_names = explode(".", $dnsHost);
-					$bottom_host_name = $host_names[count($host_names)-2] . "." . $host_names[count($host_names)-1];
+					$bottom_host_name = null;
+					$specific_country_tld = array(
+						"com.br",
+						"net.br"
+					);
+					foreach ($specific_country_tld as $countrytld) {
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

rbgarga:
@ancientz could you please sign the CLA as described at https://www.pfsense.org/about-pfsense/#cla ? Without it this patch cannot be reviewed / merged

ancientz:
@rbgarga Hi, sorry it took too long, just did the online CCLA form.

rbgarga:
The idea is good, but I believe there should be a better way to implement it including all other TLDs supported by IANA. Adding just com.br and net.br hard coded is not the way to go IMO

jim-p:
Trying to detect/parse out the hostname and TLD or subdomain parts is too error prone for country-specific TLDs and subdomains. The Namecheap code path already has code for separate domain and hostname fields. If another DynDNS provider needs them separated and has similar parsing issues, it should be converted to that style instead.

ancientz:
@rbgarga I agree, but I could not think of a simpler way to do this, considering that there might be (I'm not sure) other countries with TLDs based with two subdomains, something like .net.sc.br, and so on. I think @jim-p alternative is the best, that way solve all the problems and also make the code cleaner than It is right now.

rbgarga:
Agreed, so I'm closing this one and we can have specific PRs as needed. Thanks!

PR #3131, opened 2016-09-06
XMLRPC, generic xmlrpc_client implementation + bugfixes in voucher sync

PiBa-NL:
XMLRPC, generic xmlrpc_client implementation + bugfixes in voucher sync

### DIFF
--- a/src/etc/inc/xmlrpc_client.inc
+++ b/src/etc/inc/xmlrpc_client.inc
@@ -21,11 +21,32 @@
 
 require_once("XML/RPC2/Client.php");
 
-class pfsense_xmlrpc_client_extended {
+class pfsense_xmlrpc_client {
 	
 	private $username, $password, $url, $filenotice, $error;
 	
-	public function __construct($syncip, $port, $username, $password) {
+	public function __construct() {
+		global $config;
+		$hasync = $config['hasync'];
+
+		if (empty($hasync['username'])) {
+			$username = "admin";
+		} else {
+			$username = $hasync['username'];
+		}
+		/* if port is empty lets rely on the protocol selection */
+		$port = $config['system']['webgui']['port'];
+		if (empty($port)) {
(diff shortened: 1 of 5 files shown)


rbgarga:
@PiBa-NL code looks good. Since xmlrpc_client.inc already existed in src tree, could you please remove it from https://github.com/pfsense/pfsense/blob/master/src/etc/pfSense.obsoletedfiles#L38

src/etc/inc/xmlrpc_client.inc
```diff
@@ -0,0 +1,141 @@
...
+	function xmlrpc_method($method, $parameter = "", $timeout = 240) {
+		$resp = $this->xmlrpc_internal($method, $parameter, $timeout);
+		return $resp;
+	}
+	function get_error() {
+		return $this->error;
+	}
+}
+
+class pfsense_xmlrpc_client extends pfsense_xmlrpc_client_extended {
```
rbgarga: I was thinking about the way you did it, with 2 classes to be able to set connection data or use system default and something came to mind. What do you think about having a single class like this:
```
public function construct ($data = array()) {
    if (empty($data['username'])) {
        $this->username = $hasync['username'];
    } else {
        $this->username = $data['username'];
    }
    // ... Same for hostname, port, password
}
```

src/etc/inc/xmlrpc_client.inc
```diff
@@ -0,0 +1,137 @@
+<?php
+/*
+ * xmlrpc_client.php
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2016 Electric Sheep Fencing, LLC
```
rbgarga: Can you please just replace this line by the following to reflect recent changes? Then I'm ready to merge
```
 * Copyright (c) 2016 Rubicon Communications, LLC (Netgate)
```

PR #3133, opened 2016-09-10
Check IP Services - Info Box

NOYB:
Change warning box with dismissal to collapsible info box.

### DIFF
--- a/src/usr/local/www/services_checkip.php
+++ b/src/usr/local/www/services_checkip.php
@@ -157,10 +157,11 @@ endforeach; ?>
 </nav>
 
 <div class="infoblock">
-	<?php print_info_box(gettext('' . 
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_checkip.php
```diff
@@ -156,9 +156,11 @@
...
 </nav>
 
-<?php
-print_info_box(gettext('The first (highest in list) enabled check ip service will be used to ' . 
-						'check IP addresses for Dynamic DNS services, and ' .
-						'RFC 2136 entries that have the "Use public IP" option enabled.'));
-
-include("foot.inc");
+<div class="infoblock">
+	<?php print_info_box(gettext('' .
```
rbgarga: This empty string concatenation is not necessary ('' .)

PR #3134, opened 2016-09-10
Redundant logic

stilez:
1) if it's a numeric integer (hence non-empty [0-9]+ ) and the first char isn't "0" then the value will always be >0, so test is redundant

2) logic for gateway type is part tested in previous error check, no need to duplicate, clearer this way

3) minor standardising of error message texts (some referred to an "IPv4/6 Address Family rule" and others to an "IPv4/6 Filter rule", some refer to "a rule that applies to IPv4 and IPv6" while others refer to "combined IPv4 and IPv6 rules")

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -311,20 +311,20 @@ if ($_POST) {
 						continue;
 					}
 					if (($_POST['ipprotocol'] == "inet46") && ($_POST['ipprotocol'] != $family)) {
-						$input_errors[] = gettext("A gateway can not be assigned to a rule that applies to IPv4 and IPv6");
+						$input_errors[] = gettext("A gateway can not be assigned to a rule that applies to both IPv4 and IPv6");
 					}
 					if (($_POST['ipprotocol'] == "inet6") && ($_POST['ipprotocol'] != $family)) {
-						$input_errors[] = gettext("An IPv4 gateway group can not be assigned on an IPv6 Address Family rule");
+						$input_errors[] = gettext("An IPv4 gateway group can not be assigned to an IPv6 Filter rule");
 					}
 					if (($_POST['ipprotocol'] == "inet") && ($_POST['ipprotocol'] != $family)) {
-						$input_errors[] = gettext("An IPv6 gateway group can not be assigned on an IPv4 Address Family rule");
+						$input_errors[] = gettext("An IPv6 gateway group can not be assigned to an IPv4 Filter rule");
 					}
 				}
 			}
 		}
 		if (is_ipaddr(lookup_gateway_ip_by_name($_POST['gateway']))) {
 			if (($_POST['ipprotocol'] == "inet46") && ($_POST['gateway'] <> "")) {
-				$input_errors[] = gettext("A gateway can not be assigned to a rule that applies to IPv4 and IPv6");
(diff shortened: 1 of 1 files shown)


src/usr/local/www/firewall_rules_edit.php
```diff
@@ -42,8 +42,8 @@
 }
 
 function is_posnumericint($arg) {
-	// Note that to be safe we do not allow any leading zero - "01", "007"
-	return (is_numericint($arg) && $arg[0] != '0' && $arg > 0);
+	// Integer > 0? (Note that to be safe we do not allow any leading zero - "01", "007")
+	return (is_numericint($arg) && $arg[0] != '0');
```
NOYB: The function is to return true only for positive numeric integer that does not have any leading zeros.
What if $arg is negative?

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -311,27 +311,27 @@ function is_aoadv_used($rule_config) {
...
+			if (($_POST['ipprotocol'] == "inet46") && ($_POST['gateway'] <> "")) {
+				$input_errors[] = gettext("Gateways can not be assigned in a rule that applies to both IPv4 and IPv6.");
+			}
+			if (($_POST['ipprotocol'] == "inet6") && (!is_ipaddrv6(lookup_gateway_ip_by_name($_POST['gateway'])))) {
+				$input_errors[] = gettext("An IPv4 gateway can not be assigned in IPv6 rules.");
+			}
+			if (($_POST['ipprotocol'] == "inet") && (!is_ipaddrv4(lookup_gateway_ip_by_name($_POST['gateway'])))) {
+				$input_errors[] = gettext("An IPv6 gateway can not be assigned in IPv4 rules.");
+			}
 		}
```
NOYB: In the original the ``&& ($_POST['gateway'] <> "") `` test is applied to all three of the enclosed ``if`` statements.
So unless ``(!is_ipaddrv6(lookup_gateway_ip_by_name($_POST['gateway']))`` equates to false for an empty gateway by name, this may not be equivalent.

src/usr/local/www/firewall_rules_edit.php
stilez: is_numericint() only passes integers >= 0 or non-empty strings containing 0-9 only. So there's no need to test again for negatives, and since empty strings are excluded, if the first character isn't zero, the value can't be zero either.

src/usr/local/www/firewall_rules_edit.php
stilez: If $_POST['gateway'] doesn't exist, or is blank or not a valid gateway for any reason, then lookup_gateway_ip_by_name() isn't going to be able to return a valid IP address for it, and looks almost certainly like it returns false. Additionally all the is_ipaddr*() functions return a falsey value if the string isn't a valid IP (of the given type if specified). So any case where is_ipaddr*() allowed a TRUE value for the if() clause, $_POST['gateway'] would have had to be non-empty. I think that clause is really just a "belt and braces" safety check that the value passed to the other functions exists and isn't empty. But it doesn't matter because the rest does in fact look like it will handle that sort of thing correctly if not. 

That said it can be made more efficient. Let's do so.
is defined in gwlb.inc (about line 1037), and is_ipaddr*() are defined in util.inc.

PR #3135, opened 2016-09-12
Fix #6768 IPv6 static mapping on delegated prefixes

phil-davis:
For example, WAN receives a /48 delegated from the upstream (ISP...),
e.g. "2001:470:abcd::" pfSense then uses this as a starting point to
calculate the addresses on LAN, OPT1, OPT2 etc where they have been
specified as "track interface WAN".
Actually each local interface gets just a /64 taken out of the /48,
using the chunk specified by "IPv6 Prefix Id" for that local interface.
e.g. if "IPv6 Prefix Id" is set to "a123" on LAN, then the LAN would be:
2001:470:abcd:a123::/64

Then when we specify a static-mapped address in LAN, or other things
that live in LAN, e.g. "::4242" we mean 4242 on from the base LAN
address, so "2001:470:abcd:a123::4242"

i.e. we always have a CIDR of 64 when calculating this stuff. We do not
want the logic that was in this code that was using the upstream prefix
delegation size (like /48).

Note: The code in services.inc "worked" because var $ifname was not set,
and so $trackifname was blank, $trackcfg was blank, and so the attempted
calculation of $pdlen always came out as 64 anyway. That tricked me for
a while trying to understand why the original code in services.inc worked.
system.inc did not work, because it actually calculated $pdlen and got a
number like 48 - which we do not want here.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1346,12 +1346,11 @@ EOD;
 		}
 		$ifcfgsnv6 = get_interface_subnetv6($dhcpv6if);
 		$subnetv6 = gen_subnetv6($ifcfgipv6, $ifcfgsnv6);
-
-		if ($ifcfgv6['ipaddrv6'] == 'track6') {
-			$trackifname = $config['interfaces'][$ifname]['track6-interface'];
-			$trackcfg = $config['interfaces'][$trackifname];
-			$pdlen = 64 - $trackcfg['dhcp6-ia-pd-len'];
-		}
+		// We might have some prefix-delegation on WAN (e.g. /48),
+		// but then it is split and given out to individual interfaces
+		// (LAN, OPT1, OPT2...) as multiple /64 subnets. So the size
+		// of each subnet here is always /64.
+		$pdlen = 64;
 
 		if ($is_olsr_enabled == true) {
 			if ($dhcpv6ifconf['netmask']) {
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -393,10 +393,8 @@ function system_hosts_generate() {
 						$ipaddrv6 = $host['ipaddrv6'];
 						if ($ipaddrv6 && $host['hostname']) {
 							if ($isdelegated) {
-								$trackifname = $config['interfaces'][$dhcpif]['track6-interface'];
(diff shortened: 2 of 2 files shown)


NewEraCracker:
Hi,

Can't one just configure WAN as a /64 and multiple LAN/OPT interfaces to track6 from WAN as long as track6_prefix_id differs on each of them and ISP allows it?

(Just saying, never tested this at all)

phil-davis:
As I understand what you are saying, that is the normal case. WAN will get a /64 from the ISP for use as the WAN subnet, and then also receive a delegated range (/48 or whatever) from the upstream ISP.
The firewall/router (in this case pfSense) can then allocate /64 subnets to local LANs out of the /48 - track6_prefix_id is the way to organise which /64 of the /48 is given to which local LAN.
Then the firewall/router can also delegate chunks of the /48 down to other routers that are downstream of it, so you might give out some /56 (for example) to routers inside you network, which can then split that up as they need. pfSense has the functionality to specify all that (I just played with it to test this change - getting a /48 from he.net then delegating a /56 of it down to a test VM running pfSense and letting that test VM allocate /64s of that to its local LANs)

virgiliomi:
Thanks for the fix... however I don't think it's right to assume that the prefix length will always be /64. DHCPv6 can be used on other prefix sizes and will still work properly... only SLAAC won't work properly with anything but /64. 

If Track Interface is being used then it's safe to assume that it's a /64 since pfSense is configuring that detail. But if the interface has a different IPv6 configuration, it could be using a different prefix length and can still use DHCPv6 to assign addresses. In that case, assuming a prefix length of /64 is not correct. 

I know we all want to assume that everyone is going to use a /64 for their various internal networks, but it's just not always going to be the case, for whatever reason.

phil-davis:
The code that is using the magic "64" here is only for the case when the interface is set to "Track Interface". So it will always be a /64 in these cases.
For other "ordinary" LANs, as you say, the user specifies the CIDR, and when they add static mappings and the like, they specify the whole IPv6 address. So we do not have to merge a network part with a host part of the address.

virgiliomi:
Ok, understood now. That's why you're working on the code and I'm not. I see that now in system.inc, but the change clip for services.inc didn't give me enough context, and I didn't go digging through the whole file.

Thanks for clearing that up for me!

PR #3137, opened 2016-09-14
Secure SMTP Connection Modes - Mutually Exclusive

NOYB:
Secure SMTP Connection modes are mutually exclusive.  Select only one.


src/usr/local/www/system_advanced_notifications.php
```diff
@@ -156,6 +156,11 @@
 			unset($config['notifications']['smtp']['tls']);
 		}
 
+		if ((isset($config['notifications']['smtp']['ssl'])) && 
+			(isset($config['notifications']['smtp']['tls']))) {
```
rbgarga: 2nd line indent should be 4 spaces instead of 1 tab

PR #3138, opened 2016-09-15
Allow snmpd to bind to multiple interfaces.

skrude61:
This patch makes it possible to bind snmpd to multiple interfaces other than ALL (for example to LAN and LAN CARP VIP). Binding to all interfaces is a security issue if you don't want to expose snmp to WAN.


rbgarga:
RELENG_2_3_2 is a branch where we only push bug fixes and security fixes. This change is a new feature, please re-submit it to branch master

PR #3139, opened 2016-09-15
Enhance rules to allow (1) multiple ICMP types, (2) ICMP types on IPv4+6 rules. Transparent change.

stilez:
1. Allow ICMP types on IPv4+IPv6 rules, for those subtypes valid on both IPv4+6;
2. Allow multiple ICMP types to be selected
3. Clean up handling (eg remove need for 2 separate input selects for IPv4 and IPv6)
4. Completely backwards compatible (doesn't break $config format)

(Very minor other edit in passing - I moved an "any" to the top of a list, it's minor but one shouldn't have to search for the "any" option.)

The PR is transparent and shouldn't have compatibility issues. It's 100% compatible with current $configs because a multiple icmptype selection is held as an imploded() comma-separated list (eg "echoreq,echorep") so a single-value icmptype will be correctly handled.

I've tested the changes in both GUI and rules generation. The new code appears to work correctly as far as I can tell.  For example, with a multiple ICMPtype rule on multiple IP protocols, such as (IPv4+IPv6) with types "echorep,echoreq":
- rules.debug shows the rule correctly on two lines, one for IPv4 with "icmp-type { echoreq,echorep }" and a separate line for the ICMP6 version "icmp6-type { echoreq,echorep }".
- pfctl -sr shows this expanded to four lines, again correctly, one for each type, on each protocol.
- firewall_rules.php shows the corect title for this rule when hovered.

A major motive is that it's plain annoying to have to create "double rules" for IPv4 and IPv6 for common ICMP messages such as echoreq when it's not needed for many other kinds of rule - and it's more annoying when it's a rule on multiple icmptypes **and** on both IPv4+6!

**Code notes** - 
1. The "any" option is removed from the icmptypes list in filter.inc, because it's not an icmptype; it's instead handled in the GUI code. As usual and as before, "any icmptype" is represented in $config['filter']['rule'] by icmptype being unset.
2. There aren't separate IPv4 and IPv6 icmptype tables in filter.inc, or separate IPv4 and IPv6 GUI select boxes in firewall_rules_edit.php; there's now one table and one box. 
3. GUI JS code on the dropdown box - selecting "any" deselects all others, deselecting all selects "any", and switching between IPv4/v6/v4+6 updates the listed selections and copies over any valid existing selections to the new list.

I can't find any other files affected by the change, after searching for icmptype/icmptypes in the repo. Also icmptype handling is hard to miss and remains mostly unaffected other than (transparently) holding a longer string for multiple options, and breaking it out for GUI editing/selection + pf.conf generation. So I'm cautiously hopeful that there aren't other consequential updates needed that I missed.

That said, as filter.inc is large and filters can be complex, it's worth checking that there aren't any processing paths within filter.inc or firewall_rules_edit.php that I haven't taken note of, and that gettext() and htmlspecialchars() are correctly handled (used where needed, omitted where not needed). But as above, I'm cautiously hopeful.

_(Caveat - merging my local file changes back into the master source should be correct but it's possible I missed some. It works nicely here and I think I copied the changes over correctly to github - fingers crossed!)_


rbgarga:
After apply this patch and go to Firewall -> Rules I got following error:
```
Catchable fatal error: Argument 4 passed to Form_Select::__construct() must be of the type array, null given, called in /usr/local/www/firewall_rules_edit.php on line 1323 and defined in /usr/local/www/classes/Form/Select.class.php on line 61 Call Stack: 0.0232 235280 1. {main}() /usr/local/www/firewall_rules_edit.php:0 0.0791 1743600 2. Form_Select->__construct() /usr/local/www/firewall_rules_edit.php:1323 
```

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1881,20 +1881,12 @@ function typesel_change() {
...
 	function proto_change() {
-		if ($('#proto').find(":selected").index() < 3) {
-			portsenabled = 1;
-			hideClass('tcpflags', false);
-		} else {
-			portsenabled = 0;
-			hideClass('tcpflags', true);
-		}
+		var is_tcpudp = (jQuery.inArray($('#proto :selected').val(), ['tcp','udp', 'tcp/udp']) != -1);
+		portsenabled = (is_tcpudp ? 1 : 0);
```
stilez: New line 1885 - I've left this value as it was (ie explicit 1 or 0) for ease of review. It's very likely that this line should just read "portsenabled = is_tcpudp;". if so, please make this change.

stilez:
@rbgarba - gettext() fixes all done. I cannot reproduce your issue with the dropdown not populating (clean install 2.3.2 on VM with the diffs in this PR merged via system patches and nothing else done).  Can you recheck?

rbgarga:
The bug I mentioned is still reproducible on 2.4 snapshots with this patch applied. I just followed the same instructions and it happened.

I figured out it happens on Safari and not on Firefox / Chrome

rbgarga:
@jdillard do you have idea about what could be the cause of the issue I reported? It only happens on Safari

rbgarga:
Thanks @jdillard!

@stilez can you merge Jared's pull request to your fork so I can run a final test and get it merged?

stilez:
Can merge, but I still want to move this to use checkboxes not dropdown lists for the UI - side by side it's a lot simpler and easier on the eye.  I have a near-complete version (not touched for a while), been insanely busy the last while but I'll try to sort it out over the Christmas break. I'll do Jared's fix (thanks @jdillard !) but can you not merge this for a little longer to get me the time needed to finish the UI update?

jdillard:
No problem @stilez. As far as the PR goes, you can always submit another with the UI improvements. It would probably be better that way to keep things clean and to prevent potential code drift.

rbgarga:
An extra review from @jim-p will be necessary since it's a big change

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1216,6 +1258,7 @@ function build_if_list() {
 	'Protocol',
 	$pconfig['proto'],
 	array(
+		'any' => gettext('any'),
```
jim-p: This change is unwarranted. The protocol default should remain TCP, it is set that way on purpose (otherwise people get confused by the lack of port options being visible). It's also not relevant to the other changes being made on this PR.

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1228,25 +1271,22 @@ function build_if_list() {
 		'pim' => 'PIM',
 		'ospf' => 'OSPF',
 		'sctp' => 'SCTP',
-		'any' => gettext('any'),
```
jim-p: See above.

stilez:
@rbgarga @jim-p - Ive finished the change mentioned above (switching selector UI from dropdowns -> checkboxes and jim-p's request which I addressed a different way). I prefer it a lot, much easier on the eye and simpler. I would upload it but there's a recent issue with the Github webUI I've reported to Github staff, that means they can't be edited online and I lack offline git-fu. Once they respond and/or fix it I'll upload it. If you want I can email the files outside github in the meantime, let me know an email address if so.

rbgarga:
@stilez I would prefer to merge this version as it is after you just revert the changes @jim-p asked and then you could send a new PR with the changes.

This is  a big patch and was already reviewed, it would make things easier to just review the new changes after these ones are in place

stilez:
@rbgarga - That's good sense, go ahead. 

I have 2 very minor (as in trivial 1-line) changes to make to the reviewed version. As github UI isn't working for me, can you briefly look at them and if acceptable, put them in somehow yourself, or at worst put it in as a PR to mine (merging PRs into mine seems to work without issues). On the flip side, if you want to merge the existing as-is and do them afterwards that's fine as well.

- A (small) CSS change that significantly improves the rendering and appearance of icmptype info in the filter-rules table. At about line 693 of filter_rules.php there's a line that `sprintf()`'s the icmp info.  Change to read: 

> `echo sprintf('<br /><div style="cursor:help;padding:1px;line-height:1.1em;max-height:2.5em;max-width:180px;overflow-y:auto;overflow-x:hidden" title="%s:%s%s"><small><u>%s</u></small></div>', gettext('ICMP subtypes'), chr(13), $t, str_replace(',', '</u>, <u>',$filterent['icmptype']));`

The effect is that if multiple icmptypes are selected the table can switch to overflow + v-scrollbar automatically if needed, rather than trying to output one long line of text, which works well. It also underlines each icmptype individually which is much easier on the eye.

- An alternate fix for jim-p's point. this only affects creating new rules and I take the point. I found this a much better fix (AFAICS) - basically leave "any" at the top of the list as that's the logical place people almost always look for it if they want it, but _set the default proto to tcp for new rules_ so that ports and other expected items are displayed by default too. All that's needed for this is to add a single line at about line 294 of firewall_rules_edit.php, to read as follows:

>  /* defaults */
>	if ($_GET['if']) {
>		$pconfig['interface'] = $_GET['if'];
>	}
>	$pconfig['type'] = "pass";
>	`$pconfig['proto'] = "tcp"; // ensures port fields visible and default=tcp, for a new rule`
>	$pconfig['src'] = "any";
>	$pconfig['dst'] = "any";

@jim-p can you try this as a fix, and see if it would be acceptable for resolving your point?

src/usr/local/www/firewall_rules_edit.php
stilez: Points taken - can you take a look at an alternate fix [(8afd8c0)](https://github.com/stilez/pfsense-stilez-working/commit/8afd8c0409a9ed246a244e7a1dbdcb2ad627434b) which seems to meet all requirements neatly, and let me know if it is acceptable for resolving your point?

src/usr/local/www/firewall_rules_edit.php
stilez: @jim-p - points taken. Can you take a look at a possible neat alternate fix [(8afd8c0)](https://github.com/stilez/pfsense-stilez-working/commit/8afd8c0409a9ed246a244e7a1dbdcb2ad627434b) which seems to meet all requirements neatly, and let me know if it is acceptable for resolving your point?

stilez: @jim-p - points taken. Can you take a look at a possible neat alternate fix [(8afd8c0)](https://github.com/stilez/pfsense-stilez-working/commit/8afd8c0409a9ed246a244e7a1dbdcb2ad627434b) which seems to meet all requirements neatly, and let me know if it is acceptable for resolving your point?

stilez: Points taken - can you take a look at an alternate fix though [(8afd8c0)](https://github.com/stilez/pfsense-stilez-working/commit/8afd8c0409a9ed246a244e7a1dbdcb2ad627434b) which seems to meet all requirements neatly, and let me know if it is acceptable for resolving your point?

stilez: Points taken - can you take a look at an alternate fix though [(8afd8c0)](https://github.com/stilez/pfsense-stilez-working/commit/8afd8c0409a9ed246a244e7a1dbdcb2ad627434b) which seems to meet all requirements neatly, and let me know if it is acceptable for resolving your point?

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -295,6 +317,10 @@ function is_aoadv_used($rule_config) {
 
 	unset($input_errors);
 
+	if (!array_key_exists($_POST['ipprotocol'], $icmplookup)) {
+		$input_errors[] = gettext("The IP protocol is not recognized."); 
+	}
+
```
rbgarga: There are some extraneous characters here that broke the page

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -296,6 +296,7 @@ function is_aoadv_used($rule_config) {
 		$pconfig['interface'] = $_GET['if'];
 	}
 	$pconfig['type'] = "pass";
+	$pconfig['proto'] = "tcp"; // ensures port fields visible and default=tcp, for a new rule
```
jim-p: It seems to work but it's still a change unrelated to this PR. That belongs in a separate PR.

src/usr/local/www/firewall_rules_edit.php
stilez: Thats due to github's web UI - as mentioned above it's had some bugs introduced that makes web editing hard. More so when the extraneous chars aren't immediately visible. Will try to find and remove them!

src/usr/local/www/firewall_rules_edit.php
stilez: ok

src/usr/local/www/firewall_rules_edit.php
stilez: Fixed I think? Can you check as it's hard to be sure, the extraneous chars don't show up in the GitHub UI

stilez:
Both issues addressed, hopefully all ok now

rbgarga:
@stilez looks like there is still an issue after merge it - https://redmine.pfsense.org/issues/7105

stilez:
Looks like it's been [fixed already](https://github.com/pfsense/pfsense/commit/da2a39e2961d22a403df464534b52bf6dbf9cf01) ?

rbgarga:
Yeah, it seems to be fine now. Happened only with rules created before IPv6 support, thus with no <ipprotocol> field

PR #3140, opened 2016-09-16
Improve button label + help messages for rules -> source port

stilez:
1. Rename "srcportadv" to "srcporttoggle" - not ideal to have 2 fields both labelled "advanced options". This presentation probably works nicer
2. Self-explanatory show/hide button itself doesn't need a label. (If srcports are shown, they will have a label as usual)
2. Move explanatory text about src port normally being un-needed to below src port, and simplify main help msgs.

Mainly GUI text/labels improvement

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -1324,15 +1324,17 @@ foreach (['src' => 'Source', 'dst' => 'Destination'] as $type => $name) {
 
 	if ($type == 'src') {
 		$section->addInput(new Form_Button(
-			'btnsrcadv',
-			'Display Advanced',
+			'btnsrctoggle',
+			'',
 			null,
 			'fa-cog'
-		))->setAttribute('type','button')->addClass('btn-info btn-sm');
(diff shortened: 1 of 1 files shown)


stilez:
Sorry about the multiple title edits!

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1369,16 +1371,8 @@ function build_if_list() {
...
-	if ($type == 'src')
-		$group->setHelp('Specify the source port or port range for this rule. This is '.
-			'usually random and almost never equal to the destination port range (and '.
-			'should usually be <b>any</b>).  The "To" field may be left '.
-			'empty if only filtering a single port.');
-	else
-		$group->setHelp('Specify the destination port or port range for this rule. ' .
-			'The "To" field may be left empty if only filtering a '.
-			'single port.');
+	$group->setHelp("Specify the {$name} port or port range for this rule. " .
```
rbgarga: $name should be in lower case to avoid having messages like:
"Specify the Destination port"

Other than that, setHelp() calls are added to .po to be translated, then this message should be prepared for translation using sprintf()

stilez: done

PR #3141, opened 2016-09-16
XMLRPC, xmlrpc_client, add new setConnectionUrl(url,user,pass)

PiBa-NL:
XMLRPC, xmlrpc_client, add new setConnectionUrl(url,user,pass), for use by packages that sync to custom locations, set public/private on all functions

### DIFF
--- a/src/etc/inc/xmlrpc_client.inc
+++ b/src/etc/inc/xmlrpc_client.inc
@@ -21,6 +21,8 @@
 
 require_once("XML/RPC2/Client.php");
 
+define('PEAR_IGNORE_BACKTRACE', 1);
+
 class pfsense_xmlrpc_client {
 	
 	private $username, $password, $url, $filenotice, $error;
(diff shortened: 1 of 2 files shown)


rbgarga:
I believe it would be better if we follow do something more generic that will resolve all cases of use adding individual functions. Something like:
```
__construct() {
    $this->protocol = DEFAULT_VALUE (from config['system']['webgui']['protocol'])
    $this->port = DEFAULT VALUE (from config['system']['webgui']['port'])
    $this->username = $hasync['username']
    $this->path = 'xmlrpc.php';
    ....
}

setUsername($username) {
    $this->username = $username;
}

setPassword($password) {
    $this->password = $password;
}

setServerPath($path) {
    $this->path = $path;
}
...
```

And then deprecate setConnectionData() and always build URL based on all items $protocol://$hostname:$port/$path

What do you think about it @PiBa-NL ?

PiBa-NL:
The head, ill separate it, your right i shouldn't have put it inside this same pullrequest/branche.

As for requiring to use the 'set' function for each and every parameter. I'm not sure that will bring much advantage to keep the class interface 'simple' for someone looking at it.

I think the current:
`	$rpc_client->setConnectionData($syncip, $port, $username, $password);`
Is nicer than having several separate functions for each parameter:
`	$rpc_client->setIp($syncip);
	$rpc_client->setPort($port);
	$rpc_client->setUserName($username);
	$rpc_client->setPassword($password);`

Also looking through the current packages that actually use XMLRPC it it seems that only 1 parameter was missing from the setConnectionData function, namely the 'schema'..
Almost every package has those same 5 parameters available for the sync configuration. Only pfSense itself needs it to have 'autodetect' the scheme, which would still be possible by just leaving the scheme empty..

And using a custom 'path' well i'm currently thinking it might be 'to much' to allow that.. 
And if someone is going to specify a custom target they should probably also specify the credentials to use. Otherwise we could also 'leak' the default credentials to a custom set webserver if someone doesn't have full admin permissions, but can configure sync on one of the packages.. Ofc that can then be avoided by resetting all default values when 1 custom option is set but probably unnecessarily complicates the whole class..

So my current proposal is to drop the setConnectionUrl and make setConnectionData accept 5 parameters.  ($schema, $syncip, $port, $username, $password)
Where the schema may be left empty and it would then be guessed from pfSense settings like is currently also done..

rbgarga:
@PiBa-NL $schema is supposed to have protocol (http/https), right? If yes, I agree with your idea.

src/etc/pfSense.obsoletedfiles
```diff
@@ -34,7 +34,7 @@
 /etc/inc/smtp.inc
 /etc/inc/sysctl.inc
 /etc/inc/uuid.php
-/etc/inc/xmlrpc.inc
+#/etc/inc/xmlrpc.inc
```
rbgarga: Please just remove this line

src/etc/inc/xmlrpc.inc
```diff
@@ -0,0 +1,12 @@
+<?php
```
rbgarga: I believe we don't need to re-add this file, it's used in 1 or 2 packages

src/etc/inc/xmlrpc_client.inc
```diff
@@ -43,37 +45,36 @@ class pfsense_xmlrpc_client {
 				$port = "443";
 			}
 		}
-		$this->setConnectionData($hasync['synchronizetoip'], $port, $username, $hasync['password']);
+		$this->setConnectionData("", $hasync['synchronizetoip'], $port, $username, $hasync['password']);
 	}
-	
-	public function setConnectionData($syncip, $port, $username, $password) {
+
+	public function setConnectionData($scheme, $syncip, $port, $username, $password) {
```
rbgarga: $scheme is not used in most cases. Please move it to be last parameter.

PR #3144, opened 2016-09-19
Allow snmpd to bind to multiple interfaces.

skrude61:
This patch makes it possible to bind snmpd to multiple interfaces other than ALL (for example to LAN and LAN CARP VIP). Binding to all interfaces is a security issue if you don't want to expose snmp to WAN.


rbgarga:
@skrude61 there is something wrong with your fork, this PR contains tons of commits. I'll close it. Please submit a fresh one.

skrude61:
Hi Renato, sorry for that. Merging from RELENG_2_3_2 to master apparently did something strange.
I'm not yet familiar with GIT. I've submitted a new PR a few minutes ago.
I have another patch I would like to submit a PR for, but if I try, it would submit the SNMP patch, too. How would that be handled?

rbgarga:
@skrude61 You should create a local branch on your fork for each PR you plan to submit, and then you can send multiple PRs at the same time as described at:
https://help.github.com/articles/creating-a-pull-request-from-a-fork/

PR #3147, opened 2016-09-20
Filter out already revoked certificates from drop-down list.

skrude61:
On System / Certificate Manager / Certificate Revocation / Edit in section "Choose a Certificate to Revoke" the drop-down list "Certificate" should not list already revoked certificates.

### DIFF
--- a/src/usr/local/www/system_crlmanager.php
+++ b/src/usr/local/www/system_crlmanager.php
@@ -510,6 +510,22 @@ if ($act == "new" || $act == gettext("Save") || $input_errors) {
 		}
 	}
 
+	// build list of revoked refids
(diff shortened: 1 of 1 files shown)


jim-p:
You might want multiple CRLs for the same CA, each with different sets of revoked certificates, selected on different servers, so this doesn't make a ton of sense as-is.

And if we did want to filter them, there is a function to check that already which would greatly simplify this task. See `is_cert_revoked()` at https://github.com/pfsense/pfsense/blob/master/src/etc/inc/certs.inc#L808

skrude61:
OK, I overlooked is_cert_revoked(). And regarding your objection: I will submit a new PR.

PR #3148, opened 2016-09-20
 Do not show certificates in drop-down list that are already contained in this CRL.

skrude61:
On System / Certificate Manager / Certificate Revocation / Edit in section "Choose a Certificate to Revoke" the drop-down list "Certificate" should not list certificates already contained in this CRL.


jim-p:
Much better :+1: 
Would you mind resubmitting that as just one commit so the older changes aren't in this history?

skrude61:
Done.

PR #3149, opened 2016-09-20
Reset .tcshrc prompt foreground color to terminal default instead of hardcoding white

valneacsu:
Some information in the tcsh prompt (e.g. release, username, hostname) is hardcoded to white text instead of allowing the terminal to use its default foreground color. This makes the prompt hard to read on terminals where the background is not black or a dark color. This change fixes this problem by resetting the color to the terminal default.


pfSense-git-updates:

valneacsu:
Closing so I can re-submit from a branch.

PR #3152, opened 2016-09-22
Move hard-coded timeout "240 mins" to global, and part-fix CSRF timeout code issue

stilez:
1. Various files refer to the default timeout of 240 mins. Replacing this by a single setting in $g for ease of reference and to ensure future consistency if changed.
2. Also minor fixing/improving of timeout-related logic in _system_usermanager_settings.php:
   
   (a) use is_numericint() to simplify the timeout validation logic;
   
   (b) changing the incorrect default value offered on the GUI:   blank and zero are both permitted values, but it's actually blank, not zero, which represents the default value. However zero - which is a security risk - was the default offered to the user on the GUI. The default should have been "blank" not zero, as "empty" represents the system default.
3. Also part-fixing bug in csrf_startup() in guiconfig.inc 
   
   (a) $config[] not accessible when used so test incorrectly fails when it shouldn't. Tagged as FIXME (redmine 6803)
   
   (b) Per 2.1 release notes, CSRF timeout should match session timeout. But session timeout can be 0 (non-expiring) even though insecure and not recommended, or unset (system default), and this wasn't being fully respected.


src/usr/local/www/system_usermanager_settings.php
```diff
@@ -134,7 +131,7 @@
 			}
 		}
 
-		if (isset($_POST['session_timeout']) && $_POST['session_timeout'] != "") {
+		if (is_numericint($_POST['session_timeout'])) {
```
stilez: Note - we don't need to test further at this point, because if it isn't either a positive integer or (empty string/unset), $input_errors[] will have been nset above and we'll never get to this line anyway

src/etc/inc/globals.inc
```diff
@@ -80,8 +80,11 @@ $g = array(
 	"help_base_url" => "/help.php",
 	"pkg_prefix" => "pfSense-pkg-",
 	"default_timezone" => "Etc/UTC",
-	"language" => "en_US"
+	"language" => "en_US",
+	"default_session_timeout_mins" => 240
 );
+		// NOTE - the value of "default_session_timeout_mins" is hard-coded in guiconfig.inc, because csrf_magic is loaded early
```
rbgarga: Move this comment just above default_session_timeout_mins definition, and while on that use a comment block with /* */ keeping it on 80-columns

src/usr/local/www/guiconfig.inc
```diff
@@ -28,8 +28,22 @@
 if (!$nocsrf) {
 	function csrf_startup() {
 		csrf_conf('rewrite-js', '/csrf/csrf-magic.js');
-		$timeout_minutes = isset($config['system']['webgui']['session_timeout']) ? $config['system']['webgui']['session_timeout'] : 240;
-		csrf_conf('expires', $timeout_minutes * 60);
+
+//FIXME: 
+// This code doesn't seem to work as expected, because $config[] isn't accessible within csrf_startup() when it executes, so it mis-sets csrf timeout. 
+// Ref - https://redmine.pfsense.org/issues/6803 (1st note)
```
rbgarga: Isn't it because there is no "global $config" declaration here?

Looks like it was fixed by e0d81869f0 and then you can remove this comment

stilez: That doesn't fix it. See discussion on Redmine 6803. The problem is that $config not only isn't declared via a "global" statement - it isn't available and hasn't been rread from file at that point. So the whole reading of config.xml + any decrypting, etc, is needed, or some workaround, not just adding a "global $config" statement. The above "fix" wasn't a fix.

rbgarga:
@stilez can you please rebase?

stilez:
Conflicts removed, mergeable

src/usr/local/www/system_usermanager_settings.php
```diff
@@ -177,15 +174,19 @@
...
 	'Session timeout',
 	'number',
-	$pconfig['session_timeout'],
-	['min' => 0]
-))->setHelp('Time in minutes to expire idle management sessions. The default is 4 '.
-	'hours (240 minutes). Enter 0 to never expire sessions. NOTE: This is a security '.
-	'risk!');
+	$pconfig['session_timeout']
+))->setHelp(sprintf(gettext('Time in minutes to expire idle management sessions. Leave empty for the default timeout ' .
+	'of %s. Enter 0 to never expire sessions. NOTE: Non-expiry is a security risk!'), $d));
```
jim-p: The note text needs changed. The last two sentences could be better, maybe:
"A value of 0 will never expire idle sessions, which is a security risk"

src/usr/local/www/system_usermanager_settings.php
```diff
@@ -177,15 +174,19 @@
 
 $section = new Form_Section('Settings');
 
+list($d_hr, $d_min) = [int($g['default_session_timeout_mins']/60), $g['default_session_timeout_mins'] % 60];
```
jim-p: This seems like a lot of extra work for what it does. I'd just express it all in minutes since that's what the field takes.

src/usr/local/www/guiconfig.inc
```diff
@@ -29,8 +29,22 @@ if (!$nocsrf) {
...
+
+		if (!isset($config['system']['webgui']['session_timeout'])) {
+			csrf_conf('expires', 240 * 60);
```
jim-p: It may not be able to access $config at this point, but can it not access $g either? Seems it would be better to read the value straight from $g so it doesn't need to be put in two places, but if it's too early for that then there may not be any better choice.

Thinking more generally, there is also the option of reading it via a shell exec to /usr/local/sbin/read_xml_tag.sh at some point during the boot process and then setting it in an environment variable or in some other readable way. Doing that directly from PHP is ugly and parsing the config in PHP at that point also seems ugly since it will be read in shortly anyhow.

src/usr/local/www/system_usermanager_settings.php
stilez: Its not bad ion code. more to the point, if a user scenario leads to a longish timeout, then expressing it in minutes just isn't intuitively clear, and a timeout (like any security setting) should be easily clear on plain reading. I'll lookl at if the code can be simplified though.

src/usr/local/www/guiconfig.inc
stilez: makes very good sense but beyond me. if needed to fix bug 6803 it'll need someone with much greater knowhow than I have, to write the code change for this.

Could you?? :)

src/usr/local/www/system_usermanager_settings.php
stilez: text changed. better?

stilez:
issues in this area AFAIK - 


1. $config isn't available at the point CSRF needs to know and set the page timeout.

stilez:
@rbgarba - can you update the tags on this one, so it gets dealt with? Issues dealt with AFAIK, but points above need a look

src/etc/inc/globals.inc
```diff
@@ -85,6 +85,12 @@ $g = array(
 	"pkg_prefix" => "pfSense-pkg-",
 	"default_timezone" => "Etc/UTC",
 	"language" => "en_US",
+/*   FIXME - the value of "default_session_timeout_mins" is hard-coded 
+     in guiconfig.inc, because csrf_magic is loaded early before $config is available.
+     See bug 6803.
+     Any change to this default would need changing there as well.
+*/
```
rbgarga: Indent is broken on comment. While touching it please use the desired multiline comment style:
```
/*
 * foo
 * bar
 */

File: src/usr/local/www/guiconfig.inc [hunk 8]
```diff
@@ -29,8 +29,22 @@ if (!$nocsrf) {
 	function csrf_startup() {
 		global $config;
 		csrf_conf('rewrite-js', '/csrf/csrf-magic.js');
-		$timeout_minutes = isset($config['system']['webgui']['session_timeout']) ? $config['system']['webgui']['session_timeout'] : 240;
-		csrf_conf('expires', $timeout_minutes * 60);
+
+//FIXME: 
+// This code doesn't seem to work as expected, because $config[] isn't accessible within csrf_startup() when it executes, so it mis-sets csrf timeout. 
+// Ref - https://redmine.pfsense.org/issues/6803 (1st note)
```
rbgarga: Fix comment indent, style using /* */ and try to keep line size shorter please

src/usr/local/www/system_usermanager_settings.php
```diff
@@ -177,15 +174,19 @@
...
 	'session_timeout',
 	'Session timeout',
 	'number',
-	$pconfig['session_timeout'],
-	['min' => 0]
-))->setHelp('Time in minutes to expire idle management sessions. The default is 4 '.
-	'hours (240 minutes). Enter 0 to never expire sessions. NOTE: This is a security '.
-	'risk!');
+	$pconfig['session_timeout']
+))->setHelp(sprintf(gettext('Time in minutes to expire idle management sessions. Leave empty for the default timeout ' .
```
rbgarga: setHelp was reworked recently to accept syntax similar to sprintf, please remove sprintf and gettext from it and use:
```
-> setHelp("This string has a parameter %s", "parameter");
```

sbeaver-netgate:
Awaiting changes for 9 months. Will gladly re-open if those changes are made

PR #3157, opened 2016-09-25
Allow get_pkg_info() to fallback on "pkg info" for local pkgs if "pkg search" cannot access a local catalog copy

stilez:
_Current behaviour_

At the moment, get_pkg_info() is used to get all information on packages.  The parameter _$only_local_ is set to request info directly from the local copy of the repo catalog (using -U) without requesting the remote repo catalog or updating the local copy from the remote repo catalog.  If the calling code wants only installed pkgs, it filters the returned list of pkgs looking for _$pkg['installed'] == true_.

There's a couple of problems (or unnecessary limitations) with this, due to the behaviour of pkg search -U. 
1. if the remote catalog is requested and the request failed, the local copy will also be deleted. If this happens, then pkg search returns an error even with -U.  So even if all we wanted to know from get_pkg_info() was the names of locally installed packages, it can't be used for this (even though getting this information would not need access to a remote or local repo catalog)
2. get_pkg_info() is the main method to get a list of installed pfSense packages and their data. This behaviour of pkg search means that any time we don't have a network connection or for any reason get_pkg_info() fails to access the repo catalog remotely, we become blocked perpetually from _any_ information, listing, or action on _all_ router packages, even if the desired wouldn't need remote access or repo catalog info to complete.  We will continually be locked out from these in the GUI (and in any console pfSense utilities which use get_pkg_info()) until remote repo access is obtained again and a catalog copy can be re-acquired. 
3. This also pretty much blocks anything to do with getting installed package information or removal from occurring offline either.

_Change made_

It's likely that when code explicitly requests $local_only, it isn't expecting or requiring the local copy to be up to date.  So I've modified the code as follows:
1. removed $info argument which was never being used (in the codebase or the function itself)
2. Added a new optional parameter $installed_only to explicitly request installed pkg info only (faster if we know that no remote request will be needed)
3. Created a fallback when pkg search fails - if $local_only is set and pkg search failed, retry falling back to pkg info to at least provide info on matching installed packages. This is probably more helpful than returning an error, as the assumption with $local_only is not "latest data in repo" so no harm done.  It means the user can at least view and operate on local pkgs at all times in the GUI (although any info about uninstalled pkgs or repo queries generally will still need a means to query the remote or local repo catalog, of course).

_Tested online and offline - test method as follows:_
1. Create floating rule to block any traffic on any protocols to/from WAN
2. Following code snippet in command prompt:

<pre>
require_once("pkg-utils.inc");
require_once("util.inc");
$input_errors='';

$pkglist = get_pkg_info('all', false, false);  // VARY THE ARGS HERE

// cut down the detail in the resulting array (if any), to keep it simple for testing
if(is_array($pkglist)) {
  $brief_output = array_map(
    function($d) {
      $a = '';
      if (isset($d['name'])) { $a.= "Name: {$d['name']}  "; }
      if (isset($d['version'])) { $a.= "Version: {$d['version']}  "; }
      if (isset($d['installed_version'])) { $a.= "Installed_version: {$d['installed_version']}  "; }
      if (isset($d['installed'])) { $a.= "Installed? {$d['installed']}  "; }
      if (isset($d['broken'])) { $a.= "Broken? {$d['broken']}"; }
      return $a;
    },
    $pkglist
  );
}

echo "Input errors (if any):\n" . print_r($input_errors, true) . "\n\n";
if(is_array($pkglist)) {
  echo "Results: \n" . print_r($brief_output, true) . "\n";
} else {
  echo "Did not return an array. Textified return value was: " . $pkglist . "\n";
}
</pre>

Tested all online. if the WAN is killed, and $installed_only = false, $only-local works until the first time it's false, from then on it fails. But when offline and either $installed_only is set, or $only_local is set (ie we'd be happy with a local copy), it still works.

_Impact_

The change is transparent so it shouldn't have any untoward effect.  

get_pkg_info() isn't called in many places either, and its default behaviour is unchanged.

If get_pkg_info() falls back to, or uses, pkg info, then it has 2 effects - the function won't report newer versions that are on the repo, or other non-installed packages (that are on the repo). But in both cases that's a direct and obvious effect of not actually having access to the remote repo to get that data. At least with this fallback, you get the best info you can, and you can at least view/remove/get info on installed pkgs where previously the GUI wouldn't let you.

It will also speed up inquiries where only installed pkgs are needed, as no repo catalog access occurs.


rbgarga:
Please rebase your fork

stilez:
Sorry,. I have no idea where the conflicts reported above are happening. So I can't see my way to fixing it. I've tried to copy the current version and re-add back the changes in the original PR, but it still says there are conflicts, but the web UI wont show me where. @rbgarga - can you help me out a bit?

stilez:
Conflict may be resolved - as I'm new to GitHub "resolve conflicts" please check especially that the commits attempting to fix the conflict didn't introduce any code changes/errors

src/etc/inc/pkg-utils.inc
```diff
@@ -359,21 +360,35 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
...
 
 	if ($lock) {
 		mark_subsystem_dirty('pkg');
 	}
-	$rc = pkg_exec("search {$extra_param}-R --raw-format json-compact " . $pkgs, $out, $err);
+	if (!$installed_only) {
+		// repo catalog search (either remote or local_only)
+		$rc = pkg_exec("search {$extra_param}-R --raw-format json-compact " . $pkgs, $out, $err);
+	}
+	if (($rc != 0 && $local_only) || $installed_only) {
```
rbgarga: At this point $rc may be uninitialized when $installed_only is true, please initialize it before

src/etc/inc/pkg-utils.inc
```diff
@@ -359,21 +360,35 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
...
+		/* use pkg info if (1) installed pkg search or (2) local catalog copy search requested + failed.
+		
+		   The local repo catalog copy may be cleared if a previous call to pkg search couldn't get the remote repo catalog.
+		   
+		   If the calling code would have accepted local copy info (which isn't assumed up to date) then it makes sense to
+		   fall back on pkg info to at least return the known info about installed pkgs (pkg info should still work), instead
+		   of failing and returning no info at all. For example, this enables offline view + management of installed pkgs. */
```
rbgarga: Since there are another change to do please use following format for multi line comments and try to keep it in 80 columns:
```
/*
 * multiline comment
 * should be formatted like this
 */
```

stilez:
changes made

rbgarga:
Something wrong happened during rebase so I've imported it direct as one commit e47af756de79d4e8b0356cf22f72f62f09e9ad7d

NOYB:
The get_pkg_info() fallback using pkg info if no local copy of repo catalog commit results in this error at end of boot up for both 2.4 and 2.3.

ERROR: Error trying to get packages list. Aborting...
https://forum.pfsense.org/index.php?topic=123670.0

rbgarga:
@stilez I've reverted it for now until we can find a fix

stilez:
I tried this with 2.3.2 on a VM, which should be a prime target if there's an issue, and had no issues on rebooting, either with network adapters connected on boot or disconnected on boot. 

I used the version of pkg-utils.inc in this PR (noting there was an issue merging so @rbgarba merged it in one go above).  

@NYOB - any debug info, or a version of config.xml that triggers the issue?

NOYB:
Did not capture anything other than the console displayed error message right after generating RRD graphs.
Installed a fresh 2.4 in VirtualBox VM, selecting all the installation defaults, and had the error occurred with that as well.
Never tried in 2.3.2 because the commit was only in 2.4 and 2.3 branches.

stilez:
Found the issue - it's down to the behaviour of `pkg info`, which errors out rather than providing an empty return, if no packages matched locally (as will be the case in @NYOB's clean install). Fix is to test specifically if there are zero installed packages matching, before using `pkg info` to get the list of local matching pkgs.  

Resubmitted with fix, in PR #3418.

PR #3158, opened 2016-09-25
Simplify logic

stilez:
Nested conditions go back and forth between setting "FloatingRules", or setting "wan", or setting "FloatingRules" again.  No need for that


phil-davis:
This is when we need nicely callable functions for everything and a heap of test cases that call the functions in all sorts of ways and verify the results. Then we could refactor all this sort of code and feel more comfortable about it.
From looking at the changes, I agree that the new logic has the same effect as the previous logic.

PR #3160, opened 2016-09-26
Enable a direction on any firewall rule

stilez:
It can be very useful to set rules on outgoing (not just incoming) packets on an IF. This is already available for floating rules, but not for rules on other IF's (even though it would work). No obvious drawback or issues resulted in testing or apparent from the codebase or PF documentation. So this PR makes the "in/out/any" option visible and working on all rules, not just floating rules. Totally transparent + no config update needed. 

firewall_rules.php now shows interfaces+direction (for floating rules, which it probably should anyway as this is quite important info) or just direction (others interfaces). Care taken to ensure word wrapping is assisted on the IF's for floating rules, for narrower displays.

If direction is not explicitly set on a rule, then it's assumed to be "any" for floating rules and "in" for rules on all other interfaces (both when editing the rule and also when building the pf config in filter.inc).

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2705,14 +2705,14 @@ function filter_generate_user_rule($rule) {
 	} else {
 		$aline['type'] = $type . " ";
 	}
-	if (isset($rule['floating']) && $rule['floating'] == "yes") {
-		if ($rule['direction'] != "any") {
-			$aline['direction'] = " " . $rule['direction'] . " ";
-		}
-	} else {
-		/* ensure the direction is in */
+	if (isset($rule['direction']) && in_array($rule['direction'], ['in', 'out'])) {
+		// if explicitly in/out, use that value
+		$aline['direction'] = " " . $rule['direction'] . " ";
+	} elseif (!(isset($rule['floating']) && $rule['floating'] == "yes") && $rule['direction'] != 'any') {
+		// not floating rule and not "any", "in" or "out", so set to "in" - default for non floating rules
 		$aline['direction'] = " in ";
 	}
+		// else "any" direction, either by default or because set explicitly - nothing to add to rule
 	if (isset($rule['log'])) {
 		$aline['log'] = "log ";
 	}
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -363,7 +363,7 @@ if (isset($config['interfaces'][$if]['blockbogons'])) {
 $rulescnt = pfSense_get_pf_rules();
 
 // Update this if you add or remove columns!
-$columns_in_table = 13;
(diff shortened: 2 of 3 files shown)


jim-p:
While it is technically possible, it's not something we've deliberately not wanted to offer to the user to reduce confusion and the potential for error. If they want a directional rule, they can use floating rules.

stilez:
I'm not sure that direction will confuse users, given the other options that are provided. Also floating rules go at a specific place in the rules list, so you can't put a rule which specifies direction for a specific interface, above some rules and below others. Often floating rules just aren't very helpful, unless you make almost everything a floating rule, because you can't sequence them in with other rules, that's not their function.

I don't think it's a complicated option, to specify whether the test is applied on incoming or outgoing packets.  We don't see people confused by floating rules, AFAIK?  It isn't confusing the GUI, and it's very useful in many cases + provides an ability to add rules in sequence for a specific interface, which can't always be done easily otherwise.  Can you reconsider that it actually isn't that complicated?

That said, if it really is likely to confuse some users, can we at least move it in with the "advanced options", so the functionality is available when wanted by other users? Would that be ok?

jim-p:
I just don't see it being worth the trouble no matter where it's put, when floating rules serve the purpose fine. If you need outbound rules they won't have any bearing on the inbound rules anyhow, so having an outbound floating rule is no different than putting it in the middle of the interface ruleset.

From a user support perspective it would be a usability nightmare no matter where it's put. If rules are misbehaving it's yet another place admins and support reps have to check to see if the rule was goofed. Visual indicators can help here, but the moment someone finds it they'll think they need to use it, and they'll inevitably get the direction mixed up.

tl;dr: It's a rarely needed option that has no clear benefit to being on the interface rule tabs except for very rare use cases that are already properly covered by floating rules.

stilez:
Hmm. 
I am a bit awed that anyone who can cope with basic pfSense and basic pfSense rules, will get confused by whether the rule is looking at packets arriving at the router or leaving the router. But I would respect your experience in this - it's common knowledge how badly cluelessness can strike. But I just wouldn't have thought the incremental cluelessness - the difference between support headache level with pfSense and with pfSense + packet direction, would be much of anything. You really think "nightmare for support" is a realistic additional impact if the option is offered, or would create a support nightmare in any user where one didn't exist before?

jim-p:
It's not just an incremental change. It's a massive change in behavior for rules. It changes the whole meaning of rule tabs that has been consistent in all prior pfSense releases. In addition to the GUI change it would also require updating all the documentation on how rules work in the wiki, book, videos, and so on. It's not the kind of change we can just put behind an advanced button and hope only those who need it, see it. It was one of the primary motivators for making the floating rule tab, so that it would be less likely to confuse or be misused/abused in an isolated location.

PR #3166, opened 2016-09-30
System Update Settings - GitSync Options

NOYB:
Add GitSync options selectors to post update GitSync.
Minimal, Diff, Verbose, Dry Run

The GitSync options have utility for post update sync also.

Also applicable to RELENG_2_3_2.

Dependency Note:
The verbose option will need to be split out into "show_files" and "show_command" when the following commit is applied.
https://github.com/pfsense/pfsense/pull/3168/commits/a6fd4c0ff4141a1836ed74add1df55477d87ddd0


rbgarga:
What would be the idea of having options like --dry-run or --show-command set on GUI?

NOYB:
Testing, debugging, troubleshooting, etc.  Same reasons one might use them in the command shell.

rbgarga:
@NOYB can you please re-submit it for master branch?

NOYB:
For master branch
https://github.com/pfsense/pfsense/pull/3436

PR #3170, opened 2016-10-01
Reset .tcshrc prompt foreground color to terminal default instead of hardcoding white

valneacsu:
Some parts of the tcsh prompt (e.g. release vesrion, username, hostname) are hardcoded to white text instead of using the terminal default foreground color. This results in the prompt being hard to read on terminals where the background is not black or a dark color, as there is poor contrast between the light background and white text. This change fixes the problem by resetting the foreground color to the terminal default foreground color instead of the hardcoded white.


rbgarga:
Pushed a different fix, thanks!

PR #3172, opened 2016-10-03
Improve explanation on Outbound NAT target fields

stilez:
The fields are a bit ambiguous, making it a bit confusing for people with less experience of Outbound NAT. 

Improves the help fields on the target interface (it's outbound from the router not outbound from the source subnet as might be thought, and clarify when it's WAN), move the hint about VIP to the "target" field where it's more relevant (and explain better when to create a VIP), and clean up the message for manually entered subnets.


jim-p:
This had conflicts and I wasn't too happy with the wording. Far too much text for that page. I committed some changes that get the point across.

PR #3176, opened 2016-10-03
standardise old code ("or" -> "||")

stilez:


src/etc/inc/config.console.inc
```diff
@@ -532,8 +532,7 @@ function vlan_setup() {
 		$vlan['if'] = chop(fgets($fp));
 
 		if ($vlan['if']) {
-			if (!array_key_exists($vlan['if'], $iflist) or
-			    !is_jumbo_capable($vlan['if'])) {
+			if (!array_key_exists($vlan['if'], $iflist) || !is_jumbo_capable($vlan['if'])) {
```
rbgarga: Please keep it in 2 lines to avoid adding new lines > 80 columns

stilez: C'mon, is that really needed? :) Not least look how many lines in the code (right close to it, let alone elsewhere and throughoput all the codebase) are > 80 chars. Does anyone in this day and age really need lines kept to 80 for reasons that still apply in this day and age of large screen resolutions (which almost all devs will have)?

That said, if it really needs to be kept to 80, then yes, for sure. But to my mind I feel that it is then harder to read, not easier, if lines are artificially split midway and most of the screen width is unusable.

Let me know if you really want it!

bms: This is, quite possibly, one of my pet peeves: code that overflows 80 columns and so I can't hack it well on a netbook-type device. Also what rbgarga said about code quality. If you have too many nested conditionals, you're doing it wrong, IMO.

PR #3178, opened 2016-10-07
Workaround extension load order bug ensuring session.so always loads first

NewEraCracker:
https://forum.pfsense.org/index.php?topic=119253.0
(With slight changes from original php-session-fix.txt)

Edit: Reviewed to fix a minor logic issue.

### DIFF
--- a/src/etc/rc.php_ini_setup
+++ b/src/etc/rc.php_ini_setup
@@ -101,6 +101,25 @@ else
 	fi
 fi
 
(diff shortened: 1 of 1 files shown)


rbgarga:
This change is not necessary. I've bumped PHP port revision, what will lead to all PHP ports to be rebuilt using the new extensions framework. New packages 5.6.26_1 will be in place soon. Thanks

PR #3180, opened 2016-10-09
Bring up wireless interfaces only after setting up all the interface parameters

valneacsu:
This prevents issues when using VAPs as described in https://redmine.pfsense.org/issues/5616

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2578,11 +2578,6 @@ function interface_wireless_configure($if, &$wl, &$wlcfg) {
 
 	$wlcmd = array();
 	$wl_sysctl = array();
-	/* Disable this for now as we will bring up the interface as a last step.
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -2578,8 +2578,11 @@ function interface_wireless_configure($if, &$wl, &$wlcfg) {
 
 	$wlcmd = array();
 	$wl_sysctl = array();
-	/* Make sure it's up */
-	$wlcmd[] = "up";
+	/* Disable this for now as we will bring up the interface as a last step.
+	 * Bringing up the interface with all the other arguments causes issues when using VAPs.
+	 * Related to #5616.
+	 */
+	//$wlcmd[] = "up";
```
rbgarga: It's better to just remove the code here instead of commenting it out. We have history to see why it was removed

PR #3182, opened 2016-10-10
[theme] Compact-RED: backport changes from master

moisseev:


rbgarga:
RELENG_2_3_2 is a maintenance branch, we try to change it only when necessary. This is most a cosmetic change for a non-official theme

PR #3183, opened 2016-10-11
Register CN of OpenVPN clients in DNS Resolver

znerol:
https://redmine.pfsense.org/issues/6847

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -777,7 +777,7 @@ function openvpn_reconfigure($mode, $settings) {
 				break;
 		}
 	}
-	if (isset($config['unbound']['enable']) && isset($config['unbound']['regovpnclients'])) {
+	if ($settings['dev_mode'] === 'tun' && isset($config['unbound']['enable']) && isset($config['unbound']['regovpnclients'])) {
 		switch($settings['mode']) {
 			case 'server_tls':
 			case 'server_tls_user':
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -777,6 +777,15 @@ function openvpn_reconfigure($mode, $settings) {
 				break;
 		}
 	}
+	if (isset($config['unbound']['enable']) && isset($config['unbound']['regovpnclients'])) {
+		switch($settings['mode']) {
+			case 'server_tls':
+			case 'server_tls_user':
+				$domain = !empty($settings['dns_domain']) ? $settings['dns_domain'] : $config['system']['domain'];
+				$conf .= "learn-address \"/usr/local/sbin/openvpn.learn-address.sh $domain\"\n";
+				break;
+		}
+	}
 
 	/*
 	 * When binding specific address, wait cases where interface is in
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -782,7 +782,9 @@ function openvpn_reconfigure($mode, $settings) {
 			case 'server_tls':
 			case 'server_tls_user':
 				$domain = !empty($settings['dns_domain']) ? $settings['dns_domain'] : $config['system']['domain'];
-				$conf .= "learn-address \"/usr/local/sbin/openvpn.learn-address.sh $domain\"\n";
+				if (is_domain($domain)) {
+					$conf .= "learn-address \"/usr/local/sbin/openvpn.learn-address.sh $domain\"\n";
+				}
 				break;
 		}
 	}
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -376,6 +376,9 @@ include: {$g['unbound_chroot_path']}{$cfgsubdir}/host_entries.conf
 # dhcp lease entries
 include: {$g['unbound_chroot_path']}{$cfgsubdir}/dhcpleases_entries.conf
 
+# OpenVPN client entries
+include: {$g['unbound_chroot_path']}{$cfgsubdir}/openvpn.*.conf
+
 # Domain overrides
 include: {$g['unbound_chroot_path']}{$cfgsubdir}/domainoverrides.conf
 {$forward_conf}
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -266,6 +266,15 @@ EOF;
 	$verbosity = isset($unboundcfg['log_verbosity']) ? $unboundcfg['log_verbosity'] : 1;
 	$use_caps = isset($unboundcfg['use_caps']) ? "yes" : "no";
 
+	if (isset($unboundcfg['regovpnclients'])) {
+		$openvpn_clients_conf .=<<<EOD
+# OpenVPN client entries
+include: {$g['unbound_chroot_path']}{$cfgsubdir}/openvpn.*.conf
+EOD;
+	} else {
+		$openvpn_clients_conf = '';
+	}
+
 	// Set up forwarding if it is configured
 	if (isset($unboundcfg['forwarding'])) {
 		$dnsservers = array();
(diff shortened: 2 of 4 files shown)


pfSense-git-updates:

src/etc/inc/openvpn.inc
```diff
@@ -719,6 +719,15 @@ function openvpn_reconfigure($mode, $settings) {
 				break;
 		}
 	}
+	if ($settings['dev_mode'] === 'tun' && isset($config['unbound']['enable']) && isset($config['unbound']['regovpnclients'])) {
```
jim-p: Why lock this to mode 'tun'? It's also possible to make a tap mode OpenVPN without a bridge that has addresses managed by OpenVPN.

src/etc/inc/openvpn.inc
```diff
@@ -719,6 +719,15 @@ function openvpn_reconfigure($mode, $settings) {
...
+		switch($settings['mode']) {
+			case 'server_tls':
+			case 'server_tls_user':
+				$domain = !empty($settings['dns_domain']) ? $settings['dns_domain'] : $config['system']['domain'];
+				$conf .= "learn-address \"/usr/local/sbin/openvpn.learn-address.sh $domain\"\n";
```
jim-p: $domain should be wrapped in escapeshellarg()

src/usr/local/sbin/openvpn.learn-address.sh
```diff
@@ -0,0 +1,53 @@
...
+
+DIR="/var/unbound"
+CONF="$DIR/openvpn.client.$IP.conf"
+PIDFILE="/var/run/unbound.pid"
+
+case "$OP" in
+
+	add|update)
+		# Remove all configs which mention the FQDN
+		grep -l -null -F "local-data: \"$CN.$DOMAIN A $IP\"" $DIR/openvpn.client.*.conf | xargs -0 rm
```
jim-p: Use braces around shell variables to reduce ambiguity, for example: `${CN}.${DOMAIN}` and so on. Many instances of this throughout the script.

src/usr/local/sbin/openvpn.learn-address.sh
```diff
@@ -0,0 +1,53 @@
...
+		rm -f "$CONF"
```
jim-p: Needs a sanity check before doing rm based on a variable name alone.

src/usr/local/sbin/openvpn.learn-address.sh
```diff
@@ -0,0 +1,53 @@
...
+		(
+			echo "server:" &&
+			echo "chroot: $DIR" &&
+			echo "directory: $DIR" &&
+			echo "include: $TMPCONF"
+		) > "$TMPSRV"
+
+		chmod 644 "$TMPCONF" "$TMPSRV"
+		/usr/local/sbin/unbound-checkconf "$TMPSRV" && mv "$TMPCONF" "$CONF"
+		rm -f "$TMPCONF" "$TMPSRV"
```
jim-p: Needs a sanity check before doing rm based on a variable name alone.

src/usr/local/sbin/openvpn.learn-address.sh
```diff
@@ -0,0 +1,53 @@
...
+
+		/bin/pkill -HUP -F "$PIDFILE"
+	;;
+
+	delete)
+		# CN is not set on delete.
+		rm -f "$CONF" && /bin/pkill -HUP -F "$PIDFILE"
```
jim-p: Needs a sanity check before doing rm based on a variable name alone.

rbgarga:
I'll wait items pointed out by @jim-p to be addressed before start reviewing

src/etc/inc/openvpn.inc
znerol: According to [man openvpn](https://openvpn.net/index.php/open-source/documentation/manuals/65-openvpn-20x-manpage.html) the address script parameter ...

> can be an IPv4 address such as "198.162.10.14", an IPv4 subnet such as "198.162.10.0/24", or an ethernet MAC address (when **--dev tap** is being used) such as "00:FF:01:02:03:04".

The script cannot handle MAC addresses (hence we cannot operate in `tap` device mode) and neither IPv4 subnets. I believe we never get those when the `mode` config is one of `server_tls` or `server_tls_user` (please correct me if I'm wrong).

src/usr/local/sbin/openvpn.learn-address.sh
znerol: What sort of sanity check do you have in mind here? `TMPCONF` and `TMPSRV` are generated using `mktemp` which I assume is the correct way to securely create temporary files with unique names.

jim-p: At the very least, a test to make sure that both parameters have some value that (e.g. not empty) and that none of them are only '/'. Ideally, it should also test if the file exists before trying to use `rm -rf` on it rather than assuming it does. There are a couple other usages of rm in the script that could use the same treatment.

Basically, we need to make sure the script won't `rm -rf` the entire filesystem or more files than intended, especially if somehow the mktemp call does not return an expected result.

tl;dr: Make no assumptions, assume input is invalid until proven otherwise, especially when removing files using variable names.

src/etc/inc/openvpn.inc
jim-p: OK, that makes sense. Though you should then also reflect this limitation in the help/description of the GUI option.

src/usr/local/sbin/openvpn.learn-address.sh
znerol: The script does not use `-r` flag to `rm`, there is no recursion. Also note that `$IP` and `$CONF` should be fine now due to f4e8b87

jim-p:
Two more general suggestions that should be addressed:
1. Calls to executables (e.g. `grep`, `rm`) must use the full path to the utilities rather than relying on `$path`.
2. Variable references in shell scripts need braces in all references, e.g. `$IP` -> `${IP}`

znerol:
@jim-p even if a string only consists of exactly one variable? e.g. `/bin/pkill -HUP -F "$PIDFILE"` -> `/bin/pkill -HUP -F "${PIDFILE}"`

jim-p:
Yes, it is more clear and consistent with other code that way. Doing it half one way and half the other looks sloppy IMO.

znerol:
@jim-p I think I addressed all of the feedback now (thanks a lot, btw.). I did a smoke test on one of my hosts, and the seems to behave. Let me know if there is anything else I can do here.

src/etc/inc/unbound.inc
```diff
@@ -376,6 +376,9 @@ include: {$g['unbound_chroot_path']}{$cfgsubdir}/host_entries.conf
 # dhcp lease entries
 include: {$g['unbound_chroot_path']}{$cfgsubdir}/dhcpleases_entries.conf
 
+# OpenVPN client entries
```
jim-p: Perhaps this should be wrapped in similar tests to the case in openvpn.inc, so it isn't looking for these files when they do not or cannot exist? Or at least test for `$config['unbound']['regovpnclients']`

src/usr/local/sbin/openvpn.learn-address.sh
```diff
@@ -0,0 +1,60 @@
...
+DIR="/var/unbound"
+PIDFILE="/var/run/unbound.pid"
+
+if [ -n "${IP}" -a "$(basename ${IP})" = "${IP}" ]; then
+	CONF="${DIR}/openvpn.client.${IP}.conf"
+
+	case "${OP}" in
+
+		add|update)
+			TMPCONF=$(mktemp "${CONF}.XXXXXX")
```
jim-p: Needs full path (`/usr/bin/mktemp`) on this and the following line

src/usr/local/sbin/openvpn.learn-address.sh
```diff
@@ -0,0 +1,60 @@
...
+
+DOMAIN="${1}"
+OP="${2}"
+IP="${3}"
+CN="${4}"
+
+DIR="/var/unbound"
+PIDFILE="/var/run/unbound.pid"
+
+if [ -n "${IP}" -a "$(basename ${IP})" = "${IP}" ]; then
```
jim-p: Needs full path (`/usr/bin/basename`)

PR #3187, opened 2016-10-13
Replace underscore with hyphen in option names

NOYB:
Thanks Jorge


NOYB:
Pushed this to the wrong branch.  Deleting

PR #3189, opened 2016-10-13
Make is_inrange_v6 more robust

phil-davis:
If passed invalid IPv6 strings in the parameters, inet_pton() was
emitting warnings. Avoid this by first checking for valid IPv6
addresses.

This helps https://forum.pfsense.org/index.php?topic=119565.0 by stopping annoying "inet_pton(): Unrecognized address" warnings.


phil-davis:
Related forum post: https://forum.pfsense.org/index.php?topic=119565.0

rbgarga:
I've pushed a different version. Thanks!

PR #3190, opened 2016-10-13
DHCPV6 only check VIPs in range if range valid

phil-davis:
If the user has input invalid values into range from and to, then there
is no point checking any IPv6 VIPs to see if they fall in the range.
None of them would be "in range" because the specified range is not even
valid.


phil-davis:
Related forum post: https://forum.pfsense.org/index.php?topic=119565.0

PR #3191, opened 2016-10-13
RAM Disk Management

NOYB:
RAM Disk Management

1) Treat the RAM disk more like a permanent storage device with content managed/restored by the system and made available at boot up, before needed by any services.
 a) Handle saving and restoring RAM disk content at reboot/shutdown/boot centrally in more of a system manged fashion.
 b) If it's in the RAM disk store it gets restored early in the pfSense startup so it's available for whatever needs to use it.
 c) Services utilizing RAM disk don't need to be aware that their content is on a RAM disk, and handling content restore individually.

2) Has the benefit of eliminating some issues with the previous code as well. Such as...
 a) Restoring RRD multiple times during boot, potentially at least 3 times, by rc.newwanip, rc.newwanipv6, and rc.boot. Some even overlapping.
 b) Not removing the backups if/when not being utilized. Such as on a full install with the RAM disk option not enabled.
 c) Eliminate some duplicate code.

3) Eliminate rrd dump/restore to/from xlm file.
Backup and restore the entire /var/db/rrd/ to/from the RAM Disk Store.

4) Add logic to backup alias tables if the backup is older than the table.

5) Include the logs in the RAM disk store so they will persist across reboots.

6) Looking forward.
 a) The more centrally system managed approach and restoring the RAM disk content as early as possible will make it easier to include other content that may be useful/desirable or need to persist in the RAM disk across reboots. Some packages data for instance.  Simply `tar -czf` what needs to be restored to `/cf/conf/RAM_Disk_Store/*.tgz`.  See existing `/etc/rc.backup*.sh` for examples.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -2496,13 +2496,12 @@ function process_alias_urltable($name, $type, $url, $freq, $forceupdate=false, $
 				touch($urltable_filename);
 			}
 
-			/* If this backup is still there on a full install, but we aren't going to use ram disks, remove the archive since this is a transition. */
-			if (($g['platform'] == $g['product_name']) && !isset($config['system']['use_mfs_tmpvar'])) {
-				unlink_if_exists("{$g['cf_conf_path']}/RAM_Disk_Store{$urltable_filename}.tgz");
-			} else {
-				/* Update the RAM disk store with the new/updated table file. */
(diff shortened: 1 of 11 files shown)


rbgarga:
It supersedes #2902

NOYB:
No
#3191 is for 2_3.
#2902 is for master.

rbgarga:
Yeah, I deleted the comment seconds after I added it.

NOYB:
So that's why I couldn't find the comment.
Yeah.   They are both nearly identical.  Just a couple differences.  Don't recall the differences specifically of top of my but, as of this moment, they are very few.

rbgarga:
Such a big change for 2.3, lets keep it 2.4 only. Thanks!

PR #3194, opened 2016-10-16
FIX 6857: Configure routing early on during boot process, so remote ldap server s are available when invoking local_sync_accounts()

pruiz:
This pull-req fixes redmine issue: https://redmine.pfsense.org/issues/6857


rbgarga:
Pushed a different fix, thanks!

PR #3195, opened 2016-10-16
FIX 6856: Avoid removal of lo0's IP aliases during configuration synchronization

pruiz:
This fixes issue: https://redmine.pfsense.org/issues/6856


rbgarga:
Removing Changes requested label since there are no requested changes commented here

rbgarga:
Already fixed in https://redmine.pfsense.org/issues/7010

PR #3198, opened 2016-10-20
Fix #6864 Improve help for IP address format

phil-davis:
This is the first thing that is easily done. The 'title' text is appended after "Please match the requested format: ", so this will give the user some idea what the input format problem is. The 'title' text also appears when hovering over the field.

The next nice step would be to automatically convert to lowercase as characters are typed or pasted into the field.


phil-davis:
See #3199 for a (hopefully) improved version, that includes these "help" enhancements and also converts input automatically to lowercase.

PR #3199, opened 2016-10-20
Fix #6864 automatically convert IPv6 input to lowercase

phil-davis:
1) As the user leaves the field, or presses Save, onChange will fire and
convert the input string to lowercase. This saves the user havng to even
think about it.
2) Provide some extra text that describes the expected format, to avoid
them getting just "Please match the requested format"

### DIFF
--- a/src/usr/local/www/classes/Form/IpAddress.class.php
+++ b/src/usr/local/www/classes/Form/IpAddress.class.php
@@ -31,14 +31,19 @@ class Form_IpAddress extends Form_Input
 		switch ($type) {
 			case "BOTH":
 				$this->_attributes['pattern'] = '[a-f0-9:.]*';
+				$this->_attributes['title'] = 'An IPv4 address like 1.2.3.4 or an IPv6 address like 1:2a:3b:ffff::1 (use lowercase)';
+				$this->_attributes['onChange'] = 'javascript:this.value=this.value.toLowerCase();';
 				break;
 
 			case "V4":
 				$this->_attributes['pattern'] = '[0-9.]*';
+				$this->_attributes['title'] = 'An IPv4 address like 1.2.3.4';
 				break;
 
 			case "V6":
 				$this->_attributes['pattern'] = '[a-f0-9:]*';
+				$this->_attributes['title'] = 'An IPv6 address like 1:2a:3b:ffff::1 (use lowercase)';
+				$this->_attributes['onChange'] = 'javascript:this.value=this.value.toLowerCase();';
 				break;
 		}
 	}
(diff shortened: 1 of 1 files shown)


loos-br:
phil-davis, just remove the 'use lowercase' text now that it is not necessary anymore and it is ready to go.
thanks!

NOYB:
Other than the specified input pattern and RFC "recommendation", not a requirement.  What is the requirement for the IPv6 address to be lower cased?  If it is not case sensitive then it should be maintained and saved as entered by the user.

phil-davis:
Complying with the RFC recommendation means that configs will have lowercase IPv6 addresses, and those will pass along naturally to whatever conf file(s) they end up in, avoiding any issues with daemons/utilities that might not like uppercase in IPv6 addresses. A few months ago there was an example of some end consumer of the address (I don't remember the detail now, I guess it might be linked to the change that specified lowercase input pattern) that did not like uppercase.
@NOYB yes, we could let people input whatever upper/lowercase mix they like, then ensure that wherever an IPv6 address is read from the config, it is made lowercase before doing anything else with it. But guess it seemed easy and more consistent to (also) enforce lowercase at the point of data entry.

loos-br:
@NOYB, in a decent programming language the IPv6 address would be stored in its binary form and always represented as recommended by RFC, but php is hard...

This should be part of user data sanitizing (instead of discarding and/or refusing to accept such addresses).

NOYB:
That's just it.  It's an RFC recommendation.  Not a requirement.  It is supposed to be case-insensitive.  So store it that way.  And covert it in cases were it is incorrectly required to be case sensitive.

Build pfSense to follow specifications.

phil-davis:
Well, I am not setting policy here, just enhancing the UI input to make it easier for the user to comply with the UI validation that already exists.
If someone wants to set some different policy then sure, go ahead and allow mixed-case IPv6 addresses in the UI/config and then add back-end code to lowercase (and compress and...) IPv6 addresses as they are about to be used "for real".

NOYB:
Which facilitate it continuing to be done out of spec. rather than correcting it.

loos-br,
That's what I said.  It should not refuse to accept mixed case as it is valid.

jim-p:
While the RFC itself is a "recommendation", it is a standards track document and the verbage says "MUST" when referring to representing the addresses as lowercase. And why not just store and use them as lowercase? What is the advantage to allowing the addresses to be mixed case in the config and converting later? There is no advantage to allowing them to be stored that way, except to be lazy with input validation, which isn't an advantage.

Being consistent is best, and as long as we're being consistent, we may as well follow the RFC and maintain lower case, accepting upper/mixed is OK so long as it's stored and used as lower.

NOYB:
At least accepting upper/mixed is the biggest part of it.  Sometimes it may be useful/convenient to cut and paste addresses from other sources where it may be presented in upper/mix.

I prefer maintaining case as well, but that is secondary to input acceptance/

PR #3201, opened 2016-10-20
Improve IPv4 address validation for services_dhcp

phil-davis:
The input pattern that goes with Form_IpAddress by default allows for IPv4 and IPv6 valid characters. The back-end validation here is checking for IPv4 addresses, so it seems reasonable that the front-end input pattern checks might as well be restricted to the IPv4 valid characters. Unneeded setPattern have also been removed.
The exception is Gateway, where the special string "none" is allowed, so I just left the setPattern as it is (yes, it could be modified to only allow the letters "n" "o" and "e").

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -844,13 +844,15 @@ $group = new Form_Group('Range');
 $group->add(new Form_IpAddress(
 	'range_from',
 	null,
-	$pconfig['range_from']
+	$pconfig['range_from'],
+	'V4'
 ))->setHelp('From');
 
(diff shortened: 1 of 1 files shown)


phil-davis:
If the other proposed change to enhance the input pattern "help" text for Form_IpAddress goes ahead, then this sort of thing will also be needed, otherwise the "help" will be misleading, making users think they can put in an IPv6 address.
In any case, this front-end input pattern validation might as well be a bit tighter than it was.
If this seems "a good thing" then I will look at the other uses of Form_IpAddress and see where they need to be made specific to V4 or V6.

PR #3202, opened 2016-10-20
For DHCPv6, ddnsdomainprimary should be V6

phil-davis:
Actually, I don't know about this, but it looks odd. the DHCPv6 code is insisting that ddnsdomainprimary be an IPv4 address.
Is that really correct?
Or must it be an IPv6 address?
Or can it be either?

Please close if the existing code is actually good!


rbgarga:
This is correct the way it is. It's the upstream DNS server to be updated, and services.inc (https://github.com/pfsense/pfsense/blob/master/src/etc/inc/services.inc#L1208) is using only 'primary' configuration knob. It should be changed to use primary for v4 and primary6 for v6 if it's allowed.

PR #3206, opened 2016-10-24
Fix #6874 - dyndns: add header processing in curl

filnko:
some dyndns implementations rely on the correct HTTP header being set. the information was lost and now fixed.


filnko:
Related to this bug:
https://redmine.pfsense.org/issues/6874

MTecknology:
This patch worked for my testing. :+1:

rbgarga:
Merged manually. Thanks!

PR #3207, opened 2016-10-25
Minor changes to README

nrmodesto:


jim-p:
Changes seem OK but I'm not certain they're necessary or worth making at this time. The whole chunk of text should probably be reviewed internally from a marketing perspective.

PR #3209, opened 2016-10-27
DHCPv6 server range from-to check

phil-davis:
When track6 is on, the particular IPv6 address of the interface is not known (and could change from time to time). So $ifcfgip is set to "::", and $ifcfgsn to 64.
Unfortunately gen_subnetv6("::", 64) returns blank "". That no longer passes the checks done by recent changes to is_inrange_v6().
To fix this we can just explicitly set
```
$subnet_start = "0::";
```
for this case.

Alternatively we could add code to gen_subnetv6() to handle the special case of being passed "::" as the IP address.
Actually the underlying issue is that the Net_IPv6 compress() function seems to be compressing "0:0:0:0:0:0:0:0" to "", when I would hope it would compress to "0::" or "::0". But that requires some upstream changes, and sorting out which of "0::" and "::0" is the correct representation.

### DIFF
--- a/src/usr/local/www/services_dhcpv6.php
+++ b/src/usr/local/www/services_dhcpv6.php
@@ -340,10 +340,15 @@ if (isset($_POST['apply'])) {
 		}
 		if (!$input_errors) {
 			/* make sure the range lies within the current subnet */
-			$subnet_start = gen_subnetv6($ifcfgip, $ifcfgsn);
-			$subnet_end = gen_subnetv6_max($ifcfgip, $ifcfgsn);
-
 			if (is_ipaddrv6($ifcfgip)) {
+				if ($ifcfgip == "::") {
+					$subnet_start = "0::";
+				} else {
+					$subnet_start = gen_subnetv6($ifcfgip, $ifcfgsn);
+				}
+
+				$subnet_end = gen_subnetv6_max($ifcfgip, $ifcfgsn);
+
 				if ((!is_inrange_v6($_POST['range_from'], $subnet_start, $subnet_end)) ||
 				    (!is_inrange_v6($_POST['range_to'], $subnet_start, $subnet_end))) {
 					$input_errors[] = gettext("The specified range lies outside of the current subnet.");


phil-davis:
Ultimately it would be nicer to fix Net_IPv6:compress()
The needed code diff is this sample commit in a repo of mine that has a branch of the code that I made the same as what is currently in pfSense (from the Net_IPv6 port), and then did a commit with the fixes:
https://github.com/phil-davis/Net_IPv6/commit/638b96a253164b65c63825c38e79812b6c5f448d

That needs to be submitted upstream. But the upstream is SVN and needs tests added and the tests need to be run in an environment that has PHPunit and... If it was on GitHub and used Travis-CI to automatically do the tests etc, then I would happily do the work, because it would only take a few minutes. i.e. if an Open Source project has not moved to GitHub by now, then volunteers do not contribute to it.

phil-davis:
Another way to fixup this across the code is to provide a compress_ip6() wrapper function which handles the "0:0:0:0:0:0:0:0" case and passes others through to Net_IPv6:compress(). That will fixup any other place where a zero IPv6 address might be used, without breaking anything.
https://github.com/pfsense/pfsense/pull/3210

rbgarga:
Hi @phil-davis. I've imported your patch to pear-Net_IPv6 port as you can see at: 

https://svnweb.freebsd.org/ports?view=revision&revision=424752

and I'll make this new version available on pfSense ASAP. Do you have plans to submit it upstream any time?

rbgarga:
I've also pushed the fix to IPv6.inc on RELENG_2_3 and RELENG_2_3_2

phil-davis:
I raised a bug report https://pear.php.net/bugs/bug.php?id=21139
Hopefully someone there can easily verify the problem and fix, and apply it.

rbgarga:
Great. Thanks! I've cherry-picked FreeBSD ports commit to pfSense/FreeBSD-ports. Now I believe this PR and also #3210 can be closed, right?

phil-davis:
Yes
Closed as the proper fix is being made to the Net_IPv6 code.

PR #3210, opened 2016-10-27
Provide compress_ip6 wrapper

phil-davis:
To handle special case where
```
Net_IPv6::compress("0:0:0:0:0:0:0:0");
```
returns blank "", when it should return "::".

Change all Net_IPv6::compress() calls to use compress_ip6()


phil-davis:
Closed as the proper fix is being made to the Net_IPv6 code.

PR #3211, opened 2016-10-29
syslogd, create configured logsocket directories

PiBa-NL:
This makes sure that logsockets for packages are made also when package hasn't started yet.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1018,11 +1018,6 @@ EOD;
 		}
 		unset($syslogconf);
 
-		// Ensure that the log directory exists
-		if (!is_dir("{$g['dhcpd_chroot_path']}/var/run")) {
-			exec("/bin/mkdir -p {$g['dhcpd_chroot_path']}/var/run");
-		}
-
 		$sourceip = "";
 		if (!empty($syslogcfg['sourceip'])) {
 			if ($syslogcfg['ipproto'] == "ipv6") {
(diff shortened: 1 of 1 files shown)


src/etc/inc/system.inc
```diff
@@ -1049,14 +1044,19 @@ EOD;
...
 			}
 		}
 	}
+	
 	$syslogd_sockets = "";
 	foreach ($log_sockets as $log_socket) {
+		// Ensure that the log directory exists
+		$logpath = dirname($log_socket);
+		if (!is_dir($logpath)) {
+			exec("/bin/mkdir -p {$logpath}");
```
rbgarga: Please replace exec() call by mkdir()

src/etc/inc/system.inc
```diff
@@ -1049,14 +1044,19 @@ EOD;
...
 			}
 		}
 	}
+	
 	$syslogd_sockets = "";
 	foreach ($log_sockets as $log_socket) {
+		// Ensure that the log directory exists
+		$logpath = dirname($log_socket);
+		if (!is_dir($logpath)) {
+			mkdir($logpath, 0777, true);
```
PiBa-NL: I put it because if not specified it would be the default value for mkdir. But to make sure if intermediate paths are missing they will be created i thought to specify the recursive flag.. Its just that mode parameter sits in between..
`bool mkdir ( string $pathname [, int $mode = 0777 [, bool $recursive = false [, resource $context ]]] )`

However i will switch to safe_mkdir() which will do it all at once..

PR #3214, opened 2016-10-31
rules, allow rules to be enabled/disabled upon being XMLRPC synced to the backup node

PiBa-NL:
rules, allow rules to be enabled/disabled upon being XMLRPC synced to the backup node

### DIFF
--- a/src/etc/rc.filter_synchronize
+++ b/src/etc/rc.filter_synchronize
@@ -143,6 +143,12 @@ function carp_sync_xml($sections) {
 			if (isset ($config_copy['filter']['rule'][$x]['nosync'])) {
 				unset ($config_copy['filter']['rule'][$x]);
 			}
+			if ($config_copy['filter']['rule'][$x]['xmlrpcmodification'] == "enable") {
+				unset($config_copy['filter']['rule'][$x]['disabled']);
+			}
+			if ($config_copy['filter']['rule'][$x]['xmlrpcmodification'] == "disable") {
+				$config_copy['filter']['rule'][$x]['disabled'] = true;
+			}
 		}
 	}
 	if (is_array($config_copy['aliases']['alias'])) {
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -68,6 +68,7 @@ function is_aoadv_used($rule_config) {
 	    ($rule_config['nopfsync']) ||
 	    (($rule_config['statetype'] != "") && ($rule_config['statetype'] != "keep state")) ||
 	    ($rule_config['nosync']) ||
+	    ($rule_config['xmlrpcmodification']) ||
 	    ($rule_config['vlanprio'] != "") ||
 	    ($rule_config['vlanprioset'] != "") ||
 	    ($rule_config['sched'] != "") ||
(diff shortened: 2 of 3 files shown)


rbgarga:
@PiBa-NL Technically this patch looks OK, I just would like to understand what kind of problem you are tryng to solve with it. Could you give some examples about use cases?

PiBa-NL:
For me personally i found that for 2 pfSense instances running on ESXi i would have trouble connecting 1 particular ipsec vpn client running inside a W7 machine on the same ESXi machine as the backup box their would arrive duplicated traffic. As apparently the isp side kinda 'broadcasts' packets to both pfsense boxes (even though 1 is properly set as master by carp statuses). It would prevent me from connecting this vpn fortinet client to a remote fortigate box. When adding a floating rule to block this traffic from passing through the backup pfsense machine it started working..  I realize there might be a underlying issue here that should be solved instead but i could not put my finger on sofar.. Also there is a different option present already in pfSense to 'not sync' rules to the backup machine.. Might usable in a similar way by not passing all outbound traffic and then blocking some.. But only selectively passing traffic to the internet, and then passing some more on the master which isnt synced to backup, but i found that a difficult reasoning.. 

And i did think about perhaps replacing that 'dont sync' feature.. Would that be better?

A second usecase would be if for whatever reason the master/slave roles should be reversed or master dies, its nice to have a more complete set of rules on the backup machine to maybe promote to master and only needing to check the enabled and sync status of rules that are then present but inactive. Instead of 'missing'.

rbgarga:
I'm not sure about the risks of such feature. I'll wait for some comments from @jim-p too

jim-p:
It seems like a hammer in search of a nail. The original case you described would have problems when running from the backup node, wouldn't it? If there is a floating rule to block the traffic out there, it would still block when it's the master, so it wasn't actually solving the problem just pushing it off to fail at a different time.

For other potential use cases of this, rather than enable/disable for rules, if a rule need to be specific to a node you could qualify them with more specific sources/destinations that wouldn't match on the other node. It's best to keep the rules and other settings as identical as possible on both nodes. Nothing special needs to happen with rules when switching roles when the rules are identical.

PiBa-NL:
Then if i may ask, what is your purpose for the option "No XMLRPC Sync" ?

jim-p:
An option that's outlived its usefulness. :-) I haven't seen it used for any positive effect in a number of years, but it's still there since there may be someone out there relying on it or has an upgraded configuration with it.

PiBa-NL:
Found my issue. i had a 'somewhat' configured 'QinQ' on my LAN interface. Which apparently also sets a flag PPROMISC on the interface. (note the double 'P') Not exactly sure what either of those do. But they effectively did 'duplicate' traffic leading to multiple 'isakmp: phase 1 I agg' packets with in response multiple 'isakmp: phase 1 R agg' packets making the client not connect.. Anyhow removing the QinQ and rebooting pfSense afterwards seemingly fixed my original issue the proper way.. (without reboot the flag stayed in place..)

I'm not sure how QinQ is supposed to be used.. But if its never in combination with 'regular' lan traffic then i think my config was just plain bad..

So if you guys are convinced no one should ever use this option, im oké with closing the request ;) Thanks for your time anyhow :)

PR #3222, opened 2016-11-10
Within a dhcp pool, emit option names in dhcpd.conf.

zxvv:
I would like to use the DHCP server to configure client's MTU on a subnet.

Currently, when one sets option 26 = 9000 in the GUI, the generated
/var/dhcpd/etc/dhcpd.conf will contain something like:

subnet 172.16.4.0 netmask 255.255.255.0 {
        option custom-opt6-0 9000;
...
whereas the desired configuration (I believe) would be:

        option interface-mtu 9000;

For codes that are defined for server responses, according to [1], 
this proposed code changes services.inc to emit the option names.
There is a forum question about this, but no discussion ensued [2]. 

[1] http://www.ipamworldwide.com/ipam/isc-dhcpv4-options.html
[2] https://forum.pfsense.org/index.php?topic=116395.0

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -431,6 +431,90 @@ function services_dhcpd_configure($family = "all", $blacklist = array()) {
 	unlock($dhcpdconfigurelck);
 }
 
+/* see: http://www.ipamworldwide.com/ipam/isc-dhcpv4-options.html */
+$dhcpoptionsname[1] = "subnet-mask";
+$dhcpoptionsname[2] = "time-offset";
+$dhcpoptionsname[3] = "routers";
+$dhcpoptionsname[4] = "time-servers";
+$dhcpoptionsname[5] = "ien116-name-servers";
+$dhcpoptionsname[6] = "domain-name-servers";
+$dhcpoptionsname[7] = "log-servers";
+$dhcpoptionsname[8] = "cookie-servers";
+$dhcpoptionsname[9] = "lpr-servers";
+$dhcpoptionsname[10] = "impress-servers";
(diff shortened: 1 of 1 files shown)


src/etc/inc/services.inc
```diff
@@ -529,7 +613,11 @@ function services_dhcpdv4_configure() {
 						} else {
 							$itemtype = "text";
 						}
-						$custoptions .= "option custom-{$dhcpif}-{$poolidx}-{$itemidx} code {$item['number']} = {$itemtype};\n";
+                                                $optionname = "custom-{$dhcpif}-{$poolidx}-{$itemidx}";
+                                                if (array_key_exists($item['type'], $dhcpoptionsname)) {
+                                                	$optionname = $dhcpoptionsname[$item['type']];
+                                                }
+						$custoptions .= "option $optionname code {$item['number']} = {$itemtype};\n";
```
rbgarga: This change is not correct. This is the place where custom options are defined and not where value is attributed. Then in this place you should just skip $custoptions definition if it's a known item. Other than that, you should note there is a similar block above at line 598 that must be changed as well.

The block where value is attributed you will find at line 912 and this is where you will use values defined in $dhcpoptionsname.

As a general note, please keep indent rules using TABs with size of 8

zxvv: Thank you so much for reviewing and pointing me in the right direction.   Any preferences or direction about PHP implementation is very welcome.

I'd like to also look at letting the user choose options by name, as you suggested, but if it's OK I'll start by trying to correct the proposed code.  Again, thanks!

sbeaver-netgate:
No action in seven months

PR #3225, opened 2016-11-13
Fix #6918 Allow aliases with capital letters in rules

phil-davis:
Expand the types of Form_IpAddress so that the caller can specify
exactly what combination of IPv4, IPv6 address and alias is allowed for
the field.
Set the appropriate input pattern and hover help text.
Only toLowercase() the entered value if it has a ":" in it - i.e. it
looks like it is intended to be an IPv6 address (rather than an IPv4 or
an alias name).

### DIFF
--- a/src/usr/local/www/classes/Form/IpAddress.class.php
+++ b/src/usr/local/www/classes/Form/IpAddress.class.php
@@ -45,6 +45,23 @@ class Form_IpAddress extends Form_Input
 				$this->_attributes['title'] = 'An IPv6 address like 1:2a:3b:ffff::1';
 				$this->_attributes['onChange'] = 'javascript:this.value=this.value.toLowerCase();';
 				break;
+
+			case "ALIASV4V6":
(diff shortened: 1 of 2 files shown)


phil-davis:
This provides Form_IpAddress with the ability to know and handle the cases where a combination of IPv4 address, IPv6 address and/or alias is allowed. I added the extra cases for ALIASV4 and ALIASV6 so they are available for anything that needs them.
Form_IpAddress was being somewhat misused by the existing code, in that a setPattern was being added on to it to allow it to accept a wider range of input, and of course the other functionality in Form_IpAddress was not able to realize that.
With this change, there is no need to massage anything like that.
There may be other places where Form_IpAddress is being used but modified by setPattern - I will have a look for any of those also...

phil-davis:
Note: code in various other places that uses Form_IpAddress and allows entry of an IP address or alias will need the change to cal Form_IpAddress specifying ALIASV4V6 ...
I will look at that today.

PR #3227, opened 2016-11-15
An alias or IPv4 address cannot contain a colon

phil-davis:
I did a bit too much cut-paste without edit yesterday. At the moment, the ALIASV4 case is not being used, so it is easy to correct it without causing any change to the functional system.


phil-davis:
This is superseded by things getting reverted and #3242 which implements the new cases correctly in the first place.

PR #3228, opened 2016-11-15
IPv6 address can contain a dot

phil-davis:
When requiring the entry of an IPv6 address, the regex pattern should still allow a dot, so that an IPv6 address can be entered in the format that has an IPv4-address-like part at the end:
a:b:c:d:e:f:1.2.3.4
which is a valid way to choose to specify an IPv6 address.


rbgarga:
Hi @phil-davis, can you please rebase your fork?

phil-davis:
Closing in favor of #3241 
I am just online at the moment - easier to make the 1-char edit again and make a new pull request all from GitHub web interface.

PR #3229, opened 2016-11-15
Add cases to Form_IpAddress for host and IP

phil-davis:
combinations, so we can specify that entry of a particular IP address family or host name is required, and give the user an appropriate bit of hover text and regex pattern match.
Make first use of it for input of NTP servers.


phil-davis:
Once the ALIAS cases are finally sorted out in #3242 then this one will have conflicts anyway, after that I will sort it out and also remove the bits of JavaScript lowercase() stuff. Then it can be looked at again.

rbgarga:
Depends of #3242

phil-davis:
Closing in favour of #3248

PR #3232, opened 2016-11-15
Allow aliases where valid for Form_IpAddress

phil-davis:
For any Form_IpAddress calls where we want to accept an alias, we need
to specify the parameter 'ALIASV4V6' 'ALIASV4' or 'ALIASV6' so that the
correct regex pattern and hover text will be used, and that mixed-case
alias names will not be automagically converted to lowercase.


rbgarga:
@phil-davis, Code that added ALIASV4, ALIASV6 and ALIASV4V6 were reverted because the change in the way to deal with lowercase IPv6. Could you submit a new patch to re-add them? After that we can commit this one. Thanks!

phil-davis:
See #3242

rbgarga:
Depends of #3242

phil-davis:
Closing in favour of #3248

PR #3234, opened 2016-11-18
Add Server EKU detection (for IKE v2)

bruno-infotrad:
When using IPsec IKEv2, it is imperative that the server certificate server authentication Extended Key Usage be set. The existing certificate manager does not display this property (the "Server" property is really related to the nsCertType extension - see https://redmine.pfsense.org/issues/6877#change-29471) so I thought it would be useful to add this capability. Feel free to rewrite or modify.


jim-p:
That area is already rather cluttered and 1.3.6.1.5.5.8.2.2 is just as important for a server to have, so if you add 'TLS Web Server Authentication' you also have to look for 1.3.6.1.5.5.8.2.2 (iKEIntermediate). That whole concept needs redesigned. Probably adding a "Key Usage" column and listing all of the possible values there. That is a lot more work, however.

bruno-infotrad:
I like the idea of adding a column; why do you say this is a lot more work?
I would be willing to help, provided it aligns with the pfSense team.

jim-p:
Compared to adding one field, it's a bit more work in layout, wording, etc. Not impossible, but not trivial like this change would have been.

PR #3236, opened 2016-11-19
openvpn, startup locking sequence to prevent issues around pid file / process management

PiBa-NL:
openvpn, startup locking sequence to prevent issues around pid file / process management

fixes: https://redmine.pfsense.org/issues/6940


rbgarga:
I'm afraid it will slow down boot in cases where many tunnels are configured.

rbgarga:
6940 was marked as duplicate of https://redmine.pfsense.org/issues/6132

PR #3242, opened 2016-11-26
Add Form_IpAddress cases for Alias

phil-davis:
Callers of Form_IpAddress sometimes need to allow the user to enter an IP address (V4, V6 or either) or an Alias.
Add cases here that have the appropriate input regex patterns and message for the hover-text.

### DIFF
--- a/src/usr/local/www/classes/Form/IpAddress.class.php
+++ b/src/usr/local/www/classes/Form/IpAddress.class.php
@@ -36,11 +36,27 @@ class Form_IpAddress extends Form_Input
 
 			case "V4":
 				$this->_attributes['pattern'] = '[0-9.]*';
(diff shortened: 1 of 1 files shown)


rbgarga:
I would like to hear from @sbeaver-netgate about this one. Thoughts?

sbeaver-netgate:
I have several thought on the subject. Will post as soon as I have time, but later this week

rbgarga:
I'll assign it and other that depends of it to you then. Thanks!

phil-davis:
#3245 is a more OO way to do this.
@sbeaver-netgate perhaps you were thinking more like that? Or?

phil-davis:
Closing in favour of #3248

PR #3245, opened 2016-11-30
Add classes for IP V4 and V6 Address Forms

phil-davis:
Instead of having the $type parameter to Form_IpAddress and using a
switch statement, the class hierarchy can be extended to have a Form
class for each IpAddress combination, and put the custom message and
pattern in the specific class.

### DIFF
--- a/src/usr/local/www/classes/Form/IpAddress.class.php
+++ b/src/usr/local/www/classes/Form/IpAddress.class.php
@@ -24,25 +24,12 @@ class Form_IpAddress extends Form_Input
 {
 	protected $_mask;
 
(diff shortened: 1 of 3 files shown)


phil-davis:
This is a more OO way to do this than the parameter plus switch statement way. 
See #3242

phil-davis:
Note: If this is the way to move forward, then this PR also needs to have associated code updated to change anything that calls Form_IpAddress and specifies the $type parameter as "V4" or "V6", to use the relevant Form_IpV4Address or Form_IpV6Address class.

phil-davis:
Closing in favour of #3248

PR #3246, opened 2016-11-30
Add DHCP Server ddnsforcehostname option

overhacked:
I was frustrated recently when I discovered that the hostnames used for DDNS updates by the DHCP Server were coming from the clients' supplied host-name option instead of the configured hostnames for static leases. This is consistent with the DHCPD documentation. From the **dhcpd.conf(5)** man page:

> The DHCP server determines the client's hostname by first looking for a ddns-hostname  configuration  option,  and using that if it is present.  If no such option is present, the server looks for a valid hostname  in the  FQDN option sent by the client.  If one is found, it is used; oth- erwise, if the client sent a host-name option, that  is  used.   Other- wise,  if  there  is a host declaration that applies to the client, the name from that declaration will be used.  If none of these applies, the server will not have a hostname for the client, and will not be able to do a DNS update.

pfSense doesn't implement the ddns-hostname option, so I've added it. It can be turned on at the interface level or individually for each Static Mapping.


pfSense-git-updates:

PR #3248, opened 2016-12-03
Form_IpAddress add types remove patterns

phil-davis:
1) Add alias and host types to Form_IpAddress with the appropriate hover
text.
2) Remove the patterns - the UI of those is not so effective anyway, so
leave the validation of input to the back end.
3) Update uses of Form_IpAddress to use the appropriate Alias or Host
type as needed.
4) Remove explicit setPattern() from those uses of Form_IpAddress.

### DIFF
--- a/src/usr/local/www/classes/Form/IpAddress.class.php
+++ b/src/usr/local/www/classes/Form/IpAddress.class.php
@@ -30,17 +30,38 @@ class Form_IpAddress extends Form_Input
 
 		switch ($type) {
 			case "BOTH":
-				$this->_attributes['pattern'] = '[a-fA-F0-9:.]*';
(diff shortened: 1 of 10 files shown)


phil-davis:
Note: The following files still have setPattern attached to Form_IpAddress:
services_igmpproxy_edit.php
services_dnsmasq_domainoverride_edit
services_unbound_domainoverride_edit

They all have patterns that are a bit special - extra things that can be input in addition to IP address chars. In a sense those patterns are a documentation of the expected valid chars. The setPattern() could be removed and just a comment put in the code to remind future maintainers that these are not just Form_IpAddress, but are a bit more than that.

phil-davis:
@sbeaver-netgate Does this look like a reasonable way to rationalize the Form_IpAddress front-end, in association with your addrtolower() in the back-end?

sbeaver-netgate:
This lookd good Phil. Thank you. I'll be adding the back-end validation early next week.

PR #3249, opened 2016-12-03
Captive portal: add option to include idle time in total session time

plumbeo:
Add an option to choose whether the time spent idle by a user disconnected for exceeding the idle timeout must be included in the total session time sent to the RADIUS server or not.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -814,7 +814,9 @@ function captiveportal_prune_old() {
 			if ($lastact && (($pruning_time - $lastact) >= $uidletimeout)) {
 				$timedout = true;
 				$term_cause = 4; // Idle-Timeout
-				$stop_time = $lastact; // Entry added to comply with WISPr
+				if (!isset($config['captiveportal'][$cpzone]['includeidletime'])) {
+					$stop_time = $lastact;
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


plumbeo:
As far as I can see the RADIUS RFC doesn't specify what's the correct behavior and both alternatives make sense, maybe the current behavior is a bit more favorable for the users because the time between when they stop using the service and when they're disconnected doesn't count against their daily usage time. 

I added the option because the gateway I plan to replace with pfsense doesn't discount the idle time and I wanted to keep the accounting consistent.

tuyan:
Wouldn't this be the responsibility of the radius server to also include an extra time when calculating the session length for the next session of the same user ?

plumbeo:
It could but Freeradius, at a minimum, records and uses the Session-Time value it gets from the NAS.

The RFC just says "this attribute indicates how many seconds the user has received service for," so should the idle time be included? Arguably the user received service during that time, even if she (not having explicitly disconnected from the network) elected not to use it, but not counting it makes sense too, and the comment mentioned a Wispr specification I couldn't find.

Given that at least one vendor (HP, former Colubris) decided not to include it, I suppose making it an option can't hurt.

rbgarga:
I don't see a problem to have this option. My only concern is it's changing the default behavior leaving it disabled by default. Please change it to on by default

plumbeo:
@rbgarga hi, the default behaviour is not to include the idle time and the patch changes this when the option is enabled, so shouldn't it stay like it is?

rbgarga:
@plumbeo I just need more coffee, sorry

PR #3250, opened 2016-12-04
Do not truncate IPv6 address in NTP widget (Bug #4815)

doktornotor:


rbgarga:
Merged with wrong PR # in commit log, sorry.

PR #3251, opened 2016-12-06
Backport Cloudflare and Gratis plus passwords in base64 DynDNS changes

phil-davis:
Note: corresponding change to upgrade_config.inc to come in master to
correctly implement the upgrade_155_to_156 code, that is master only and
will become upgrade_158_to_159

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -108,6 +108,7 @@ if (isset($id) && isset($a_dyndns[$id])) {
 }
 
 if ($_POST) {
+	global $dyndns_split_domain_types;
 	unset($input_errors);
 	$pconfig = $_POST;
 
@@ -128,7 +129,7 @@ if ($_POST) {
 		$reqdfieldsn[] = gettext("Password");
 		$reqdfields[] = "username";
 		$reqdfieldsn[] = gettext("Username");
-		if ($pconfig['type'] == "namecheap") {
+		if (in_array($pconfig['type'], $dyndns_split_domain_types)) {
 			$reqdfields[] = "domainname";
 			$reqdfieldsn[] = gettext("Domain name");
 		}
@@ -181,7 +182,7 @@ if ($_POST) {
 		$dyndns['type'] = $_POST['type'];
 		$dyndns['username'] = $_POST['username'];
 		if ($_POST['passwordfld'] != DMYPWD) {
-			$dyndns['password'] = $_POST['passwordfld'];
+			$dyndns['password'] = base64_encode($_POST['passwordfld']);
 		} else {
 			$dyndns['password'] = $a_dyndns[$id]['password'];;
 		}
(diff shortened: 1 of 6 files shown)


phil-davis:
See forum https://forum.pfsense.org/index.php?topic=122216.0

phil-davis:
See #3252 for the change to upgrade_config in master that fixes another issue in master, and will also help anyone who gets to 15.8 in 2.3.3-DEVELOPMENT and then later upgrades to 2.4 (master).

jim-p:
I'll have to look at this closer tomorrow but I am not fond of the idea of mixing up the upgrade code in this way.

phil-davis:
The upgrade path needs to be "single-threaded", so things in master that do not get back-ported end up having to be always made later upgrade steps than the things that have been back-ported to an older release.
In this case it works out happily, because the glxsb upgrade code needs a fix anyway.

src/usr/local/www/services_dyndns_edit.php
```diff
@@ -491,6 +492,9 @@ function setVisible(service) {
 				hideInput('ttl', false);
 				break;
 			case "namecheap":
+			case "cloudflare-v6":
+			case "cloudflare":
```
CarlGill: Can we Remove this.   There already are lines controlling Cloudflare, and by adding these here the other lines don't get read in.

If this was intentional than can we remove the other cloudflare stanza?

jim-p:
Though the intention was not to duplicate, what fields specifically do you not see that should be active for cloudflare? We can certainly remove the extraneous references but it doesn't look like there would actually be a problem the way it is now.

The domain name field is supposed to be there now. The current code should show the "proxied" block when it wouldn't have otherwise. I don't see any fields that should be there which are hidden. Which fields are not showing that you think should be there?

src/usr/local/www/services_dyndns_edit.php
phil-davis: I don't understand what you mean?
This switch statement only mentions cloudflare here, not in any other cases. In fact "cloudflare" is not specifically mentioned in this file other than here and in help text.

jim-p: The block just under this also has cloudflare inside:

https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/services_dyndns_edit.php#L472

https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/services_dyndns_edit.php#L489

phil-davis: I see - the code in this commit did not have that (I was clicking the View button from here) but in master and RELENG_2_3 the other case got added by a different commit. I guess the necessary merge resolution missed the mark a bit.

phil-davis: I suspect something like https://github.com/pfsense/pfsense/pull/3717 is what is needed to integrate the 2 changes that have caused this. But I don't have Cloudflare, so @CarlGill or someone will need to confirm.
The change should allow the "proxied" field to appear in the cloudflare case.

CarlGill:
@jim-p  So I can confirm the "proxied" value is no longer needed with the latest API.  And We could safely remove the code added to handle the previous issues related to the Proxy value.

Previously, if you sent an update call to the cloudflare API with the Proxied field being blank, it would default False on the cloudflare side, and remove the proxied attribute from the dynamic DNS record.  This seems to be fixed to a more sane default of "What ever is already there".

Also, after reviewing the Case Defaults, it looks like we can safely remove the duplicate stanzas I was worried about as the only different was the Proxied value.

I will submit a PR later to remove the unused code.

BTW, is there an effort to remove Trolls from the IRC room?

src/usr/local/www/services_dyndns_edit.php
CarlGill: Copying Comment from parent thread.

My changes are no longer needed following Cloudflare API changes.  The pfSense proxy field was a needed workaround due to how Cloudflare handled the lack of the Proxied parameter in the API call.  API changes have corrected their default behavior.

I can confirm everything works as intended with the latest pfSense release.  Only remaining changes are cleaning up unused Proxy code, and some helpful Documentation and Tips on the page.

PR #3252, opened 2016-12-06
Fix unset glxsb

phil-davis:
This bit of upgrade_config code needs a global reference to $config


phil-davis:
Note: This also helps for the backported changes in #3251 that implement 15.7 and 15.8 in RELENG_2_3. People can go to 15.8 on 2.3.3-DEVELOPENT, then when they later upgrade to 2.4 (master) they will correctly also get this unset glxsb code run as 15.9

PR #3255, opened 2016-12-06
Allow NAT 1:1 destination self on edit

phil-davis:
1) Add a NAT 1:1 rule with destination "This Firewall (self)" and save (goodness knows why you would want to do this?)
2) Edit the rule
The Destination dropdown displays "any" rather than the special net that was selected for the rule.

This fixes the UI problem.
Maybe "This Firewall (self)" actually should not be in the Destination dropdown list?


rbgarga:
The correct fix is to remove self from destination drop-down. Are you planing to work on this? If not let me know. Thanks!

phil-davis:
See #3260

PR #3257, opened 2016-12-06
Fix #6224 NAT edit - preserve user selections when input errors

phil-davis:
1) Edit a NAT Port Forward rule, change the destination type to "Network", but do not input any network address/mask.
2) Press Save, an input error is reported telling that the network destination address/mask is required - good.
    However, the destination type is no longer "Network" - it goes to "xxx address" where "xxx" is the selected interface.

The problem is that JS dst_change() is being called on every page load. That is good when starting a new entry (the Destination gets set to WAN address to match the default interface of WAN), or when starting to edit an existing entry (the code runs without messing anything up). But when data has been $_POSTed and there are input errors, we just want to redisplay whatever the user had selected, not go trying to fix up correlations between fields.

### DIFF
--- a/src/usr/local/www/firewall_nat_edit.php
+++ b/src/usr/local/www/firewall_nat_edit.php
@@ -1236,7 +1236,13 @@ events.push(function() {
 
 	hideSource(!srcenabled);
 	ext_change();
+<?php
+if (!$_POST) {
+?>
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: This fixes the issue, but it does seem a bit of a kludge to me, jumping out of JS to wrap some PHP test around a single line of JS. But I couldn't think of a way to teach dst_change() to understand the case when 'dst' is empty because the user has selected "Single Host" or "Network" and left the address field blank, vs when the field is blank because this is a new entry being added.

PR #3259, opened 2016-12-07
Feature #3151 Disable gateway monitoring actions

phil-davis:
without disabling gateway monitoring.

This allows the user to continue to monitor the gateway with dpinger, so
they can see how it is performing, but for the system not to take any
real action if the latency/loss exceeds the given limits.

A typical use case for this would be on a single-WAN system. There is no
failover option, so there is no point taking any real action when the
latency/loss is high. Having stuff try to failover (and stop/start
stuff...) is just disruptive.

In this case the user could have disabled monitoring completely, but then
they get no feedback abut gateway performance.

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -115,7 +115,9 @@ function start_dpinger($gateway) {
 	$params .= "-B {$gateway['gwifip']} ";	/* Bind src address */
 	$params .= "-p {$pidfile} ";		/* PID filename */
 	$params .= "-u {$socket} ";		/* Status Socket */
-	$params .= "-C \"{$alarm_cmd}\" ";	/* Command to run on alarm */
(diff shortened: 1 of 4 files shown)


phil-davis:
Note: I added the "hghdelay" and "highloss" gateway statuses because it looked misleading when the gateway was reported "down" "Offline" in red on the dashboard or Gateways Status displays when action_disabled. In that case the gateway is just outside the high range of the monitoring parameters, but the system is still trying to use it.

dennypage:
I like it.

franciscodimattia:
On version 2.6.0, even if "Disable Gateway Monitoring Action" is checked, the interface is shown as Offline in the dashboard and a notification is sent to my email saying that the interface was omited from the load balancing group due to packet loss.
Couldn't check if it was really taken down because it is my production router.

PR #3260, opened 2016-12-08
Remove destination self from NAT 1:1

phil-davis:
It is not needed and caused problems if used and was not remembered on
edit anyway.


phil-davis:
This is the replacement for #3255

PR #3261, opened 2016-12-08
Specify IP to set for zoneedit

phil-davis:
Documentation of the zoneedit dynamic DNS change API:
https://support.zoneedit.com/Knowledgebase/Article/View/4/1/changes-to-dynamic-dns

Format is:
https://dynamic.zoneedit.com/auth/dynamic.html?host=<your_hostname>&dnsto=<your_ip>

The dyndns.class code for "zoneedit" does not specify "dnsto" at all. So I guess Zoneedit will be assuming the IP address that the request came from, which will be whatever is the current default gateway of pfSense - which probably works OK for ordinary failover of a primary to secondary, but not when the user wants the dynamic DNS name to point to something else.

This is a possible fix for Redmine 6992. It needs testing by someone with the relevant configuration and Zoneedit account, e.g. the OP of the issue.
https://redmine.pfsense.org/issues/6992

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -485,7 +485,7 @@
 					if ($this->_dnsPort) {
 						$port = ":" . $this->_dnsPort;
 					}
-					curl_setopt($ch, CURLOPT_URL, "{$server}{$port}?host=" .$this->_dnsHost);
+					curl_setopt($ch, CURLOPT_URL, "{$server}{$port}?host=" . $this->_dnsHost . '&dnsto=' . $this->_dnsIP);
 					break;
 				case 'dyns':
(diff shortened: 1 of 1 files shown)


phil-davis:
@rbgarga Did you manage to test it?
I actually do not know if it really works, I just followed the Zoneedit FAQ doc.

rbgarga:
No, I don't have an account to validate it, what is a common situation when we deal to dyndns. But I've checked recommended clients source code and they all use this parameter.

https://sourceforge.net/projects/zoneclient/files/
https://github.com/egiles/dns-zoneedit/blob/master/lib/DNS/ZoneEdit.pm#L71

Now I've updated the ticket and submitter will be able to help us testing it

PR #3262, opened 2016-12-09
Added dhcp6c no-release and user configurable DUID

marjohn56:
dhcp6c no-release requires modified WIDE-DHCP6.

Configurable DUID creates dhcp6_duid and stores it in /conf. On dhcp6
configure it copies that file, if it exists, to var/db. This also
corrects a problem with the duid being lost if using a RAM disk.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3865,6 +3865,10 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		return;
 	}
 
+	if (file_exists("/conf/dhcp6c_duid")) { // Using user defined DUID.
+		exec("cp -f /conf/dhcp6c_duid /var/db/");
+	}
+	
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
 
@@ -3974,10 +3978,11 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$rtsoldscript .= "/bin/sleep 1\n";
 	}
 	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
+	$noreleaseOption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
 
 	/* add the start of dhcp6c to the rtsold script if we are going to wait for ra */
 	if (!isset($wancfg['dhcp6withoutra'])) {
-		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption} -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
+		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption} {$noreleaseOption} -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
 		$rtsoldscript .= "/usr/bin/logger -t rtsold \"Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
 	}
 	/* Add wide-dhcp6c shell script here. Because we can not pass a argument to it. */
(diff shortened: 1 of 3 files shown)


rbgarga:
This PR depends of https://github.com/pfsense/FreeBSD-ports/pull/224

rbgarga:
It would be preferrable if you split these changes in different Pull Requests or, at least, in 2 different commits

marjohn56:
OK, I'll do that ASAP.

src/etc/inc/interfaces.inc
```diff
@@ -3865,6 +3865,10 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		return;
 	}
 
+	if (file_exists("/conf/dhcp6c_duid")) { // Using user defined DUID.
```
rbgarga: You should never hardcode /conf, instead you need to use (from /etc/inc/globals.inc) global var $g['conf_path']

src/etc/inc/interfaces.inc
```diff
@@ -3865,6 +3865,10 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
...
+		exec("cp -f /conf/dhcp6c_duid /var/db/");
```
rbgarga: PHP has a function to copy files, you should use copy() instead of a shell session

src/etc/inc/util.inc
```diff
@@ -1000,7 +1000,21 @@ function is_macaddr($macaddr) {
 
 	return true;
 }
-
```
rbgarga: Please keep blank line separating 2 functions

src/etc/inc/util.inc
```diff
@@ -1000,7 +1000,21 @@ function is_macaddr($macaddr) {
...
+		return false;
+	}
+	for ($i = 0; $i < 16; $i++) {
+		if (ctype_xdigit($values[$i]) == false)
+			return false;
+		if (hexdec($values[$i]) < 0 || hexdec($values[$i]) > 255)
+			return false;
+	}
+	write_dhcp6_duid($dhcp6duid);
+	return true;
```
rbgarga: It always returns true, even if write_dhcp6_duid() fails

src/etc/inc/util.inc
```diff
@@ -2474,5 +2488,21 @@ function get_smart_drive_list() {
 	sort($disk_list);
 	return $disk_list;
 }
+/* Write the DHCP6 DUID file */
```
rbgarga: Please keep blank line separating 2 functions and no blank lines between comment and function prototype

src/etc/inc/util.inc
```diff
@@ -2474,5 +2488,21 @@ function get_smart_drive_list() {
...
 
+function write_dhcp6_duid($dhcp6duid) {
+	// Create the hex array from the dhcp6duid config entry and write to file
+
+	$temp = str_replace(":","",$dhcp6duid);
+	$duid_binstring = pack("H*",$temp);
+	if ($fd = fopen("/conf/dhcp6c_duid", "wB")){
+		fwrite($fd, $duid_binstring);
+		fclose($fd);
+		return 1;
```
rbgarga: Why return 1 here?

src/etc/inc/util.inc
```diff
@@ -2474,5 +2488,21 @@ function get_smart_drive_list() {
...
+	}
+	else {
```
rbgarga: It has a return on if() statement, so it doesn't need to add log_error inside else. if can simply be closed here

src/etc/inc/util.inc
rbgarga: Return true sounds better IMO

src/etc/inc/util.inc
```diff
@@ -2474,5 +2488,21 @@ function get_smart_drive_list() {
...
+		log_error(gettext("Error: cannot open duidfile for writing."));
+		return 1;
```
rbgarga: It should return false in this case

src/usr/local/www/interfaces.php
```diff
@@ -274,7 +274,7 @@ function remove_bad_chars($string) {
 		$pconfig['type6'] = "slaac";
 		break;
 	case "dhcp6":
-		$pconfig['dhcp6-duid'] = $wancfg['dhcp6-duid'];
+		$pconfig['dhcp6duid'] = $wancfg['dhcp6duid'];
```
rbgarga: Why are you changing parameter name here?

rbgarga:
@marjohn56  I did the review anyway, so you can consider these changes when splitting changes

src/etc/inc/interfaces.inc
marjohn56: OK will do.

src/etc/inc/util.inc
marjohn56: Will do.

src/etc/inc/interfaces.inc
marjohn56: OK, will change that. I assumed /conf is always conf and would never change.

rbgarga:
@marjohn56 I was also thinking about it and you don't need to write DUID file to /conf. You already have it on config.xml, so you should write it direct to /var/db when it's necessary

marjohn56:
So at the start of interface_dhcpv6_configure() where I do the copy just call the write_dhcp6_duid() at that point?

rbgarga:
@marjohn56 yes

marjohn56:
OK, I’ll run it up that way and see if I can break it.

rbgarga:
@marjohn56 if you are splitting in 2 different PRs, consider to submit this custom DUID first since it doesn't depend of https://github.com/pfsense/FreeBSD-ports/pull/224

src/usr/local/www/interfaces.php
marjohn56: because one of the functions got upset when trying to pass 'dhcp6-duid', it thinks it's a minus

src/etc/inc/util.inc
marjohn56: Just copying what I'd seen elsewhere, :) 
I'll use true and false where possible in future.

rbgarga:
Closed in favor of #3278

PR #3263, opened 2016-12-09
Do NOT block DHCP traffic with DHCP Relay enabled (Bug #4558, Bug #6996)

doktornotor:
The traffic is getting blocked since $dhcrelaycfg is nowhere defined.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -3397,6 +3397,7 @@ EOD;
 				}
 				/* allow access to DHCP relay on interfaces */
 				if (isset($config['dhcrelay']['enable'])) {
+					$dhcrelaycfg =& $config['dhcrelay'];
 					$dhcifaces = explode(",", $dhcrelaycfg['interface']);
 					foreach ($dhcifaces as $dhcrelayif) {
 						if ($dhcrelayif == $on) {
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -3397,8 +3397,7 @@ EOD;
 				}
 				/* allow access to DHCP relay on interfaces */
 				if (isset($config['dhcrelay']['enable'])) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/filter.inc
```diff
@@ -3397,6 +3397,7 @@ EOD;
 				}
 				/* allow access to DHCP relay on interfaces */
 				if (isset($config['dhcrelay']['enable'])) {
+					$dhcrelaycfg =& $config['dhcrelay'];
 					$dhcifaces = explode(",", $dhcrelaycfg['interface']);
```
rbgarga: I don't see a reason to set a reference to $config['dhcrelay']. IMO this variable is unnecessary and $config['dhcrelay'] can be used direct on explode() call

doktornotor: Well, the only reason is that it's being done everywhere else. And the variable is being completely unnecessary everywhere else as well.

```
$ grep -Rni dhcrelaycfg /etc/inc | grep "dhcrelaycfg ="
/etc/inc/services.inc:1726:     $dhcrelaycfg =& $config['dhcrelay'];
/etc/inc/services.inc:1860:     $dhcrelaycfg =& $config['dhcrelay6'];
/etc/inc/upgrade_config.inc:2556:       $dhcrelaycfg =& $config['dhcrelay'];
```

The scope of this PR is a oneliner fix for an issue I keep hitting for years. Frankly, I do not care much about how's it done but would love to have the thing finally working properly,

rbgarga:
I've pushed a fix. Thanks!

rbgarga:
Oh, sorry, after read your answer I thought you were not doing the change. So I went ahead and did it.

doktornotor:
Well as said, as long at it finally works, it doesn't matter. Thanks. I tested it again with ```$dhcifaces = explode(",", $config['dhcrelay']['interface']);``` and it worked as well, so I hope nothing else breaks here.

PR #3266, opened 2016-12-10
IPv6, allow DHCP6 client to use a prefix size of /59

PiBa-NL:
This seemed to fix the issue of getting weird subnet sizes on the LAN like /63 instead of /64 which prevented slaac from working for a user ( JVWJ ) i talked with on irc.


PiBa-NL:
Ill add it to the services_dhcpv6.php to. @rbgarga & @jim-p , Other question though, should every size between /48 and /64 be possible? Because adding /59 now sounds like we might add /49 tomorrow if any isp chooses to provide that.. Or for the moment just add /59 and add other sizes as we see them happen 'in the world'?

PR #3271, opened 2016-12-13
DHCP server edit when DHCP Relay enabled

phil-davis:
Part of Redmine #6997
1) Display the DHCP Server settings even when DHCP Relay is enabled, but
disable the "enable" checkbox in this case - so the user cannot enable
DHCP Server when DHCP Relay is enabled, but can view and modify the DHCP
Server parameters (ready for the day they want to use it).
2) Actually validate the POSTed DHCP Server parameters, even if it is
disabled. Previously you could put rubbish values in fields and they
would be "happily" saved when DHCP Server was disabled. We want users to
be able to setup DHCP Server ready-to-enable and knowing that they have
put validated stuff in.
3) When DHCP Server is disabled, allow Range From and Range To to be
left empty. When first setting up parameters (e.g. with the UI page up
for the first time) it is OK for the user to be able to just press
"Save". e.g. they are just pressing Save as a way to get out (as if it
is a Cancel button). We do not need to force them to enter range from/to
IP addresses.

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -232,220 +232,226 @@ if (isset($_POST['save'])) {
 	$pconfig['numberoptions'] = $numberoptions;
 
 	/* input validation */
+
+	// Note: if DHCP Server is not enabled, then it is OK to adjust other parameters without specifying range from-to.
 	if ($_POST['enable'] || is_numeric($pool) || $act == "newpool") {
 		$reqdfields = explode(" ", "range_from range_to");
 		$reqdfieldsn = array(gettext("Range begin"), gettext("Range end"));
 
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: GitHub shows loads of diff lines because a long block of code was unindented. Actually there are not so many real changes.

phil-davis:
Note: With DHCP Server disabled, the user can also add/delete/modify pools and static mappings to their hearts content, and that gets validated as they go. Then all 3 pieces - main DHCP Server settings, Pool settings and Static mappings are all valid and ready-to-go, so the user can then check the Enable box (assuming they have disabled DHCP Relay) and Save.

rbgarga:
@phil-davis FYI, you can add '?w=0' at the end of URL and it will show only what have changed skipping space only changes, like diff -w.

https://github.com/pfsense/pfsense/pull/3271/files?w=0

PR #3273, opened 2016-12-13
Status DHCP Leases show CID and other enhancements

phil-davis:
1) If there is a Client Id specified then show it (otherwise it is possible to define a Static Mapping that has only Cliwent Id and all other fields empty, which would look like an empty row in the displayed table).
2) Ignore empty static lease entries - when an interface has no static mappings the "static" section of the DHCP server settings can be "<static></static>" which was resulting in a single-element that was empty being displayed as an empty row.
3) If there are no leases to display, then make a row with the message "No leases to display" so the user knows the system at least tried.
4) If there are no leases in use, then make a row with the message "No leases are in use" so the user knows the system at least tried.

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -275,25 +275,33 @@ if (count($pools) > 0) {
 	asort($pools);
 }
 
+$got_cid = false;
+
 foreach ($config['interfaces'] as $ifname => $ifarr) {
 	if (is_array($config['dhcpd'][$ifname]) &&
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: This also includes the fix in https://github.com/pfsense/pfsense/pull/3272 
Whatever I do, it will conflict with that. So please commit #3272 and I can then sort this out, including any comments/suggestions people may have.

rbgarga:
@phil-davis #3272 was merged, please rebase

phil-davis:
New PR #3277 
That was easier, because this is just an online branch so I am not sure if I can do rebase... online in GitHub yet.

PR #3277, opened 2016-12-13
Status DHCP Leases show CID and other enhancements

phil-davis:
1) If there is a Client Id specified then show it (otherwise it is possible to define a Static Mapping that has only Client Id and all other fields empty, which would look like an empty row in the displayed table).
2) Ignore empty static lease entries - when an interface has no static mappings the "static" section of the DHCP server settings can be "" which was resulting in a single-element that was empty being displayed as an empty row.
3) If there are no leases to display, then make a row with the message "No leases to display" so the user knows the system at least tried.
4) If there are no leases in use, then make a row with the message "No leases are in use" so the user knows the system at least tried.

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -275,25 +275,33 @@ if (count($pools) > 0) {
 	asort($pools);
 }
 
+$got_cid = false;
+
 foreach ($config['interfaces'] as $ifname => $ifarr) {
 	if (is_array($config['dhcpd'][$ifname]) &&
 	    is_array($config['dhcpd'][$ifname]['staticmap'])) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/status_dhcp_leases.php
```diff
@@ -469,6 +502,7 @@ function remove_duplicate($array, $field) {
 			</thead>
 			<tbody>
 <?php
+if (count($dhcp_leases_subnet_counter)) {
```
rbgarga: Please change this if statement to use ':' and 'endif;' instead of brackets to make it easy to read and follow style used in this block

phil-davis:
Style of "if" statement changed to match surrounding code as requested.

PR #3278, opened 2016-12-14
User Configurable DUID

marjohn56:
commit 5e8db21

A DUID may be entered in WAN DHCP6 config. The DUID is checked for
validity. On call interface_dhcpv6_configure() a check is made to see if
there is a duid stored in the configuration, if so then the duid is
written to file var/db/dhcp6c_duid. A direct write to /var/db is used as I could not get path from vardb_path to work.

commit 576b5ea

Some ISP's will release the allocated PD or address on receipt of the
release signal and will then issue a new PD or address on next request.
This is not always wanted. dhcp6c sends this signal by default. A
modification to dhcp6c has the PR #229 in pfsense/FreeBSD-ports. The -n
option is added to the start up of dhcp6c if this option is set.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3870,10 +3870,7 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 
 	if (!empty($wancfg['dhcp6-duid']) {
 	// Write the DUID file
-		$dhcp6duid = $wancfg['dhcp6-duid'];
-		if(!write_dhcp6_duid($wancfg['dhcp6-duid'])) {
-			log_error(gettext("Failed to write user DUID file!"));
-		}
+		write_dhcp6_duid($wancfg['dhcp6-duid']));
 	}
 
 	if ($wancfg['adv_dhcp6_config_file_override']) {
@@ -3982,11 +3979,10 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$rtsoldscript .= "/bin/sleep 1\n";
 	}
 	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
-	$noreleaseOption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
 	
 	/* add the start of dhcp6c to the rtsold script if we are going to wait for ra */
 	if (!isset($wancfg['dhcp6withoutra'])) {
-		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption} {$noreleaseOption} -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
+		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption}-c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
 		$rtsoldscript .= "/usr/bin/logger -t rtsold \"Starting dhcp6 client for interface {$interface}({$wanif})\"\n";
 	}
 	/* Add wide-dhcp6c shell script here. Because we can not pass a argument to it. */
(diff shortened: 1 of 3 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3867,7 +3867,15 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
-
+
```
rbgarga: Remove TAB from blank line

src/etc/inc/interfaces.inc
```diff
@@ -3867,7 +3867,15 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
-
+	
+	if ($wancfg['dhcp6-duid'] != '') {
```
rbgarga: Use if (!empty()) check here to avoid errors when this array key doesn't exist

src/etc/inc/interfaces.inc
```diff
@@ -3867,7 +3867,15 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
...
+	// Write the DUID file
```
rbgarga: Fix indent

src/etc/inc/interfaces.inc
```diff
@@ -3867,7 +3867,15 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
...
+		$dhcp6duid = $wancfg['dhcp6-duid'];
```
rbgarga: There is no need to set $dhcp6duid here, you can use $wancfg['dhcp6-duid'] direct on function call

src/etc/inc/interfaces.inc
```diff
@@ -3974,10 +3982,11 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 		$rtsoldscript .= "/bin/sleep 1\n";
 	}
 	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
-
+	$noreleaseOption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
+	
 	/* add the start of dhcp6c to the rtsold script if we are going to wait for ra */
 	if (!isset($wancfg['dhcp6withoutra'])) {
-		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption} -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
+		$rtsoldscript .= "/usr/local/sbin/dhcp6c {$debugOption} $noreleaseOption -c {$g['varetc_path']}/dhcp6c_{$interface}.conf -p {$g['varrun_path']}/dhcp6c_{$wanif}.pid {$wanif}\n";
```
rbgarga: Add variable inside a { }

src/etc/inc/util.inc
```diff
@@ -2474,5 +2488,19 @@ function get_smart_drive_list() {
...
+	// Create the hex array from the dhcp6duid config entry and write to file
+				
+	$temp = str_replace(":","",$duidfile);
+	$duid_binstring = pack("H*",$temp);
+	if ($fd = fopen("/var/db/dhcp6c_duid", "wb")){
+		fwrite($fd, $duid_binstring);
+		fclose($fd);
+		return true;
+	}
+	else {
```
rbgarga: This else { block can be removed, if code went into if () it will return anyway

src/etc/inc/util.inc
```diff
@@ -2474,5 +2488,19 @@ function get_smart_drive_list() {
...
+		return false;
```
rbgarga: Maybe it would be nice to check if duid passed on parameter is valid adding a is_duid() call inside this function

rbgarga:
Depends of https://github.com/pfsense/FreeBSD-ports/pull/229

marjohn56:
Use if (!empty()) check here to avoid errors when this array key doesn't exist

 

As that call’s not used elsewhere in interfaces.inc I used what is used elsewhere. :)

 

So that becomes if (!empty($wancfg['dhcp6-duid']) ?

 

Regards,

 

Martin

marjohn56:
In src/etc/inc/interfaces.inc <https://github.com/pfsense/pfsense/pull/3278#pullrequestreview-13095448> :

> @@ -3867,7 +3867,15 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 
        $wanif = get_real_interface($interface, "inet6");
        $dhcp6cconf = "";
-
+       
+       if ($wancfg['dhcp6-duid'] != '') {
+       // Write the DUID file
+              $dhcp6duid = $wancfg['dhcp6-duid'];

There is no need to set $dhcp6duid here, you can use $wancfg['dhcp6-duid'] direct on function call

 

If I use if(!write_dhcp6_duid($wancfg['dhcp6-duid']) then it errors.

src/etc/inc/interfaces.inc
marjohn56: That's what happens when you've only had 4 hours sleep! Sorry

marjohn56:
If we add the is_duid() check then the log error for write fail would need to move into this function as a return false could indicate file error or duid error, unless you can think of a better way.

 

Regards,

 

Martin

src/etc/inc/interfaces.inc
marjohn56: dhcp6c is now PR #230

PR #3283, opened 2016-12-16
Add support for chrooted SCP to User Manager

doktornotor:
This makes it possible to assign a chrooted SCP privilege to users, as discussed in Bug #7012. 

NOTE: I'd rather not overcomplicate this given the fact that this is quite a fringe use case, so the chroot setup is left as a manual exercise to admins for now; however there's ```/usr/local/etc/rc.d/scponlyc``` script shipped with the package which should pretty much automate this in case someone needs such setup.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -474,6 +474,8 @@ function local_user_set(& $user) {
 		$user_shell = "/bin/tcsh";
 	} elseif (userHasPrivilege($user, "user-copy-files")) {
 		$user_shell = "/usr/local/bin/scponly";
+	} elseif (userHasPrivilege($user, "user-copy-files-chroot")) {
+		$user_shell = "/usr/local/sbin/scponlyc";
 	} elseif (userHasPrivilege($user, "user-ssh-tunnel")) {
 		$user_shell = "/usr/local/sbin/ssh_tunnel_shell";
 	} elseif (userHasPrivilege($user, "user-ipsec-xauth-dialin")) {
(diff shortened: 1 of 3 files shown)


src/etc/inc/auth.inc
```diff
@@ -474,6 +474,8 @@ function local_user_set(& $user) {
 		$user_shell = "/bin/tcsh";
 	} elseif (userHasPrivilege($user, "user-copy-files")) {
 		$user_shell = "/usr/local/bin/scponly";
+	} elseif (userHasPrivilege($user, "user-copy-files-chroot")) {
```
jim-p: I'm thinking maybe this check should be above the test for user-copy-files, that way if the user somehow has both set, the more secure option of the two is used. That way they could be in a group with the weaker privilege but have the stricter privilege applied.

doktornotor: Well, I don't care either way, certainly can be done... I've been thinking about adding some input validations there for the "conflicts", but that got nowhere since I couldn't figure out where it should be put at all.

jim-p: Validation is probably a lost cause there since it could be groups that conflict. Better to warn them and let it go and, like this, in cases of conflict have a good default behavior between the two. There's a case to be made either way. Group with the strict permission and a user with the looser one to override that way, but given the choice I'd prefer to go with the stricter choice when there is a conflict.

doktornotor: Yep, actually thinking about it the conflicts are somehow legit here (considering the group vs. user privs case).

PR #3285, opened 2016-12-17
Captive portal: add button to disconnect all users

plumbeo:
Add a function to disconnect all logged in users and a button to call it in the captive portal status page.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -347,14 +347,9 @@ EOD;
 		@unlink("{$g['varetc_path']}/captiveportal-{$cpzone}-error.html");
 		@unlink("{$g['varetc_path']}/captiveportal-{$cpzone}-logout.html");
 
-		captiveportal_radius_stop_all();
+		captiveportal_disconnect_all(10); // NAS-Request
 
 		captiveportal_filterdns_configure();
-
-		/* send Accounting-Off to server */
-		if (!platform_booting()) {
-			captiveportal_send_server_accounting(true);
-		}
 
 		/* remove old information */
 		unlink_if_exists("{$g['vardb_path']}/captiveportal{$cpzone}.db");
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -1082,6 +1082,32 @@ function captiveportal_disconnect_client($sessionid, $term_cause = 1, $logoutRea
 	}
 }
 
+/* remove all clients */
+function captiveportal_disconnect_all($term_cause = 6, $logoutReason = "DISCONNECT") {
+	global $g, $config, $cpzone, $cpzoneid;
+
+	$radiussrvs = captiveportal_get_radius_servers();
+	$cpdb = captiveportal_read_db();
+
+	$unsetindexes = array();
+
+	foreach ($cpdb as $cpentry) {
+		if (empty($cpentry[11])) {
+			$cpentry[11] = 'first';
+		}
+		$radiusservers = $radiussrvs[$cpentry[11]];
+
+		captiveportal_disconnect($cpentry, $radiusservers, $term_cause);
+		captiveportal_logportalauth($cpentry[4], $cpentry[3], $cpentry[2], $logoutReason);
+		$unsetindexes[] = $cpentry[5];
+	}
+	unset($cpdb);
+
+	if (!empty($unsetindexes)) {
+		captiveportal_remove_entries($unsetindexes);
+	}
+}
+
 /* send RADIUS acct stop for all current clients */
 function captiveportal_radius_stop_all() {
 	global $config, $cpzone;
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1923,14 +1923,14 @@ function system_reboot_sync() {
 }
 
 function system_reboot_cleanup() {
-	global $config, $cpzone;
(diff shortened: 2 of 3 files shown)


plumbeo:
If this patch is accepted, would it be ok to replace the calls to captiveportal_radius_stop_all with calls to the new function with the appropriate parameters? 
captiveportal_radius_stop_all sends basic accounting-stop packets that don't include the session traffic and doesn't record the end of the sessions in the system log.

plumbeo:
I just pushed a tentative patch that replace calls to captiveportal_radius_stop_all with calls to the new captiveportal_disconnect_all. In doing that I also fixed a bug (missing zoneid) in system_reboot_cleanup() so with this patch the RADIUS accounting when rebooting or disabling a zone is correct (including AcctInputOctets/AcctOutputOctets) and the end of the sessions is logged in the system log. In theory captiveportal_radius_stop_all could be removed completely.
The patch also comments out a couple of calls to captiveportal_send_server_accounting(). I'm not sure what that function was expected to do in these codepaths, but right now it just sends accounting off packets to all clients, including clients that are not in the zone that's being disabled, messing up with the ability of the RADIUS server to know how many clients are active.

src/etc/inc/captiveportal.inc
```diff
@@ -1082,6 +1082,34 @@ function captiveportal_disconnect_client($sessionid, $term_cause = 1, $logoutRea
...
+/* remove all clients */
+function captiveportal_disconnect_all($term_cause = 6, $logoutReason = "DISCONNECT") {
+	global $g, $config, $cpzone, $cpzoneid;
+
+	$radiussrvs = captiveportal_get_radius_servers();
+	$cpdb = captiveportal_read_db();
+
+	$unsetindexes = array();
+
+	$stop_time = time();
```
jim-p: This is unnecessary, see next note

src/etc/inc/captiveportal.inc
```diff
@@ -1082,6 +1082,34 @@ function captiveportal_disconnect_client($sessionid, $term_cause = 1, $logoutRea
...
+
+	foreach ($cpdb as $cpentry) {
+		if (empty($cpentry[11])) {
+			$cpentry[11] = 'first';
+		}
+		$radiusservers = $radiussrvs[$cpentry[11]];
+
+		captiveportal_disconnect($cpentry, $radiusservers, $term_cause, $stop_time);
```
jim-p: Omit $stop_time here and captiveportal_disconnect() will set it to time() automatically, no need to make the manual setting yourself since it will be identical.

src/etc/inc/system.inc
```diff
@@ -1923,14 +1923,15 @@ function system_reboot_sync() {
...
+	global $config, $cpzone, $cpzoneid;
 
 	mwexec("/usr/local/bin/beep.sh stop");
 	require_once("captiveportal.inc");
 	if (is_array($config['captiveportal'])) {
 		foreach ($config['captiveportal'] as $cpzone=>$cp) {
-			captiveportal_radius_stop_all();
-			captiveportal_send_server_accounting(true);
+			$cpzoneid = $cp[zoneid];
+			captiveportal_disconnect_all(7); // Admin-Reboot
```
jim-p: Why rely on globals being set when you could pass that as a function parameter instead so there is much less room for error?

src/etc/inc/captiveportal.inc
```diff
@@ -347,14 +347,14 @@ EOD;
...
+		captiveportal_disconnect_all(10); // NAS-Request
 
 		captiveportal_filterdns_configure();
 
 		/* send Accounting-Off to server */
-		if (!platform_booting()) {
+		/*if (!platform_booting()) {
 			captiveportal_send_server_accounting(true);
-		}
+		}*/
```
jim-p: Why retain the commented out code, why not remove it instead?

src/etc/inc/system.inc
```diff
@@ -1923,14 +1923,15 @@ function system_reboot_sync() {
...
+			//captiveportal_send_server_accounting(true);
```
jim-p: Why retain the commented out code, why not remove it instead?

src/etc/inc/captiveportal.inc
plumbeo: The idea was to ensure that the stop time was the same for all the sessions, since, in theory, if there were hundreds or thousands of active sessions, it could change between the first and the last one. Thinking about it again it's probably better to just remove it and use the real disconnection time, I'll change it.

jim-p: That seems reasonable, if you decide to keep that behavior, please add a comment to that effect so the intent is shown in the code. I'd tend to lean toward letting each one naturally pick it up as it happens though.

src/etc/inc/captiveportal.inc
plumbeo: That's the reason I pushed the patch as an RFC, I wasn't sure that I understood why the function was even called. I planned to remove it in the final patch, after giving another look at the function, unless someone pointed me to a reason to keep it.

src/etc/inc/system.inc
plumbeo: Mostly because that's what all the other functions in captiveportal.inc do. This is the only place where it makes sense to use it but the globals aren't already set, so I wasn't sure that it was worth it to change it to use explicit parameters and then having it working in a different way.

PR #3286, opened 2016-12-17
Fix capitalization of confirmation messages for btn-danger buttons an…

plumbeo:
…d fa-trash icons

Value and title were converted to lower case but not textContent. Consequently some pop-up messages included capitalized words, for example in the Log Settings page ("Are you sure you wish to Reset Log Files?").

### DIFF
--- a/src/usr/local/www/js/pfSense.js
+++ b/src/usr/local/www/js/pfSense.js
@@ -144,7 +144,7 @@ $(function() {
 	// the element value
 	$('.btn-danger, .fa-trash').on('click', function(e){
 		if (!($(this).hasClass('no-confirm'))) {
-			var msg = $.trim(this.textContent);
+			var msg = $.trim(this.textContent).toLowerCase();
 
 			if (!msg)
 				var msg = $.trim(this.value).toLowerCase();


NOYB:
Why the aversion to capital letters?  That's the way it is intended to be.  Sentences are meant to have capital letters.  Poor form to force all things to lower case without regard of cause/need.
Does not LGTM.  Restrictive without cause/need.

jdillard:
The message gets appended to the sentence "Are you sure you wish to"...

jdillard:
@NOYB Do you think it would be better to drop the toLowerCase() and change the sentence to something like:

`q = 'Are you sure you wish to: '+ msg + '?'; `

NOYB:
That is already how it is, but with msg being forced to lower case.  Is it not?  Maybe I'm not following what you mean.

Not really a big deal.  Just don't like the globally forcing of case were it is not necessary, as it prevents the use of capital letters when/were they are appropriately desired.  That's my beef with it.

My initial thought was that it was lower casing the entire sentence.  So maybe only half as bad first thought.  But still don't really like the idea of globally forced.

PR #3287, opened 2016-12-17
Captive portal: fix sorting by last activity time in status page

plumbeo:
Given that the last activity time of each user is not saved in the database, the display code correctly calculates it on the fly when populating the user list. To the contrary the ordering code just sorts the users by the sixth column of the database ('sessionid'), which results in a random order.

Retrieve each user's last activity time when needed and add it to the array, then display the saved values and use them for sorting the list.

### DIFF
--- a/src/usr/local/www/status_captiveportal.php
+++ b/src/usr/local/www/status_captiveportal.php
@@ -66,33 +66,8 @@ if ($_GET['act'] == "del" && !empty($cpzone) && isset($cpzoneid) && isset($_GET[
 	exit;
 }
 
-function clientcmp($a, $b) {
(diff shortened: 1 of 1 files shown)


plumbeo:
I just pushed a second commit that fixes sorting by ip address.

plumbeo:
@sbeaver-netgate thank you for pointing me to that! I converted the table to a sortable table and, since the new code sorts the actual contents of the table instead of reading the db every time, the conversion also fixed sorting by last activity without the need to change anything else.

sbeaver-netgate:
Happy days :) Thank you.

PR #3289, opened 2016-12-18
Improve RADVD router mode help text (Todo #6889)

doktornotor:
Improve RA router mode descriptions as suggested by Corey Boyle in https://redmine.pfsense.org/issues/6889


jim-p:
I like the detail but it turns the drop-down into a big wall of text. I think it will end up confusing people more this way. Perhaps the detail could be worked into the description text underneath instead of in the drop-down?

doktornotor:
Shrug; yeah it could be worked into the description text underneath, turning it into an even bigger wall of text. I definitely give up on that, trying to format text with  setHelp0 sucks goats nuts to put it midly, not to mention the gettext mess resulting thereof.

rbgarga:
@sbeaver-netgate done

PR #3290, opened 2016-12-18
Add enable link to Status > UPnP & NAT-PMP error message if disabled (Todo #6689)

doktornotor:


rbgarga:
@doktornotor please do not update .pot file manually. When it's required, just add a note on PR and I can run script to re-generate it after merge

doktornotor:
Oh, you have a script now? Great to know, I'd rather never touch the giant blob again, it's a big pain.

rbgarga:
@doktornotor yes, it's on ./tools/scripts/update_pot.sh :)

doktornotor:
@rbgarga - Grrrh... should have known earlier. Definitely will avoid touching the pot thing in future, thanks for the pointer.

src/usr/local/www/status_upnp.php
```diff
@@ -52,7 +52,7 @@
     !$config['installedpackages']['miniupnpd']['config'][0]['iface_array'] ||
     !$config['installedpackages']['miniupnpd']['config'][0]['enable']) {
 
-	print_info_box('UPnP is currently disabled.', 'danger');
+	print_info_box(gettext('UPnP is currently disabled. Go to <a href="pkg_edit.php?xml=miniupnpd.xml">Services > UPnP &amp; NAT-PMP</a> to enable.'), 'danger');
```
rbgarga: Please remove html content from gettext() string and use sprintf

PR #3293, opened 2016-12-19
Fix #6927 1:1 NAT validate address family

phil-davis:
Ensure that all the manually-entered addresses come from the same
address family - i.e. they are all either IPv4 or IPv6 addresses.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2481,14 +2481,20 @@ function get_smart_drive_list() {
 //	$label: the label used by the GUI to display this value. Required to compose an error message
 //	$err_msg: pointer to the callers error message array so that error messages can be added to it here
 //	$alias: are aliases permitted for this address?
+// Returns:
+//	4 - if $addr is a valid IPv4 address
+//	6 - if $addr is a valid IPv6 address
+//	1 - if $alias=true and $addr is an alias
+//	false - otherwise
+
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: I enhanced the return values from validateipaddr() so the caller can know if the validated address is IPv4 or IPv6 (or an alias if that check was requested). That makes it a better replacement for is_ipaddr().
Without doing that, this code would have called is_ipaddr() in order to know if the address was IPv4 or IPv6, and also called validateipaddr() in order to get the $input_errors array stuff done (for the case of an invalid address). That looked rather silly when I simply did minimal integration of the validate_ipaddr() stuff into the code I had.

phil-davis:
Another note: The "magic numbers" returned by is_ipaddr() and validateipaddr() could be defined as named constants - perhaps using the IP address types constants that were recently added to globals.inc
and extending those to include "ALIAS".
That re-factoring needs a bit of thought. And then go through the system to change everything that makes special use of the "4" and "6" and change it to use the defined constants.

rbgarga:
@phil-davis I like this solution and it worked fine on my tests. I would like to ask you only to split it in 2 commits just to keep the history sane. Could you do it please?

sbeaver-netgate:
I'll extend the address family defines I recently added to globasls.inc to include ALIAS once this is committed.

phil-davis:
validateipaddr() changes are in PR #3298

phil-davis:
1:1 NAT validation is in #3299

PR #3294, opened 2016-12-19
User Configurable DUID

marjohn56:
Replaces PR #3278

User may enter a DUID in WAN DHCP6 section. DUID validity is checked and
on WAN interface configure the  DUID is written to /var/db/dhcp6c_duid.

This also solves the issue of users using RAM disks who lose the DUID on
reboot.

This PR is identical to PR 3278, just tidied up.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3867,6 +3867,10 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
+	if (!empty($wancfg['dhcp6-duid']) {
+	// Write the DUID file
+		write_dhcp6_duid($wancfg['dhcp6-duid']));
+	}
 
 	if ($wancfg['adv_dhcp6_config_file_override']) {
 		// DHCP6 Config File Override
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2536,4 +2536,38 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
 
 	return false;
 }
+/* returns true if $dhcp6duid is a valid duid entrry */
+function is_duid($dhcp6duid) {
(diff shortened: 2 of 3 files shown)


jim-p:
Wouldn't it make more sense to have this as a system-configurable option (System > Advanced, Networking tab, perhaps), and then have a per-WAN override to use instead? The idea behind a DUID is that it was supposed to be a value for the entire system, not a per-interface value, which would be an IAID attached to the DUID. I could see it being useful to override the DUID on a WAN, but it should be configurable for the system itself.

marjohn56:
In my case my ISP uses the DUID to tie the allocated PD to my system, if I change my DUID I get a different PD. I see your point though about multiple WAN interfaces may all require thier own DUIDs, but at present dhcp6c writes to the /var/db as a file dhcp6c_duid, it does not differentiate between WAN ports. We'd have to make dhcp6c look for a different filename as well as it writing a different filename if there were multiple instances.

jim-p:
So if it is not a per-interface value, why present it on each interface? I can see that being offered for convenience, but it's a system value, not an interface value the way it's being stored in this PR. The way this is coded, if you have a second WAN that has a different DUID set, it would clobber the first, wouldn't it?

marjohn56:
Because I hadn't thought if it that way.. :)

But can you have multiple instances of dhcp6c running, dhcp6c creates it's own DUID if one is not there, and it calls it dhcp6c_duid, and stores it in /var/db. How do we get around that, recode dhcp6c to write the file as dhcp6c_duid_wan* for example, then it could be per interface.

marjohn56:
Having spent some time looking at dhcp6c in more detail this morning, and the pending future change to ISC then jim-p is correct. This should move into the System->Advanced section with an option in the WAN configuration to use the system defined DUID. At present, for those who only need a fixed DUID to get around their issues, then just using the DUID is fine. The option of IAID can then then be added to the WAN DHCP advanced options when the client supports it. I will move the DUID to the System->Advanced section and issue a new PR when ready.

rbgarga:
Superseded by #3297

PR #3296, opened 2016-12-20
1:1 NAT validate address family example

phil-davis:
This is what the code would look like with a simple integration of the validateipaddr() changes. Notice how it still needs to use is_ipaddr() to find out the address family, but then also calls validateipaddr() to make the $input_errors stuff happen.
e.g. lines 176 to 179


phil-davis:
Take your pick from this or #3293 or re-engineer it some other way.

rbgarga:
#3293 is better solution

PR #3298, opened 2016-12-20
validateipaddr return address family

phil-davis:
Enhanced the return values from validateipaddr() so the caller can know if the validated address is IPv4 or IPv6 (or an alias if that check was requested). That makes it a better replacement for is_ipaddr().


sbeaver-netgate:
I have updated globals.inc to make it easier to use the defined values here:

// IP address types
define("IPV4", 4);
define("IPV6", 6);
define("IPV4V6", 2);
define("ALIAS", 1);

phil-davis:
I think that's it! Do you want me to make another PR with just a single commit?

sbeaver-netgate:
Nope. I can merge it as submitted. Thanks!

PR #3299, opened 2016-12-20
Fix #6927 1:1 NAT validate address family

phil-davis:
Ensure that all the manually-entered addresses come from the same
address family - i.e. they are all either IPv4 or IPv6 addresses.


phil-davis:
Depends on #3298

PR #3300, opened 2016-12-20
Captive portal: add button to disconnect all users

plumbeo:
I somehow managed to delete my branch so #3285 was auto-closed and I can't re-open it.

I removed the commented code since it definitely looked like it wasn't doing anything useful and stopped setting a common disconnection time for all users letting the function use the actual disconnection time.

I'm still not sure if I should change the function to avoid using globals, it would the only one of a group of related captive portal functions to do that.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -1082,6 +1082,32 @@ function captiveportal_disconnect_client($sessionid, $term_cause = 1, $logoutRea
 	}
 }
 
+/* remove all clients */
+function captiveportal_disconnect_all($term_cause = 6, $logoutReason = "DISCONNECT") {
+	global $g, $config, $cpzone, $cpzoneid;
+
+	$radiussrvs = captiveportal_get_radius_servers();
+	$cpdb = captiveportal_read_db();
+
+	$unsetindexes = array();
+
+	foreach ($cpdb as $cpentry) {
+		if (empty($cpentry[11])) {
+			$cpentry[11] = 'first';
+		}
+		$radiusservers = $radiussrvs[$cpentry[11]];
+
+		captiveportal_disconnect($cpentry, $radiusservers, $term_cause);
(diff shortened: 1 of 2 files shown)


src/etc/inc/captiveportal.inc
```diff
@@ -347,15 +347,10 @@ EOD;
...
 		@unlink("{$g['varetc_path']}/captiveportal-{$cpzone}-logout.html");
 
-		captiveportal_radius_stop_all();
+		captiveportal_disconnect_all(10); // NAS-Request
 
 		captiveportal_filterdns_configure();
 
-		/* send Accounting-Off to server */
-		if (!platform_booting()) {
-			captiveportal_send_server_accounting(true);
```
jim-p: Why remove this part? It seems to be unrelated to the other changes, unless I'm overlooking something.

plumbeo: @jim-p I found during testing that if many zones share a radius server and one of the zones is disabled, that function sends a stop message to the server and the server stops the accounting for all the clients it knows, including clients in other zones who are still getting service. 
At that point the RADIUS server loses track of the number of active sessions and their length and it can't enforce Simultaneous-Use limits or other limits any more.

jim-p: Seems then it would be better to test for the proper condition to send or not send, rather than remove it. At the very least I'd move that to a separate bug/PR since it is not directly related to this change.

plumbeo: There's not much we can test, basically only "is this zone the only active zone using RADIUS accounting with this specific RADIUS server?" and even if the answer is yes, by that time we've already sent Accounting-Stop packets for all the clients so Accounting-Off buys us nothing. 
Given that Accounting-Off is optional I'd just not bother sending it, maybe only when rebooting, after all clients have been cleanly disconnected.

plumbeo: Given that they're not directly related, I'll move the last two commits to a new PR, so this one can be closed, if it's ok.

PR #3301, opened 2016-12-20
Fix CPU update period in System Information widget

doktornotor:
This only updates after the refresh period set in System - General Setup - Dashboard update period; should not be hardcoded.

@rbgarga - This probably needs the update_pot.sh magic.


src/usr/local/www/widgets/widgets/system_information.widget.php
```diff
@@ -225,7 +225,8 @@
 					<div id="cpuPB" class="progress-bar progress-bar-striped" role="progressbar" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" style="width: 0%">
 					</div>
 				</div>
-				<span id="cpumeter"><?=gettext('(Updating in 10 seconds)')?></span>
+				<?php $update_period = (isset($config['widgets']['period']) && !empty($config['widgets']['period'])) ? $config['widgets']['period'] : "10"; ?>
```
rbgarga: empty() already does isset() so you can use just it

PR #3297, opened 2016-12-20
User Define DUID - Moved to System->Advanced->Networking

marjohn56:
Moved the Configuration of the DUID to System->Advanced->Networking.

DUID validity is checked for content and length. then creates a system
config variable called global-v6duid.
Can replace PR #3294

In Interfaces WAN an option exists to use the global DUID ( which has
been defined or it flags an error ). On WAN interface configure the
write of the DUID file takes place by calling the function in util.inc

PR #3303, opened 2016-12-21
User Defined DUID

marjohn56:
Replaces PR #3297

DUID is configured in System->Advanced->Networking. Checks are made for
validity and length.

User may then opt to use that DUID in WAN->DHCP6. If selected, the
dhcp6c duid file is then written to /var/db/ on call of
interface_dhcpv6_configure.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3868,7 +3868,8 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
 
-	if (isset($wancfg['dhcp6-duid'])) {
+	if (isset($wancfg['dhcp6-duid'])  &&
+    	    !empty($config['system']['global-v6duid'])) {
 	// Write the DUID file
 		if(!write_dhcp6_duid($config['system']['global-v6duid'])) {
 			log_error(gettext("Failed to write user DUID file!"));
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3868,6 +3868,13 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
 
+	if (isset($wancfg['dhcp6-duid'])) {
+	// Write the DUID file
+		if(!write_dhcp6_duid($config['system']['global-v6duid'])) {
+			log_error(gettext("Failed to write user DUID file!"));
+		}
+	}
+	
 	if ($wancfg['adv_dhcp6_config_file_override']) {
 		// DHCP6 Config File Override
 		$dhcp6cconf = DHCP6_Config_File_Override($wancfg, $wanif);
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2559,24 +2559,22 @@ function is_duid($dhcp6duid) {
 }
 
 /* Write the DHCP6 DUID file */
-function write_dhcp6_duid($duidfile) {
+function write_dhcp6_duid($duidstring) {
 	// Create the hex array from the dhcp6duid config entry and write to file
 	global $g;
  	
- 	if(is_duid($duidfile))
- 	{
- 		$temp = str_replace(":","",$duidfile);
- 		$duid_binstring = pack("H*",$temp);
- 		if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "wb")) {
- 			fwrite($fd, $duid_binstring);
-			fclose($fd);
- 			return;
- 		}
- 		else {
- 			log_error(gettext("Error: attempting to write DUID file - File write error"));
- 			return;
- 		}
- 	}
- 	log_error(gettext("Error: attempting to write DUID file - Invalid DUID detected"));
+ 	if(!is_duid($duidstring)) {
+		log_error(gettext("Error: attempting to write DUID file - Invalid DUID detected"));
+		return false;
+	}
+	$temp = str_replace(":","",$duidstring);
+	$duid_binstring = pack("H*",$temp);
+	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "wb")) {
+		fwrite($fd, $duid_binstring);
+		fclose($fd);
+		return true;
+	}
+	log_error(gettext("Error: attempting to write DUID file - File write error"));
+	return false;
 }
 ?>
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2542,4 +2542,41 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
 
 	return false;
 }
+
+/* returns true if $dhcp6duid is a valid duid entrry */
+function is_duid($dhcp6duid) {
+	$values = explode(":", $dhcp6duid);
+	if (count($values) != 16 || strlen($dhcp6duid) != 47) {
+		return false;
+	}
+	for ($i = 0; $i < 16; $i++) {
+		if (ctype_xdigit($values[$i]) == false)
+			return false;
+		if (hexdec($values[$i]) < 0 || hexdec($values[$i]) > 255)
+			return false;
+	}
+	return true;
+}
+
+/* Write the DHCP6 DUID file */
+function write_dhcp6_duid($duidfile) {
+	// Create the hex array from the dhcp6duid config entry and write to file
+	global $g;
+ 	
+ 	if(is_duid($duidfile))
+ 	{
+ 		$temp = str_replace(":","",$duidfile);
+ 		$duid_binstring = pack("H*",$temp);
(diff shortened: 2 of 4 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3868,6 +3868,13 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 	$wanif = get_real_interface($interface, "inet6");
 	$dhcp6cconf = "";
 
+	if (isset($wancfg['dhcp6-duid'])) {
```
rbgarga: Looks like you forgot to bring fixes from other PR

This check should be:
```
if (isset($wancfg['dhcp6-duid']) &&
    !empty($config['system']['global-v6duid'])) {
```

src/etc/inc/interfaces.inc
```diff
@@ -3868,6 +3868,13 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
...
+	// Write the DUID file
```
rbgarga: Fix indent

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,41 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+		if (ctype_xdigit($values[$i]) == false)
+			return false;
+		if (hexdec($values[$i]) < 0 || hexdec($values[$i]) > 255)
+			return false;
+	}
+	return true;
+}
+
+/* Write the DHCP6 DUID file */
+function write_dhcp6_duid($duidfile) {
```
rbgarga: $duidfile is not a good name for this variable

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,41 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+	// Create the hex array from the dhcp6duid config entry and write to file
+	global $g;
+ 	
+ 	if(is_duid($duidfile))
```
rbgarga: { should be moved to this line to respect style

rbgarga: This check should be if (!is_duid()) and then log_error and return false

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,41 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+ 	{
+ 		$temp = str_replace(":","",$duidfile);
+ 		$duid_binstring = pack("H*",$temp);
+ 		if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "wb")) {
+ 			fwrite($fd, $duid_binstring);
+			fclose($fd);
+ 			return;
```
rbgarga: return true here

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,41 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+ 		}
+ 		else {
```
rbgarga: Else is unnecessary, if code flow went to previous if it will return true and never reach it

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,41 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+ 			log_error(gettext("Error: attempting to write DUID file - File write error"));
+ 			return;
```
rbgarga: Return false here

src/usr/local/www/interfaces.php
```diff
@@ -274,7 +274,10 @@ function remove_bad_chars($string) {
 		$pconfig['type6'] = "slaac";
 		break;
 	case "dhcp6":
-		$pconfig['dhcp6-duid'] = $wancfg['dhcp6-duid'];
+		if(empty($config['system']['global-v6duid'])) {
+			unset($wancfg['dhcp6-duid']);
+		}
```
rbgarga: This shouldn't be done. Individual interface option should not be automatically changed in GUI if global option is not set.

src/usr/local/www/interfaces.php
```diff
@@ -794,6 +797,9 @@ function remove_bad_chars($string) {
 	if (($_POST['spoofmac'] && !is_macaddr($_POST['spoofmac']))) {
 		$input_errors[] = gettext("A valid MAC address must be specified.");
 	}
+	if (($_POST['dhcp6-duid'] && empty($config['system']['global-v6duid']))) {
+		$input_errors[] = gettext("No System DUID has been entered. Please enter a valid DUID in System->Advanced->Networking");
+	}
```
rbgarga: It should be a warning. It can be a static warning together with field description and can also include a link to system -> advanced -> network to make it easy for users

src/usr/local/www/system_advanced_network.php
```diff
@@ -83,6 +84,20 @@
 			unset($config['system']['prefer_ipv4']);
 		}
 
+		if (!empty($_POST['global-v6duid'])) {
+			strtolower(str_replace("-", ":", $_POST['global-v6duid']));
```
rbgarga: Result of strtolower is not being attributed to any variable, so it's useless

src/usr/local/www/system_advanced_network.php
```diff
@@ -83,6 +84,20 @@
...
+			if (!is_duid($_POST['global-v6duid'])) {
+				$input_errors[] = gettext("A valid DUID must be specified");
+			} else {
+				$config['system']['global-v6duid'] = $_POST['global-v6duid'];
+			}
+		}
```
rbgarga: else goes in the same line of }

src/usr/local/www/system_advanced_network.php
```diff
@@ -83,6 +84,20 @@
...
+		else {
+			unset($config['system']['global-v6duid']);
+			// nothing there so clear the WAN setting if its set
+			unset($config['interfaces']['wan']['dhcp6-duid']);
```
rbgarga: Don't do it, just ignore the setting on interfaces.inc

rbgarga:
Please, do not close this PR and send another. Lets keep all changes here to avoid bringing already fixed items back again

src/usr/local/www/interfaces.php
marjohn56: Please explain what you mean by 'global option', What I am trying to prevent is the user setting 'dhcp6-duid' when there  is no valid 'global-v6duid'. I would have preferred the option of masking the select box, if that is possible, thus making it impossible to select the option in the first place.

src/etc/inc/interfaces.inc
marjohn56: Think I have?

src/etc/inc/util.inc
marjohn56: done

src/etc/inc/util.inc
marjohn56: done

src/usr/local/www/interfaces.php
marjohn56: Could you explain further what you mean  by global option? What I am trying to prevent is the user selecting this option without entering a valid DUID; the wrong or no DUID may, in the case of a RAM disk cause dhcp6c to write a new DUID, this is what I'm trying to avoid as much as anything.

rbgarga:
@marjohn56 I believe I'm misunderstanding part of this change.

* The custom DUID is added to a global field on system->advanced->network
* There is also a boolean field on each interface to define if that interface should use custom DUID
* If a custom DUID is defined and interface X has this boolean field set, write_dhcp6_duid() is called and /var/db/dhcp6_duid file is created

Here is my question. What happens if custom DUID is defined and there are 2 interfaces configured to use DHCP6, one with dhcp6-duid field set and other unset? If it's a single file /var/db/dhcp6_duid instead of a file for each interface?

marjohn56:
You may not have read the comment from jimp on my earlier version, which although allowed for interfaces to define separate DUID’s, would in fact have caused the same

chcp6c_duid to be overwritten. Doing it this way allows the user to select to use a global DUID, only one Is allowed because as jimp pointed out, we should be using IAID’s to differentiate between the interfaces. Currently pfSense used dhcp6c which does not do IAID’s, ISC can if ever that gets implemented. 

 

The only other  option is to modify dhcp6c to read and write to a specific DUID file, such as dhcp6_duid_re0 etc, that would have been my preferred choice.

 

At this point though, most users who have one V6 interface just want a DUID they can store and that does not change on reboot.

 

What are your thoughts?

rbgarga:
In this case, my opinion is we shouldn't add configuration item in interface.php, only the global item in system_advanced_network.php. If user add a custom DUID there, then it will be used.

At interfaces.php we can add a message near the place user chose dhcp6 advanced options saying "if you want to define a custom DUID, do it <here>" with a link to system_advanced_network.php.

What do you think @jim-p ?

jim-p:
Just the global option with a hint/link is good and simple.

marjohn56:
I will make it so captain. Should I also add a warning that setting it with multiple WAN v6 may cause problems, or words to that effect?

marjohn56:
Close this PR and I'll put another in, it should be quick as I''ll only need to change two files.

PR #3304, opened 2016-12-21
User Defined DUID

marjohn56:
User may define a DUID to use in System->Advanced->Networking. The
entered DUID is validated for composition and length, if valid it is
stored in the config.xml. On call of wan_dhcp6_configure() the DUID is
written to file to be read by dhcp6c on launch.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2542,4 +2542,40 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
 
 	return false;
 }
+
+
+/* returns true if $dhcp6duid is a valid duid entrry */
+function is_duid($dhcp6duid) {
+	$values = explode(":", $dhcp6duid);
+	if (count($values) != 16 || strlen($dhcp6duid) != 47) {
+		return false;
+	}
+	for ($i = 0; $i < 16; $i++) {
+		if (ctype_xdigit($values[$i]) == false)
+			return false;
+		if (hexdec($values[$i]) < 0 || hexdec($values[$i]) > 255)
+			return false;
+	}
+	return true;
+}
+
+/* Write the DHCP6 DUID file */
+function write_dhcp6_duid($duidstring) {
+	// Create the hex array from the dhcp6duid config entry and write to file
+	global $g;
+ 	
+ 	if(!is_duid($duidstring)) {
+		log_error(gettext("Error: attempting to write DUID file - Invalid DUID detected"));
+		return false;
+	}
+	$temp = str_replace(":","",$duidstring);
+	$duid_binstring = pack("H*",$temp);
+	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "wb")) {
+		fwrite($fd, $duid_binstring);
+		fclose($fd);
+		return true;
+	}
+	log_error(gettext("Error: attempting to write DUID file - File write error"));
+	return false;
+}
 ?>
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2543,7 +2543,6 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
 	return false;
 }
 
-
 /* returns true if $dhcp6duid is a valid duid entrry */
 function is_duid($dhcp6duid) {
 	$values = explode(":", $dhcp6duid);
@@ -2578,19 +2577,20 @@ function write_dhcp6_duid($duidstring) {
 	log_error(gettext("Error: attempting to write DUID file - File write error"));
 	return false;
 }
+
 /* returns duid string from 'vardb_path']}/dhcp6c_duid' */
 function get_duid_from_file()
 {
 	global $g;
 	
+	$duid_ASCII = "";
 	$count = 0;
 	
 	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "r")) {
 		if(filesize("{$g['vardb_path']}/dhcp6c_duid")==16) {
-			$buffer = fread($fd,filesize("{$g['vardb_path']}/dhcp6c_duid"));					
-			while($count < 16) {				
-				$a = $buffer[$count];
-				$duid_ASCII .= bin2hex($a);
+			$buffer = fread($fd,16);					
+			while($count < 16) {
+				$duid_ASCII .= bin2hex($buffer[$count]);
 				$count++;
 				if($count < 16) {
 					$duid_ASCII .= ":";
@@ -2600,7 +2600,7 @@ function get_duid_from_file()
 		fclose($fd);
 	}
 	//if no file or error with read then the string returns blanked DUID string
-	if($count != 16) {
+	if(!is_duid($duid_ASCII)) {
 		return "--:--:--:--:--:--:--:--:--:--:--:--:--:--:--:--";
 	}
 	return($duid_ASCII);	
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2586,14 +2586,10 @@ function get_duid_from_file()
 	$count = 0;
 	
 	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "r")) {
-		
 		if(filesize("{$g['vardb_path']}/dhcp6c_duid")==16) {
-			
-			$string = fread($fd,filesize("{$g['vardb_path']}/dhcp6c_duid"));
-					
-			while($count < 16) {
-				
-				$a = $string[$count];
+			$buffer = fread($fd,filesize("{$g['vardb_path']}/dhcp6c_duid"));					
+			while($count < 16) {				
+				$a = $buffer[$count];
 				$duid_ASCII .= bin2hex($a);
 				$count++;
 				if($count < 16) {
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2578,4 +2578,35 @@ function write_dhcp6_duid($duidstring) {
 	log_error(gettext("Error: attempting to write DUID file - File write error"));
 	return false;
 }
+/* returns duid string from 'vardb_path']}/dhcp6c_duid' */
+function get_duid_from_file()
+{
+	global $g;
+	
+	$count = 0;
+	
+	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "r")) {
+		
+		if(filesize("{$g['vardb_path']}/dhcp6c_duid")==16) {
+			
+			$string = fread($fd,filesize("{$g['vardb_path']}/dhcp6c_duid"));
+					
+			while($count < 16) {
+				
+				$a = $string[$count];
+				$duid_ASCII .= bin2hex($a);
+				$count++;
+				if($count < 16) {
+					$duid_ASCII .= ":";
+				}
+			}
+		}
+		fclose($fd);
+	}
+	//if no file or error with read then the string returns blanked DUID string
+	if($count != 16) {
+		return "--:--:--:--:--:--:--:--:--:--:--:--:--:--:--:--";
+	}
+	return($duid_ASCII);	
+}
 ?>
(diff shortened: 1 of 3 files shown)


marjohn56:
Simples... Oh that's a UK thing you may not understand! :)

src/usr/local/www/system_advanced_network.php
```diff
@@ -106,8 +119,8 @@
 		if ($_POST['disablelargereceiveoffloading'] == "yes") {
 			$config['system']['disablelargereceiveoffloading'] = true;
 		} else {
-			unset($config['system']['disablelargereceiveoffloading']);
-		}
+			unset($config['system']['disablelargereceiveoffloading']); hi
```
jim-p: What's this?

src/usr/local/www/system_advanced_network.php
```diff
@@ -186,6 +199,15 @@
...
 	'IPv6 will be used. If this option is selected, IPv4 will be preferred over IPv6.');
 
+$section->addInput(new Form_Input(
+	'global-v6duid',
+	'DHCP6 DUID',
+	'text',
+	$pconfig['global-v6duid'],
+	['placeholder' => 'xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx']
+	))->setWidth(9)->sethelp('Enter the DUID to use here. Use this option if using RAM Disk, as the DUID will be lost on reboot. The existing DUID may be found in var/db/dhcp6_duid.' .'<br />' .
+			'NOTE: Do not use this option with multiple DHCP6 WAN interfaces.');
```
jim-p: Is this note still relevant? If it's a system-wide DUID it should be safe to use for any number of interfaces.

src/usr/local/www/system_advanced_network.php
```diff
@@ -186,6 +199,15 @@
 ))->setHelp('By default, if IPv6 is configured and a hostname resolves IPv6 and IPv4 addresses, '. 
 	'IPv6 will be used. If this option is selected, IPv4 will be preferred over IPv6.');
 
+$section->addInput(new Form_Input(
+	'global-v6duid',
+	'DHCP6 DUID',
+	'text',
+	$pconfig['global-v6duid'],
+	['placeholder' => 'xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx']
+	))->setWidth(9)->sethelp('Enter the DUID to use here. Use this option if using RAM Disk, as the DUID will be lost on reboot. The existing DUID may be found in var/db/dhcp6_duid.' .'<br />' .
```
jim-p: Might be nice to print and/or set the placeholder to the current DUID rather than suggest the user dig into a system file to find it.

src/usr/local/www/system_advanced_network.php
marjohn56: Not in my commit jim-p, don't know where that has come from.

src/usr/local/www/system_advanced_network.php
marjohn56: O.K. If the system is virginal, there will not be any existing DUID, but that's obvious. If we set the placeholder, do we need to add a checkbox to say use the DUID? Not sure how the placeholder works, do you need to physically enter what is there as the placeholder in order for it to take?

src/usr/local/www/system_advanced_network.php
jim-p: It was introduced by the second commit on this PR, aff01dbd97c10731b98bc70c32ac19822c95d304

src/usr/local/www/system_advanced_network.php
jim-p: If they leave the field blank it will use the existing DUID, which is what the placeholder is at the time. There wouldn't be a way to make it editable from there but usually it wouldn't need to be. If someone is changing it they likely would be changing it to a much different value.

marjohn56: But what happens if they don't want to use the Global DUID?

marjohn56:
Looks like I was tying an email at the same time... :(

 

Corrected and updated.

src/usr/local/www/system_advanced_network.php
marjohn56: I only have one WAN interface so I cannot be sure about this, and I know what we said about DUID the other day, but the RFC does say that DUID's should be interface unique. When running dhcp6c on multiple WAN interfaces I assume multiple dhcp6c clients are launched, or is that not the case?

If it is the case then the solution is to go back to the by WAN setting of DUID's and I'll modify dhcp6c to use specific DUIDs for each interface. We already pass an the argument to dhcp6c anyway so it would be very simple to do.

src/usr/local/www/system_advanced_network.php
jim-p: If they want to change it, then they enter in whatever they want. It's a convenience there to show the user what the system is using now rather than making them dig for it. The placeholder value is displayed in the field grayed out to indicate the "default" value of the field.

src/usr/local/www/system_advanced_network.php
jim-p: The RFC doesn't say that. None that I've seen anyhow. While the DUID can be generated using an interface address as part of the value, DUID is not intended to be related to any interface. It's a client identifier that uniquely identifies a _device_ and not an interface. Identifying specific interfaces is where IAIDs come in, which we've been over.

marjohn56: I will bow to your greater knowledge on that then.

Problem is, on RAM disks the DUID disappears on reboot, a new one is created, but that only happens when dhcp6 is fired up, I can see all sorts of 'I've got no DUID' type questions.  I'll have a look at pulling the duid from the file and inserting it in the placeholder. If nothing else I'm finding out more about pfSense..

jim-p: That easy, if there is no file, just print the x's.

marjohn56:
True.. :)

src/usr/local/www/system_advanced_network.php
marjohn56: True..

marjohn56: Completed.. Existing DUID now displays in placeholder. I have put more verbage there to try and let users know that unless they enter the duid and save it, it is not stored in the config and therefore can change, especially when using RAM disks.

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,67 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+	$duid_binstring = pack("H*",$temp);
+	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "wb")) {
+		fwrite($fd, $duid_binstring);
+		fclose($fd);
+		return true;
+	}
+	log_error(gettext("Error: attempting to write DUID file - File write error"));
+	return false;
+}
+/* returns duid string from 'vardb_path']}/dhcp6c_duid' */
```
rbgarga: Add a blank line between functions

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,67 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+function get_duid_from_file()
+{
+	global $g;
+	
+	$count = 0;
+	
+	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "r")) {
+		if(filesize("{$g['vardb_path']}/dhcp6c_duid")==16) {
+			$buffer = fread($fd,filesize("{$g['vardb_path']}/dhcp6c_duid"));
```
rbgarga: At this point you are sure filesize() returned 16 so you don't need to call it again. Instead use 16

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,67 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+	}
+	log_error(gettext("Error: attempting to write DUID file - File write error"));
+	return false;
+}
+/* returns duid string from 'vardb_path']}/dhcp6c_duid' */
+function get_duid_from_file()
+{
+	global $g;
+	
+	$count = 0;
```
rbgarga: Initialize $duid_ASCII

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,67 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+	
+	if ($fd = fopen("{$g['vardb_path']}/dhcp6c_duid", "r")) {
+		if(filesize("{$g['vardb_path']}/dhcp6c_duid")==16) {
+			$buffer = fread($fd,filesize("{$g['vardb_path']}/dhcp6c_duid"));					
+			while($count < 16) {				
+				$a = $buffer[$count];
+				$duid_ASCII .= bin2hex($a);
```
rbgarga: There is no need to set this variable $a, you can pass $buffer[$count] as bin2hex() parameter. And also, $a is not a meaningful variable name

src/etc/inc/util.inc
```diff
@@ -2542,4 +2542,67 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+				$count++;
+				if($count < 16) {
+					$duid_ASCII .= ":";
+				}
+			}
+		}
+		fclose($fd);
+	}
+	//if no file or error with read then the string returns blanked DUID string
+	if($count != 16) {
```
rbgarga: Instead of check for $count it would be better to check using is_duid() to be sure it's a correct value

src/usr/local/www/system_advanced_network.php
```diff
@@ -83,6 +84,18 @@
...
 
+		if (!empty($_POST['global-v6duid'])) {
+			$_POST['global-v6duid'] = strtolower(str_replace("-", ":", $_POST['global-v6duid']));
+			if (!is_duid($_POST['global-v6duid'])) {
+				$input_errors[] = gettext("A valid DUID must be specified");
+			} else {
+				$config['system']['global-v6duid'] = $_POST['global-v6duid'];
+			}
+		}
+		else {
```
rbgarga: else goes in the same line of } as in:
```
} else {
```

marjohn56: Done

src/usr/local/www/system_advanced_network.php
```diff
@@ -186,6 +199,17 @@
 ))->setHelp('By default, if IPv6 is configured and a hostname resolves IPv6 and IPv4 addresses, '. 
 	'IPv6 will be used. If this option is selected, IPv4 will be preferred over IPv6.');
 
+$section->addInput(new Form_Input(
+	'global-v6duid',
+	'DHCP6 DUID',
+	'text',
+	$pconfig['global-v6duid'],
+	['placeholder' => $duid]
+	))->setWidth(9)->sethelp('The current DUID is displayed above. You may enter a new DUID whuch will be used on the next WAN interface UP event.' .'<br />' .
```
phil-davis: s/whuch/which

marjohn56:
Hurrah! :)

chrcoluk:
I noticed an issue.

The code enforces a 16 character length DUID.

However if you start dhcp6 client in debug mode (so it prints current DUID to log), the DUID pfsense generates is only 14 characters long.  So lets say you want a persistent DUID and you paste in the most recent generated pfsense DUID, the box will report the DUID is invalid when its not, as DUID does not have a minimum length of 16 in the specs.

I did notice the ghosted out example DUID started of with 4 extra characters of 0e:00 which I entered and it worked, but those 4 characters are not part of the dynamic DUID generation.

Also its more logical to place this box in the WAN IPv6 section not generic networking as it only applies on wan_dhcp6.

phil-davis:
A DUID can be generated in a number of ways, and is variable length up to 128 octets:
https://tools.ietf.org/html/rfc3315#page-19
https://tools.ietf.org/html/rfc6355

marjohn56:
@chrcoluk It;s not a ghosted out example., The DUID is read in from the existing DUID generated by dhcp6c, If you open the DUID in a hex editor you would see that. And we had a very long discussion as to where the DUID should be. As it's a GLOBAL option, i.e.  it applies system wide then it lives where it lives, otherwise you could have multiple WAN's with difference DUID's, and that''s incorrect.

@phil-davis Spot on!

marjohn56:
just had a look at dhcp6c, it only accepts and generates a type 1 DUID - Link-layer address plus time (DUID-LLT).

phil-davis:
See #3365 for fixes to where the input is validated, and a format_duid() helper function that helps massage various possible user input into the required format.

PR #3305, opened 2016-12-21
Captive portal: convert tables to sortable tables

plumbeo:
Convert the used vouchers table from the old GET order method to a sortable table, and make the other tables (allowed IPs, hostnames, zones) sortable too.


rbgarga:
Already merged. Thanks!

PR #3306, opened 2016-12-22
Fix #7031 Allow interfaces that use DHCP for OpenVPN

phil-davis:
even though the interface (or gateway group) has not yet actually
received an IP address.
This is useful when setting up a new system that is currently offline.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -5972,6 +5972,52 @@ function get_failover_interface($interface, $family = "all") {
 	return get_real_interface($interface, $family);
 }
 
+/****f* interfaces/interface_has_dhcp
+ * NAME
+ *   interface_has_dhcp - determine if the interface or gateway group uses DHCP
+ * INPUTS
+ *   interface or gateway group name
+ *   family - 4 (check for IPv4 DHCP) or 6 (check for IPv6 DHCP)
+ * RESULT
+ *   true - if the interface uses DHCP/DHCP6, or the name is a gateway group which has any member that uses DHCP/DHCP6
+ *   false - otherwise (DHCP/DHCP6 not in use, or the name is not an interface or gateway group)
+ ******/
+function interface_has_dhcp($interface, $family = 4) {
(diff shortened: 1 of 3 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -5972,6 +5972,52 @@ function get_failover_interface($interface, $family = "all") {
...
+								return true;
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	return false;
```
rbgarga: I don't like to add new code containing so deep indentation. I didn't test but I believe it could be simplified as:
```
function interface_has_dhcp($interface, $family = 4) {
	global $config;

	if ($config['interfaces'][$interface]) {
		if (($family == 4) &&
		    ($config['interfaces'][$interface]['ipaddr'] == "dhcp")) {
			return true;
		} elseif (($family == 6) &&
		    ($config['interfaces'][$interface]['ipaddrv6'] == "dhcp6")) {
			return true;
		} else {
			return false;
		}
	}

	if (!is_array($config['gateways']['gateway_group'])) {
		return false;
	}

	if ($family == 6) {
		$dhcp_string = "_DHCP6";
	} else {
		$dhcp_string = "_DHCP";
	}

	foreach ($config['gateways']['gateway_group'] as $group) {
		if ($group['name'] != $interface ||
		    !is_array($group['item'])) {
			continue;
		}
		foreach ($group['item'] as $item) {
			$item_data = explode("|", $item);
			if (substr($item_data[0],
			    -strlen($dhcp_string)) == $dhcp_string) {
				return true;
			}
		}
	}

	return false;
}
```

phil-davis:
Personally I like coding positive conditions on which to do something, and put the something inside the if statement, rather than using "bail out early" conditions (continue,...) to avoid the following code.
But in this case that did result in a large amount of nesting/indent.
I went through refactoring each bit towards what you suggested and ended up with the code you have above, convincing myself at each stage that the real logic was the same. Then tested and it works as good as the original.

PR #3308, opened 2016-12-23
Captive portal: add a pop-up with more details about the session to the active sessions list

plumbeo:
Add a popup with more details (traffic, duration of the session, time left before the session ends or the idle timeout expires, if applicable) to the list of active sessions in the captive portal status page.

schermata

### DIFF
--- a/src/usr/local/www/status_captiveportal.php
+++ b/src/usr/local/www/status_captiveportal.php
@@ -36,6 +36,56 @@ require_once("filter.inc");
 require_once("shaper.inc");
 require_once("captiveportal.inc");
 
+function print_details($cpent) {
+	global $config, $cpzone, $cpzoneid;
+
+	printf("<a data-toggle=\"popover\" data-trigger=\"hover focus\" title=\"%s\" data-content=\" ", gettext("Session details"));
+
+	/* print the duration of the session */
+	$session_time = time() - $cpent[0];
+	printf(gettext("Session duration: %s") . "<br>", convert_seconds_to_dhms($session_time));
+
+	/* print the time left before session timeout or session terminate time or the closer of the two if both are set */
+	if (!empty($cpent[7]) && !empty($cpent[9])) {
+		$session_time_left = min($cpent[0] + $cpent[7] - time(),$cpent[9] - time());
+		printf(gettext("Session time left: %s") . "<br>", convert_seconds_to_dhms($session_time_left));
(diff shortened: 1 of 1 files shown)


src/etc/inc/util.inc
```diff
@@ -1850,6 +1850,16 @@ function format_bytes($bytes) {
 	}
 }
 
+function format_time($seconds) {
```
jim-p: We already have convert_seconds_to_dhms() in pfsense-utils.inc

src/usr/local/www/status_captiveportal.php
```diff
@@ -126,9 +172,9 @@
 		<table class="table table-striped table-hover table-condensed sortable-theme-bootstrap" data-sortable>
 			<thead>
 				<tr>
+					<th><?=gettext("Username")?></th>
 					<th><?=gettext("IP address")?></th>
 					<th><?=gettext("MAC address")?></th>
-					<th><?=gettext("Username")?></th>
```
jim-p: Why move this? Portals without authentication will have no meaningful username.

src/etc/inc/util.inc
plumbeo: Completely missed that, I found only format_bytes() and format_number() in utils.inc and looked for something similarly named. I'll fix the patch.

src/usr/local/www/status_captiveportal.php
plumbeo: They still show 'unauthenticated' in the column and it looked nicer to me to have the usernames/vauchers on the left side. I'll switch it back.

plumbeo:
I pushed a new version that doesn't change the order of the columns (even though I still like better the version with the username first) and uses convert_seconds_to_dhms().

Speaking of columns, given that when MAC filtering is disabled the column is useless and only wastes space, I pushed a second commit that hides it in that case.

jim-p:
Though it's tempting to change lots of things in the same file, we prefer to keep unrelated changes separate so they can be tested and approved independently. Let's keep this one just for the extra info and make a separate PR for the mac filtering fix.

plumbeo:
I moved the commit that hid the MAC column to a new PR (#3310) and left only the one that shows the popup.

rbgarga:
@plumbeo please rebase your fork to resolve conflicts

plumbeo:
Fork rebased.

PR #3309, opened 2016-12-24
Customize DHCP Server on interfaces message

phil-davis:
A suggested in comments for https://redmine.pfsense.org/issues/6930

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -743,13 +743,17 @@ if (is_subsystem_dirty('staticmaps')) {
 $tab_array = array();
 $tabscounter = 0;
 $i = 0;
+$have_small_subnet = false;
 
 foreach ($iflist as $ifent => $ifname) {
 	$oc = $config['interfaces'][$ifent];
 
 	/* Not static IPv4 or subnet >= 31 */
-	if (!is_ipaddrv4($oc['ipaddr']) ||
-	    empty($oc['subnet']) || $oc['subnet'] >= 31) {
+	if ($oc['subnet'] >= 31) {
+		$have_small_subnet = true;
+		continue;
+	}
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_dhcp.php
```diff
@@ -764,7 +768,12 @@ function build_pooltable() {
 }
 
 if ($tabscounter == 0) {
-	print_info_box(gettext("The DHCP Server can only be enabled on interfaces configured with a static IPv4 address. This system has none."));
+	if ($have_small_subnet) {
+		$sentence2 = gettext("This system's static IPv4 interfaces have subnet 31 or 32.");
```
jim-p: Maybe something like this instead:
`sprintf(gettext("The static IPv4 subnet configured on this interface has a CIDR mask of %s, which does not contain enough addresses."), htmlspecialchars($oc['subnet']));`

src/usr/local/www/services_dhcp.php
```diff
@@ -764,7 +768,12 @@ function build_pooltable() {
...
+	} else {
+		$sentence2 = gettext("This system has none.");
+	}
+	print_info_box(gettext("The DHCP Server can only be enabled on interfaces configured with a static IPv4 address with subnet less than 31.") . " " . $sentence2);
```
jim-p: The way CIDR masks and subnet sizes work can confuse some people (larger mask numerically makes a smaller subnet capacity), so I try to avoid referencing it in that way.

Maybe "The DHCP Server requires a static IPv4 subnet large enough to serve addresses to clients." and then in the non-static case the second sentence could be "This interface is not configured with a static IPv4 address."

phil-davis:
The interface reported in the message is an example of one on the system that has a static IPv4 address, but the subnet is too small. There could be more than one like that, but it doesn't seem worth it to make an array of them and report on the whole array. So this will just pick whichever happens to be found last in the list as the example.

PR #3313, opened 2016-12-27
Check for static routes wholly within interface subnet

phil-davis:
The existing code checks for an interface subnet and static route subnet
that are both exactly the same, and puts up an input error for that.
Actually if the static route subnet is contained in the interface
subnet, then we should put up an input error - because that is pretty
dumb to try and static route some part of a local interafce subnet away
out some gateway. Probably it won't break things (the routing will
probably deliver the local packets direct to the local interface anyway,
effectively ignoring the static route). But it seems good to let the
user know that what they are doing is nonsense.
Or is there some use case that I ave not thought about?

Note: If the static route covers a range bigger than an interface
subnet, that is OK. Packets for addresses within the interface are
delivered locally, packets for the rest of the static routed network
will be routed to the specified gateway.

### DIFF
--- a/src/usr/local/www/interfaces.php
+++ b/src/usr/local/www/interfaces.php
@@ -695,7 +695,12 @@ if ($_POST['apply']) {
 
 			foreach ($staticroutes as $route_subnet) {
 				list($network, $subnet) = explode("/", $route_subnet);
-				if ($_POST['subnet'] == $subnet && $network == gen_subnet($_POST['ipaddr'], $_POST['subnet'])) {
+				// If the proposed interface subnet is equal in size, or greater in size (CIDR is smaller) than the static route subnet
+				// and the network address of the static route subnet is in the proposed interface subnet, then there is a problem.
+				// The static routed addresses are contained in the proposed interface subnet, which will be messy!
+				// Note: If the static routed range is bigger than the proposed interface subnet, that is OK.
+				// Addresses in the interface are delivered locally, and outside use the static route.
+				if (($_POST['subnet'] <= $subnet) && ip_in_subnet($network, gen_subnet($_POST['ipaddr'], $_POST['subnet']) . "/" . $_POST['subnet'])) {
 					$input_errors[] = gettext("This IPv4 address conflicts with a Static Route.");
 					break;
 				}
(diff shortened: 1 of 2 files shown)


phil-davis:
Well I am surprised. It would be confusing to me if I routed away a subset of some local interface subnet. Then some time later I might give a fixed IP to some local device in the local interface subnet that happens to be in the static routed range, and I will wonder why the local device does not work (it will send packets up to the router, but replies will be directed out the static route).

jim-p:
That context isn't normally how it's used. Consider a situation where someone is stuck with a /24 on WAN but wants to use a portion of that internally. They could (ugly as it is), setup a proxy ARP for, say, a /26 inside the /24 and then use that on a local interface, VPN, routed across an internal link, etc. There is a potential that through bad network and address management that might break, but that doesn't make it less valid, but more error-prone.

Point is, it can work and does work, but has the potential to be misused. So we can't deny it, but printing a cautionary message somewhere could help avoid unintentional mistakes.

jim-p:
JS might be too intrusive for that. It is valid, and someone using it intentionally might find that annoying. We could probably do something simple like set a flag for this condition and then change the type of message box printed for $savemsg at line 1704 so it shows up differently.

sbeaver-netgate:
Let's just the same mechanism as we use for input errors:


$input_warnings[] = "The interface settings you have selected are unusual, but have been applied anyway";

..
..

if ($input_warnings) {
	print_input_errors($input_warnings, "info");  // New optional argument to specify text attributes. Defaults to "danger"
}

Steve
--

phil-davis:
I'll have a play tomorrow, past my bed time now ;)

phil-davis:
For pages that re-display themselves after Save and present the "Apply" box, this is relatively easy to implement as a extension of print_input_errors() - showing "input warnings" above the "Apply" box.
But many pages do a "save" and go straight back to their "main list" - e.g. press Save in static_routes_edit and if there are no input errors, it saves to the config and directs back to static_routes.php, where the Apply box/button shows. At that point the context of the entry just edited is gone. The "warning" text would have to be stored away somewhere to be read back and displayed by whichever "subsystem dirty" detection thought it was relevant.
It would be nice to have some infrastructure to handle notifying users of "unusual but not forbidden" settings - but seems a bit tricky to implement!

phil-davis:
Leave this for now - I will think about if it is worth it!

rbgarga:
Any news about it @phil-davis ?

phil-davis:
Needs more thought. I keep working on other stuff that requires complete passes through the code base - breadcrumb links, setHelp and gettext formatting...

PR #3315, opened 2016-12-27
Captive portal: rework logging and RADIUS accounting when disabling a zone or rebooting

plumbeo:
Use captiveportal_disconnect_all() instead of captiveportal_radius_stop_all() to disconnect all users, sending complete RADIUS accounting and logging the end of the sessions in the system log.
Since several zones can share the same RADIUS server, send an Accounting-Off packet only when rebooting, not when disabling a zone.

Requires #3300

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -347,14 +347,9 @@ EOD;
 		@unlink("{$g['varetc_path']}/captiveportal-{$cpzone}-error.html");
 		@unlink("{$g['varetc_path']}/captiveportal-{$cpzone}-logout.html");
 
-		captiveportal_radius_stop_all();
+		captiveportal_radius_stop_all(10); // NAS-Request
 
 		captiveportal_filterdns_configure();
-
-		/* send Accounting-Off to server */
-		if (!platform_booting()) {
-			captiveportal_send_server_accounting(true);
-		}
 
 		/* remove old information */
 		unlink_if_exists("{$g['vardb_path']}/captiveportal{$cpzone}.db");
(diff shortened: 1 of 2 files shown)


plumbeo:
After some more tests I verified that calling captiveportal_send_server_accounting() before rebooting is not a problem because even if the Accounting-Off packet can cause the server to set a spurious stop time for sessions in a different zone, eventually we're going to call captiveportal_disconnect_all() for each zone, updating the accounting information with the correct values (at least on FreeRadius 2.2.8).

When disabling a zone instead I'm pretty sure it's only safe when there's a single zone, or the zones never share RADIUS servers, and given that in many cases we won't be able to send it, I think it's better just not to send the packet at all and avoid adding more complications.

rbgarga:
@plumbeo please rebase your fork

plumbeo:
@rbgarga fork rebased

plumbeo:
During tests I found that disconnecting all users can take a long time (70 seconds for ~400 active users here), could this interfere with the reboot page (that has a 90 seconds countdown)?

jim-p:
The reboot countdown timer will reset and try again if it expires. It could accidentally redirect the user back to the dashboard if the reboot hasn't happened yet, though.
If it's going to take that much time, maybe we need to consider having this as an option rather than mandatory. There are people out there that have thousands of users on captive portal at the same time, and if the processing time increases that significantly, the fix may be worse than the problem it's attempting to correct.

plumbeo:
@jim-p yes, that was worrying me too. Reproducing the problem is hard on a test vm but my best explanation is that the delay is due to the removal of each user's rules, one at a time.
Since that's not something we need to do when rebooting, and there's code that takes care of it in a more efficient manner when disabling a zone, I went back to captiveportal_radius_stop_all(), extended it to log the disconnections and fixed it so that now it sends complete accounting data, including a selectable termination cause.
The new patch seems to work and it took only 3 seconds to disconnect ~380 users on a gateway that was showing the problem with the earlier version of the patch.

PR #3316, opened 2016-12-28
Captive portal: fix race conditions when disconnetting users

plumbeo:
While I was doing more tests I found a problem with captiveportal_disconnect_all(). There's a longer explanation in the commit message but basically the problem is that both captiveportal_disconnect_all() and captiveportal_prune_all() read the user database, do a lot of work that with many users can take some time, then update the database, so if they run at the same time one of the two is going to clobber the other.

The first commit papers over the first issue I found by anticipating the update of the database in captiveportal_disconnect_all, so that if captiveportal_prune_all() runs when c_d_a() is working it won't try to disconnect again users that are already being disconnected. That's unfortunately not a solution for c_p_a() because when it starts we have no idea of who's going to be disconnected so we need to wait.

The second commit tries to use locking to avoid the problem once and for all. I found that some functions and modules were using lock()/unlock()/try_lock() from /etc/inc/util.inc, while rc.prunecaptiveportal was using some code that was probably older but did more or less the same things, so I converted it and then used the new lock in c_d_a().

This isn't tested, I didn't even run it, I wanted first to know if something like this is acceptable or if there's another better way to fix the problem that I'm not seeing.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -1089,7 +1089,11 @@ function captiveportal_disconnect_all($term_cause = 6, $logoutReason = "DISCONNE
 	$radiussrvs = captiveportal_get_radius_servers();
 	$cpdb = captiveportal_read_db();
 
-	$unsetindexes = array();
+	/* remove immediately the active users from the database to avoid races */
+	$unsetindexes = array_column($cpdb,5);
+	if (!empty($unsetindexes)) {
+		captiveportal_remove_entries($unsetindexes);
+	}
 
 	foreach ($cpdb as $cpentry) {
 		if (empty($cpentry[11])) {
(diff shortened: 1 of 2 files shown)


plumbeo:
Fixed a couple of obvious mistakes, did more tests and pushed a new version that looks to be working OK.

jim-p:
Seems OK in theory but I'd like to know better how to reproduce the original problem so we have a way to test and see if it fixes the issue.

plumbeo:
@jim-p I'm finding very hard to reproduce the issue on my test VMs, where disconnect_all() takes a few seconds even with hundreds of fake users (created with a script that logins, sends and receives some data, then changes ip and repeat). I could only reproduce it reliably on a gateway with real users and long-lived sessions with hundreds of states each (that's why I suspect that the cleanup causes the delay), so in the end I added a 1s sleep inside the loop in disconnect_all(), to make sure that it lasted enough for prune_old() to start (and vice versa).

PR #3318, opened 2016-12-28
Fix get_notices parameter in xmlrpc.php

phil-davis:
This looks to me like the var name here should be $category
Note: I don't see where this get_notices() inside class pfsense_xmlrpc_server is ever unused anyway - so maybe get_notices() can just be deleted from here?

### DIFF
--- a/src/usr/local/www/xmlrpc.php
+++ b/src/usr/local/www/xmlrpc.php
@@ -550,10 +550,10 @@ class pfsense_xmlrpc_server {
 		if (!function_exists("get_notices")) {
 			require_once("notices.inc");
 		}
-		if (!$params) {
+		if (!$category) {
 			$toreturn = get_notices();
 		} else {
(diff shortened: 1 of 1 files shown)


rbgarga:
@phil-davis did you check on packages? Maybe it's used there. If it's not, I agree it should be removed

phil-davis:
Nothing comes up when using GitHub search for get_notices in FreeBSD_ports or the old pfsense-packages repos. I checked the search functionality, and searching for just "notices" brings up plenty of stuff, so the search works.

phil-davis:
Having the 2 commits here means that the fix will be visible (in case someone cares in future) and then the removal of the code.

jim-p:
IIRC this was, at one time, planned to be used by a central management system to fetch alerts. Likely irrelevant now, but might want to run this by @sbeaver-netgate for good measure.

PR #3319, opened 2016-12-28
Fix #3454 Do not show Mark All as Read button when no priv

phil-davis:
If the user does not have access to index.php then the "Mark All as Read" button for the notices popup does not work for them anyway, so do not show it.
This fixes the obvious UI inconsistency - where the user has a button that they press, but it is not effective.


phil-davis:
Note: After this fix, it would also be good to think about under what circumstances the notices should be shown to users who have restricted privs, and which users should be just able to see the notices but not clear them.

PR #3320, opened 2016-12-28
Remove unused print_notices and print_notice_box

phil-davis:
When looking into how the notices features work these days, I found that these functions are no longer used anywhere.
Maybe they should be deleted?


rbgarga:
@phil-davis Just to be sure, did you check agains packages?

phil-davis:
Not found in FreeBSD_ports or in the old pfsense-packages.

PR #3321, opened 2016-12-28
Fix #7043 Do not display crash reporter link when no crash_reporter a…

phil-davis:
…ccess

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -78,7 +78,7 @@ if ($g['disablecrashreporter'] != true) {
 			if (isAllowedPage("/crash_reporter.php")) {
 				$savemsg .= sprintf(gettext("Click <a href='crash_reporter.php'>here</a> for more information."));
 			} else {
-				$savemsg .= sprintf(gettext("Contact the system administrator for more information."));
+				$savemsg .= sprintf(gettext("Contact a firewall administrator for more information."));
 			}
 			$class = "warning";
 		}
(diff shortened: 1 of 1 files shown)


src/usr/local/www/index.php
```diff
@@ -74,7 +74,12 @@
 		}
 
 		if ($x > 0) {
-			$savemsg = sprintf(gettext("%s has detected a crash report or programming bug. Click <a href='crash_reporter.php'>here</a> for more information."), $g['product_name']);
+			$savemsg = sprintf(gettext("%s has detected a crash report or programming bug."), $g['product_name']) . " ";
+			if (isAllowedPage("/crash_reporter.php")) {
+				$savemsg .= sprintf(gettext("Click <a href='crash_reporter.php'>here</a> for more information."));
+			} else {
+				$savemsg .= sprintf(gettext("Contact the system administrator for more information."));
```
jim-p: Looks good, I would prefer the wording "the system administrator" be changed to "a firewall administrator" but it's not critical.

phil-davis:
Yes, I was thinking to write "network administrator", but "firewall administrator" is even better.

jim-p:
We have many references throughout pfSense and the book that refer to the same thing multiple ways, such as "firewall", "system", "box", "router", and so on. In the book as I go I've been trying to change them all to "firewall" where it makes more consistent sense. It's not something I've been compelled to change en masse, however (yet).

PR #3322, opened 2016-12-28
Add privs to control display of notices

phil-davis:
This is a suggested way to allow control of the display and clearing of
notices. The use case is:
1) A user with minimal page privs (e.g. can just change their password, or access a few status pages or...) should not be automatically able to see notices or clear them. Because notices might contain critical system information which discloses some problem with the system, and clearing them would prevent a full firewall administrator from seeing them.

For this implementation:
a) Users with all pages (admin, anyone in the admins group or with pages-all priv) will see notices and can clear them.
b) Users with user-view-notices can see the notices but not clear them.
c) Users with user-view-clear-notices can see the notices and also clear them.
d) Other users do not see notices.

In its current implementation, this is not totally backward-compatible.
Users who have only a few page privs will see notices prior to upgrade, then after upgrade they will not be able to see notices. The firewall administrator will have to grant them user-view-notices or user-view-clear-notices if they wish these users to see and/or clear notices.

It would be possible to automatically add these privs to existing users as a config upgrade step, but actually I suspect that in 99% of cases the firewall admin would not really want such users to see/clear notices.

Discussion welcome...

### DIFF
--- a/src/etc/inc/priv/user.priv.inc
+++ b/src/etc/inc/priv/user.priv.inc
@@ -38,6 +38,14 @@ $priv_list['user-config-readonly'] = array();
 $priv_list['user-config-readonly']['name']  = gettext("User - Config: Deny Config Write");
 $priv_list['user-config-readonly']['descr'] = gettext("If present, ignores requests from this user to write config.xml.");
 
+$priv_list['user-view-notices'] = array();
+$priv_list['user-view-notices']['name']  = gettext("User - Notices: View");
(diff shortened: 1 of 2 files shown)


phil-davis:
I raised issue https://redmine.pfsense.org/issues/7051 for this feature, so if approved it will have trace-ability through Redmine for making release notes.

doktornotor:
This code is a complete disaster with LDAP/RADIUS auth when the server is unreachable. Basically, if there are any notices pending, it makes GUI unusable.

https://redmine.pfsense.org/issues/7253

PR #3324, opened 2016-12-29
Report problems applying changes

phil-davis:
At present various code keeps track of the return value of functions
that are called to apply changes ($retval). Then calls:
```get_std_save_message($retval)```
But get_std_save_message() always returns the same text "The changes
have been applied successfully."

So why is there are parameter to get_std_save_message()?

If we fix this up, returning different text to indicate that there is a
problem, then the subsequent print_info_box() that actually puts up the
message needs to pass the relevant severity also (success, warning,...).
So to minimize code that needs to be changed/duplicated in many places,
I have made print_apply_result_box() which takes $retval and sorts out
both the message text and severity all in 1 go.

system_routes has the typical changes that would be needed in the
top-level code to implement this.

Without changing top-level code, the changes in guiconfig.inc would
result in the problem text being reported when there is a problem, but
in the "success" color.

### DIFF
--- a/src/usr/local/www/interfaces.php
+++ b/src/usr/local/www/interfaces.php
@@ -407,11 +407,14 @@ if (isset($wancfg['wireless'])) {
 
 }
 
+$changes_applied = false;
+
 if ($_POST['apply']) {
 	unset($input_errors);
 	if (!is_subsystem_dirty('interfaces')) {
 		$input_errors[] = gettext("The settings have already been applied!");
 	} else {
+		$retval = 0;
 		unlink_if_exists("{$g['tmp_path']}/config.cache");
 		clear_subsystem_dirty('interfaces');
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by phil-davis** [comment]
This is up for discussion, including:
1) The exact text of the message when there is a problem - is there always some evidence of the problem in System Logs somewhere?
2) Should the severity for the message box be "warning" or "danger" or?
3) Anything else?

File: src/usr/local/www/interfaces.php [hunk 1]
```diff
@@ -1695,11 +1698,10 @@ function build_gatewayv6_list() {
 					gettext("Don't forget to adjust the DHCP Server range if needed after applying."));
 }
 
-if ($savemsg) {
```
phil-davis: This savemsg code has been sitting unused, because the code for doing "Apply" never set up $savemsg.

phil-davis:
The changes for interfaces.php are an example of a place where there is currently no "applied successfully" message displayed after pressing "Apply".
The changes in this example make the message happen and in the way that is integrated with the guiconfig.inc changes in this PR.

phil-davis:
If the underlying infrastructure for this is agreed, I am happy to go through the GUI PHP files and make the small changes needed, like in the interfaces.php and system_routes.php examples.

rbgarga:
I like the idea. What do you think @sbeaver-netgate and @jim-p ?

phil-davis:
OK, I will look through the top-level UI code that uses get_std_save_message(), make the needed changes and push it...

phil-davis:
Closing in favor of #3343 which has the (current) best/full solution.

PR #3325, opened 2016-12-29
new blank rules - move "any" to top of list

stilez:
@jim-p wanted this split out from PR #3139 as it wasn't related to that PR.

Users should not have to search for the "any option in the list of protocols. This moves "any" at the logical place people look for it (top of list not 2/3 down it at random. keeps default = tcp and ensures  extra ports etc fields visible.

Will probably need rebasing after previous PR merged, but will deal with that at the time.

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -275,6 +275,7 @@ if (isset($id) && $a_filter[$id]) {
 		$pconfig['interface'] = $_GET['if'];
 	}
 	$pconfig['type'] = "pass";
+	$pconfig['proto'] = "tcp"; // for new blank rules, default=tcp, also ensures ports fields are visible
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
Change looks good, but perhaps gettext("Any") (capital 'A') would be better still. All the other options in the selector have caps.

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1217,7 +1217,7 @@ function build_if_list() {
 	'Protocol',
 	$pconfig['proto'],
 	array(
-		'any' => gettext('any'),
+		'any' => gettext('Any'),
```
NOYB: The word "any" is not a proper noun, abbreviation or acronym.  Don't think it should typically be capitalized in this context of a protocols list.

PR #3326, opened 2016-12-29
Fix ntp gps status

jskyboo:
- trim sat in use
- gps ok for GPGLL
- lat & lon direction for GPGGA & GPGLL
- sat in use text position fixed


jskyboo:
Yes if your configuration is set to read from GPRMC(or All and GPRMC is always sent first) you will not see any changes as the problems were mostly with parsing GPGLL and GPGGA. GPGLL wasn't setting gps ok correctly(not showing any of the data it was parsing) and both GPGLL and GPGGA were not reading the lat and lon direction correctly as it was hardcoded to only read the lat and lon direction from GPRMC. The lat and lon direction is only a problem on the ntp status page, the widget works fine(except for not showing GPGLL info at all). As for the trimming of sat in use value and fixing it's alignment, that value is only read from GPGGA so you will only see the field when set to GPGGA. But if your gps can output both GPGGA and GPGLL you should be able to test the changes by setting your NMEA Sentences config to GGA and GLL, respectively.

PR #3327, opened 2016-12-30
Fix inconsistent English in $input_errors[]

stilez:
Some errors use **"cannot"**, others use **"can not"**.  Cannot is probably better here. 

So make all $input_errors[] consistent.

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -572,7 +572,7 @@ if ($_POST) {
 	}
 	if ($_POST['os']) {
 		if ($_POST['proto'] != "tcp") {
-			$input_errors[] = gettext("OS detection is only valid with the TCP protocol.");
+			$input_errors[] = gettext("OS detection is only available for TCP.");
 		}
 		if (!in_array($_POST['os'], $ostypes)) {
 			$input_errors[] = gettext("Invalid OS detection selection. Please select a valid OS.");
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -333,11 +333,11 @@ if ($_POST) {
 			foreach ($config['gateways']['gateway_group'] as $gw_group) {
 				if ($gw_group['name'] == $_POST['gateway'] && $_POST['ipprotocol'] != $a_gatewaygroups[$_POST['gateway']]['ipprotocol']) {
 					if ($_POST['ipprotocol'] == "inet46") {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/firewall_rules_edit.php
```diff
@@ -572,7 +572,7 @@ function is_aoadv_used($rule_config) {
 	}
 	if ($_POST['os']) {
 		if ($_POST['proto'] != "tcp") {
-			$input_errors[] = gettext("OS detection is only valid with protocol TCP.");
+			$input_errors[] = gettext("OS detection is only valid with the TCP protocol.");
```
jim-p: The "P" in TCP is "protocol", this wording reads more redundantly. Maybe cut it out and say "... only valid for TCP".

stilez:
I wonder if "cannot/can not" comes across differently in US v. UK English?  In UK English, "can not" in this kind of notification feels awkward and stilted, rather than just formal. "Cannot" feels informative rather than informal.

jim-p:
It's possible. Sometimes I read them exactly the same, other times I read them with more emphasis on the 'not'. They have identical definitions and "cannot" is a contraction, which is informal and something we try to avoid. I don't think it's worth changing.

raburton:
Apologies for jumping in a random PR, but I do like grammar! In modern English "cannot" is the preferred version of "can not", as @stilez says using "can not" is stilted in most uses. As for it being a contraction, technically maybe (but there is no loss of letters so you could argue otherwise), it's certainly not considered an informal version. "Can't" on the other hand certainly would be and I'd disagree with using that. Nice summary of the opinions of various authoritative bodies (well, OED plus some American ones): http://www.dailywritingtips.com/cannot-or-can-not/

jim-p:
Given the last reference on the page -- "except when you want to be emphatic" (which we do, it's an error), and the "bottom line" section, I'd still leave them alone rather than change them.

raburton:
Fair enough, but in terms of consistency "can not" appears in just 6 code files, "cannot" appears in 105 so looks to be the more common within the project. In firewall_rules_edit.php it's quite mixed, "cannot" is used 13 times and "can not" 11 times, all within the same type of error messages.

stilez:
The "emphatic" refers to emphasising the negation - "you can **not** play until you do your homework", "Flowers can **not** go on the table". Here we don't really need such emphasis - it's a statement of fact rather than a command , that IPv4 addresses can't go in an IPv6 rule, and so on, and we're informing the user of the point, rather than demanding of them to emphatically avoid the opposite. But it's all semantics!

jim-p:
It's all the same. Too much time wasted arguing over nothing.

PR #3328, opened 2016-12-31
Convert ICMP types selector from dropdown to checkbox list in fw_rule…

stilez:
_(See also PR #3139: dropdown boxes used for ICMP type selection, with this to go in a separate PR)_

This converts the UI for firewall rules icmp types to checkboxes. It has a few benefits:

* The dropdown list only showed up to about 4 items. Therefore when IP protocol was changed, it wasn't immediately obvious which (if any) items in the list were still set and which weren't, without manually scrolling the list, as ICMP types vary according to the selected IP Protocol.  With checkboxes it's immediately clear.

* After changing IP protocol the dropdown box also often appeared scrolled/jumped to an apparently-arbitrary position (or changed appearance) because its contents had changed. Checkboxes don't cause this minor UI confusion.

* Dropdown boxes used extra javascript to handle any/none and changes to selected items in the list. With the UI in this PR the code is much simplified.  For example we can skip repopulating the valid ICMP types on IP protocol change, and we don't need any on_change events related to ICMP type selection when a user selects or deselects a specific ICMP type vs. "any".

Code notes for this PR:

* The form fields are a list of checkboxes, returned as an array `$_POST['icmptype_selection']`. This needs a small amount of conversion to and from string format (in $config) but that's trivial and the original code needed it too.

* There is no need for an "any" option any more.  If you want a rule on ICMP it has to be on some or all ICMP types, so an empty icmptype selection is disallowed, and "select/deselect all" buttons are provided for ease instead of "any".  Probably more intuitive?

* When IP protocol changes, the visibility of the various icmptype checkboxes is changed but the checkboxes remain otherwise unchanged. In all, this seems to work better and smoother. It also makes it easier to visually check that as the IP protocol is changed or when incorrect input is provided, the GUI handles it roughly as a user might expect without surprises.

* Logically as ICMP types change according to the selected IP protocol, one can either (1) insert/delete so that the HTML contains those (and only those) icmp types valid for the selected IP protocol, or (2) show/hide the icmptype inputs depending on theiur validity, then filter the form data to exclude data from checkboxes we know would have been hidden due to the choice of IP protocol.<br/><br/>In this PR I switched to the second approach instead. It gives much simpler code and a better result on the UI. For example, upon more than one change to the selected IP protocol, especially after select/deselect all, the UI still displays the intuitively expected result.  It also means that we don't need to trap all the on_change() events to handle a special "any" item or changes to icmptype selection, or to recreate/repopulate the input fields.


phil-davis:
It needs something to reload the user selections of ICMP types after input_errors.
e.g. start with a rule like the attached - ICMP with a couple of type selected in the checkboxes. Edit the rule, change the IP address "1.2.3.4" to some invalid string and press Save. An input error is shown about the invalid IP address (good), but the ICMP types come back all checked.
icmprule

phil-davis:
I guess it will need similar to https://github.com/pfsense/pfsense/pull/3334 which sets the necessary format of ICMP types after "$pconfig = $_POST"

stilez:
PR #3334 fixed it for me.   I've merged the few lines of code from there into this PR.  The variable needed renaming to $icmptype_selection otherwise it's the same added code and should fix the issue mentioned above.

As there were conflicts due to other PRs, it's worth a quick check to ensure that nothing went in (or got removed) other than the few lines intended.

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1287,14 +1299,50 @@ function build_if_list() {
...
+		($data['valid6'] ? ' icmptype_valid_inet6' : '') .
+		(($data['valid4'] && $data['valid6']) ? ' icmptype_valid_inet46' : '');
+	$html .= "<span class='{$extraclasses}' style='padding-right:10px'>{$chkbox}</span>\n";
+}
+
+$buttonall = (new Form_Button(
+	'btn_icmp_selectall',
+	'Select all',
+	null,
+	''
```
sbeaver-netgate: The link and the icon both default to null so

$buttonall = (new Form_Button(
	'btn_icmp_selectall',
	'Select all'
));

is sufficient.

stilez: Done. Also same for "deselectall" checkbox, and removed a completely useless `? true:false' clause a bit below them.

stilez:
@sbeaver-netgate - high praise indeed, which you might want to know, meant a lot to receive unexpectedly. The feedback both means a lot, and helps as I gradually move towards working on substantial chunks and functionality.

jim-p:
That giant wall of checkboxes is a huge step backward, IMO. Even formatted it takes up way too much space. How about using the control that @sbeaver-netgate came up with for the NCP list on OpenVPN clients and servers? It would move the selected types to a separate box to make it more obvious which ones are selected.

stilez:
That's a nice idea. The "selectable/selected" style of lists? Hadn't thought of that. It would also solve the UI issues of a single multi select dropdown, too. I'll knock up a PR for that in  the next few days to compare and see what reviewers think

stilez:
As an aside, maybe @sbeaver-netgate (or anyone else) could convert that ad-hoc code into a bootstrap class that creates a "choose from list A and add/remove from list B" form element, as this is probably something the GUI could use fairly widely with benefit if done?

stilez:
Now done the control as suggested by @jim-p, see PR #3479.  

It does indeed look better, good call!

rbgarga:
Close in favor of #3479

PR #3330, opened 2016-12-31
Improve labels + help text for IPv6 tunneling options

stilez:
* Move the help message to be on the group as a whole not just the IP input field
* Clarify the IP field (label only states "IP address", narrative explains tunelling, but without further field info, the user is left unaided to figure what address should be entered and should it be IPv4, IPv6 or either? This isn't made clear.
* Group title uses the term "tunneling" and this is probably the most commonly understandable term, so keep it consistent and don't switch to "encapsulation" halfway through. (We use the term "encapsulation" in the help text already so it's there for purists)

### DIFF
--- a/src/usr/local/www/system_advanced_network.php
+++ b/src/usr/local/www/system_advanced_network.php
@@ -158,23 +158,25 @@ $section->addInput(new Form_Checkbox(
 ))->setHelp('NOTE: This does not disable any IPv6 features on the firewall, it only '.
 	'blocks traffic.');
 
+
 $group = new Form_Group('IPv6 over IPv4 Tunneling');
 $group->add(new Form_Checkbox(
 	'ipv6nat_enable',
 	'IPv6 over IPv4 Tunneling',
-	'Enable IPv4 NAT encapsulation of IPv6 packets',
+	'Enable IPv6 over IPv4 tunneling',
 	$pconfig['ipv6nat_enable']
 ));
 
 $group->add(new Form_Input(
 	'ipv6nat_ipaddr',
-	'IP address',
+	'IP address of tunneling peer',
 	'text',
 	$pconfig['ipv6nat_ipaddr']
-))->setHelp('Enable IPv4 NAT encapsulation of IPv6 packets. <br/>This provides an '.
-	'RFC 2893 compatibility mechanism that can be used to tunneling IPv6 packets over '.
-	'IPv4 routing infrastructures. If enabled, don\'t forget to add a firewall rule to '.
-	'permit IPv6 packets.');
+))->setHelp('The IPv4 address of the tunneling peer');
+
+$group->setHelp('These options create an RFC 2893 compatible mechanism for IPv4 NAT encapsulation of IPv6 packets, that can be used ' .
+	'to tunnel IPv6 packets over IPv4 routing infrastructures. A firewall rule to allow passing of IPv6 packets ' .
+	'must also be <a href="firewall_rules.php">created</a>.');
 
 $section->add($group);
 
--- a/src/usr/local/www/system_advanced_network.php
+++ b/src/usr/local/www/system_advanced_network.php
@@ -159,7 +159,7 @@ $section->addInput(new Form_Checkbox(
 	'blocks traffic.');
 
 
-$group = new Form_Group('IPv6 over IPv4 Tunneling');
+$group = new Form_Group('IPv6 over IPv4');
 $group->add(new Form_Checkbox(
 	'ipv6nat_enable',
 	'IPv6 over IPv4 Tunneling',
--- a/src/usr/local/www/system_advanced_network.php
+++ b/src/usr/local/www/system_advanced_network.php
@@ -169,14 +169,14 @@ $group->add(new Form_Checkbox(
 
 $group->add(new Form_Input(
 	'ipv6nat_ipaddr',
-	'IP address of tunneling peer',
+	'Tunnel Peer\'s IP Address',
 	'text',
 	$pconfig['ipv6nat_ipaddr']
 ))->setHelp('The IPv4 address of the tunneling peer');
 
-$group->setHelp('These options create an RFC 2893 compatible mechanism for IPv4 NAT encapsulation of IPv6 packets, that can be used ' .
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_advanced_network.php
```diff
@@ -158,23 +158,25 @@
 ))->setHelp('NOTE: This does not disable any IPv6 features on the firewall, it only '.
 	'blocks traffic.');
 
+
 $group = new Form_Group('IPv6 over IPv4 Tunneling');
 $group->add(new Form_Checkbox(
 	'ipv6nat_enable',
 	'IPv6 over IPv4 Tunneling',
-	'Enable IPv4 NAT encapsulation of IPv6 packets',
+	'Enable IPv6 over IPv4 tunneling',
```
jim-p: I'd rather not repeat the text from the label exactly three times there. I prefer the more descriptive text in this field, and it matches what is used in the longer description below.

src/usr/local/www/system_advanced_network.php
```diff
@@ -158,23 +158,25 @@
...
 	$pconfig['ipv6nat_enable']
 ));
 
 $group->add(new Form_Input(
 	'ipv6nat_ipaddr',
-	'IP address',
+	'IP address of tunneling peer',
```
jim-p: "tunnel" is probably better here, so "IP Address of Tunnel Peer" or even "Tunnel Peer IP Address"

src/usr/local/www/system_advanced_network.php
```diff
@@ -158,23 +158,25 @@
...
 	'text',
 	$pconfig['ipv6nat_ipaddr']
-))->setHelp('Enable IPv4 NAT encapsulation of IPv6 packets. <br/>This provides an '.
-	'RFC 2893 compatibility mechanism that can be used to tunneling IPv6 packets over '.
-	'IPv4 routing infrastructures. If enabled, don\'t forget to add a firewall rule to '.
-	'permit IPv6 packets.');
+))->setHelp('The IPv4 address of the tunneling peer');
+
+$group->setHelp('These options create an RFC 2893 compatible mechanism for IPv4 NAT encapsulation of IPv6 packets, that can be used ' .
+	'to tunnel IPv6 packets over IPv4 routing infrastructures. A firewall rule to allow passing of IPv6 packets ' .
```
jim-p: "allow passing of" reads oddly, just "pass" should be good here, so "... rule to pass IPv6 packets ..."

It could also use clarification if it's an IPv6 rule or an IPv4 rule using the IPv6 protocol, which can confuse people in this context.

src/usr/local/www/system_advanced_network.php
```diff
@@ -158,23 +158,25 @@
...
+	'must also be <a href="firewall_rules.php">created</a>.');
```
jim-p: Not too crazy about the passive voice here. Maybe this: "is required." with both of those words linked.

src/usr/local/www/system_advanced_network.php
stilez: It's going to repeat either way with both wordings - either repeating the section label or repeats the help text that immediately follows. My preference is based on the thought that "nat encapsulation" is a more technical term some users will know and some won't (and is in the help text already for those needing it), whereas "tunneling" is immediately obvious to everyone and therefore more easy to recognise if the user is looking for it.

Maybe just "Enable IPv6 over IPv4" rather than either, and trust to the help text and section label to explain the rest of it?

jim-p: Then make the left-side label "IPv6 over IPv4" and the checkbox label "Enable IPv6 over IPv4 Tunneling". I'd still prefer to see the longer checkbox label since it reads much less redundantly. The other text is far enough away the repetition isn't bad, whereas here it's 3x in a row (the group title, option label, and checkbox label).

src/usr/local/www/system_advanced_network.php
stilez: I went for the latter which you sound like you prefer, but added the missing possessive "s" which is correct. Change if you prefer.

src/usr/local/www/system_advanced_network.php
stilez: done, as best I understand you. Is that what you meant?

jim-p: That works, too. It breaks it up a bit either way, so long as they aren't exactly identical.

src/usr/local/www/system_advanced_network.php
```diff
@@ -158,23 +158,25 @@
...
 	$pconfig['ipv6nat_ipaddr']
-))->setHelp('Enable IPv4 NAT encapsulation of IPv6 packets. <br/>This provides an '.
-	'RFC 2893 compatibility mechanism that can be used to tunneling IPv6 packets over '.
-	'IPv4 routing infrastructures. If enabled, don\'t forget to add a firewall rule to '.
-	'permit IPv6 packets.');
+))->setHelp('The IPv4 address of the tunneling peer');
+
+$group->setHelp('These options create an RFC 2893 compatible mechanism for IPv4 NAT encapsulation of IPv6 packets, ' .
+	'that can be used to tunnel IPv6 packets over IPv4 routing infrastructures. A firewall rule to pass IPv6 packets ' .
+	'is <a href="firewall_rules.php">also necessary</a>.');
```
jim-p: why "also"? No other requirement is listed. Just "necessary" or "required" is fine.

src/usr/local/www/system_advanced_network.php
```diff
@@ -158,23 +158,25 @@
...
 	'IPv6 over IPv4 Tunneling',
-	'Enable IPv4 NAT encapsulation of IPv6 packets',
+	'Enable IPv6 over IPv4 tunneling',
 	$pconfig['ipv6nat_enable']
 ));
 
 $group->add(new Form_Input(
 	'ipv6nat_ipaddr',
-	'IP address',
+	'Tunnel Peer\'s IP Address',
```
jim-p: Using the possessive form may technically be more correct but it looks awkward and it reads fine the other way. If it bothers you, go with "IP address of Tunnel Peer" instead.

stilez:
@jim-p @rbgarba is this one okay now? It's still tagged as changes requested

PR #3331, opened 2016-12-31
Enhance security note for privilege assignment pages

stilez:
As commented by Phil Davis on Redmine 2247 (note 3), it might be sensible for some of the privileges to note on the Privilege assignment pages, that they effectively grant root access privileges to any users or groups to whom they are assigned.

Rather than add an extra field to the PRIVS section to state "does it grant root access", I've added a brief note under the privilege selector boxes for users and groups, noting the point and highlighting the most likely permission where it may be overlooked - namely that WebCfg-All pages includes effective root access via command prompt and file editing. 

The note is short, but should be enough for any user setting privileges to get the idea and take it into account when they assign privileges.

### DIFF
--- a/src/usr/local/www/system_groupmanager_addprivs.php
+++ b/src/usr/local/www/system_groupmanager_addprivs.php
@@ -153,7 +153,9 @@ $section->addInput(new Form_Select(
 	build_priv_list(),
 	true
 ))->addClass('multiselect')
-  ->setHelp('Hold down CTRL (PC)/COMMAND (Mac) key to select multiple items.');
+  ->setHelp('Hold down CTRL (PC)/COMMAND (Mac) key to select multiple items.<br /><br />' .
+	   'Security note: Some privileges in this list expose root or other highly privileged access if access is granted.<br />' .
+	    'In particular, <i>WebCfg-All pages</i> exposes root access via the command prompt and edit file pages.');
 
 $section->addInput(new Form_Select(
 	'shadow',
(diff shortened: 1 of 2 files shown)


phil-davis:
I was also looking at similar - I added a text section at the bottom of the page listing all the privs that effectively grant "super user": #3333 
Which way is best???

stilez:
If we're discussing "best ways", my own personal sense would be that, given the privs file is built by a very simple script, it's trivial and neat to add extra priv fields in those files where enhanced privileges will result, so they can be automatically recognised and highlighted programmatically. 

For example:

> ##|+PRIV
> ##|*IDENT=page-diagnostics-edit
> ##|*NAME=Diagnostics: Edit File
> ##|*DESCR=Allow access to the 'Diagnostics: Edit File' page.
> ##|*MATCH=diag_edit.php*
> ##|*MATCH=browser.php*
> ##|*MATCH=vendor/filebrowser/browser.php*
> ##|*WARN=The %s page effectively gives root privilege to any user or group granted access. Please take care when granting this privilege.
> ##|-PRIV

or via a standardised message using:

> ##|*WARN=standard-warning-root

because the message will probably be quite standardised for these pages (allowing standard-* as a macro for a few hard-coded warnings would allow a small list of standardised warnings + custom warnings when needed)

Then as the user selects various privileges, if any contain a WARN field, the appropriate warning is displayed in a field above the "Save" button so they are made aware.

In short, handling via the PRIVS section seems much simpler and more certain of being up to date, than a plain text-box warning?

phil-davis:
Yes, I agree. That way it will build the lists/warnings automagically for the future. I will have a look in the next few hours.

phil-davis:
See #3337 for an attempt at this.

rbgarga:
If we go with #3337 I believe this change is not necessary. Thoughts @jim-p ?

stilez:
Closing in favour of #3341  as the enhanced approach in that PR seems to have been acceptable

PR #3333, opened 2016-12-31
Add privilege information

phil-davis:
to user and group priv forms


phil-davis:
This duplicated code really should go into an include file, so the string of gettext() can all be put together in one place.

phil-davis:
Closing in favor of #3337

PR #3338, opened 2017-01-01
Improve - System - General - DNS Server Settings descriptions

doktornotor:
- Add missing DNS Resolver references 
- Nuke PPTP VPN mentions
- Use letter case consistently.

The "Disable DNS Forwarder" option naming is rather unfortunate as well (would be much better with something like "Use local DNS server"), but changing that would require documentation changes.

### DIFF
--- a/src/usr/local/www/services_unbound.php
+++ b/src/usr/local/www/services_unbound.php
@@ -317,7 +317,9 @@ $section->addInput(new Form_Checkbox(
 	'DNS Query Forwarding',
 	'Enable Forwarding Mode',
 	$pconfig['forwarding']
-));
+))->setHelp(sprintf('If this option is set, DNS queries will be forwarded to the upstream DNS servers defined under'.
+					' %sSystem: General Setup%s or those obtained via DHCP/PPP on WAN'.
+					' (if DNS Server Override is enabled there).','<a href="system.php">','</a>'));
 
 $section->addInput(new Form_Checkbox(
 	'regdhcp',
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_unbound.php
```diff
@@ -317,7 +317,9 @@ function build_if_list($selectedifs) {
 	'DNS Query Forwarding',
 	'Enable Forwarding Mode',
 	$pconfig['forwarding']
-));
+))->setHelp(sprintf('If this option is set, DNS queries will be forwarded to the upstream DNS servers defined under'.
+					' %sSystem: General Setup%s or those obtained via DHCP/PPP on WAN'.
```
jim-p: We've been using the format "Menu Name > Menu Selection" in documentation so we should use that in the GUI as well.

src/usr/local/www/system.php
```diff
@@ -391,7 +391,8 @@ function is_timezone($elt) {
 	))->setHelp(($i == 4) ? 'Address':null);
 
 	$help = "Enter IP addresses to be used by the system for DNS resolution. " .
-		"These are also used for the DHCP service, DNS forwarder and for PPTP VPN clients.";
+		"These are also used for the DHCP service, DNS Forwarder and DNS Resolver " .
+		"(if it has DNS Query Forwarding enabled).";
```
jim-p: This reads a bit odd. How about "... and DNS Resolver when it has DNS Query Forwarding enabled"

PR #3339, opened 2017-01-02
OpenVPN Client 2.4 save correct protocol

phil-davis:
This needs to be fixed the same as vpn_openvpn_server.php


phil-davis:
@jim-p without this, the text description is saved to the config rather than the code, which then breaks the ovpn conf file that is written.

jim-p:
I thought I caught all of those, but the one slippy by, thanks!

PR #3343, opened 2017-01-03
Report problems applying changes

phil-davis:
1) Strictly keep track of the accumulating $retval from calls to various
functions that apply changes.
2) Use new function print_apply_result_box() to print a suitable message
in a suitable severity based on $retval

### DIFF
--- a/src/usr/local/www/firewall_aliases.php
+++ b/src/usr/local/www/firewall_aliases.php
@@ -48,15 +48,8 @@ if ($_POST) {
 		$retval = 0;
 
 		/* reload all components that use aliases */
-		$retval = filter_configure();
+		$retval |= filter_configure();
 
-		if (stristr($retval, "error") <> true) {
-			$savemsg = get_std_save_message($retval);
-			$class = "success";
-		} else {
-			$savemsg = $retval;
-			$class = "danger";
-		}
 		if ($retval == 0) {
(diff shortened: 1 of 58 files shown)


phil-davis:
Notes:
1) I have tested a few samples, I will do more testing and report back (there are 50+ places in the GUI to make a change, press save, apply and see that at least the success message comes out)
2) status_monitoring.php in the ports repo should also be updated to the new scheme (although get_std_save_message() is quite backward-compatible - status_monitoring should not actually break)

phil-davis:
All working for me after fixing up the system_syslogd_start and vpn_ipsec_settings bits in the 2 commits above.

phil-davis:
@rbgarga @jim-p Any chance to merge this before someone else makes a conflicting change in one of these 58 files?

jim-p:
We are still checking it out and talking it over internally. We're hesitant to make such a wide-reaching change without testing each page.

phil-davis:
No problem. I have tried all the pages myself, making sure they run (i.e. no syntax errors) and give the success message OK after saving. The exceptions are:
1) Traffic shaper wizard - didn't run through that.
2) Interfaces assign - didn't work through creating an interface mismatch to generate those interface mismatch messages.
3) In most cases there is no easy (or possible) way to make the save/apply fail, so actually it's hard to demonstrate/test the message for the failure case.

PR #3344, opened 2017-01-03
Add require_state_filter. Implements #7069.

derelict-pf:
By default, the entire state table is displayed when entering Diagnostics > States. This option requires a filter to be entered before the states are displayed. Useful for systems with large state tables.


rbgarga:
Yeah, good idea @jim-p. Can you move it to system.php @derelict-pf ?

rbgarga:
That's something wrong with this PR, your fork seems to be broken. Please re-submit

PR #3345, opened 2017-01-03
Remember authentication between HTTP requests with $_SESSION

scherma:
Well my indentation seems to be messed up, but if you don't mind letting that slide, storing the result of ldap_get_groups and radius_get_groups will mean the LDAP/RADIUS servers don't get hammered with dozens of requests for every page.

### DIFF
--- a/src/etc/inc/priv.inc
+++ b/src/etc/inc/priv.inc
@@ -3,356 +3,273 @@
  * priv.inc
  *
  * part of pfSense (https://www.pfsense.org)
- * Copyright (c) 2004-2016 Electric Sheep Fencing, LLC
+ * Copyright (c) 2004-2016 Rubicon Communications, LLC (Netgate)
  * Copyright (c) 2005-2006 Bill Marquette <bill.marquette@gmail.com>
  * Copyright (c) 2006 Paul Taylor <paultaylor@winn-dixie.com>.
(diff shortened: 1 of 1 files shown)


jim-p:
There is a reason we don't do this. What if an admin revoked the privileges of a lesser admin, but they were still logged in? If it's cached in the session, the ex-admin would retain access.

Also there are a ton of invalid whitespace changes throughout the PR.

scherma:
Have been discussing this in IRC as the issue came when I discovered a single page load creates 200+ lines of ldap log output (and continuous logs on this scale on a page with widgets) which is pretty painful when you want to do log collection. I considered setting an expiry check as well for exactly the reason you suggested, where it would clear the stored data after a short time, e.g. 5 minutes; however I have been discussing this in the IRC channel and it was suggested there that it wouldn't be particularly important to do this. Quite happy to write that up and submit a revised version, what do you think?

jim-p:
If there was a re-auth timer then maybe we can take a closer look, but either way it can't have all the bogus whitespace changes. The LDAP log output may be annoying but being a security product we have to put security first.

scherma:
That's very sensible, next time I'll listen to my instinct more and people in IRC less :) sorry about the whitespace, it's something to do with my vim config. Will make sure revised version doesn't have that issue.

PR #3352, opened 2017-01-05
Minor help text change vpn_openvpn_client

phil-davis:
Make the same change as done for server in https://github.com/pfsense/pfsense/commit/54c90c3527b0a7c596a616cbf6fc58fd6663646a


jim-p:
I went the other way, I intended to have the longer description on both.

PR #3354, opened 2017-01-06
Redmine #7089 Enhance Etc/GMT timezone descriptions

phil-davis:
These work opposite to the way mere mortals expect.
Read: https://github.com/eggert/tz/blob/master/etcetera


phil-davis:
Also see #3355 for some words that could be added to the setHelp.

phil-davis:
@rbgarga The commits you just made based on this look good. 👍

rbgarga:
I've pushed a modified version in e5f4c8297. Thanks

PR #3355, opened 2017-01-06
Redmine #7089 Add extra help about timezone

phil-davis:
We could also add some text like this to encourage users to choose the "Continent/City" time zones.

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -456,7 +456,8 @@ $section->addInput(new Form_Select(
 	'Timezone',
 	$pconfig['timezone'],
 	array_combine($timezonelist, $timezonelist)
-))->setHelp('Select the timezone or location within the timezone to be used by this system.');
+))->setHelp('Select the timezone or location within the timezone to be used by this system. '.
+	'Usually choose a "Continent/City". Only choose a special or "Etc" entry if you understand why you need to use it.');
 
 $section->addInput(new Form_Input(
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system.php
```diff
@@ -456,7 +456,8 @@ function is_timezone($elt) {
 	'Timezone',
 	$pconfig['timezone'],
 	array_combine($timezonelist, $timezonelist)
-))->setHelp('Select the timezone or location within the timezone to be used by this system.');
+))->setHelp('Select the timezone or location within the timezone to be used by this system. '.
+	'Usually choose a "Continent/City". Only choose a special or "Etc" entry if you understand why you need to use it.');
```
jim-p: I'd flip this a bit to strengthen the point, this would change both lines:
`'Select a geographic region name (Continent/Location) to determine the timezone for the firewall. '`
and
`'<br/>Choose a special or "Etc" zone only in cases where the geographic zones do not properly handle the clock offset required for this firewall.'`

Thoughts?

phil-davis: Yeh, that sounds good. @rbgarga already merged this PR, so @jim-p you can easily make those changes yourself straight on master.

jim-p: Ah, I saw it was deleted and wasn't looking at the commit log, figured you had something else cooking. I've pushed the change now.

PR #3356, opened 2017-01-06
Allow hiding of gateways in Gateways Widget

phil-davis:
Sometimes there are gateways that I do not care about, so it is handy to
be able to save some space on the dashboard by hiding them.
This uses similar code structure taken from service_status widget. I
made the Save and Clear buttons and all that stuff use the same styles
also. I guess that the way it was all done in the Services widget is the
more "modern" way.

### DIFF
--- a/src/usr/local/www/widgets/widgets/gateways.widget.php
+++ b/src/usr/local/www/widgets/widgets/gateways.widget.php
@@ -41,9 +41,24 @@ if ($_POST) {
 	if (!is_array($user_settings["widgets"]["gateways_widget"])) {
 		$user_settings["widgets"]["gateways_widget"] = array();
 	}
+
 	if (isset($_POST["display_type"])) {
 		$user_settings["widgets"]["gateways_widget"]["display_type"] = $_POST["display_type"];
 	}
+
+	if (is_array($_POST['gatewaysfilter'])) {
+		$validNames = array();
+		$a_gateways = return_gateways_array();
+
+		foreach ($a_gateways as $gname => $gateway) {
+			array_push($validNames, $gname);
+		}
+
+		$user_settings["widgets"]["gateways_widget"]["gatewaysfilter"] = implode(',', array_intersect($validNames, $_POST['gatewaysfilter']));
+	} else {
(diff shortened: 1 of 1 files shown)


phil-davis:
There are some other widgets where it would be useful to be able to filter the list, e.g.:
Dynamic DNS
Interfaces
and maybe IPsec, OpenVPN and Wake-on-LAN

These things can have long lists in a big system. Allowing filtering would let people customize their dashboard to see just the part of the system they care about. For me it would usually just be filtering out test items that I have set up on a production system but do not matter if they are up or down.

phil-davis:
Firstly, whatever is done here, the corresponding changes will be needed in the Services widget to keep the UI consistent.
Underneath, the user_settings config will need to store the items that are NOT to be displayed. That way an empty user_settings will cause all items to display. So there will be a bit of shuffling between the true/false of the UI and the "flipped" list stored in the user_settings.

The positive principle sounds good, a compact table list of gateway names, each with a checkbox. All the checkboxes would be checked by default, and unchecked means "do not display" and that gateway is written to the user_settings list of hidden/filtered gateways.

The little note "Click the wrench icon to specify which gateways are displayed" could just be displayed only if there are hidden gateways - that will avoid wasting a bit of dashboard space for those who keep the defaults.

Nearly 9pm Friday night here - @sbeaver-netgate if you can do some layout hackwork that would be fine.

Of course no rush for this, it was just an idea I had thought about in the past and got around to playing with today!

sbeaver-netgate:
I have merged this PR so that I can use the config code you have made. I'll drop in the table stuff later tonight.

phil-davis:
@sbeaver-netgate the code you did works fine.
I notice that if I uncheck all the boxes (i.e. try to have no gateways shown in the Gateways widget) then Save, all the gateways are shown. And when I open the settings tool again, all the gateways are checked. I suppose that is sensible - there is not much point letting the user have the widget on the dashboard with nothing in it.

phil-davis:
@sbeaver-netgate tomorrow I will look at updating the Services widget to use the same checkbox controls system. Then I can think about other widgets that could usefully have this kind of filtering of the displayed data.

sbeaver-netgate:
Thank you Phil. That would be good.

PR #3357, opened 2017-01-06
Allow flexible translation of GMT text

phil-davis:
The "s" cannot be hardcoded, because not all languages will append an ASCII "s" to make the plural.
I think this will give translators enough flexibility to make it work.
They can translate "hour" and "hours" and can move the 3 parameters and "GMT" around in the constructed sentence.

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -112,7 +112,7 @@ foreach ($timezonedesc as $idx => $desc) {
 
 	switch ($direction) {
 	case '-':
-		$direction_str = gettext('AHEAD');
+		$direction_str = gettext('AHEAD of');
 		break;
 	case '+':
 		$direction_str = gettext('BEHIND');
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system.php
```diff
@@ -122,10 +122,10 @@ function is_timezone($elt) {
 	}
 
 	$hr_offset = substr($desc, 8);
-	$plural = $hr_offset == "1" ? "" : "s";
+	$hour_word = $hr_offset == "1" ? gettext("hour") : gettext("hours");
 
 	$timezonedesc[$idx] = $desc . " " . sprintf(gettext(
-	    "(%s hour%s %s GMT)"), $hr_offset, $plural, $direction_str);
+	    '(%1$s %2$s %3$s GMT)'), $hr_offset, $hour_word, $direction_str);
```
rbgarga: Good catch, I believe we could be better using ngettext(), that is a function designed to deal with plural cases. What do you think?
http://php.net/manual/en/function.ngettext.php

rbgarga: @phil-davis let me know if you are going to work on it, otherwise I'll

phil-davis: @rbgarga I will have a play...

phil-davis:
See #3358 for a way to do ngettext()

phil-davis:
I assume we will go with some sort of ngettext() - so closing here.

PR #3359, opened 2017-01-06
Allow admin to clear notices

phil-davis:
even if the admin account is not in the admins group.

### DIFF
--- a/src/usr/local/www/head.inc
+++ b/src/usr/local/www/head.inc
@@ -425,7 +425,7 @@ $allow_clear_notices = false;
 if (are_notices_pending()) {
 	// Evaluate user privs to determine if notices should be displayed, and if the user can clear them.
 	$user_entry = getUserEntry($_SESSION['Username']);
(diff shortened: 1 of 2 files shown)


phil-davis:
I refactored some stuff in priv.inc to avoid repeating the code that checks UID 0 for a 3rd time.

src/usr/local/www/head.inc
```diff
@@ -425,7 +425,7 @@ $allow_clear_notices = false;
 if (are_notices_pending()) {
 	// Evaluate user privs to determine if notices should be displayed, and if the user can clear them.
 	$user_entry = getUserEntry($_SESSION['Username']);
-	if (userHasPrivilege($user_entry, "user-view-clear-notices") || userHasPrivilege($user_entry, "page-all")) {
+	if (isAdminUID($username) || userHasPrivilege($user_entry, "user-view-clear-notices") || userHasPrivilege($user_entry, "page-all")) {
```
jim-p: This should be `$_SESSION['Username']`

phil-davis:
I had just quickly tested that existing stuff still worked, then went to sleep.
Should be fixed now.

PR #3360, opened 2017-01-06
Captive portal: make captiveportal_disconnect_all() faster

plumbeo:
captiveportal_disconnect_all() removes the users one at a time and in some cases, when many hundreds of users are connected, can take up to several dozens of seconds to complete.
Instead of looping through all users, send all the accounting information, reset the user database and delete all the active rules and reinit them. Use locking to prevent new users from logging in until the function ends.


plumbeo:
I could only test a slightly different version of the patch on 2.3.2 (captiveportal_init_rules in 2.3 doesn't delete existing rules) but I'm fairly certain it should work as expected, even the locking part (tested adding a 60s delay before releasing the lock).

PR #3361, opened 2017-01-06
openvpn, write config and wait for pidfile inside lock_ex

PiBa-NL:
openvpn, make sure config is written and not overwritten while starting openvpn, and wait for pid of child process to be written before exiting function

This prevents openvpn from getting into a 'unmanageable' state where pid file does not match the running process.

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1245,11 +1245,9 @@ function openvpn_restart($mode, $settings) {
 		$i = 0;
 		$pid = "--";
 		while ($i < 3000) {
-			if (file_exists($pfile) ) {
+			if (isvalidpid($pfile)) {
 				$pid = rtrim(file_get_contents($pfile));
-				if (is_numericint($pid)) {
-					break;
-				}
+				break;
 			}
 			usleep(1000);
 			$i++;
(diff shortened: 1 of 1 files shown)


src/etc/inc/openvpn.inc
```diff
@@ -1238,8 +1240,24 @@ function openvpn_restart($mode, $settings) {
...
+		$pid = "--";
+		while ($i < 3000) {
+			if (file_exists($pfile) ) {
+				$pid = rtrim(file_get_contents($pfile));
+				if (is_numericint($pid)) {
+					break;
+				}
+			}
+			usleep(1000);
+			$i++;
```
rbgarga: Run through this loop 3000 times with usleep(1000) inside looks like too much to me, it can run for a looooong time.

Other than that, isvalidpid() does basically the same operation you are doing and it also checks if process is running. Any reason for not using it?

PiBa-NL: Should take no longer than 3 seconds total give or take a little overhead. Ill change it to using isvalidpid() though didn't realize that existed.

rbgarga: yeah, you are right, I misread the coded

PR #3362, opened 2017-01-06
traffic graph, keep updating traffic graphs also on inactive browser tab

PiBa-NL:
traffic graph, keep updating traffic graphs in webgui also when switching to a different tab in the browser


PiBa-NL:
Doesn't seem like this is the correct fix.. Closing.

PR #3364, opened 2017-01-07
Example breadcrumb links

phil-davis:
This is a way that the breadcrumbs could be populated with links, allowing the user to click backwards from where they came from.

a) In most cases the first thing in our breadcrumbs list is not somewhere that can be clicked-to - e.g "Interfaces" or "Services", the name of a top-level menu hat does not have its own page. So no link to put with that breadcrumb.
b) There seems no point putting a link to the place you are at currently (the end of the breadcrumb list).

So there are only links to be put in when there are 3 or more entries in the breadcrumb chain.

Also, I used $_SERVER['SERVER_ADDR'] to make a link that the remote client browser should be able to use to reach the pfSense. SERVER_NAME would sometimes not work here, because the client browser could be running from somewhere that cannot resolve the server name.
But SERVER_ADDR could also be some private IP address that is not reachable from the client browser - e.g. the pfSense itself is behind some NAT that has a port-forward through to the webGUI. How is that supposed to be done? (giving a client browser a proper full link back to something on the current server)

This is just an initial way to demonstrate what could be done for Redmine 7099. Plenty of discussion to come about what breadcrumb elements to put links on, exactly how those links work, and the best underlying architecture/infrastructure to use to code it efficiently.

### DIFF
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -500,10 +500,20 @@ function genhtmltitle($title, $links=array()) {
 		$bc = '<ol class="breadcrumb">';
 
 		foreach ($title as $idx => $el) {
-			if (strlen($links[$idx]) > 0) {
-				$bc .= '<li><a href="/' . $links[$idx] . '">' .$el.'</a></li>';
+			$href = $links[$idx];
+			if (strlen($href) > 0) {
+				if ($href == '@self') {
+					$href = $_SERVER['SCRIPT_NAME'];
+					if (strlen($_SERVER['QUERY_STRING']) > 0) {
+						$href .= '?' . $_SERVER['QUERY_STRING'];
+					}
+				}
+				if (substr($href, 0, 1) != '/') {
+					$href = '/' . $href;
+				}
+				$bc .= '<li><a href="' . $href . '">' . $el . '</a></li>';
 			} else {
-				$bc .= '<li>'.$el.'</li>';
+				$bc .= '<li>' . $el . '</li>';
 			}
 		}
 
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -510,7 +510,7 @@ function genhtmltitle($title, $links=array()) {
 				if (substr($href, 0, 1) != '/') {
 					$href = '/' . $href;
 				}
-				$bc .= '<li><a href="' . $href . '">' . $el . '</a></li>';
+				$bc .= '<li><a href="' . htmlentities($href) . '">' . $el . '</a></li>';
 			} else {
 				$bc .= '<li>' . $el . '</li>';
 			}
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -492,15 +492,19 @@ function gentitle($title) {
 	}
 }
 
-function genhtmltitle($title) {
+function genhtmltitle($title, $links=array()) {
 
 	// If the array contains only one element, there are no breadcrumbs, so don't
 	// add anything else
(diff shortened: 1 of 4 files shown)


src/usr/local/www/guiconfig.inc
```diff
@@ -492,15 +492,19 @@ function gentitle($title) {
...
 	// If the array contains only one element, there are no breadcrumbs, so don't
 	// add anything else
 	if (count($title) > 1) {
 		$bc = '<ol class="breadcrumb">';
 
-		foreach ($title as $el) {
-			$bc .= '<li>'.$el.'</li>';
+		foreach ($title as $idx => $el) {
+			if (strlen($links[$idx]) > 0) {
+				$bc .= '<li><a href="https://' . $_SERVER['SERVER_ADDR'] . '/' . $links[$idx] . '">' .$el.'</a></li>';
```
stilez: Do we need to manually add "https" or server name here. Currently When pages link to other pages, they mostly just use "PAGENAME.php for the target, and leave the server IP/URL, and the protocol (http/https) as it was. (also will this assumption of https break anything for non https usage?)

jim-p: As above, this should be a relative link or it needs to check the protocol and GUI port and form a full absolute URL.

phil-davis: @jim-p Your are correct - I had been thinking that the remote client would not know what "/services_dhcp.php" is when clicking a link. But there is "magic" somewhere (I guess in the client browser) that fills in the protocol, URL and port bits based on the current page.
I pushed a fix to that.

phil-davis:
I added some code to allow the caller to specify "@self" and the underlying code will generate the reference to the current page, and append the query string (if any). That saves every caller needing to work out the logic of how to do that.
Now the DHCP Server and Static Mapping Edit pages have examples of how this could work, including a link to self for the last entry in the breadcrumb chain.

src/usr/local/www/guiconfig.inc
```diff
@@ -492,15 +492,29 @@ function gentitle($title) {
...
 	if (count($title) > 1) {
 		$bc = '<ol class="breadcrumb">';
 
-		foreach ($title as $el) {
-			$bc .= '<li>'.$el.'</li>';
+		foreach ($title as $idx => $el) {
+			$href = $links[$idx];
+			if (strlen($href) > 0) {
+				if ($href == '@self') {
+					$href = $_SERVER['SCRIPT_NAME'];
```
phil-davis: Pushed a change.

phil-davis:
Next improvement done as suggested by @jim-p

src/usr/local/www/guiconfig.inc
```diff
@@ -492,15 +492,28 @@ function gentitle($title) {
...
+			if (strlen($href) > 0) {
+				// For convenience, if the caller specifies '@self' then make a link
+				// to the current page, including any query string.
+				if ($href == '@self') {
+					$href = $_SERVER['REQUEST_URI'];
+				}
+				if (substr($href, 0, 1) != '/') {
+					$href = '/' . $href;
+				}
+				$bc .= '<li><a href="' . $href . '">' . $el . '</a></li>';
```
jim-p: Almost forgot a big one, this needs to be wrapped in htmlentities(), so `htmlentities($href)`

phil-davis:
Done - getting closer!

phil-davis:
Once there is agreement on the underlying code for this, I will make a PR with single commit for guiconfig.inc and head.inc - that is backward compatible anyway so adds the support without breaking any existing breadcrumb lists.
Then I can do the hack-work to add links for pages.

phil-davis:
See #3366 for single commit.

PR #3365, opened 2017-01-08
Helper format_duid() for DUID input

phil-davis:
format_duid() takes an input string and tries to be nice to the user, massaging the input string as best it can to try and put it into the format that is expected for putting into the config. This should be a convenience for users, specially if they paste a 14-component DUID from a log file into here.

The validation code that set $input_errors[] was in the section that was actually applying the changes. If you input a non-valid string, that would be reported, but also "The changes have been applied successfully."

The validation code has been moved higher up so that non-valid conditions are correctly reported and save is not done.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2570,9 +2570,11 @@ function format_duid($dhcp6duid) {
 	if (count($values) == 14) {
 		array_unshift($values, "0e", "00");
 	}
-	foreach ($values as $idx => $value) {
-		$values[$idx] = str_pad($value, 2, "0", STR_PAD_LEFT);
-	}
+
+	array_walk($values, function(&$value) {
+		$value = str_pad($value, 2, '0', STR_PAD_LEFT); 
+	});
+
 	return implode(":", $values);
 }
 
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -2566,19 +2566,16 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
  * This function does not validate the input. is_duid() will do validation.
 */
 function format_duid($dhcp6duid) {
-	$formatted_dhcp6duid = strtolower(str_replace("-", ":", $dhcp6duid));
-	$values = explode(":", $formatted_dhcp6duid);
+	$values = explode(":", strtolower(str_replace("-", ":", $dhcp6duid)));
 	if (count($values) == 14) {
-		$formatted_dhcp6duid = '0e:00:' . $formatted_dhcp6duid;
-		$values = explode(":", $formatted_dhcp6duid);
+		array_unshift($values, "0e", "00");
 	}
 	foreach ($values as $idx => $value) {
 		if (strlen($value) == 1) {
 			$values[$idx] = "0" . $value;
 		}
 	}
-	$formatted_dhcp6duid = implode(":", $values);
-	return $formatted_dhcp6duid;
+	return implode(":", $values);
 }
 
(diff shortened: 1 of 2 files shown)


marjohn56:
Nice!

src/etc/inc/util.inc
```diff
@@ -2553,7 +2553,35 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+ * "0e:00:00:01:00:01:nn:nn:nn:nn:nn:nn:nn:nn:nn:nn"
+ *
+ * This function does not validate the input. is_duid() will do validation.
+*/
+function format_duid($dhcp6duid) {
+	$formatted_dhcp6duid = strtolower(str_replace("-", ":", $dhcp6duid));
+	$values = explode(":", $formatted_dhcp6duid);
+	if (count($values) == 14) {
+		$formatted_dhcp6duid = '0e:00:' . $formatted_dhcp6duid;
+		$values = explode(":", $formatted_dhcp6duid);
```
rbgarga: It would be simpler to just add desired items in the begin of array:
```
array_unshift($values, "0e", "00");
```

src/etc/inc/util.inc
```diff
@@ -2553,7 +2553,35 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+	}
+	foreach ($values as $idx => $value) {
+		if (strlen($value) == 1) {
+			$values[$idx] = "0" . $value;
+		}
+	}
+	$formatted_dhcp6duid = implode(":", $values);
+	return $formatted_dhcp6duid;
```
rbgarga: Return result of implode instead of assign it to a variable

phil-davis:
Comments addressed. In the end $formatted_dhcp6duid var is not needed at all.

src/etc/inc/util.inc
```diff
@@ -2553,7 +2553,35 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+*/
+function format_duid($dhcp6duid) {
+	$formatted_dhcp6duid = strtolower(str_replace("-", ":", $dhcp6duid));
+	$values = explode(":", $formatted_dhcp6duid);
+	if (count($values) == 14) {
+		$formatted_dhcp6duid = '0e:00:' . $formatted_dhcp6duid;
+		$values = explode(":", $formatted_dhcp6duid);
+	}
+	foreach ($values as $idx => $value) {
+		if (strlen($value) == 1) {
```
jim-p: This length test/add 0 block could be replaced by a single call to `str_pad()` or a `sprintf()` call with a format that would always output two digits.

phil-davis:
str_pad() used. The code is getting more compact.

src/etc/inc/util.inc
```diff
@@ -2553,7 +2553,30 @@ function validateipaddr(&$addr, $type, $label, &$err_msg, $alias=false) {
...
+ * This function does not validate the input. is_duid() will do validation.
+*/
+function format_duid($dhcp6duid) {
+	$values = explode(":", strtolower(str_replace("-", ":", $dhcp6duid)));
+	if (count($values) == 14) {
+		array_unshift($values, "0e", "00");
+	}
+	foreach ($values as $idx => $value) {
+		$values[$idx] = str_pad($value, 2, "0", STR_PAD_LEFT);
+	}
```
rbgarga: Final optimization IMO should be replace this foreach by array_walk:
```
array_walk($values, function(&$value) {
    $value = str_pad($value, 2, '0', STR_PAD_LEFT); 
});
```

PR #3366, opened 2017-01-09
Breadcrumb links support

phil-davis:
Pages can set $pglinks as an array of the links to the relevant pages in the $pgtitle breadcrumb list.
- blank entries in $pglinks array will result in no link for that entry in the chain.
- "@self" entries will generate a reference to the page itself, including any query string. This will usually be used for the last entry in the breadcrumb list, providing an easy place to click and reload/refresh the page.

### DIFF
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -492,15 +492,28 @@ function gentitle($title) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


phil-davis:
This is the backend support for Redmine 7099

rbgarga:
Superseded by #3369

PR #3367, opened 2017-01-09
Require state filter. Implements #7069.

derelict-pf:
Moved to System > General. Also implemented modular field generation function like the rest there, moved the config variable under webgui, and removed the underscores.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -366,6 +366,26 @@ function gen_user_settings_fields(&$section, $pconfig) {
 	gen_pagenamefirst_field($section, $pconfig['pagenamefirst']);
 }
 
+/****f* pfsense-utils/gen_requirestatefilter_field
+ * NAME
+ *   gen_requirestatefilter_field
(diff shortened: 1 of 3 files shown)


src/etc/inc/pfsense-utils.inc
```diff
@@ -366,6 +366,26 @@ function gen_user_settings_fields(&$section, $pconfig) {
...
+function gen_requirestatefilter_field(&$section, $value) {
+	$section->addInput(new Form_Checkbox(
+		'requirestatefilter',
+		'Require State Filter',
+		'Do not display state table without a filter',
+		$value
+	))->setHelp('By default, the entire state table is displayed when entering '.
+		'Diagnostics > States. This option requires a filter to be entered '.
+		'before the states are displayed. Useful for systems with large state tables.');
+}
```
rbgarga: I didn't get the idea of having it on a function

jim-p: Ditto but several other fields on the page are also functions so it looks like it was done to match the style already in use on the page.

PR #3369, opened 2017-01-10
Breadcrumbs detail

phil-davis:
This implements the detail of links for breadcrumbs. I enhanced genhtmltitle() so that if the $links parameter is not passed then it will generate a default link, with the last entry in the page title list having a link to self and the other items having no link. This default scenario is what is needed for "ordinary" pages that are just a simple single page (lots of the Status and Diagnostics pages are like this). It means that I did not have to touch the source code of those pages - the default action of genhtmltitle() produces the correct link.


phil-davis:
Note: This contains the code/commit from #3366 - so if this PR is accepted there is no need for #3366

sbeaver-netgate:
Thanks Phil. Nicely done and long overdue. What is the point of breadcrumbs without hyperlinks?

rbgarga:
It supersedes #3366

PR #3371, opened 2017-01-11
Display client id in DHCP Mapping list

phil-davis:
Do the same as in status_dhcp_leases.php - if there is something entered in the Client Id field then display a Client Id column.
Otherwise for someone using Client Id the list can look confusing, specially if there is nothing in the MAC Address, IP Address or Hostname. A row can just look like an empty row, with maybe a description(if one was entered).

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -1369,6 +1369,17 @@ print($form);
 // DHCP Static Mappings table
 
 if (!is_numeric($pool) && !($act == "newpool")) {
+
+	// Decide whether display of the Client Id column is needed.
+	$got_cid = false;
+	if (is_array($a_maps)) {
+		foreach ($a_maps as $map) {
+			if (!empty($map['cid'])) {
+				$got_cid = true;
+				break;
+			}
+		}
+	}
 ?>
 
 <div class="panel panel-default">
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_dhcp.php
```diff
@@ -1379,6 +1390,13 @@ function build_pooltable() {
 					<tr>
 						<th><?=gettext("Static ARP")?></th>
 						<th><?=gettext("MAC address")?></th>
+<?php
+	if ($got_cid) {
```
rbgarga: Please use if (...): / else: / endif; style in cases where HTML content is printed inside the conditional

src/usr/local/www/services_dhcp.php
```diff
@@ -1402,6 +1420,15 @@ function build_pooltable() {
 						<td ondblclick="document.location='services_dhcp_edit.php?if=<?=htmlspecialchars($if)?>&amp;id=<?=$i?>';">
 							<?=htmlspecialchars($mapent['mac'])?>
 						</td>
+<?php
+			if ($got_cid) {
```
rbgarga: Please use if (...): / else: / endif; style in cases where HTML content is printed inside the conditional

phil-davis:
See #3372 
It is a bit hard to know what standard to apply. status_dhcp_leases.php uses the if{} syntax (where Icopied it from) and other place in services_dhcp.php also use if{}
```if (is_array($a_maps)) {```

PR #3373, opened 2017-01-11
Redmine #5549 Allow variable number of DNS Servers

phil-davis:
1) Modify system.php to allow the user to add as many DNS server entries as they like.
2) Most of the backend code already loops through the dnsserver array (currently of 1..4) to implement writing to conf files - so none of that code needs touching.
3) Modify system.inc code that manages static routes for DNS Servers that have a gateway specified - make it loop through however many there are.

I think this is all that is needed.

Note: There are other places that have limited lists of DNS servers - e.g. DHCP will only let the user specify up to 4 DNS servers to be given to clients. That is a different issue and I think is OK right now (not sure what the max number of DNS servers is that a client can/must accept)

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -198,14 +198,12 @@ function system_resolvconf_generate($dynupdate = false) {
 	}
 
 	/* setup static routes for DNS servers. */
-	for ($dnscounter=1; $dnscounter<5; $dnscounter++) {
+	$dnscounter = 1;
+	$dnsgw = "dns{$dnscounter}gw";
+	while (isset($config['system'][$dnsgw])) {
 		/* setup static routes for dns servers */
-		$dnsgw = "dns{$dnscounter}gw";
-		if (isset($config['system'][$dnsgw])) {
-			if (empty($config['system'][$dnsgw]) ||
-			    $config['system'][$dnsgw] == "none") {
-				continue;
-			}
+		if (!(empty($config['system'][$dnsgw]) ||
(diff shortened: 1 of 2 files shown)


phil-davis:
Yes, I see that. I think the problem is I let the fields for each DNS be dns1, dns2, dns3,... and I think the code that handles Delete Row expects that the first row is zero-based - dns0
I will fix it...

phil-davis:
In the commit I made the GUI processing code use 'dns0', 'dns1'... and 'dnsgw0', 'dnsgw1'... because that zero-based naming with the number at the end of the field name is what works with the underlying repeatable row code.
'dns0', 'dns1'... ends up as a 'dnsservers' array in the config anyway, that just has the default key-numbering of PHP.
'dnsgw0', 'dnsgw1'... has to become 'dns1gw', 'dns2gw'... in the config, to preserve the existing format of the DNS Gateway keys. That is a bit more annoying!
This works as far as I can tell. It could be simplified if we change the way DNS Gateways are stored in the config, propagate that through all code that uses it, and provided config upgrade code.

phil-davis:
Hmmm - I hope that is it!

phil-davis:
Once the design and code for this is reviewed and it is all tested and ready to go, I can easily make another PR with just a single commit - to avoid having all these little "bits-and-pieces" commits end up in the history.

phil-davis:
See #3376 for the same code in a single commit.

rbgarga:
Closing in fabvor of #3376

PR #3375, opened 2017-01-12
Remove excess loops in DynDNS Widget

phil-davis:
I was looking at this code and saw that every time around the outer loop (for each DynDNS entry) the code would call get_configured_interface_with_descr() and return_gateway_groups_array() and spin through various returned arrays looking for matches that could be looked up directly with the array key that we already have.

So I have made it more efficient, but still having the same output.

### DIFF
--- a/src/usr/local/www/widgets/widgets/dyn_dns_status.widget.php
+++ b/src/usr/local/www/widgets/widgets/dyn_dns_status.widget.php
@@ -99,6 +99,9 @@ if ($_REQUEST['getdyndnsstatus']) {
 	exit;
 }
 
+$iflist = get_configured_interface_with_descr();
+$groupslist = return_gateway_groups_array();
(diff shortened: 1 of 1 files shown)


src/usr/local/www/widgets/widgets/dyn_dns_status.widget.php
```diff
@@ -125,32 +128,19 @@
...
-				break;
-			}
+
+		if (isset($groupslist[$dyndns['interface']])) {
+			print($dyndns['interface']);
 		}
 		?>
 		</td>
 		<td>
+		<?=htmlspecialchars($dyndns_providers[$dyndns['type']]);?>
```
rbgarga: Please add a if (isset()) before access the array just to be sure

phil-davis:
I made it display the underlying 'type' value if the array key is not set.

PR #3378, opened 2017-01-13
Add username to logout menu and icon

phil-davis:
I find it difficult to work out who I am logged in as. Other software often provides some indication of who is the current logged-in user. An easy way to do this is to add it to the "Log out" text. Then it can be seen either by hovering over the "Log out" icon, or in the System menu.

I added it to both places. But maybe it should just be associated with the icon, and not the System->Logout menu entry?

I also notice that the System menu has "Logout" (all one word), and the hover text for the icon is "Log out" (2 words). These should be consistent - but which one is preferred?

### DIFF
--- a/src/usr/local/www/head.inc
+++ b/src/usr/local/www/head.inc
@@ -216,7 +216,7 @@ $ext_menu_path_data = read_ext_menu_path_data();
 
 // System
 $system_menu = array();
-$system_menu[] = array(gettext("Logout"), "/index.php?logout");
+$system_menu[] = array(gettext("Logout") . " " . $_SESSION['Username'], "/index.php?logout");
 $system_menu[] = array(gettext("Advanced"), "/system_advanced_admin.php");
 $system_menu[] = array(gettext("Update"), "/pkg_mgr_install.php?id=firmware");
 $system_menu[] = array(gettext("General Setup"), "/system.php");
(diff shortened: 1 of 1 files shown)


phil-davis:
Maybe something like this?

src/usr/local/www/head.inc
```diff
@@ -497,7 +498,7 @@ if (are_notices_pending()) {
 				?>
 					<li class="dropdown">
 						<a href="/index.php?logout">
-							<i class="fa fa-sign-out" title="<?=gettext("Log out")?>"></i>
+							<i class="fa fa-sign-out" title="<?=gettext("Logout") . " (" . $_SESSION['Username'] . "@" . htmlspecialchars($system_url) . ")"?>"></i>
```
jim-p: I was thinking more like `$_SERVER['REMOTE_ADDR']` to show where the user was logged in from (that's what we record in the config when they make changes)

phil-davis:
I am happy for you guys to decide the fine detail of what information to display, make the change directly and close this PR. That will save me messing about making more commits that are just copies of what you are suggesting anyway.
My objective here was just to have somewhere in the UI where I can see what user I am logged in as - I don;t care so much about the exact implementation.

jim-p:
After using yours for a bit I like the way it works. Using the user's own IP address without more context would probably be more confusing. Maybe I'll make a widget just to show the admin login info. The system info widget is getting overcrowded.

PR #3380, opened 2017-01-13
Check for duplicate Wake on LAN entries

phil-davis:
Currently it is posible to enter an interface+MAC combination that is already in the WoL list. That seems silly.


NOYB:
Why does it need to be prevented?  What harm is there?

phil-davis:
Nothing actually goes wrong, but without the check it lets users put unneeded crud into their config.
I noticed it when putting in test entries to play with when adding the filtering checkboxes to the WoL widget, it looked odd having multiple entries in the list that looked the same.

PR #3384, opened 2017-01-15
traffic-graphs, setting to keep updating them while invisible

PiBa-NL:
traffic-graphs
- setting to keep updating them while invisible
- allow showing different graphs to be shown on different browser tabs (dont use localstorage for graphs to query)
- show interface name in graph instead of realname

### DIFF
--- a/src/usr/local/www/js/traffic-graphs.js
+++ b/src/usr/local/www/js/traffic-graphs.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-function draw_graph(refreshInterval, then) {
+function draw_graph(refreshInterval, then, backgroundupdate) {
 
 	d3.select("div[id^=nvtooltip-]").remove();
 	d3.select(".interface-label").remove();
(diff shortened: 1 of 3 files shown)


jdillard:
So far everything looks good. The only catch is that it will crash the browser tab is left open for a long enough period of time (typically hours). I wonder if the message should be changed to reflect that (versus mentioning cpu usage) or if that option should be marked as "beta".

PiBa-NL:
Might be that if left open long enough it will crash the browser, i did not check that in detail. However i think that would then also happen with the graphs as they are currently present right if kept on a 'active' window? Still worth fixing of-course assuming its a leak in the Javascript somewhere and not a browser bug.., but that is imho outside of the purpose of this pull-request.. If adding the text 'beta' helps getting this pulled ill of-course do so.

jdillard:
It's not a show stopper. I think it is more a browser issue, but if a known issue when running javascript in the background for a long period of time. That was one of the reasons for have it not run in the background and then start again when the tab became active. At least now people have the option to choose the lesser of the two evils that works for them. 

Thanks for make the PR! I think we can deal with any fallout that may come from tabs crashing and deal with them accordingly. I'll get this merged tomorrow.

PiBa-NL:
Oké thanks, i tried 'profiling' the Javascript used, but couldn't find any obvious leak so far. Ill try a little more, but if its a browserbug i suppose there is little we can do besides report it to the relevant vendor..

jdillard:
Same here. I have it set to 'clear graphs' currently and I am leaving it open in a separate browser to see if that crashes, but haven't had the time to do anything more advanced yet. At some point I wanted to see if it was possible to compare timestamps and reset the graphs every 45min or so if run in the background to see that might help. Regardless, we should probably start the discussion about that problem/solution in a new redmine ticket.

jdillard:
Here is the redmine ticket to discuss potential issues for the inactive tab issue: https://redmine.pfsense.org/issues/7157

PR #3387, opened 2017-01-16
Fix #7126 RFC2136 links

phil-davis:
It needs 3 counters - the row counter so the AJAX can put the responses back in each row, plus the counters that index into Dynamic DNS and RFC2136 config entries for double-click to edit.


phil-davis:
Note: This is the fix for prior to adding the filtering to the widget.
I will do a fix for after the filtering now...

PR #3391, opened 2017-01-16
Status / Interfaces - Relinquish DHCP Lease

NOYB:
Send DHCP release message to the server and update leases, etc. so that a DHCP Discovery will be used instead of a DHCP Request for the previous address when renewing.

Fixes #4209, #6070

Releasing DHCP on WAN interface should send a release
https://redmine.pfsense.org/issues/4209
01/13/2015

DHCP release does not send proper DHCP Release packet to DHCP server
https://redmine.pfsense.org/issues/6070
04/04/2016
duplicate of #4209

### DIFF
--- a/src/sbin/dhclient-script
+++ b/src/sbin/dhclient-script
@@ -352,7 +352,7 @@ BOUND|RENEW|REBIND|REBOOT)
 	fi
 	;;
 
-EXPIRE|FAIL)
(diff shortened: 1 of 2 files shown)


rbgarga:
Waiting for https://github.com/pfsense/pfsense/pull/3390

PR #3392, opened 2017-01-17
Add Filter to OpenVPN Widget

phil-davis:
Final part of Redmine 7122


phil-davis:
This one was easier than I expected, the backend already keeps a unique 'vpnid' numbered field across all OpenVPN instances/types, so it was easy to use as the unique key to store in the list of items to be filtered.

rbgarga:
Looks like get_openvpnent_key() is unnecessary in this case, no? vpnid key can be used direct from arrays

phil-davis:
True, by the time I got to the end of coding the unique id to use became trivial.

PR #3393, opened 2017-01-17
Do not nag people about passwords mismatch when notifications are disabled (Bug #7129)

doktornotor:
These fields tend to get "helpfully" pre-filled with random crap by built-in browser password managers or extensions such as LastPass. No need to bother people with these when the feature is disabled.

### DIFF
--- a/src/usr/local/www/system_advanced_notifications.php
+++ b/src/usr/local/www/system_advanced_notifications.php
@@ -129,7 +129,10 @@ if ($_POST) {
 			if ($_POST['smtppassword'] == $_POST['smtppassword_confirm']) {
 				$config['notifications']['smtp']['password'] = $_POST['smtppassword'];
 			} else {
-				$input_errors[] = gettext("SMTP passwords must match");
+				if ($_POST['disable_smtp'] != "yes") {
+					// Bug #7129 - do not nag people about passwords mismatch when SMTP notifications are disabled
+					$input_errors[] = gettext("SMTP passwords must match");
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


doktornotor:
A couple of notes:

- how on earth do I make the checkbox checked by default? (That'd be what about 99% of pfSense users need)
- WTF is this DMYPWD thing? ```if ($_POST['password'] != DMYPWD) {```

rbgarga:
I don't like the idea of adding these kind of exceptions. Extensions like lastpass can be disabled for specific pages. Lets see what other developers think about it

doktornotor:
Well I don't like them either, but it's just incredibly annoying. The other alternative would be disabling the rest of the fields when the feature is disabled, except that it's the exact opposite of what is desired here: _"Check this option to disable growl notifications **but preserve the settings below**."_

jim-p:
I don't like exceptions like this either but the way browsers and password managers spam form fields I'm not sure there is a better choice for usability.

Sadly, LastPass and many of its users feel that allowing web developers to control autocomplete behavior is a bad thing, and that they know better than web developers what they want to do with a form. Search their forum and you'll find people claiming that to allow this would be some sort of apocalyptic event where sites (gasp!) disable saving passwords on login forms. Even when a browser or LastPass guesses wrong and fills in the wrong thing without a bunch of manual work.

There isn't a good way I know of for the server side of things indicate that a page isn't a login page and that the client shouldn't be trying to autofill the form. I know it's possible, since there are still sites out there that somehow manage to evade autofill, but some of the odd JS-based methods of avoiding autocomplete are worse than the original problem.

> WTF is this DMYPWD thing? if ($_POST['password'] != DMYPWD) {

The "DMYPWD" bit is a part of the standard password/confirm password control that we have which does not put the user's actual password back onto the page after it has been saved, for security.

doktornotor:
Well, I did the same with SMTP so that the code is consistent on the page. Other than that, it seems I've sort of rediscovered https://redmine.pfsense.org/issues/5087 -- but I recall this wasn't an issue here at that time except for Chrome, apparently things are getting gradually worse with browsers/password managers :(

PR #3394, opened 2017-01-17
RTSOLD lock creation, dhcp6c launch & kill changes,

marjohn56:
Added lock file creation and check to RTSOLD script creation. This is to
prevent mutliple launches of dhcp6c, this appears to happen when multiple
RA's are received in rapid succession at the start of a session. Once
created dhcp6c cannot be launched again until the lock file is deleted,
this is done within the kill_dhcp6_client process locking the two
together.

The kill -9 used to kill the dhcp6c client is changed to -15, -9 causes
the process to exit without sending a release if required, and if the
timing is just rignt can cause the pid file to be left behind; -15
allows for a graceful exit and if the release flag is not set then it
sends and waits for the release confirnation.

The launch of dhcp6c when in dhcp6withoutRA is moved to its own
function, as much as anything this makes the code tidy around the bottom
of interface_dhcpv6_configure().

These changes require the use of a new dhcp6c client, the PR for this is
#258 and this has also been PR'd upstream. The new client contains an
extra flag '-x' . This flag is used to prevent the execution of the same
script by dhcp6c and RTOSOLD at the same time, the flag causes dhcp6c to
not execute the script on receipt of a reply to a request. In
dhcp6cwithoutRA mode then RTSOLD will execute the script after dhcp6c
has been launched and when the RA has been received. This should also
prevent a problem where some ISP's BNG's are tardy when it comes to
replying to the RS.

This has been tested to death on ISP's that require dhcp6withoutRA and I
have tested it using a second pfSense unit as a sub off my main..

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3022,13 +3022,40 @@ function find_dhcp6c_process($interface) {
 }
 
 function kill_dhcp6client_process($interface) {
+	global $g;
+
 	if (empty($interface) || !does_interface_exist($interface)) {
 		return;
 	}
 
 	if (($pid = find_dhcp6c_process($interface)) != 0) {
-		mwexec("kill -9 {$pid}");
-		sleep(1);
+		/* Kill -9 caused the pid to get left behind, also if we need a */
+		/* relase sent then it needs to be -15, this then allows dhcp6c */
+		/* to send the release, it will also clean up after itself */
+		// Debug - next line remove if not required
+		mwexec("/usr/bin/logger -t info 'shutting down dhcp6c process'");
+		mwexec("kill -15 {$pid}"); 
+		sleep(2); //Allow dhcp6c to send releae and exit gracefully if needed.	
+	}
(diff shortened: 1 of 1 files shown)


rbgarga:
Depends of https://github.com/pfsense/FreeBSD-ports/pull/258

marjohn56:
It appears upstream is not too keen on pulling this, Rather than have a separate client for pfsense I am testing something else. 

What I am trying to achieve is the stopping of the run of the wan script and thus newwanipv6 on the initial and only initial dhcp6c trigger. Having thought a bit more about it I can achieve the same thing by creating a temporary file that gets created in the wan6 dhcp configure and gets deleted on the first run of the wan script. This way any subsequent calls to run the script by dhcp6, rebind etc are not affected. On that initial call RTSOLD will run the script but the temporary file will already have been deleted by the run of dhcp6c. Either way, the script will only call newwanipv6 once.

marjohn56:
Please see Redmine #5993

marjohn56:
Need to close this one now as a lot of alterations have been made. New PR keeping the good stuff to follow.

PR #3399, opened 2017-01-19
Allow control of Gold menu item display

phil-davis:
Mentioned in forum post https://forum.pfsense.org/index.php?topic=123520.0
At the moment, if the user has "all pages" or "all privs"... then the Gold menu entry is displayed. Otherwise it is not (just an empty Gold menu header is shown.

This is a way to let the display of "Gold" menu items be controlled, similar to the existing "Help" pages priv. It makes the menu-priv system a bit more consistent, whatever you think of the existence of the Gold menu.


jim-p:
We talked this over internally and decided it would be better to go a different way: The Gold menu and its entries should always show, as well as the Help menu items because they are external links. See b29aa6e57b941577c2f9faa724134d5bf1fe3482

phil-davis:
Yep, that works fine.

PR #3400, opened 2017-01-19
Fix #7139 Accessing help about this page

phil-davis:
from a user that does not have admin or all page privilege.


phil-davis:
In priv.inc cmp_page_matches() it was testing a match between "*help.php" and a Request URI string like "help.php?page=services_dhcp.php" - they do not match and so access to the help.php page was denied.

The only page with a name like "*help.php" is "help.php" itself, so the leading "*" is not needed.

PR #3401, opened 2017-01-19
Fix #7140 User is redirected away when help is their first page priv

phil-davis:


phil-davis:
This seems a bit of a hack here, having to hardcode a test for "help.php".
But what is better - sort the help.php entry to the end of the list? And then still have to handle the case where it is the only entry in the allowed pages list.

phil-davis:
See #3412 for a way to handle this entirely in help.php - inspired by @jim-p comment above.

jim-p:
Closing in favor of #3412, I just merged that one it will show up momentarily. Thanks!

PR #3406, opened 2017-01-20
Mechanism to change required field from JavaScript

phil-davis:
Sometimes a field is only required based on the value of some other field.
In many cases the required field/fields are hidden and only displayed when they are required (e.g. as different IPv4 Configuration Type is selected for an interface, a whole relevant block of input fields is shown and other blocks are hidden.) For those cases it is easy to hard-code the required fields (e.g. pppoe_username in the diff of interfaces.php). Whenever the field is visible it is required.
But sometimes the field is optional, but becomes required depending on the value of other fields - e.g. pppoe_idletimeout is required when pppoe_dialondemand is checked. So the element-required class needs to be added to or removed from the pppoe_idletimeout label as pppoe_dialondemand changes.

Here is one way to do it:
1) Add a unique id to the span tag of the label of the group item, so it can be found and referenced - I made and id starting with "pfgui-label-span-".
2) Do this for every label, so the labels can be referenced if some class needs to be added to them later.
3) Add JS to reference the id and addClass('element-required') removeClass('element-required') as the user does stuff in the UI.

There might be much cleaner ways to do this - happy to get suggestions.

### DIFF
--- a/src/usr/local/www/classes/Form/Group.class.php
+++ b/src/usr/local/www/classes/Form/Group.class.php
@@ -154,8 +154,11 @@ EOT;
 			// 'element-required'. Text decoration can then be added in the CSS to indicate that this is a
 			// required field
 			if (substr($title, 0, 1 ) === "*" ) {
-				 $title = '<span class="element-required">' . substr($title, 1) . '</span>';
+				$special_class = ' class="element-required"';
+				$title = substr($title, 1);
 			}
+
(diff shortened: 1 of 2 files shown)


sbeaver-netgate:
See https://redmine.pfsense.org/issues/7083 notes.

I added a setRequired() function to the JS which I think might be a little cleaner/more consistent way to do this.

phil-davis:
@sbeaver-netgate Thanks for the setRequired() function. It works.
See #3409 for an example of how this kind of thing would be coded now. Put discussion there.

PR #3407, opened 2017-01-20
Captive portal: allow to use RADIUS accounting with every authentication method

plumbeo:
This patch series allows to enable RADIUS accounting for every authentication method (or lack thereof).

The first patch, functional on its own, adds a new RADIUS context for accounting and an option in the captive portal zone's settings page to use the new context instead of sending the accounting information to the primary RADIUS authentication context. To preserve the old behaviour the new setting that controls where the accounting packets are sent is set on upgrade to the correct value to keep sending them to the authentication server if accounting was already enabled.

The second patch changes the settings page to show the Accounting section independently of the Authentication section, hides and shows the opportune RADIUS options accordingly and makes sure that the appropriate contexts are always recorded when a user logs in to the captive portal.

The "Fix typo" commit fixes an unrelated typo that I spotted casually, it's here because opening a separate PR for it didn't look worth it.


plumbeo:
@jim-p using a different server for authentication and accounting is less useful, but not unheard of (saw a case where the authentication server was integrated in an appliance and didn't support accounting).
Being able to send the accounting to a RADIUS server even when not doing RADIUS authentication, or not authenticating at all, can be very important, it's basically the only way to have decent usage statistics. I started working on this because a few weeks ago to set up a unauthenticated captive portal (welcome page with usage terms, then free navigation) I had to write a custom login page with hard-coded username and password to make the user do a fake authentication and finally get the accounting information.

WRT 2.5, is 2.4 supposed to be out so soon that this one can't go in? The captive portal right now isn't completely functional, both idle timeout and traffic counters don't work and there's something wrong with RADIUS MAC authentication too, so it looks like there's a lot of work needed, and not particularly simple stuff.

jim-p:
It may be useful to you but it's a large change and very few (any?) others would utilize it. Given the size of it and potential impact, we need to focus on the parts of 2.4 that, as you have noticed, are still not functional. Adding more complexity to this code during that time would be harmful and slow the process down. Later on, perhaps.

plumbeo:
@jim-p given that every firewall/captive portal/wireless controller I've seen supported separate authentication and accounting servers (Cisco, Ruckus, HP, Aruba, Juniper, Nomadix, etc.) I'm reasonably sure I'm not the only one who finds this feature useful, nor the only one who would use it if it was available... anyway, I'll open a new PR after 2.4 is out.

plumbeo:
I checked the release history and since it looks very unlikely that 2.5 will be out before the fist half of 2018 and I think this is a useful feature, I wanted to make the case for this PR in a more organized fashion.

1) Having separate authentication and accounting management and servers is a standard feature available in almost all devices who have a RADIUS backend, see for example Cisco WLC (http://www.cisco.com/c/en/us/td/docs/wireless/controller/7-4/configuration/guides/consolidated/b_cg74_CONSOLIDATED/b_cg74_CONSOLIDATED_chapter_0101100.html) and ASA (http://www.cisco.com/c/en/us/td/docs/security/asa/asa91/configuration/general/asa_91_general_config/aaa_overview.html), Juniper (https://www.juniper.net/techpubs/en_US/junos14.1/topics/concept/radius-accounting-understanding-qfx-series.html), Aruba (https://community.arubanetworks.com/t5/Controller-less-WLANs/How-to-configure-separate-radius-and-accounting-servers-in-IAP/ta-p/214987), HP (http://h20628.www2.hp.com/km-ext/kmcsdirect/emr_na-c02566413-2.pdf#G7.1147705). 
The rationales for using separate servers are many: performance (the authentication servers are generally only used at the start of a session while the accounting servers can receive packets for all the duration, so separating them can help when there are many users), flexibility (one can use a central authentication server, and have multiple accounting servers, for example one for each remote facility), integration with devices that don't support the full RADIUS protocol (for example, they proxy the requests to an internal LDAP server but can't do accounting).

2) Accounting isn't needed only when doing RADIUS authentication. Right now when using vauchers or doing unauthenticated access, usage data is available internally, so it's possible to see the amount of traffic sent and received, or see a graph of the number of active users, but this information, that can be critical when planning and managing a captive portal, isn't accessible from the outside nor it can be easily retrieved and aggregated from multiple devices (short of processing the text logs, which is not exactly the optimal solution). RADIUS is the de facto standard for accounting and billing, so using it to export these data makes easy to integrate pfSense with any pre-existing tools already used to process and analyze them. Again, this isn't something that someone with a single smallish firewall and a dozen users is likely going to need (though it could be useful even in this case), but it could easily become important on more complicated deployments.

3) The code changes in this PR are apparently complex, but in reality they're pretty straightforward. The first patch has the bulk of them, and it only does two things: it creates a new field in the active users database ('acctctx') and uses it to save a new RADIUS context similar to the existing ones ('first' and 'second') and it makes sure that all code paths that include a call to RADIUS_AUTHENTICATION() and RADIUS_ACCOUNTING_{START,STOP}() use the appropriate context, adding a new parameter to some functions if needed; the rest is just boilerplate code to add the necessary configuration options and ensure that all the required fields are filled and those that aren't needed are hidden.
The second patch is even simpler: the most significant changes are three lines that add the accounting context to the code paths followed when there's no or local/voucher authentication, while the bulk of the code is just javascript to show or hide the appropriate configuration options in the settings page.

4) I tested all the possible combinations of authentication and accounting modes and I couldn't find any functional difference unless RADIUS accounting is enabled. I'd argue that even if the new functionality isn't used the code after this PR is more deterministic and more likely to work correctly than the old one. The old code did a lot of stuff like arbitrarily setting the "first" RADIUS context in random places when the field in the database was empty. That more or less worked, but it doesn't really take a lot to find weird corner cases: for example, if you enable RADIUS authentication, then switch for a while to "no auth", let some users log in and reenable RADIUS authentication, suddenly you start getting RADIUS accounting packets and trying to reauthenticate on RADIUS "unauthenticated" users. This isn't a contrived scenario, that's what anyone would do if he needed to temporarily give access to everyone on a network controlled by a captive portal and then restrict it again. 
Similar issues caused problems in the past and were "fixed" using workarounds like checking that the username was "unauthenticated" (but what if a real user use that as the username, or if someone decides to localize that string without thinking of the implications?) or PR #2127, that arbitrarily changed the RADIUS context of vaucher users to prevent reauthentication issues. With this PR merged whenever a user logs in the session is recorded in the active sessions' database with the right values of authentication and accounting context, that is, no matter what happens next, the code will never send unasked for authentication or accounting requests.

The next steps after this would be to use the authentication context recorded in the database to make decisions instead of using the attribute trick or checking the username or a runtime value (or, alternatively, making sure that when things change that require it, all the users are disconnected, though that's not how things work now and someone could be relying on the old behaviour) and start using explicit parameters instead of relying so much globals (it already caused bugs like the one fixed in PR #3315), but to be honest I feel a bit hesitant about spending the time needed to do that.

@jim-p @rbgarga

sbeaver-netgate:
Closing since no action in 11 months. Please revise/resubmit if still desired.

PR #3408, opened 2017-01-20
Add all protocols listed /etc/protocols for rules editing

iz3gme:


rbgarga:
Some protocols require specific validation, it's not that simple like just adding all of them to a list.

Also, Pull Requests must be filled against branch 'master' instead 'RELENG_2_3_2'

iz3gme:
If you give me more details i'll try to implement needed validations (or exclude those protocols) and resubmit to 'master'

PR #3410, opened 2017-01-20
RTSOLD lock creation, dhcp6c launch & kill changes #3

marjohn56:
Added lock file creation and check to RTSOLD script creation. This is to prevent multiple launches of dhcp6c, this appears to happen when multiple RA's are received in rapid succession at the start of a
session. Also it seems that on startup interface_dhcpv6_configure() appears to get called multiple times, the locks prevent multiple instances in that case also. Once created dhcp6c cannot be launched again until the lock file is deleted, this is done within the kill_dhcp6_client process locking the two together.

The kill value used to kill the dhcp6c client is now variable.  The value -9 causes the process to exit without sending a release if even required, and if the timing is just right can cause the pid file to be
left behind; -15 allows for a graceful exit and if the release flag is not set then it sends and waits for the release confirmation, the value will now switch between those two values depending on the configuration option 'No Release'. If no release is true then -9 is used. Any left behind pid is removed by an 'rm *' statement at the end of the kill routine. This change will make it possible to stop the use of the -n flag, thus allowing for dcp6c to send a release manually, if so required; a future update will add this if needed.

The launch of dhcp6c when dhcp6withoutRA is selected is moved to its own function, as much as anything this makes the code tidier around the bottom of interface_dhcpv6_configure().

A completely new method of implimenting dhcp6wihtoutRA is used! 

Initially this PR was going to use a modified dhcp6c, but as that met some resistance from upstream I went back to square one and looked at what needed to be achieved to get the same results but without the need for a modified client, the result is a complete re-work of the dhcpwithoutRA mode. The origin of this dates back to the beginning of Redmine #5993 and the need for dhcp6 before RA, the patch was a workaround to get pfSense working with my ISP. Subsequent patches have been workarounds to fix problems with the workaround.

 In default mode RTSOLD launches dhcp6c, in the new method In dhcp6wihtoutRA mode dhcp6 launches RTSOLD.

In simple terms its dhcp6->rtsold->WAN6_configure. 

The scripts are created dynamically and change depending on the mode, in default mode there are no changes. In dhcp6withoutRA the scripts change and a new dhcp6c_wan_dhcp6withoutra_script is created. The dhcp6c.conf file is changed to call this script.

It has been taken into account that the domain-name & server variables created by dhcp6 and passed to the original script it calls are passed on. In default mode this is simple as it calls the dhcp6c_*_script which and the vars are passed directly rc.newwanipv6, in dhcp6withoutRA its RTSOLD that calls the dhcp6c_*_script, so in order to make this change work the variables are echoed to the tmp folder and retrieved by the dhcp6withoutRA created version of dhcp6c_wan_script when that calls rc.newwanip. As my ISP does not pass DNS or Domains in the dhcp6 reply I cannot test this directly, but dummy values are being passed so I am assuming it works correctly.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3052,8 +3052,8 @@ function kill_dhcp6client_process($interface,$norelease) {
 		/* to send the release, it will also clean up after itself */
 		// Debug - next line remove if not required
 		log_error("shutting down dhcp6c process");
-		$killtype = isset($norelease) ? "-9" : "-15";
-		mwexec("kill {$killtype} {$pid}"); 
+		$sig = (isset($norelease) ? SIGKILL : SIGTERM);
+		posix_kill($pid,$sig); 
 		if(!isset($norelease)) {
 			sleep(2); //Allow dhcp6c to send releae and exit gracefully if needed.
 		}
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1300,7 +1300,7 @@ function interface_bring_down($interface = "wan", $destroy = false, $ifacecfg = 
 	switch ($ifcfg['ipaddrv6']) {
 		case "slaac":
 		case "dhcp6":
-			kill_dhcp6client_process($realif);
+			kill_dhcp6client_process($realif,isset($ifcfg['dhcp6norelease']));
 			unlink_if_exists("{$g['varetc_path']}/dhcp6c_{$interface}.conf");
 			unlink_if_exists("{$g['varetc_path']}/dhcp6c_{$interface}_script.sh");
 			unlink_if_exists("{$g['varetc_path']}/rtsold_{$realifv6}_script.sh");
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -3039,14 +3039,44 @@ function find_dhcp6c_process($interface) {
...
 	}
 
 	if (($pid = find_dhcp6c_process($interface)) != 0) {
-		mwexec("kill -9 {$pid}");
-		sleep(1);
+		/* Kill -9 caused the pid to get left behind, also if we need a */
+		/* relase sent then it needs to be -15, this then allows dhcp6c */
+		/* to send the release, it will also clean up after itself */
+		// Debug - next line remove if not required
+		mwexec("/usr/bin/logger -t info 'shutting down dhcp6c process'");
```
rbgarga: Why are you opening a shell session to execute logger binary instead of use functions like log_error() that will end up having the same effect?

src/etc/inc/interfaces.inc
```diff
@@ -3039,14 +3039,44 @@ function find_dhcp6c_process($interface) {
...
+		$killtype = isset($norelease) ? "-9" : "-15";
+		mwexec("kill {$killtype} {$pid}");
```
rbgarga: It would be better do use posix_kill()

src/etc/inc/interfaces.inc
```diff
@@ -3039,14 +3039,44 @@ function find_dhcp6c_process($interface) {
...
+}
+
+function run_dhcp6client_process($interface, $wancfg) {
+	global $g;
+
+	$debugOption = isset($wancfg['dhcp6debug']) ? "-D" : "-d";
+	$noreleaseOption = isset($wancfg['dhcp6norelease']) ? "-n" : "";
+	// Only run this if the lock does not exist. In theory the lock being there in this mode means the user has selected dhcp6withoutRA while a session is active in the other mode
+	// It should not happen as the process should have been killed and the lock deleted.
+	if(!file_exists("/tmp/dhcp6c_{$interface}_lock"))
```
rbgarga: Please use lock() to manage locks instead of touching and deleting files.

While here, move below bracket to this line to fix style

src/etc/inc/interfaces.inc
```diff
@@ -3929,7 +3959,13 @@ function interface_dhcpv6_configure($interface = "wan", $wancfg) {
 
 			$dhcp6cconf .= "\trequest domain-name-servers;\n";
 			$dhcp6cconf .= "\trequest domain-name;\n";
+			
+			/* dhcp6c will run different scripts depending on whether dhcpwithoutra is set or unset. */
+			if (isset($wancfg['dhcp6withoutra'])) {
+			$dhcp6cconf .= "\tscript \"{$g['varetc_path']}/dhcp6c_{$interface}_dhcp6withoutra_script.sh\"; # we'd like nameservers and RTSOLD to do all the work\n";
```
rbgarga: Is it really necessary to have a different script for this? IMO it would be better to use a parameter to control this set

src/etc/inc/interfaces.inc
marjohn56: Because it was a debug line, hence the comment above it and I missed deleting it when tidying. up

src/etc/inc/interfaces.inc
marjohn56: It would, but several months ago it was found posix_kill() was not killing dhcp6c as it should, hence it being the way it is and has been for quite a while.

src/etc/inc/interfaces.inc
marjohn56: Yes as two scripts are used, one called by dhcp6 ( dhcp6_*_withoutra_script.sh ) and RTSOLD which calls ( dhcp6_*_script.sh) the later being reused, otherwise there would be three scripts.

src/etc/inc/interfaces.inc
marjohn56: I may be being dense here, but looking at lock($lock, $op = LOCK_SH), it refers to it as locking configuration files, also it's appears to be a wrapper for touch(). In the same way unlock() also refers to configuration, are these the functions in util.inc you are referring to?

rbgarga: @marjohn56 Yes, I'm talking about those functions from util.inc. You should note those functions use flock() and they are the lock mechanism used in other places in pfSense, if you grep for LOCK_EX on the code you will find many examples. Please use them to deal with lock/unlock instead of touch/unlink file

marjohn56: O.K., that would work from within the php script itself, but what about the lock file that's created by the RTSOLD script at line 4069, that's inside a shell script,  Also  the lock file is looked for in the shell script at line 4062.

src/etc/inc/interfaces.inc
marjohn56: Just checked this and posix_kill() is now working fine, Must have been some other issue that was causing the kill problem a while back. I have replaced it with posix_kill()

rbgarga:
I've found some problems during tests and fixed them myself. Merged, thanks!

marjohn56:
Oh OK, I'm looking at Redmine #7145 at the moment, I have an updated client beyond what I posted there and I was going to add any script changes to this, but you beat me to it!

May I ask what problems you found?

rbgarga:
@marjohn56 sure, functional problem was a couple of variables not being properly escaped, fixed by 31677494e659768a12d134641bbd4453d2c0a03c and the other were style fixes (broken indent, spaces at EOL, too long lines), fixed by e8256d16e57d2d39c94ae5f2ecd9ab3780b334d0

marjohn56:
Nuts.. I found those in my own scripts when I started playing with dhcp6c REASON stuff, should have changed them on the PR. My apologies and thanks.

 

Probably going to change those script creation routines again as the REASON stuff is working nicely. I’ll wait for jim-p to comment on Redmine #7145 before doing anything else.

rbgarga:
Cool @marjohn56. When you are doing new changes please try to follow the recommended style as documented at https://doc.pfsense.org/index.php/Developer_Style_Guide

marjohn56:
I do try.. :)

 

It’s just that in my day job all our code is written in the format.

 

Function type foo()

{

 

}

 

Rather than KR style.

PR #3411, opened 2017-01-21
Corrected how some Super Micro A1SAi motherboards are shown on the dashboard

sirjorj:
This evening I noticed that my Super Micro C2550-based router was reporting the system correctly under 'CPU' but not 'System'.

I got curious, dug in, and found where the information was produced and displayed.  Well, it was assuming that all systems in the A1SAi family are C2758, which isn't true.  I have VERY little php experience, but I was able to figure out what was going on and made a small change.  I included model numbers for a few other of boards, but only tested on the C2550.

There are probably several better ways to do this, but "It works on my machine"(tm)  :)

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -2356,8 +2356,17 @@ function system_identify_specific_platform() {
 			break;
 		case 'SYS-5018A-FTN4':
 		case 'A1SAi':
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

sirjorj:
You really need to know my address to accept this small fix?

sbeaver-netgate:
No CLA provided in five months

PR #3413, opened 2017-01-22
Fix #7147 Use compressed IPv6 for comparisons

phil-davis:
The IPv6 of an interface may have been entered in full form, e.g.
2001:470:dead:beef:0:0:0:1
Which does not compare directly with the compressed form:
2001:470:dead:beef::1
So compress IPv6 addresses in functions before using them.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -2728,6 +2728,10 @@ function where_is_ipaddr_configured($ipaddr, $ignore_if = "", $check_localip = f
 
 	$isipv6 = is_ipaddrv6($ipaddr);
 
+	if ($isipv6) {
+		$ipaddr = Net_IPv6::compress($ipaddr);
+	}
+
 	if ($check_subnets) {
 		$cidrprefix = intval($cidrprefix);
 		if ($isipv6) {
@@ -2786,7 +2790,7 @@ function where_is_ipaddr_configured($ipaddr, $ignore_if = "", $check_localip = f
 	}
 
 	if ($check_localip) {
-		if (!is_array($config['l2tp']) && !empty($config['l2tp']['localip']) && (strcasecmp($ipaddr, $config['l2tp']['localip']) == 0)) {
+		if (!is_array($config['l2tp']) && !empty($config['l2tp']['localip']) && (strcasecmp($ipaddr, Net_IPv6::compress($config['l2tp']['localip'])) == 0)) {
 			$where_entry = array();
 			$where_entry['if'] = 'l2tp';
 			$where_entry['ip_or_subnet'] = $config['l2tp']['localip'];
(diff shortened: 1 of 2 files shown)


phil-davis:
Note1: There may be a whole lot more places where we should do a compress before using an IPv6 address, for consistency to ensure that any comparisons between IPv6 addresses will succeed. Or we should use some compare_ipv6_addresses() function whenever doing comparisons and that function can compress the addresses before doing a string comparison or...
Note2: The example in the bug report uses a "compressed" address that is not valid according to the combination of standards;
https://tools.ietf.org/html/rfc4291#section-2.2
https://tools.ietf.org/html/rfc5952
Net_IPv6::compress("2001:470:dead:beef:1:2:0:3") returns the input string (no compression possible)
2001:470:dead:beef:1:2:0:3
does not compress to:
2001:470:dead:beef:1:2::3
because rfc5952 says that you are not allowed to compress a single zero.

Unfortunately 2001:470:dead:beef:1:2::3 is accepted by the GUI validation as an IPv6 address. So I guess the Net_IPv6 package is being inconsistent in allowing this as a valid IPv6 address, but not compressing to it. I will have a look at that separately.

phil-davis:
The underlying issue is really that Net_IPv6 is being "nice" about uncompressing a string that is thought to be IPv6 "text" format. It will uncompress any single "::" sequence, e.g in the following the "::" is actually only compressing a sequence of 2 "0" fields. According to the standard, the 3 "0" fields at the end are the ones that should have been compressed.
```
require_once('pfsense-utils.inc');
$ipaddr = "2001:2::1:0:0:0";
var_dump($ipaddr);
$x = is_ipaddrv6($ipaddr);
var_dump($x);
$cip = Net_IPv6::compress($ipaddr);
var_dump($cip);
$cip = Net_IPv6::compress($ipaddr, true);
var_dump($cip);
$uip = Net_IPv6::uncompress($ipaddr);
var_dump($uip);
$cip = Net_IPv6::compress($uip);
var_dump($cip);
```
and the output is:
```
/tmp/o753Ay:7:
string(15) "2001:2::1:0:0:0"
/tmp/o753Ay:9:
bool(true)
/tmp/o753Ay:11:
string(15) "2001:2::1:0:0:0"
/tmp/o753Ay:13:
string(14) "2001:2:0:0:1::"
/tmp/o753Ay:15:
string(18) "2001:2:0:0:1:0:0:0"
/tmp/o753Ay:17:
string(14) "2001:2:0:0:1::"
```
So an address like this is accepted by validation. If we do an ordinary compress() on it, it stays the same. If we "force" compress by compress($ip, true), then the underlying code does uncompress() and then compresses the full address, resulting in the expected canonical form.
That is a bit of a quirk of the Net_IPv6 code, that if the user has "pre-compressed" the address into some "reasonable" form then Net_IPv6 does not fuss about it. However it means that 2 "compressed" addresses that validate as IPv6 addresses can be different strings.

phil-davis:
See #3414 for what I think is a better way to start on this.

PR #3414, opened 2017-01-22
Fix #7147 IPv6 compress

phil-davis:
1) Make a function text_to_compressed_ip6() that forces compression even on IPv6 addresses that already "look" compressed. e.g. : 2001:0:0:2:0:0:0:1 might be written by someone as 2001::2:0:0:0:1 but should really be written 2001:0:0:2::1
2) Use text_to_compressed_ip6() everywhere that currently uses Net_IPv6::compress()
3) Use text_to_compressed_ip6() in a few new places to make sure that where_is_ipaddr_configured() works properly, and thus anyone that calls it will also work OK.

Then there are plenty more places where text_to_compressed_ip6() could be called to make sure that unusual-looking IPv6 address strings are converted to a standardized form before being used for comparison or written to conf files of utilities.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3830,7 +3830,7 @@ function interface_6to4_configure($interface = "wan", $wancfg) {
 	foreach ($stfbrbinarr as $bin) {
 		$stfbrarr[] = dechex(bindec($bin));
 	}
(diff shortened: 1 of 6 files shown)


phil-davis:
If you care enough, then read some discussion I put earlier in #3413

PR #3418, opened 2017-01-22
Allow get_pkg_info() to fallback on "pkg info" for local pkgs if "pkg search" cannot access a local catalog copy (resubmit)

stilez:
Resubmit of PR #3157 with fix.

The issue was down to the behaviour of `pkg info`, which errors out rather than providing an empty return, if no packages matched locally (as will be the case in @NYOB's clean install). The fix is to test specifically if there are zero installed packages matching, before using `pkg info` to get the list of local matching pkgs.

For details of the underlying issue being fixed, and the issue that this resubmit fixes, see the original PR.

_Amended code_

This PR is identical to #3157 apart from the new lines 395-402, which wrap the old lines in a `pkg info -e` call with dummy args to skip the `pkg info` call if it'll hit "no matches".

_Detail_

`pkg info` gives an error "No packages match [$pkgs]" if none match, whereas `pkg search` doesn't (usually!).

The simplest fix, which seems to work nicely, is that if no repo copy exists or we're going to inquire about local pkgs, wrap the `pkg info` call with a test using `pkg info -e`, to detect whether pkg info will hit a "no packages exist" issue or not. If `pkg info -e` returns 0 then it's safe to call `pkg info`,  as there's at least one matching installed pkg so we won't hit this issue (any other errors are allowed to occur and be detected as usual). Otherwise skip the main `pkg info` (which leaves $out and other variables at the correct values they need anyway, for null match/none found).

### DIFF
--- a/src/etc/inc/pkg-utils.inc
+++ b/src/etc/inc/pkg-utils.inc
@@ -337,11 +337,13 @@ function get_package_internal_name($package_data) {
 }
 
 // Get information about packages.
-function get_pkg_info($pkgs = 'all', $only_local = false) {
+function get_pkg_info($pkgs = 'all', $local_only = false, $installed_only = false) {
+
 	global $g, $input_errors;
 
 	$out = '';
 	$err = '';
+	$rc = 0;
 
 	unset($pkg_filter);
 	if (is_array($pkgs)) {
@@ -350,7 +352,7 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
 	}
 
 	if ($pkgs == 'all') {
-		$pkgs = $g['pkg_prefix'];
+		$pkgs = $g['pkg_prefix'] . '*'; // Allows same prefix to work with both pkg search + pkg info
 	}
 
 	if (!function_exists('is_subsystem_dirty')) {
@@ -359,21 +361,46 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
 
 	/* Do not run remote operations if pkg has a lock */
 	if (is_subsystem_dirty('pkg')) {
-		$only_local = true;
+		$local_only = true;
 		$lock = false;
 	} else {
 		$lock = true;
 	}
 
 	$extra_param = "";
-	if ($only_local) {
+	if ($local_only) {
 		$extra_param = "-U ";
 	}
 
 	if ($lock) {
 		mark_subsystem_dirty('pkg');
 	}
-	$rc = pkg_exec("search {$extra_param}-R --raw-format json-compact " . $pkgs, $out, $err);
+
+	if (!$installed_only) {
+		// repo catalog search (either remote or local_only)
+		$rc = pkg_exec("search {$extra_param}-R --raw-format json-compact " . $pkgs, $out, $err);
+	}
+	if ($installed_only || ($local_only && $rc != 0)) {
+		/* use pkg info if (1) installed pkg search or (2) local catalog copy search requested + failed.
+		 *
+		 * The local repo catalog copy may be cleared if a previous call to pkg search couldn't get the
+		 * remote repo catalog.
+		 *
+		 * If the calling code would have accepted local copy info (which isn't assumed up to date) then it
+		 * makes sense to fall back on pkg info to at least return the known info about installed pkgs (pkg
+		 * info should still work), instead of failing and returning no info at all. For example, this
+		 * enables offline view + management of installed pkgs.
+		 */
+
+		// pkg info errors if none match, unlike pkg search, so test it will work beforehand...
+		$tmp1 = $tmp2 = '';
+		$tmprc = pkg_exec("info -e -R --raw-format json-compact " . $pkgs, $tmp1, $tmp2);
+		
+		if ($tmprc == 0) {
+			// ok, packages match, so pkg info can be safely called
+			$rc = pkg_exec("info -R --raw-format json-compact " . $pkgs, $out, $err);
+		} // else we already have empty values for $out etc which are correct if none matched
+	}
 	if ($lock) {
 		clear_subsystem_dirty('pkg');
 	}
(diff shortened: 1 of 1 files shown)


src/etc/inc/pkg-utils.inc
```diff
@@ -359,21 +361,46 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
...
+		 *
+		 * If the calling code would have accepted local copy info (which isn't assumed up to date) then it
+		 * makes sense to fall back on pkg info to at least return the known info about installed pkgs (pkg
+		 * info should still work), instead of failing and returning no info at all. For example, this
+		 * enables offline view + management of installed pkgs.
+		 */
+
+		// pkg info errors if none match, unlike pkg search, so test it will work beforehand...
+		$tmp1 = $tmp2 = '';
+		$tmprc = pkg_exec("info -e -R --raw-format json-compact " . $pkgs, $tmp1, $tmp2);
```
rbgarga: You can use is_pkg_installed() to check if a package is installed, it works with wildcards

stilez: Thanks, however I think here it's best not to do that.

The aim in this snippet is very specific - to identify in advance whether `pkg info` will return a "no matching packages" error for the exact parameters in `$pkgs`.  It's not _per se_ trying to determine if a package(s) are installed, rather if, given the specific args in the code, will an error be returned due to no match.

Unless is_pkg_installed() returns exactly the same as `pkg info` for the given arg string, it can't be used here, and if it's doing exactly the same then it's probably wrapping pkg info and a direct call to pkg info here will be more direct and add clarity.

tl;dr as we aren't trying to detect whether packages are installed, and we're trying to detect if pkg info will return an error specifically due to no matches on arg string, the best approach is to use pkg info itself to do that test, not a wrapper or other internal function.

stilez: Thanks, however I think here it's best not to do that.

The aim in this snippet is very specific - to identify in advance whether `pkg info` will return a "no matching packages" error for the exact parameters in `$pkgs`.  It's not _per se_ trying to determine if a package(s) are installed, rather if, given the specific args in the code, will an error be returned due to no match.

Unless is_pkg_installed() returns exactly the same as `pkg info` for the given arg string, it can't be used here, and if it's doing exactly the same then it's probably wrapping pkg info and a direct call to pkg info here will be more direct and add clarity.

tl;dr as we aren't trying to detect whether packages are installed, and we're trying to detect if pkg info will return an error specifically due to no matches on arg string, the best approach is to use pkg info itself to do that test, not a wrapper or other internal function.

rbgarga: pkg info will return 'no matching packages' if the expression used doesn't match any installed packages. Those final parameters '-R --raw-format json-compact' are only used to define output format and will not change anything when used together with -e, that is the parameter used to identify if there are or not installed packages matching the desired pattern.

In the end you would need to use 'pkg info -e pattern' what is the same of calling is_pkg_installed(pattern)

stilez: done. Though to me it's less transparent.... :)

src/etc/inc/pkg-utils.inc
```diff
@@ -359,21 +361,72 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
...
+		 * enables offline view + management of installed pkgs.
+		 */
+
+		// pkg info errors if none match, unlike pkg search, so test it will work beforehand
+		// is_package_installed() is a wrapper for pkg info -e <pattern> which is what we need here.
+		if (is_package_installed($pkgs)) {
+			// ok, 1 or more packages match, so pkg info can be safely called to get the pkg list
+			$rc = pkg_exec("info -R --raw-format json-compact " . $pkgs, $out, $err);
+		} // else we already have empty values for $out etc which are correct if none matched
+	}
```
rbgarga: Looks like there are some copy/paste leftover here

stilez: I don't think that's leftover. Remember that the function get_pkg_info() returns data for the packages matching `pattern`. The point of the PR is that if a remote repo can't be accessed and/or local installed pkg info would be enough, at least we can fall back on getting local installed pkg info by using pkg info not pkg search. But you still have to actually execute the pkg info call to get that info (in those circumstances and once you've checked it won't error out by pre-calling with the -e option).

src/etc/inc/pkg-utils.inc
```diff
@@ -359,21 +361,46 @@ function get_pkg_info($pkgs = 'all', $only_local = false) {
...
 	if ($lock) {
 		mark_subsystem_dirty('pkg');
 	}
-	$rc = pkg_exec("search {$extra_param}-R --raw-format json-compact " . $pkgs, $out, $err);
+
+	if (!$installed_only) {
+		// repo catalog search (either remote or local_only)
+		$rc = pkg_exec("search {$extra_param}-R --raw-format json-compact " . $pkgs, $out, $err);
+	}
+	if ($installed_only || ($local_only && $rc != 0)) {
```
stilez: Can you hold off any merge for a day or so, busy weekend but I want to recheck this line and something about the initial args

rbgarga: Sure, I'll wait to hear from you

stilez: Done, see note below

stilez:
@rbgarga - Minor cleanup/logic fix done, thank you. 

The internal names of the args for get_pkg_info() were the issue bothering me.  It was too easy to confuse what exactly `local_only` (local _what_?) and `installed_only` mean, and mis-using them. (For example at first sight `$local_only = true` could mean any of: "don't update local catalog copy", "only check local catalog copy", or "only report local installed pkgs"). So I ended up renaming the args, now it's much clearer what they signify. I also refactored the 2 nested if () statements and explanatory comment, and did a couple of very minor code cleanups in the same function. I'm happier now.

rbgarga:
@stilez can you please also change the call to get_pkg_info() in pkg_mgr_installed.php to use ('all', false, true) parameter? It would make it to show installed packages when system cannot update pkg metadata

rbgarga:
@stilez just ignore my last comment

PR #3419, opened 2017-01-22
Required fields in NAT pages

phil-davis:
This is a good one to highlight and discuss the issues about what should be marked as "required".

Example 1) Fields with dropdown lists - e.g. Interface, Protocol
They are "required" by the back-end but the UI already enforces selecting something from the valid list. The user can (and often does) leave them at the provided default selections. And there is no way that the user can not select/input in these fields.
**Should we mark these fields as required?**
It seems to me that it looks a bit odd to have the "required" underlining on fields further down, but not on Interface and Protocol. Without Interface and Protocol being underlined, some users might ask why? how are those fields not required?

Example 2) firewall_nat_edit.php
"Destination" is a combination of selecting from a dropdown, plus sometimes entering an address/mask. A default value is selected that does not require the user to enter anything else - they can just keep the default. So in that sense the form group is not "required". But if the user selects "Single host or alias" or "Network" then they are required to enter an address/mask.
If the answer to (1) is "yes", then this field will be marked required. If "no" then I guess we should put JavaScript here that responds to changing the selection in the dropdown?

There are similar issues in firewall_nat_1to1_edit.php and firewall_nat_out_edit.php - and will be similar on other UI pages.

### DIFF
--- a/src/usr/local/www/firewall_nat_1to1_edit.php
+++ b/src/usr/local/www/firewall_nat_1to1_edit.php
@@ -466,12 +466,12 @@ $section->addInput(new Form_Select(
 
 $section->addInput(new Form_IpAddress(
 	'external',
-	'External subnet IP',
+	'*External subnet IP',
 	$pconfig['external']
 ))->setHelp('Enter the external (usually on a WAN) subnet\'s starting address for the 1:1 mapping. ' .
 			'The subnet mask from the internal address below will be applied to this IP address.');
 
-$group = new Form_Group('Internal IP');
+$group = new Form_Group('*Internal IP');
 
 $group->add(new Form_Checkbox(
 	'srcnot',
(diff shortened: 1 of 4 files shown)


phil-davis:
@sbeaver-netgate thoughts/opinion?

sbeaver-netgate:
Agree with @Jimp I think it is desirable to mark selectors as required.

phil-davis:
Good. I made a ToDo on Redmine https://redmine.pfsense.org/issues/7160
I am happy to go through the GUI pages and make the (mostly mindless) changes - somehow carthartic for a few hours!
I will add a commit to this PR to add the extra required fields for the NAT UI pages.

phil-davis:
Other fields marked "required" as per discussion. Ready for review again.

PR #3421, opened 2017-01-22
Interface description should be required

phil-davis:
I was blanking out stuff to see what input error messages came about fields that are required. I was surprised that the Interface Description would save empty. The system did not immediately explode (I was playing with a bonus interface I had created in a VM) but I don't think it is desirable to have interfaces with no description.
The change here also only does the other checks on the description if it has actually been entered - no point checking a whole lot of other dependent stuff if the description is not even entered.


rbgarga:
@phil-davis please rebase your fork

phil-davis:
Rebased

PR #3422, opened 2017-01-22
Interface Group Name should be required

phil-davis:
Do not allow it to be blank. Do not bother with all the other checks on the group name if it is blank.
(At the moment I can enter an Interface Group with blank name)


rbgarga:
@phil-davis please rebase your fork

phil-davis:
Rebased

PR #3423, opened 2017-01-22
QinQ Edit fix row add delete

phil-davis:
1) Make the row delete button actually work
2) Do not allow ranges that go backwards like 25-20
3) Remember the POSTed stuff to redisplay to the user if there were input errors
4) Ranges or numbers entered could overlap - just save a list of the unique members. e.g. If the user enters 5-15 and 10-20


phil-davis:
Closing in favor of #3424

PR #3424, opened 2017-01-22
QinQ Edit fix row add delete

phil-davis:
1) Make the row delete button actually work
2) Do not allow ranges that go backwards like 25-20 or just "-" or 1-sided ranges "-42" "42-"
3) Remember the POSTed stuff to redisplay to the user if there were input errors
4) Ranges or numbers entered could overlap - just save a list of the unique members. e.g. If the user enters 5-15 and 10-20

### DIFF
--- a/src/usr/local/www/interfaces_qinq_edit.php
+++ b/src/usr/local/www/interfaces_qinq_edit.php
@@ -105,32 +105,48 @@ if ($_POST) {
 	}
 
 	$members = "";
-	$isfirst = 0;
 
 	// Read the POSTed member array into a space separated list translating any ranges
 	// into their included values
-	foreach ($_POST['members'] as $memb) {
+	$membercounter = 0;
+	$membername = "member{$membercounter}";
+	$valid_members = array();
+
+	while (isset($_POST[$membername])) {
 		// Might be a range
-		$member = explode("-", $memb);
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: This stuff never had any validation of the maximum tag number (is it 4096?) and nothing to limit the number of members (can you really have 4096 members?) I have not tried to do anything about that.
If someone knows what the maximum should be, then that can also be added to the validation.

jim-p:
0 and 4095 are reserved so the valid range for any VLAN tag is 1-4094. And while it's possible in theory to have that many, I don't know of anyone who has tried.

phil-davis:
Tag value validation added.

phil-davis:
Note: The original code allows a QinQ to be entered with no member tags.
That seems wrong to me, or is there some weird use case that I have not thought of?

rbgarga:
@phil-davis I've added a new function to validate integer ranges delimited by : or - as you can see at https://github.com/pfsense/pfsense/commit/ee956d19a8936016be899dfcadac048a69b1ed75

It would be nice to use it to validate the range

jim-p:
Hmm, I don't think it's valid to not have a member tag, we should probably prevent that as well.

phil-davis:
@rbgarga @jim-p  Ready for review again.

PR #3427, opened 2017-01-23
firewall_nat_edit function name typo

phil-davis:
This unrelated fix is in a big diff listing of other stuff related to a proposed RADIUS enhancement https://github.com/pfsense/pfsense/pull/3407/files#diff-f080be89fa12ff3c28d7467b12985d09
The accidental typo should be fixed, rather than getting stuck in that other PR.


phil-davis:
Note: all this code section looks dodgy also, but that is a separate issue.

PR #3428, opened 2017-01-23
Fix #7151 Misleading setHelp for Interface Group Name

phil-davis:
Might as well use the same text as is given to the user in the error message already in this code.


src/usr/local/www/interfaces_groups_edit.php
```diff
@@ -55,6 +55,8 @@
 
 $interface_list = get_configured_interface_with_descr();
 $interface_list_disabled = get_configured_interface_with_descr(false, true);
+$ifname_allowed_chars_text = gettext("Only letters (A-Z), digits (0-9), '-' and '_' are allowed.");
```
jim-p: Looks good to me, though we should probably add on "The name cannot end with a digit."

phil-davis: Done

PR #3430, opened 2017-01-23
Required fields in Interfaces pages

phil-davis:
These are IMHO the non-contentious fields that should be marked required in the various Interfaces GUI pages.
See #3419 for discussion about what else we might mark as required.

### DIFF
--- a/src/usr/local/www/interfaces_ppps_edit.php
+++ b/src/usr/local/www/interfaces_ppps_edit.php
@@ -574,7 +574,7 @@ $section = new Form_Section('PPP Configuration');
 
 $section->addInput(new Form_Select(
 	'type',
-	'Link Type',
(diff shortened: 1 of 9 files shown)


src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -617,16 +617,25 @@ function build_link_list() {
 	[]
 ))->setHelp('Select to fill in service provider data.');
 
+$username_label = 'Username';
+$password_label = 'Password';
```
rbgarga: These requires explicit call to gettext()

phil-davis:
Good catch - fixed.

phil-davis:
Remaining required fields added. This is ready for review again.

PR #3433, opened 2017-01-24
Feature #7159 Add GPS initialization command auto correct tool

jskyboo:
Corrects malformed NMEA sentences by calculating and appending the checksum and adding missing special characters "$" and "*"

### DIFF
--- a/src/usr/local/www/services_ntpd_gps.php
+++ b/src/usr/local/www/services_ntpd_gps.php
@@ -122,6 +122,30 @@ function parse_initcmd(&$nmeaset, $initcmd) {
 	}
 }
 
+function NMEAChecksum($cmd) {
+	$checksum = 0;
+	for ($i=0; $i<strlen($cmd); $i++) {
+		$checksum = ($checksum ^ ord($cmd[$i]));
+	}
+	return strtoupper(str_pad(dechex($checksum), 2, '0', STR_PAD_LEFT));
+}
+
+function autocorrect_initcmd($initcmd) {
+	$cmds = '';
+	$split_initcmd = preg_split('/[\s]+/', $initcmd);
+	foreach ($split_initcmd as $line) {
+		if (!strlen($line)) {
+			continue;
+		}
+		$begin = ($line[0] == '$') ? 1 : 0;
+		$astpos = strrpos($line, '*');
+		$end = ($astpos !== false) ? $astpos : strlen($line);
+		$trimline = substr($line, $begin, $end-$begin);
+		$cmds = $cmds . '$' . $trimline . '*' . NMEAChecksum($trimline) . "\r\n";
+	}
+	return $cmds;
+}
+
 if ($_POST) {
 	unset($input_errors);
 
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_ntpd_gps.php
```diff
@@ -449,6 +483,13 @@ function build_nmea_list() {
 	base64_decode($pconfig['initcmd'])
 ))->setHelp('Commands entered here will be sent to the GPS during initialization. Please read and understand the GPS documentation before making any changes here.');
 
+$section->addInput(new Form_Checkbox(
+	'autocorrect_initcmd',
+	null,
+	'Auto correct malformed initialization commands. (default: checked).',
```
jim-p: This box is not being checked by default, and using the set/unset method of enable/disable makes doing so difficult since you don't know if it was never set or deliberately disabled. It's much simpler to store a text value such as "enabled" and "disabled" and then check that and let unset also mean enabled.

jim-p:
Working block (before):

`$PUBX,40,GSV,0,0,0,0*59`
`$PUBX,40,GLL,0,0,0,0*5C`
`$PUBX,40,ZDA,0,0,0,0*44`
`$PUBX,40,VTG,0,0,0,0*5E`
`$PUBX,40,GSV,0,0,0,0*59`
`$PUBX,40,GSA,0,0,0,0*4E`
`$PUBX,40,RMC,0,0,0,0*46`
`$PUBX,40,GGA,0,0,0,0`
`$PUBX,40,TXT,0,0,0,0`
`$PUBX,41,1,0007,0003,9600,0`
`$PUBX,40,ZDA,1,1,1,1`


Autocorrected block:

`$PUBX,40,GSV,0,0,0,0*59`
`$PUBX,40,GLL,0,0,0,0*5C`
`$PUBX,40,ZDA,0,0,0,0*44`
`$PUBX,40,VTG,0,0,0,0*5E`
`$PUBX,40,GSV,0,0,0,0*59`
`$PUBX,40,GSA,0,0,0,0*4E`
`$PUBX,40,RMC,0,0,0,0*47`
`$PUBX,40,GGA,0,0,0,0*5A`
`$PUBX,40,TXT,0,0,0,0*43`
`$PUBX,41,1,0007,0003,9600,0*10`
`$PUBX,40,ZDA,1,1,1,1*44`

src/usr/local/www/services_ntpd_gps.php
jskyboo: The default values get set when switching GPS types, so if you have already set that then you will have to change the type to reset the default. I could change that behavior but that is how the other "defaults" are set.

jskyboo:
As for the ublox commands, I don't have a ublox to test with yet(one is on the way), but I would suspect it is related to one of the last two commands, though not sure immediately why. Which GPS model are you working with? Could you try? Also what is your baud rate set to(the config option not the PUBX command)?

`$PUBX,40,GSV,0,0,0,0*59
$PUBX,40,GLL,0,0,0,0*5C
$PUBX,40,ZDA,0,0,0,0*44
$PUBX,40,VTG,0,0,0,0*5E
$PUBX,40,GSV,0,0,0,0*59
$PUBX,40,GSA,0,0,0,0*4E
$PUBX,40,RMC,0,0,0,0*47
$PUBX,40,GGA,0,0,0,0*5A
$PUBX,40,TXT,0,0,0,0*43
$PUBX,41,1,0007,0003,9600,0
$PUBX,40,ZDA,1,1,1,1*44`

`$PUBX,40,GSV,0,0,0,0*59
$PUBX,40,GLL,0,0,0,0*5C
$PUBX,40,ZDA,0,0,0,0*44
$PUBX,40,VTG,0,0,0,0*5E
$PUBX,40,GSV,0,0,0,0*59
$PUBX,40,GSA,0,0,0,0*4E
$PUBX,40,RMC,0,0,0,0*47
$PUBX,40,GGA,0,0,0,0*5A
$PUBX,40,TXT,0,0,0,0*43
$PUBX,41,1,0007,0003,9600,0*10
$PUBX,40,ZDA,1,1,1,1`

jim-p:
Not sure what model it is, it's been around for years and has no identifying markings on the case. It shows up as a USB/Serial adapter

`ugen0.2: <Prolific Technology Inc.> at usbus0`
`uplcom0: <Prolific Technology Inc. USB-Serial Controller D, class 0/0, rev 1.10/4.00, addr 2> on usbus0`

Even changing the *46 to *47 breaks the GPS, but it seems to not want checksums on any of those later lines. Baud rate is 9600 all-around (in the GUI, in the init strings, and in /etc/remote on the gps0 line). It works fine on reboot with my original init block, but not with any of the altered/"corrected" ones. Maybe it's just this GPS...

jskyboo:
I've been wondering why that default init looked so mangled and suspected it was to support a GPS that does not handle proper format. So you say it is the switching of $PUBX,40,RMC,0,0,0,0*46 to $PUBX,40,RMC,0,0,0,0*47 that seems to break it?

jim-p:
Not just that but it seems putting a checksum on any of the lines that doesn't have a checksum causes it to fail in some way or another.

jskyboo:
So the GPS doesn't like the checksum on the later commands but has no problem with it for the earlier commands? What happens if you enable GGA does it output GGA or just not support it? $PUBX,40,GGA,1,1,1,1 or $PUBX,40,GGA,1,1,1,1*5A

PR #3435, opened 2017-01-24
Remaining required fields for Diag pages

phil-davis:
Note: There is a small policy discussion still to have. There are 3 categories of required selectors:
1) Required but a valid default is not selected automatically for a new entry - these definitely should be marked "required".
2) Required and "important" - i.e. the user probably should think about if the default value is actually what they want. These should also be marked "required" - stuff like "protocol" or "IP address type/family".
3) Required, the default value selected is 99% useful and the "care factor" is low. For example "Maximum number of pings" for Diagnostics->Ping. Technically these are required, but there is no need to draw user attention to them, in the same way that loads of checkboxes are "required" (they must be on or off) but we don't draw the user's attention to those.

My suggested policy is (1) and (2) are marked "required", (3) is not.

### DIFF
--- a/src/usr/local/www/diag_authentication.php
+++ b/src/usr/local/www/diag_authentication.php
@@ -88,7 +88,7 @@ foreach (auth_get_authserver_list() as $auth_server) {
 
 $section->addInput(new Form_Select(
 	'authmode',
-	'Authentication Server',
+	'*Authentication Server',
 	$pconfig['authmode'],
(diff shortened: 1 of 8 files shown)


phil-davis:
I just accidentally push some changes here for firewall rules, schedule and virtual IP. Those changes are also good to go IMO, so no problem. I will try not to touch this branch again.

PR #3437, opened 2017-01-24
Required fields for shaper

phil-davis:
This is a first effort for this. Not really sure exactly what is required in what context. In a lot of places the user can fill in nothing, press Save and it is accepted. Goodness knows what actually happens underneath.

### DIFF
--- a/src/etc/inc/shaper.inc
+++ b/src/etc/inc/shaper.inc
@@ -802,13 +802,13 @@ class altq_root_queue {
 		));
 
 		$section->addInput(new Form_StaticText(
-			'Name',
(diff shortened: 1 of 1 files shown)


phil-davis:
Yes, for now (or for a long time into the future!) the basic version will have to do. I will make a PR that has just the Name fields.
Actually in the more complex UI pages like these, it would be helpful for the user to know what they have to enter and what they can ignore. But it needs someone who really knows all the valid configurations to specify exactly what is required or not. Then in this case it may be a bit complex to implement!

phil-davis:
See #3438

PR #3440, opened 2017-01-25
Captive Portal auth method is required

phil-davis:
When creating a new Captive Portal Zone the user can fail to select any of the Authorization Method radio buttons. No default radio button is selected - that is probably good, as the user has to think and click to decide which authorization method they will use.
But the form validation should check that some valid value is POSTed when the CP is enabled.


phil-davis:
Note: This is adding missing functionality to existing code. I will provide the "marking fields required" changes in another PR that has lots of services*.php files.

PR #3444, opened 2017-01-25
DNS Forwarder Host Override - host is required

phil-davis:
At the moment you can leave the Host field empty and it saves.
I assume it should be required, or am I missing something?


jim-p:
Host isn't required if you're making an `@` record for just the domain.  `example.com` is a valid DNS record just as much as `www.example.com`

phil-davis:
Thanks, I hadn't thought about that.

PR #3453, opened 2017-01-31
Captive portal: add support for setting traffic quotas

plumbeo:
This patch series add support for traffic quotas to the captive portal.

Patch 1 ("always use the RADIUS-provided session timeout value if the option is enabled") isn't directly related but fixes an issue where the RADIUS attribute could be ignored if there was a shorter hard timeout configured; now when the option is set the RADIUS attribute is always honored.

Patch 2 ("add a traffic quota option") adds an option to set the maximum amount of traffic (sum of downloads and uploads) that a user is allowed to generate during a session. As it happens with the hard timeout setting, the client will still be able to reconnect immediately.

Patch 3 ("add option to retrieve the traffic quota value from RADIUS") allows to get the value of the traffic quota from a RADIUS server. The value obtained is recorded in a new column in the active sessions' database, requiring some minor changes to use the right index to retrieve the radius context in other parts of the code. Since the feature uses a new vendor-specific attribute and I couldn't find a pfSense dictionary, I created a new one (see also the next patch), using Netgate's SMI Network Management Private Enterprise Code as per RFC. I defined 5 attributes, respectively to set upload quota, download quota, total quota, max upload bandwidth and max download bandwidth, even though only the total quota one is used in this patch series (I think the existing code should do the right thing for the bandwidth limits but I didn't test it). I hope this is OK.

Patch 4 ("Add RADIUS dictionary with definitions of the new pfSense vendor-specific attributes") is the new FreeRADIUS compatible dictionary file for pfsense I was talking about earlier. The directory in which I placed it likely isn't the right place, in theory the dictionary, once approved, should be submitted to FreeRADIUS for inclusion and in the meantime added to the FreeRADIUS pkg available in pfSense. Till then tho it will have to be distributed in some other way to allow users who want to use the feature to manually add it to their FreeRADIUS server configuration.


src/etc/inc/radius.inc
```diff
@@ -696,6 +696,26 @@ class Auth_RADIUS extends PEAR {
                     }
                 }
 
+                elseif ($vendor == 13644) { /* Netgate */
+                    switch ($attrv) {
+                    case 1: /* pfSense-Max-Input-Octets */
+                        $this->attributes['maxbytesup'] = radius_cvt_int($datav);
+                        break;
+                    case 2: /* pfSense-Max-Output-Octets */
+                        $this->attributes['maxbytedown'] = radius_cvt_int($datav);
```
phil-davis: This could be consistent with maxbytesup
s/maxbytedown/maxbytesdown/

plumbeo: that's a typo, thank you for spotting it. I fixed it locally, it will be in the next version of the series with any other change required.

plumbeo: On second thought it would probably be better just not to add pfSense-Max-{Input,Output}-Octets if they're not used and do nothing. Somebody could try to use them.

rbgarga:
It's such a big change fot 2.4 at this point, will be revisited after 2.4.0 is released

jim-p:
Seems OK, but definitely needs testing. Please fix conflicts/rebase and then we can check it deeper.

plumbeo:
@jim-p I rebased the branch. I didn't test it again but the fixes for the conflicts were straightforward enough.

src/etc/inc/captiveportal.inc
```diff
@@ -2104,8 +2119,14 @@ function captiveportal_reapply_attributes($cpentry, $attributes) {
 		$dwfaultbw_up = $dwfaultbw_down = 0;
 	}
 	/* pipe throughputs must always be an integer, enforce that restriction again here. */
-	$bw_up = round(!empty($attributes['bw_up']) ? intval($attributes['bw_up'])/1000 : $dwfaultbw_up, 0);
-	$bw_down = round(!empty($attributes['bw_down']) ? intval($attributes['bw_down'])/1000 : $dwfaultbw_down, 0);
+	if (isset($config['captiveportal'][$cpzone]['radiusperuserbw'])) {
+		$bw_up = round(!empty($attributes['bw_up']) ? intval($attributes['bw_up'])/1000 : $dwfaultbw_up, 0);
+		$bw_down = round(!empty($attributes['bw_down']) ? intval($attributes['bw_down'])/1000 : $dwfaultbw_down, 0);
+	} else {
+		$bw_up = round($dwfaultbw_up,0)
```
jim-p: Missing `;` at EOL.

plumbeo:
@jim-p commits aa56cd2 and ed82e9d respectively automatically enable retrieving the values from RADIUS after an upgrade and print a customised string ("SESSION TIMEOUT", "IDLE TIMEOUT" or "QUOTA EXCEEDED" according to the reason for which the user was disconnected.

jim-p:

Thanks!

plumbeo:
Redmine Feature #8202 https://redmine.pfsense.org/issues/8202

sbeaver-netgate:
Please rebase to resolve conflicts so that we can test more deeply

src/etc/inc/globals.inc
```diff
@@ -74,7 +74,7 @@ $g = array(
 	"disablecrashreporter" => false,
 	"crashreporterurl" => "https://crashreporter.pfsense.org/crash_reporter.php",
 	"debug" => false,
-	"latest_config" => "17.3",
+	"latest_config" => "17.4",
```
sbeaver-netgate: This is out of date. Please check the current latest_config version and update accordingly.

src/etc/inc/upgrade_config.inc
```diff
@@ -5493,6 +5493,22 @@ function upgrade_172_to_173() {
 	}
 }
 
+/*
+ * Automatically enable retrieving captive portal bandwidth limits from RADIUS for each captive portal
+ * instance already enabled
+ */
+function upgrade_173_to_174() {
```
sbeaver-netgate: This is out of date. Please check the current latest_config version and update accordingly. I think you will need 175_to_176, but please check immediately before pushing the change.

plumbeo:
@sbeaver-netgate I just pushed a rebased version without conflicts

sbeaver-netgate:
Please rebase and resubmit to resolve conflicts.

plumbeo:
@sbeaver-netgate rebased

PR #3464, opened 2017-02-02
Add mising gettext() calls

phil-davis:
As reported in forum https://forum.pfsense.org/index.php?topic=124734.msg689931#msg689931 , sprintf() inside a setHelp() method call means that the text string is not picked up in the ".pot" file for translation. In some places these already have an explicit gettext() put around them. These are the remaining places with a setHelp(sprintf()) sequence that need a gettext() added.

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -1435,7 +1435,7 @@ foreach (['src' => 'Source', 'dst' => 'Destination'] as $type => $name) {
 		(isset($portValues[ $pconfig[$type .'endport'] ]) ? null : $pconfig[$type .'endport'])
 	))->setHelp('Custom');
 
-	$group->setHelp(sprintf('Specify the %s port or port range for this rule. The "To" field may be left empty if only filtering a single port.',strtolower($name)));
+	$group->setHelp(sprintf(gettext('Specify the %s port or port range for this rule. The "To" field may be left empty if only filtering a single port.'), strtolower($name)));
 
 	$group->addClass(($type == 'src') ? 'srcprtr':'dstprtr');
 	$section->add($group);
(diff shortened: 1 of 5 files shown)


rbgarga:
What about if we change SetHelp() and other methods that contain strings to be translated to work with multiple parameters and do sprintf and gettext internally?

What do you think @sbeaver-netgate and @phil-davis ?

rbgarga:
Then we would be able to call
```
->SetHelp("This is %s string", "foo");
```

sbeaver-netgate:
Good idea.

phil-davis:
Sounds good, if it can be implemented fairly quickly so that the enthusiastic translators get the strings soon.

sbeaver-netgate:
I will try to get to it today

rbgarga:
Cool @sbeaver-netgate, you can check at https://github.com/pfsense/pfsense/blob/master/tools/scripts/update_pot.sh#L53 all methods we are collecting strings to create pot

sbeaver-netgate:
The setHelp method already accommodates this, albeit in a slightly unconventional way. setHelp accepts a second argument, which is an array of args that are passed to sprintf. The first argument does get the gettext treatment. e.g.:
`$section->addInput(new Form_Input(
	'descr',
	'Description',
	'text',
	$pconfig['descr']
))->setHelp('A description may be entered here for %d administrative reference %s (not parsed).', array(23, "fish"));`

Produces:
 A description may be entered here for 23 administrative reference fish (not parsed).

So we can either keep it as it is, or I could revise it to use a more conventioanl format, as in:
`->sethelp("Do this %d times, and do it %s", 4, "slowly");`

sbeaver-netgate:
The setHelp() methods have been updated to no longer require the use of sprintf, and to apply gettext() to the first argument.

PR #3465, opened 2017-02-02
System Information Widget Filter

phil-davis:
I think this would be useful for people to be able to filter which rows of information are displayed on the System Information widget. I have used the same coding style as in the other widget filtering done recently.

It is one part of a solution for Redmine 7182.

Some users might want to see all (or many) of the rows, but the total length of the widget is too long, so there is still a use case for being able to split the data across multiple widgets and put them in different columns.

### DIFF
--- a/src/usr/local/www/widgets/widgets/system_information.widget.php
+++ b/src/usr/local/www/widgets/widgets/system_information.widget.php
@@ -30,6 +30,26 @@ require_once('notices.inc');
 require_once('system.inc');
 include_once("includes/functions.inc.php");
 
+$sysinfo_items = array(
+	'name' => 'Name',
+	'system' => 'System', 'version' => 'Version',
+	'version' => 'Version',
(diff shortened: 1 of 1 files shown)


src/usr/local/www/widgets/widgets/system_information.widget.php
```diff
@@ -30,6 +30,26 @@
...
+	'state_table_size' => 'State Table Size',
+	'mbuf_usage' => 'MBUF Usage',
+	'temperature' => 'Temperature',
+	'load_average' => 'Load Average',
+	'cpu_usage' => 'CPU Usage',
+	'memory_usage' => 'Memory Usage',
+	'swap_usage' => 'Swap Usage',
+	'disk_usage' => 'Disk Usage'
+	);
+
```
sbeaver-netgate: Should we gettext() these?

PR #3467, opened 2017-02-03
Fix argument passed to DNSstuff #7200

maxxer:


maxxer:
As noted on the issue it has been apparently fixed in 7829c6d52, as it's not reproducible on current development

PR #3469, opened 2017-02-04
Fix firewall_nat_out_edit dependency on English

phil-davis:
Reported in forum https://forum.pfsense.org/index.php?topic=125004.0
The problem is that the JavaScript is comparing the descriptive text in each dropdown entry to English strings like "Host" "Subnet" "Other" "Source". But those might be translated at run-time in the list, and so it does not match.
Instead of using the element text, I have used the value. I have put a char at the front of each value that records which "target type" it is - a [H]ost, [S]ubnet, [I]p or [O]ther. Then the JavaScript can reliably look at the first char of the value to determine what to do.
When the value is POSTed, the code strips off that bonus 1st char. When it is loaded on edit, the appropriate 1st char is added to the front again.
Thus there is no need to change what is stored in the config.
Is there some better way to do this?

### DIFF
--- a/src/usr/local/www/firewall_nat_out_edit.php
+++ b/src/usr/local/www/firewall_nat_out_edit.php
@@ -94,6 +94,20 @@ if (isset($id) && $a_out[$id]) {
 	$pconfig['dstport'] = $a_out[$id]['dstport'];
 	$pconfig['natport'] = $a_out[$id]['natport'];
 	$pconfig['target'] = $a_out[$id]['target'];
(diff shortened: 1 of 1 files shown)


phil-davis:
Thanks for committing. This does work, but I felt like it was a "bit of a hack" to put those codes in front of the value at run-time and make sure to strip them off when saving to the config. I feel like there should be a nicer way to store other metadata about options in a select, other than just in value and text.

PR #3470, opened 2017-02-04
system_advanced_firewall setHelp strings

phil-davis:
More fixes related to setHelp strings and the underlying sprintf() behavior.
Redmine 7202

This looks like it will need a bit of a look over the system to find these issues, or think a bit about how to deal with strings in the underlying methods.


phil-davis:
@sbeaver-netgate did this directly in commit https://github.com/pfsense/pfsense/commit/1e0c91d35b299c6650bfb215835a18338006af64

PR #3471, opened 2017-02-04
system_advanced_admin remove setHelp array

phil-davis:


phil-davis:
@sbeaver-netgate any chance to commit this little (and I hope obvious) fix. Then it will be in master and in my git clone. It will save me wondering if I already fixed it, as I go through looking at other stuff like this that might need fixing.

sbeaver-netgate:
Done

PR #3472, opened 2017-02-04
system_advanced_admin remove href from setHelp strings

phil-davis:
I think that giving long strings that contain various HTML tags in-line through for translation is going to be error-prone. Actually we do not want the translators to accidentally also translate href like system_usermanager or accidentally miss out some quotes or brackets. It seems safer to substitute these in using %s formatting.

This is an example of the sort of changes that could be made.

### DIFF
--- a/src/usr/local/www/system_advanced_admin.php
+++ b/src/usr/local/www/system_advanced_admin.php
@@ -324,8 +324,10 @@ $group->add(new Form_Checkbox(
 	'https'
 ))->displayAsRadio();
 
-$group->setHelp($certs_available ? '':'No Certificates have been defined. A certificate is required before SSL can be enabled. '.
-	'<a href="system_certmanager.php">'. gettext("Create or Import").'</a> '.' a Certificate.');
+if (!$certs_available) {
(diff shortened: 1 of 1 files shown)


phil-davis:
Opinions please.

system_advanced_admin.php also has an example where just a pair of bold tags is in the setHelp string (look for "null modem"). Those could also be separated out and the tags substituted in.

Maybe the policy should be that the translators will "never" (as much as we can) see strings to translate that have HTML tag stuff in them.

sbeaver-netgate:
I have been dealing with those by making the whole string an positional param. e.g.:

->setHelp("%s", '<span name="warning"></span>');

Our locale scripts only harvest the first argument to setHelp, so this effectively eliminates translation.

phil-davis:
For cases where all the text is tag, and so there is no "real" text, that is good.
I am talking about cases where there is "real" text, but there is href or bold or other highlighting sort of stuff, like in the example changes here. In that case it seems to me that it will be easier (less chance of translators accidentally breaking the HTML, less confusion for translators who do not understand what HTML is all about) if the HTML is substituted into the string, rather than being handed to translators in the string.

sbeaver-netgate:
Absolutely. Only translatable text in the first argument, HTML, variables, config stuff etc as printf style args.

PR #3473, opened 2017-02-04
Bug #7206 route53

jxmx:
This code resolves the issue in Bug #7206 and is important for the release of 2.4.0.


jxmx:
I tried to fix the locale files as best I can. There's only one with a translation (I think) that I couldn't fix and that was zh_Hans_CN/LC_MESSAGES/pfSense.po. I updated the Spanish one using Google Translate and some rudimentary Spanish knowledge although I'm sure a native speaker would do a much better job.

rbgarga:
@jxmx please do not touch .po or .pot files directly, they are managed by a external tool and any changes will be lost. Could you please re-submit removing these changes?

jxmx:
I've updated the bugfix branch with a rollback of those changes. Do you need a new pull request or can you use the existing one?

rbgarga:
It's OK this way

PR #3474, opened 2017-02-04
Add a '.' to the help text for consistency

lukehamburg:
Was about to submit a full patch for the unescaped ` %'s ` but @sbeaver-netgate [beat me to it](https://redmine.pfsense.org/issues/7202#note-9) by a couple of minutes. So here's my minuscule contribution!


phil-davis:
And I already had a PR for that, so I closed it.
@sbeaver-netgate if you are around, can you commit all this little crud. Then we will all be working off the same page, rather than having to remember which bits have already been fixed in pending PRs.

PR #3475, opened 2017-02-05
Remove HTML from translated shaper setHelp

phil-davis:
and sections must not be translated - examples of IP address formats 255.255.255.255 and so on.

### DIFF
--- a/src/etc/inc/shaper.inc
+++ b/src/etc/inc/shaper.inc
@@ -1455,28 +1455,28 @@ class priq_queue {
 			null,
 			null,
 			!empty($this->GetRed())
-		))->setHelp('<a target="_new" href="https://web.archive.org/web/20160404153707/http://www.openbsd.org/faq/pf/queueing.html#red">' . gettext('Random Early Detection') . '</a>');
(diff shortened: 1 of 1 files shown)


src/etc/inc/shaper.inc
```diff
@@ -3874,14 +3875,14 @@ EOD;
...
-		))->setHelp('IPv4 mask bits' . '<br />' . '255.255.255.255/?');
+		))->setHelp('IPv4 mask bits%1$s%2$s', '<br />', '255.255.255.255/?');
 
 		$group->add(new Form_Select(
 			'maskbitsv6',
 			null,
 			$mask['bitsv6'],
 			array_combine(range(128, 1, -1), range(128, 1, -1))
-		))->setHelp('IPv6 mask bits' . '<br />' . '<span style="font-family:consolas">ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff/?</span>');
+		))->setHelp('IPv6 mask bits%1$s%2$s', '<br />', '<span style="font-family:consolas">ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff/?</span>');
```
phil-davis: As a separate issue, why is a different font chosen for the IPv6 address example?

PR #3479, opened 2017-02-05
Convert ICMP subtypes selector control in fw_rules_edit.php as suggested, and refactor jscript a little

stilez:
See PR #3328 for background.

This PR converts the new ICMP subtypes selector in `firewall_rules_edit.php` from a single multi-select list box to a double selector, as suggested by @jim-p, which leaves it clearer and more obvious which items are selected, without adding a list of checkboxes.

It does indeed improve the form. 

I hope no "junk" or merge conflicts have made it into this, I have worked from a copy-paste of the current code, but mistakes are always possible.

I also refactored the javascript by collecting all functions and all on load activities together ("like with like") for convenience of devs in future. However all code except that directly controlling ICMP subtype selection should be unchanged.

The only functional change to the form is that it now returns icmptype as a comma separated string via a hidden control, rather than as an array of elements from the list box (as in the example suggested by @jim-p in the original PR). This made the code simpler, because the source data for the "selected" list box was already generated from a JS array anyhow, and the data in $config is in that format.


src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1435,7 +1487,7 @@ function build_if_list() {
 		(isset($portValues[ $pconfig[$type .'endport'] ]) ? null : $pconfig[$type .'endport'])
 	))->setHelp('Custom');
 
-	$group->setHelp('Specify the %s port or port range for this rule. The "To" field may be left empty if only filtering a single port.', strtolower($name));
+	$group->setHelp('Specify the " . strtolower($name) . " port or port range for this rule. The "To" field may be left empty if only filtering a single port.');
```
phil-davis: This is a merging problem - there should be no diff in this line.

stilez: Fixed, thanks

stilez:
Self-note as being dragged away - to double check that if existing icmp rule has icmptype unset or empty string, it interprets this as "all subtypes selected" rather than "none selected". I assumed this as existing code worked, but now feel I want to double-check,

rbgarga:
@stilez please rebase your fork to make this patch ok to apply for testing

stilez:
@rbgarga - done

stilez:
thanks @jim-p. can you take a look at the review point I mentioned above and reassure me if its catching all it needs to, or any fixes needed? (Ii want to be sure old configs won't have issues if icmptype is blank or unset on existing/associated rules, obviously.)

jim-p:
I didn't try that case, but I tried setting a few other flags before applying the patch and it seemed to read those OK. 

It seems to treat an old "any" selection as none being selected, and then gives an error on save. It could allow none to be selected and still treat that as any/all, since selecting nothing makes no sense, and someone who doesn't care isn't likely to want to select all before saving.

stilez:
I'll look at and fix that but not today, thanks jim

rbgarga:
Reset review while we wait new commit

stilez:
@rbgarga  @jim-p  - Done. 3 places where icmptype is pulled from old data, clearly should be doing the same kind of check on all 3 of them as well, so I moved the check into a separate function and added detection of "any" to it. Should work now - does it? :)

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1090,6 +1084,17 @@ function build_if_list() {
 	return($iflist);
 }
 
+// Expands missing/blank/"all" icmptypes to a list of specific types. Doesn't check validity.
+function expand_icmptype($icmptype, $ipprotocol) {
+	if (isset($icmptype) && is_string($icmptype) && strlen($icmptype) > 0 && $icmptype != "any") {
```
rbgarga: $icmptype will always be set since it's a parameter, this isset() call is unnecessary.

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1090,6 +1084,17 @@ function build_if_list() {
...
+		// $icmptype appears to contain specific icmptype(s)
+		return $icmptype;
+	} else {
+		// $icmptype has no (or blank) icmptype,or is "all"  meaning "all/any", select all subtypes
+		return implode(',', $icmptypes[$ipprotocol]['icmptypes']);
```
rbgarga: If I got it right $icmptypes is a variable defined out of this function scope, then it's missing a global declaration

src/usr/local/www/firewall_rules_edit.php
stilez: I just checked - if the source variable is unset (as icmptypes can conceivably be) the function is passed NULL as the value, not <undefined>. So this should be is_null() not isset(), thanks.

src/usr/local/www/firewall_rules_edit.php
stilez: Yes

stilez:
2 x review points fixed

stilez:
I nreed to look at that last again. Just setting to empty string isn't necessarily a fix (I did it in a rush). (And if it's referring to an existing rule, or existing associated rule, how can ['ipprotocol'] _not_ be set and valid....?)

stilez:
Try that :)

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1088,6 +1084,29 @@ function build_if_list() {
...
+	if (is_string($icmptype) && strlen($icmptype) > 0 && $icmptype != "any") {
+		// $icmptype appears to contain specific icmptype(s)
+		$result = $icmptype;
+	} elseif (is_string($ipprotocol) && array_key_exists($ipprotocol, $icmptypes)) {
+		// $icmptype has no (or blank) icmptype,or is "all"  meaning "all/any", select all subtypes
+		$result = implode(',', $icmptypes[$ipprotocol]['icmptypes']);
+	} elseif ($invalid_proto_action != 'unset') {
+		// ipprotocol undefined or invalid, force value to whatever we were passed, should be inet4/6/46
+		$result = implode(',', $icmptypes[$invalid_proto_action]['icmptypes']);
+	? else {
```
jim-p: `Parse error: syntax error, unexpected '?' in /usr/local/www/firewall_rules_edit.php on line 1098 `
That `?` should be a `{`

stilez: Typo fixed. Grumble grumble

stilez:
Typo fixed. Grumble grumble

stilez:
I'm really not happy with "select none means all types".

It means having multiple ways to enter and represent the same thing in the config, and also in the UI, which is rarely a good idea, and increases complexity and risk of code error.

Also, the only options supported in rules previously were "all types" or "one specific type only", so empty/unset==all made sense and it was an obvious dropdown box and config choice. With multiple choice, some quick way to choose "all types" is needed.  But I don't think defaulting to "empty selection means all selected" makes sense or is the way to go, and it's very counterintuitive in the UI.  

Instead this UI offers a very standard, very intuitive "select all" button. One click gets all valid types, and it's clear all types are selected because the left hand list of unselected types is empty. I think we could break from the past UI to that extent, and it's much clearer at a glance. I just don't think that a single click on "select all" is likely to be any kind of problem or annoyance, at most it'll take a few moments for existing users to notice and click "select all" to select all, the first time they see the rules page. 

Can you try it a bit more - I just think it's a cleaner approach this way, both code and UI.

jim-p:
An existing 'any' icmp type rule has nothing selected when loaded, so the behavior must be preserved. Other areas operate the same way (e.g. selecting no interfaces in some places binds to all). That's the preferred behavior, making the default what people expect/want.

stilez:
I hope this is a reasonable approach - I've fixed it so thsat if there wasn't an icmptype selection (which is the only way the SELECT box can be created with an empty list), then all icmptypes are pre-selected and it's displayed as all selected. If something was selected then the code change has no effect.

I don't have a copy of a sample old config to test this against, if this works then good, if not can you attach a sample config I can debug against to this PR, or email it to me? Thanks.

stilez:
Issues fixed I think?

jim-p:
I haven't had a chance to test it again yet, been wrapped up in 2.3.3 stuff and other things. I'll get to it today.

jim-p:
Getting closer...
It doesn't work with a rule that had "any" selected previously
`<icmptype>any</icmptype>`

stilez:
@jim-p try that. Now recognises and handles correctly, the case you gave, of <icmptype>any</icmptype> in an older ICMP rule.

stilez:
@jim-p :

1. Done, 

2. Intended behaviour, due to the way that each ipproto has different but overlapping sets of icmptypes that could be valid. 

When the user switches to a new ipprotocol, one could add logic that says "if all types were selected for the old ipproto, then automatically select all types valid for the new ipproto after the switch". Would that be the sort of thing you're after?

jim-p:
1. I had a rule with 'any' from before this change, editing the rule and saving it lost the 'any' and now shows all types in the list. A new rule works the same way, so the 'any' detection bit doesn't seem to work for me.

2. Yeah that sounds good, if it was 'any' before then switching protocols should have everything selected.

stilez:
@jim-p I'm away the weekend and VM store is fiddling in the background. Found one other place to simplify (see previous commit) and will deal with above shortly

jim-p:
No worries, take your time

sbeaver-netgate:
Please resolve conflicts and review the state of this PR

jim-p:

Thanks!

sbeaver-netgate:
No resolution in 10 months. Please resubmit if still required.

PR #3481, opened 2017-02-06
Make Backup & Restore buttons more robust

phil-davis:
I had set the UI to Turkish and then tried to 'Download configuration as XML' and it did not work. 'Download configuration as XML' is not yet translated in Turkish, but the single word 'Download' is translated. So the assumption in the code that the translation of 'Download' is contained somewhere in the translation of 'Download configuration as XML' is not true.

These should use matches of the translations of the same string as used in the button.

### DIFF
--- a/src/usr/local/www/diag_backup.php
+++ b/src/usr/local/www/diag_backup.php
@@ -169,15 +169,20 @@ if ($_POST['apply']) {
 	exit;
 }
 
+$restore_config_text = gettext('Restore configuration');
+$reinstall_packages_text = gettext('Reinstall Packages');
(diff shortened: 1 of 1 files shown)


phil-davis:
Actually it would be even nicer if the value that is posted in $_POST['submit'] could be a fixed coded string, rather than being the language-dependent string that is displayed on the UI. Another model-view-separation thing.

phil-davis:
See #3486 for a version that gets rid of comparing $_POST with the translated button text. I am guessing that would be the preferred way to do this. Please choose which PR to take and close the other.

rbgarga:
Close it in favor of #3486

PR #3483, opened 2017-02-06
Small enhancements to services_dhcp_edit.php

lukehamburg:
- added `ndp` call to get MAC addr if remote client is connected via IPv6
- automatically hide **Copy MAC** button if arp/ndp returns null
- switch to `exec()` instead of backticks for calls to `arp`

### DIFF
--- a/src/usr/local/www/services_dhcp_edit.php
+++ b/src/usr/local/www/services_dhcp_edit.php
@@ -30,6 +30,10 @@
 ##|*MATCH=services_dhcp_edit.php*
 ##|-PRIV
 
+function ipVersion($addr) {
+	return strpos($addr, ":") === false ? 4 : 6;
+}
+
 function staticmapcmp($a, $b) {
 	return ipcmp($a['ipaddr'], $b['ipaddr']);
 }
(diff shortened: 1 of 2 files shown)


lukehamburg:
Pushed another change, adding the **Copy DUID** convenience button to System > Advanced > Networking.

Inspiration from @marjohn56 's patch https://github.com/pfsense/pfsense/commit/59a31862f9bdbcc000734d8914e52f4e88872e97 but I prefer this simpler method since it's already used elsewhere in pfSense.

src/usr/local/www/services_dhcp_edit.php
```diff
@@ -385,8 +389,14 @@ function staticmaps_sort($ifgui) {
 
 // Get our MAC address
 $ip = $_SERVER['REMOTE_ADDR'];
-$mymac = `/usr/sbin/arp -an | grep '('{$ip}')' | cut -d" " -f4`;
-$mymac = str_replace("\n", "", $mymac);
+
+if ((ipVersion($ip))=='4') {
```
rbgarga: is_ipaddr($ip) will return 4 if it's IPv4, 6 if it's IPv6 or false if it's not a valid IP address. You don't need to add this ipVersion() function here

lukehamburg:
Ah, wasn't aware of that but now I see it. I will study `util.inc` to learn what functions are available. Will submit a new patch shortly.

lukehamburg:
Sent updated PR https://github.com/pfsense/pfsense/pull/3489

PR #3484, opened 2017-02-06
Do not pass empty string to gettext

phil-davis:
For example, system.php DNS Servers. Switch to a non-English language. Have more than 1 DNS Server specified. The setHelp text for the last row comes out fine, because it is "real text". For all but the last row, it does setHelp('') and when in non-English it spews out a bunch of text about the POT file...

So I tried a bunch of other form elements and used blank titles, checkbox descriptions and so on. They all behave the same.

Often sending blank for a title, description or whatever is a silly thing to do, and we could just recode all callers so they never bother to make the call if the text is blank. But it seems better to make sure to allow blank text in the calling parameter and handle it better.

### DIFF
--- a/src/usr/local/www/classes/Form/Checkbox.class.php
+++ b/src/usr/local/www/classes/Form/Checkbox.class.php
@@ -56,9 +56,9 @@ class Form_Checkbox extends Form_Input
 	{
 		$input = parent::_getInput();
 
-		if (empty($this->_description))
-			return '<label class="chkboxlbl">'. $input .'</label>';
+		if (!empty($this->_description) || is_numeric($this->_description))
+			return '<label class="chkboxlbl">'. $input .' '. htmlspecialchars(gettext($this->_description)) .'</label>';
 
-		return '<label class="chkboxlbl">'. $input .' '. htmlspecialchars(gettext($this->_description)) .'</label>';
+		return '<label class="chkboxlbl">'. $input .'</label>';
 	}
 }
--- a/src/usr/local/www/classes/Form/Button.class.php
+++ b/src/usr/local/www/classes/Form/Button.class.php
@@ -32,6 +32,12 @@ class Form_Button extends Form_Input
 
 	public function __construct($name, $title, $link = null, $icon = null)
 	{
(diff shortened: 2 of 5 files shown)


src/usr/local/www/classes/Form/Checkbox.class.php
```diff
@@ -56,9 +56,9 @@ protected function _getInput()
 	{
 		$input = parent::_getInput();
 
-		if (empty($this->_description))
-			return '<label class="chkboxlbl">'. $input .'</label>';
+		if (!empty($this->_description) || is_numeric($this->_description))
```
phil-davis: I switched this logic around to use the exact same boolean construction as in other Form class files. That should make it easier for future readers to understand, compare and debug.

phil-davis: This fixes the case where the checkbox description is "0". Prior to this change, the description "0" was not displayed.
(Don't you love the PHP empty() function?!)

doktornotor: @phil-davis - Due to exact same idiocy with ```empty()```, there's ```if ($foo != '') { }``` used in zillions of places.

phil-davis: Maybe we should patch the pfSense FreeBSD PHP port so that empty() behaves differently/better :)
To my mind "empty" should mean a zero-length string or the special value "null". Everything else is "something" - text, a number (even if it is zero), a boolean value true or false...
Or maybe a new function nothing() that does what empty() should do.

rbgarga: I guess you were probably kidding but anyway, no, we are not patching pfSense PHP to behave different from the original. PHP is what it is :)

doktornotor: Actually, the ```nothing()``` thing sounds like a good candidate for ```util.inc``` or something, 🤣

phil-davis: Yes, I was kidding about patching PHP!

PR #3487, opened 2017-02-06
Internationalization graph.php

phil-davis:
1) A few extra strings to be translated.
2) The scale_type code values 'up' and 'follow' need to be kept as those strings in the JS scale_type variable. The text to be translated needs to be in a different variable that can be in translated form for display. Otherwise things go wrong when in a different language.
e.g. switch to Brazilian Portuguese, then click the Autoscale. The first times it switches then gets "stuck".

### DIFF
--- a/src/usr/local/www/graph.php
+++ b/src/usr/local/www/graph.php
@@ -108,7 +108,7 @@ print('<?xml version="1.0" encoding="UTF-8"?>' . "\n");?>
 		<text id="graph_out_txt" x="20" y="16" <?=$attribs['out']?>> </text>
 		<text id="ifname" x="<?=$width?>" y="8" <?=$attribs['graphname']?> text-anchor="end"><?=htmlspecialchars($ifname)?></text>
 		<text id="switch_unit" x="<?=$width*0.55?>" y="5" <?=$attribs['switch_unit']?>><?=gettext("Switch to bytes/s"); ?></text>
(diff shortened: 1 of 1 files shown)


rbgarga:
The way you are implementing it will not add strings 'up' and 'down' to pot file. I couldn't see where a gettext() call to explicit strings is made.

src/usr/local/www/graph.php
```diff
@@ -188,13 +189,14 @@ function init(evt) {
...
-	SVGDoc.getElementById('switch_unit').firstChild.data = '<?=gettext("Switch to"); ?> ' + unit + '/s';
+	SVGDoc.getElementById('switch_unit').firstChild.data = (unit == 'bits') ? '<?=gettext("Switch to bits/s"); ?>' : '<?=gettext("Switch to bytes/s"); ?>';
 	unit = (unit == 'bits') ? 'bytes' : 'bits';
 }
 
 function switch_scale(event) {
-	scale_type = (scale_type == 'up') ? '<?=gettext("follow"); ?>' : '<?=gettext("up"); ?>';
-	SVGDoc.getElementById('switch_scale').firstChild.data = 'AutoScale (' + scale_type + ')';
+	scale_type = (scale_type == 'up') ? 'follow' : 'up';
+	scale_type_text = (scale_type == 'up') ? '<?=gettext("up"); ?>' : '<?=gettext("follow"); ?>';
```
phil-davis: @rbgarga The gettext() calls here should get the strings in the POT file for translation, similar to what was at old line 196.
And so the gettext() at line 181 should have the possibility of finding a translation.

rbgarga: Ooops, just missed that. Thanks!

PR #3488, opened 2017-02-06
Revert PERL related default options from net-mgmt/net-snmp

dbaio:
Hi @jim-p, 
net-mgmt/net-snmp is a dependency from Zabbix Proxy and those options are important for it deal with traps.
Could this be reverted ?

And thanks for your/netgate work on net-mgmt/net-snmp. It will be very useful.

Regards.


jim-p:
It should be OK, I was trying to keep the dependencies minimal.

dbaio:
Ok then, thanks.

PR #3489, opened 2017-02-06
Improved services_dhcp_edit.php and adds Copy DUID button

lukehamburg:
Reworked using switch statement and builtin is_ipaddr() function from util.inc as suggested by @rbgarga


lukehamburg:
Sorry - this got messy because I accidentally committed my original patch to my 'master' branch, so https://github.com/pfsense/pfsense/pull/3489/commits/677116cdda4e1c0368fbba8c00342e8d9d0ad616 is now hanging around. I think it's best if this PR is deleted, I will make new branches and send separate clean patches.

PR #3492, opened 2017-02-06
enhancements to services_dhcp_edit.php

luckman212:
- added ndp call to get MAC addr if remote client is connected via IPv6
- automatically hide `Copy MAC` button if arp/ndp returns null
- switch to exec() instead of backticks for calls to arp
- uses builtin is_ipaddr() function from util.inc

_(supersedes https://github.com/pfsense/pfsense/pull/3483 and https://github.com/pfsense/pfsense/pull/3489)_

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1811,22 +1811,25 @@ function alias_expand_urltable($name) {
 	return null;
 }
 
-/* obtain MAC address given an IP address by looking at the ARP table */
+/* obtain MAC address given an IP address by looking at the ARP/NDP table */
 function arp_get_mac_by_ip($ip) {
-	mwexec("/sbin/ping -c 1 -t 1 " . escapeshellarg($ip), true);
-	$arpoutput = "";
-	exec("/usr/sbin/arp -n " . escapeshellarg($ip), $arpoutput);
-
-	if ($arpoutput[0]) {
-		$arpi = explode(" ", $arpoutput[0]);
-		$macaddr = $arpi[3];
+	unset($arp_ndp_output);
+	switch (is_ipaddr($ip)) {
+		case 4:
+			mwexec("/sbin/ping -c 1 -t 1 " . escapeshellarg($ip), true);
+			exec("/usr/sbin/arp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $4}'", $arp_ndp_output);
+			break;
+		case 6:
+			mwexec("/sbin/ping6 -c 1 -X 1 " . escapeshellarg($ip), true);
+			exec("/usr/sbin/ndp -n " . escapeshellarg($ip) . " | /usr/bin/tail -n+2 | /usr/bin/awk '{print $2}'", $arp_ndp_output);
+			break;
+	}
+	if ($arp_ndp_output[0]) {
+		$macaddr = $arp_ndp_output[0];
 		if (is_macaddr($macaddr)) {
 			return $macaddr;
-		} else {
-			return false;
 		}
 	}
-
 	return false;
 }
 
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1813,24 +1813,23 @@ function alias_expand_urltable($name) {
 
 /* obtain MAC address given an IP address by looking at the ARP/NDP table */
 function arp_get_mac_by_ip($ip) {
-	unset($arp_ndp_output);
+	unset($macaddr);
+	$retval = 1;
 	switch (is_ipaddr($ip)) {
 		case 4:
 			mwexec("/sbin/ping -c 1 -t 1 " . escapeshellarg($ip), true);
-			exec("/usr/sbin/arp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $4}'", $arp_ndp_output);
+			$macaddr = exec("/usr/sbin/arp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $4}'", $output, $retval);
 			break;
 		case 6:
 			mwexec("/sbin/ping6 -c 1 -X 1 " . escapeshellarg($ip), true);
-			exec("/usr/sbin/ndp -n " . escapeshellarg($ip) . " | /usr/bin/tail -n+2 | /usr/bin/awk '{print $2}'", $arp_ndp_output);
+			$macaddr = exec("/usr/sbin/ndp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $2}'", $output, $retval);
 			break;
 	}
-	if ($arp_ndp_output[0]) {
-		$macaddr = $arp_ndp_output[0];
-		if (is_macaddr($macaddr)) {
-			return $macaddr;
-		}
+	if ($retval == 0 && is_macaddr($macaddr)) {
+		return $macaddr;
+	} else {
+		return false;
 	}
-	return false;
 }
 
 /* return a fieldname that is safe for xml usage */
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1812,7 +1812,7 @@ function alias_expand_urltable($name) {
 }
 
 /* obtain MAC address given an IP address by looking at the ARP/NDP table */
-function arp_get_mac_by_ip($ip, $do_ping = false) {
+function arp_get_mac_by_ip($ip, $do_ping = true) {
 	unset($macaddr);
 	$retval = 1;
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_dhcp_edit.php
```diff
@@ -385,8 +385,19 @@ function staticmaps_sort($ifgui) {
...
+		$mymac = exec("/usr/sbin/arp -an | grep '('{$ip}')' | head -n 1 | cut -d' ' -f4");
+		$mymac = str_replace("\n", "", $mymac);
+        break;
+    case 6:
+		$mymac = exec("/usr/sbin/ndp -na | /usr/bin/grep '{$ip}' | /usr/bin/head -n 1 | /usr/bin/awk '{ print $2 }'");
+		$mymac = str_replace("\n", "", $mymac);
+		break;
+	default:
+		unset($mymac);
+}
```
doktornotor: @rbgarga - like ```arp_get_mac_by_ip()``` from util.inc? :)

rbgarga: Ouch, I was blink when I looked around. Yes, but it needs to be changed to work with IPv6 at least

lukehamburg: Guys I am at work another 2 hrs but I will work on that tonight and hopefully roll up these suggested changes.  Thanks.

luckman212:
I tested this latest commit https://github.com/pfsense/pfsense/pull/3492/commits/dd83f869b79a858bd74c7a8bb4adcd49217445b0 against latest snap 2.4.0.b.20170207.2344 seems to work ok for me.

I also had a look at some other files that invoke `arp` and found some problems:

**/usr/local/www/diag_arp.php**
- hardcoded references to 8.8.8.8 / 8.8.4.4 for DNS lookups? (~L295)
- `$awk` var defined on L76 but then it is inconsistently used, e.g. not referenced on L176
- seems to be a terribly inefficient loop to check if host is "online":  `arp -an` is run N times (N=num of rows) dumping the whole arp table each time into awk and then further into `wc -l` (yuck). The way `status_dhcp_leases.php` does it by building and searching an array seems to be much better. I'll see if I can submit a PR for that.

**/usr/local/www/interfaces.php**
- I see that L1774 has the "Copy MAC" button disabled; so I assume that function is being removed from that page? In that case there is no need for $mymac var to be set nor the calls to arp/grep/head/cut etc in L1661-1664
- or, if you intend to bring back the function, then I can submit a PR to pull in the fixes from services_dhcp_edit.php

**/usr/local/www/status_dhcp_leases.php**

- looks ok

src/etc/inc/util.inc
```diff
@@ -1811,22 +1811,25 @@ function alias_expand_urltable($name) {
...
+	unset($arp_ndp_output);
+	switch (is_ipaddr($ip)) {
+		case 4:
+			mwexec("/sbin/ping -c 1 -t 1 " . escapeshellarg($ip), true);
+			exec("/usr/sbin/arp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $4}'", $arp_ndp_output);
+			break;
+		case 6:
+			mwexec("/sbin/ping6 -c 1 -X 1 " . escapeshellarg($ip), true);
+			exec("/usr/sbin/ndp -n " . escapeshellarg($ip) . " | /usr/bin/tail -n+2 | /usr/bin/awk '{print $2}'", $arp_ndp_output);
+			break;
```
rbgarga: exec returns last line of command output, and also fill all output lines to second parameter, if it's present. It could be even simpler if you consider exec's return, so you don't need to call tail. Also, it would be nice to check for command return code. I thought about something like:
```php
unset($macaddr);
$rc = 1;
switch (is_ipaddr($ip)) {
case 4:
	mwexec("/sbin/ping -c 1 -t 1 " . escapeshellarg($ip), true);
	$macaddr = exec("/usr/sbin/arp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $4}'", $output, $rc);
	break;
case 6:
	mwexec("/sbin/ping6 -c 1 -X 1 " . escapeshellarg($ip), true);
	$macaddr = exec("/usr/sbin/ndp -n " . escapeshellarg($ip) . " | /usr/bin/awk '{print $2}'", $output, $rc);
	break;
}

return ($rc == 0 && is_macaddr($macaddr));
```

src/usr/local/www/services_dhcp_edit.php
```diff
@@ -385,8 +385,9 @@ function staticmaps_sort($ifgui) {
 
 // Get our MAC address
 $ip = $_SERVER['REMOTE_ADDR'];
-$mymac = `/usr/sbin/arp -an | grep '('{$ip}')' | cut -d" " -f4`;
-$mymac = str_replace("\n", "", $mymac);
+
+unset($mymac);
```
rbgarga: unset is unnecessary here since you are attributing a value to the same variable in the next line

luckman212:
Thanks @rbgarga - I just pushed another attempt. Your new `arp_get_mac_by_ip()` function seems good, but verbatim it just returned the result code of the AND (which was usually just `1`).  I had to make a small modification so $macaddr was passed back.  Hope this one is ok.

src/usr/local/www/services_dhcp_edit.php
```diff
@@ -385,8 +385,7 @@ function staticmaps_sort($ifgui) {
 
 // Get our MAC address
 $ip = $_SERVER['REMOTE_ADDR'];
-$mymac = `/usr/sbin/arp -an | grep '('{$ip}')' | cut -d" " -f4`;
-$mymac = str_replace("\n", "", $mymac);
+$mymac = arp_get_mac_by_ip($ip);
```
rbgarga: The new version of arp_get_mac_by_ip() looks good but I was thinking about it and there is still a difference from the current code. arp_get_mac_by_ip() always run a ping to make sure the address will be in the table, and this is not desired in this case because if IP from $_SERVER['REMOTE_ADDR'] is from a local network it will be there.

That ping will introduce 1s delay every time this page is used.

I was thinking it would be nice to add a second parameter to arp_get_mac_by_ip(), default to true, and when it's false don't call ping/ping6.

luckman212:
Yes I agree 100% -- I did not like seeing the `ping` in there either but, it was there before I started working on it and I figured people wiser than I had put it there for a reason.

But there are too many drawbacks to having it there- e.g. as you said if the client is not on the same L2 network the ping will either fail (causing delay) or possibly return the wrong MAC.

Even if it is a local client, if there is a software firewall in place blocking icmp, the ping will also fail, producing a delay.

I will try harder to come up with a better solution. I think there is a compare_subnet() type function, maybe a call to that to determine if the client is local is better than a ping.

Doing away with the ping altogether is probably the best choice.  I cannot think of a case when it would really improve the results. If pfSense is being accessed locally then the MAC is going to be in the arp/ndp tables already by the time PHP processes that page load, so an extra ping is probably not helping anything.

luckman212:
The only place I could find `arp_get_mac_by_ip()` in use is in **services_dhcp_edit.php**. I did notice the `ip_in_subnet()` function and thought about looping through active interfaces and testing each to see if the REMOTE_ADDR was inside one of these, but that seemed like a lot of wasted CPU cycles for very little benefit.

I liked your idea of the additional parameter to control the ping though, as that might be useful in future situations where a ping is desired, even if it might incur a delay. So I added a `do_ping` param but inverted the behavior so it defaults to 'false' ("do not ping"). If set to 'true' then the ping/ping6 will be executed. I confirmed this worked using Wireshark to monitor the traffic while I refreshed the page.

I think (and hope) this commit is the most sensible solution for now.

src/etc/inc/util.inc
```diff
@@ -1811,23 +1811,29 @@ function alias_expand_urltable($name) {
...
-	if ($arpoutput[0]) {
-		$arpi = explode(" ", $arpoutput[0]);
-		$macaddr = $arpi[3];
-		if (is_macaddr($macaddr)) {
-			return $macaddr;
-		} else {
-			return false;
-		}
+/* obtain MAC address given an IP address by looking at the ARP/NDP table */
+function arp_get_mac_by_ip($ip, $do_ping = false) {
```
rbgarga: I prefer to keep the old behavior as default just to avoid having surprises on edge cases. Please change default $do_ping value to true.

luckman212:
@rbgarga I just pushed that change.  I understand your reasoning but I did search for any other places that were calling `arp_get_mac_by_ip()` and could not find any. Hope this patch is good to go now.

PR #3495, opened 2017-02-06
fbegin.inc and fend.inc are no longer used

phil-davis:
As far as I can tell these are no longer in use.
And I suspect there is a file somewhere where these should be listed so that they will be cleaned up on upgrades etc - but I can;t remember where it is.

### DIFF
--- a/src/etc/pfSense.obsoletedfiles
+++ b/src/etc/pfSense.obsoletedfiles
@@ -927,6 +927,8 @@
 /usr/local/www/edit.php
 /usr/local/www/exec.php
 /usr/local/www/exec_raw.php
+/usr/local/www/fbegin.inc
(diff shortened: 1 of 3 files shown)


rbgarga:
yes, it's src/etc/pfSense.obsoletedfiles

rbgarga:
@phil-davis did you check all packages on pfSense/FreeBSD-ports to make sure nobody is using it?

doktornotor:
These are still lingering in 
- Suricata
- Snort
- tinc

Will do some PRs shortly.

doktornotor:
Should be all, at least to GitHub search.

phil-davis:
Added to obsolete files list. This should be good to go once the FreeBSD-ports changes have been merged/released.

PR #3498, opened 2017-02-07
interfaces_ppps_edit text format and validation

phil-davis:
1) Remove embedded HTML from strings.
2) Make "The MTU is too big" message actually come out. The code around line 300 was rubbish (maybe from before bootstrap?).
3) Show friendly description rather than 'wan' 'lan' opt1' in interface link list.
4) If any specific link parameters are set, then open the advanced section on page load.

Basically every time I tried to do something to display/test the various error messages I found some other issue. So this became a general "keep fixing it" change.

### DIFF
--- a/src/usr/local/www/interfaces_ppps_edit.php
+++ b/src/usr/local/www/interfaces_ppps_edit.php
@@ -294,26 +294,20 @@ if (isset($_POST) && is_array($_POST) && count($_POST) > 0) {
 			}
 		}
 
-		// Loop through fields associated with an individual link/port and make an array of the data
-		$port_fields = array("localip", "gateway", "subnet", "bandwidth", "mtu", "mru", "mrru");
+		// Loop through each individual link/port and check max mtu
 		foreach ($_POST['interfaces'] as $iface) {
-			foreach ($port_fields as $field_label) {
-				if (isset($_POST[$field_label . $iface]) &&
-				    strlen($_POST[$field_label . $iface]) > 0) {
-					$port_data[$field_label][] = $_POST[$field_label . $iface];
-					$pconfig[$field_label][$iface] = $_POST[$field_label . $iface];
-					$parent_array = get_parent_interface($iface);
-					$parent = $parent_array[0];
-					$friendly = convert_real_interface_to_friendly_interface_name($parent);
-					if ($field_label == "mtu" && isset($config['interfaces'][$friendly]['mtu']) &&
-					    $_POST[$field_label . $iface] > ($config['interfaces'][$friendly]['mtu'] - 8)) {
-						$input_errors[] = sprintf(gettext("The MTU (%d) is too big for %s (maximum allowed with current settings: %d)."),
-						    $_POST[$field_label . $iface], $iface, $config['interfaces'][$friendly]['mtu'] - 8);
-					}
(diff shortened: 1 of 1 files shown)


src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -294,26 +294,20 @@
...
 		}
 
-		// Loop through fields associated with an individual link/port and make an array of the data
-		$port_fields = array("localip", "gateway", "subnet", "bandwidth", "mtu", "mru", "mrru");
+		// Loop through each individual link/port and check max mtu
 		foreach ($_POST['interfaces'] as $iface) {
-			foreach ($port_fields as $field_label) {
-				if (isset($_POST[$field_label . $iface]) &&
-				    strlen($_POST[$field_label . $iface]) > 0) {
-					$port_data[$field_label][] = $_POST[$field_label . $iface];
```
phil-davis: Note: $port_data is actually set up correctly further down, this line was crud that was never being executed.

src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -294,26 +294,20 @@
...
-					$pconfig[$field_label][$iface] = $_POST[$field_label . $iface];
```
phil-davis: Note: $pconfig = $_POST is above this, and already did what this was attempting to do. This line was crud that was never being executed.

src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -294,26 +294,20 @@
 			}
 		}
 
-		// Loop through fields associated with an individual link/port and make an array of the data
-		$port_fields = array("localip", "gateway", "subnet", "bandwidth", "mtu", "mru", "mrru");
```
phil-davis: We do not need to loop through all these fields any more, the only field that has validation is 'mtu'.

rbgarga:
@phil-davis can you split it in 2 different PRs? RELENG_2_3 doesn't have the new SetHelp() format but other changes are useful to be cherry-picked there

phil-davis:
Yep, will do.

phil-davis:
See #3501

src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -294,26 +294,20 @@
...
-					    $_POST[$field_label . $iface] > ($config['interfaces'][$friendly]['mtu'] - 8)) {
-						$input_errors[] = sprintf(gettext("The MTU (%d) is too big for %s (maximum allowed with current settings: %d)."),
-						    $_POST[$field_label . $iface], $iface, $config['interfaces'][$friendly]['mtu'] - 8);
-					}
+			if (isset($_POST['mtu'][$iface]) &&
+			    strlen($_POST['mtu'][$iface]) > 0) {
+				$parent_array = get_parent_interface($iface);
+				$parent = $parent_array[0];
+				$friendly = convert_real_interface_to_friendly_interface_name($parent);
+				if (isset($config['interfaces'][$friendly]['mtu']) &&
```
phil-davis: Note: This actually is better (needs to be) a !empty() rather than isset(). That has been fixed in the later PR.

PR #3501, opened 2017-02-07
interfaces_ppps_edit fixes that will work on 2.3.3 also

phil-davis:
1) Make "The MTU is too big" message actually come out. The code around line 300 was rubbish (maybe from before bootstrap?).
2) Show friendly description rather than 'wan' 'lan' opt1' in interface link list.
3) If any specific link parameters are set, then open the advanced section on page load.

I tested on a 2.4 system, then applied these diff on a 2.3.3 system and it worked there also. So this should backport fine.


phil-davis:
As requested in #3498 this is the changes that work with 2.3.3 also.
After this is merged I will make another PR on top for the setHelp changes that are not compatible with 2.3.3

PR #3505, opened 2017-02-08
Part fix #7233 keep correct staticmap_array_index

phil-davis:
staticmap_array_index needs to be incremented even for entries that were skipped for display because they (for whatever reason) have no MAC or CID set.

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -300,8 +300,9 @@ foreach ($config['interfaces'] as $ifname => $ifarr) {
 				$slease['online'] = in_array(strtolower($slease['mac']), $arpdata_mac) ? $online_string : $offline_string;
 				$slease['staticmap_array_index'] = $staticmap_array_index;
 				$leases[] = $slease;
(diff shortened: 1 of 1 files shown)


src/usr/local/www/status_dhcp_leases.php
```diff
@@ -300,8 +300,9 @@ function remove_duplicate($array, $field) {
 				$slease['online'] = in_array(strtolower($slease['mac']), $arpdata_mac) ? $online_string : $offline_string;
 				$slease['staticmap_array_index'] = $staticmap_array_index;
 				$leases[] = $slease;
-				$staticmap_array_index++;
 			}
+
+			$staticmap_array_index++;
```
rbgarga: I was wondering if foreach from line 284 couldn't be replaced by:
```
foreach ($config['dhcpd'][$ifname]['staticmap'] as $idx => $static) {
```
and then $idx would be our staticmap_array_index

phil-davis: Yes, that should work. I will resubmit.

phil-davis:
Closing in favor of #3507

PR #3509, opened 2017-02-08
Pass command line argv to playback file

mlowijs:


jim-p:
What are you trying to solve here? Playback files can already access arguments using `$command_split` or  `$argv` as a global.

mlowijs:
Is that so? I don't seem to be able to access the `$argv` variable in the playback file, it is undefined (I renamed the `pfSsh.php` file to `pfSsh.php.old`:

    [2.3.1-RELEASE][admin@router.sycade.int]/usr/local/sbin: ./pfSsh.php.old run mytestfile someArg1 someArg2

    Starting the pfSense developer shell....

    NULL

The playback file (mytestfile):
`var_dump($argv);`

mlowijs:
Oh right, my PHP is a little rusty, I forgot the `global $argv` 🤦‍♂️

PR #3513, opened 2017-02-08
setHelp strings for services*

phil-davis:


phil-davis:
I added another commit, because now that a string in services_dyndns_edit has %s parameters, other genuine "%" characters for display need to be escaped. So my first merge integration actually got it wrong.

PR #3524, opened 2017-02-10
Improve/add help messages warning about bad security choices

seanm:


sbeaver-netgate:
Requested changes not addressed

PR #3530, opened 2017-02-12
Update version string at end of boot RELENG_2_3

phil-davis:
When there is an upgrade, the echo here was outputting a stale value of the version. For example, on first upgrade from 2.3.3-DEVELOPMENT to 2.3.3-RC the console had:

pfSense (pfSense) 2.3.3-DEVELOPMENT amd6 Sat Feb 11 14:24:27 CST 2017
Bootup complete

FreeBSD/amd64 (myhost.localdomain) (ttyv0)

*** Welcome to pfSense 2.3.3-RC (amd64 full-install) on myhost ***

That is a bit confusing for users to be sure which version it is at this point.

### DIFF
--- a/src/etc/rc
+++ b/src/etc/rc
@@ -67,20 +67,25 @@ export HOME PATH
 # Set our operating platform
 PLATFORM=`/bin/cat /etc/platform`
 
-# Set our current version
-version=`/bin/cat /etc/version`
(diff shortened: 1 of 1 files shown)


phil-davis:
pfsense233rc-upgrade

PR #3532, opened 2017-02-12
Fix #7254 Fw rules interface selector

phil-davis:
This at least fixes the dropdown selector.
https://redmine.pfsense.org/issues/7254


phil-davis:
There is probably some better way to do this? @sbeaver-netgate

phil-davis:
Closing in favor of #3533 which has just a single commit.

PR #3534, opened 2017-02-12
Make breadcrumb links work

phil-davis:
To make breadcrumb links work again with the new "POST not GET" code, something like this can be done.
1) In guiconfig.inc set breadcrumb links to "usepost"
2) In each place that uses the "@self" reference, instead we have to build the link with URL parameters based on what POST parameters are needed for each page. I have done that for firewal_rules and firewall_rules_edit as examples.

This gives an idea of what would be required.

Of course it still means that the user cannot bookmark a link directly to the rules of a particular interface...

### DIFF
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -310,12 +310,13 @@ foreach ($iflist as $ifent => $ifname) {
 foreach ($tab_array as $dtab) {
 	if ($dtab[1]) {
 		$bctab = $dtab[0];
+		$bclink = $dtab[2];
 		break;
 	}
 }
 
 $pgtitle = array(gettext("Firewall"), gettext("Rules"), $bctab);
-$pglinks = array("", "firewall_rules.php", "@self");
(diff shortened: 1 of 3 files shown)


sbeaver-netgate:
Thanks Phil. This is an interesting solution. At the moment I am inclined to revert the ?act=new type links to $_REQUEST['act'] but I am still thinking/testing

phil-davis:
Yes, I think that any link that retrieves data for edit/display should work like it used to.
?id=nnn
?if=opt1
?tab=xyz

That way:
1) The breadcrumbs will work easily to refresh the page being edited.
2) Users can bookmark the link with parameters that will bring them to the needed interface, tab etc.
3) Users can scroll back through history and find the same things that they viewed previously

phil-davis:
You can add the "discussion" label to this one, because actually it is just an example and I don't think we will end up doing this.

phil-davis:
The GET/POST stuff has been revised so that the existing breadcrumbs should all work. I don't think that this is needed now?

sbeaver-netgate:
Agreed

PR #3537, opened 2017-02-12
Fix #7252 OpenVPN connect time

phil-davis:


phil-davis:
Or cherry-pick https://github.com/pfsense/pfsense/commit/6f17547aaa073ca460c8bbd8dc1dc5c12922b76d
I guess that the new OpenVPN version was added to 2.3.3 but this bit of code to handle the enhanced status output was not back-ported.

jim-p:
Closing this since we're moving 2.3.3 back to OpenVPN 2.3 so it should remove the need for this change.

PR #3538, opened 2017-02-12
Use cached groups in get_user_privileges

phil-davis:
This should help with https://redmine.pfsense.org/issues/7253
Calls to userHasPrivilege(), which calls get_user_privileges() will be able to complete out of the cache. So this should be no more of a pain when the LDAP or RADIUS server is not responding than existing calls to getAllowedPages().
This code to utilitize a cache of allowed groups is taken from the existing code in getAllowedPages() and massaged a bit to use the way $user is passed in here.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -305,7 +305,7 @@ function get_user_privileges(& $user) {
 
 	if ($authcfg['type'] == "ldap") {
 		if (isset($_SESSION["ldap_allowed_groups"]) &&
-		    (time() <= $_SESSION["auth_check_time"]+ $recheck_time)) {
+		    (time() <= $_SESSION["auth_check_time"] + $recheck_time)) {
 			$allowed_groups = $_SESSION["ldap_allowed_groups"];
 		} else {
 			$allowed_groups = @ldap_get_groups($user['name'], $authcfg);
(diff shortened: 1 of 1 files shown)


phil-davis:
I have tested that local users still work, but I am not in a position to test LDAP and RADIUS (with the LDAP or RADIUS server up and down). But theoretically this should help.

phil-davis:
There might even be a case for storing the $privs (that is built and returned by get_user_privileges()) somewhere globally that can be directly returned if not stale, rather than building it every time from the user privs and group privs. But I guess that is just a CPU performance thing.

doktornotor:
OK, this is much better, except that it still spews warning garbage on every page.

```
/usr/local/www/head.inc:447 0.0745 3491880 4. get_user_privileges() /etc/inc/auth.inc:378 Warning: Invalid argument supplied for foreach() in /etc/inc/auth.inc on line 362 Call Stack: 0.0001 233232 1. {main}() /usr/local/www/system_authservers.php:0 0.0545 2948016 2. include('/usr/local/www/head.inc') /usr/local/www/system_authservers.php:441 0.0747 3490392 3. userHasPrivilege() /usr/local/www/head.inc:450 0.0747 3491880 4. get_user_privileges() /etc/inc/auth.inc:378
```

phil-davis:
This code makes the assumption that ldap_get_groups() returns a value that is either false/empty or an array of the groups (an empty array of no groups is fine).
But ldap_get_groups() has a bug where it can return a string rather than an array.
See https://github.com/pfsense/pfsense/pull/3539 for a fix of this underlying bug that exposed this.

I will also fix the code here to protect itself...

phil-davis:
Made a branch in my local git with a single clean commit and push it up as PR #3540

PR #3542, opened 2017-02-13
Allow saving of CA when in non-English language

phil-davis:
1) Switch to Brazilian
2) Add a CA, try to "Salvar". It does not save.

I wonder how many other places there are like this in the code?

### DIFF
--- a/src/usr/local/www/system_camanager.php
+++ b/src/usr/local/www/system_camanager.php
@@ -153,7 +153,7 @@ if ($act == "expkey") {
 	exit;
 }
 
-if ($_POST['save'] == 'Save') {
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
Quite a few. I am attempting to weed them out as I do the GET/POST thing

phil-davis:
Note: Interestingly this is only a problem in 2.4.
In 2.3.3 the code is different and does not do the extra check on the value of $_POST['save'], and it works OK.
I didn't bother checking the history to see why the more specific check is now needed.

sbeaver-netgate:
Because $POST is no longer the exclusive domain of saving. It is used by deletes as well.

PR #3545, opened 2017-02-14
Only save valid widget locations in config

phil-davis:
Some widgets create extra panels, e.g. the widgets that now have the filter functionality. Those panels are processed in the ".each" at line 424. They do not have an id in the form "widget-*" and when the old code tries to find the "*" part it gets "undefined". This results in the layout being saved like:
```
<sequence>interface_statistics:col1:open,undefined:col1:close,system_information:col2:open,undefined:col2:close,picture:col3:open,rss:col3:open,ntp_status:col2:open</sequence>
```

This PR puts extra checks in place so that the "undefined" stuff does not get written into the widget sequence string. With the fixed code (and me moving widgets on my dashboard), the sequence looks like:
```
<sequence>picture:col1:open,interface_statistics:col1:open,system_information:col2:open,ntp_status:col2:open,rss:col3:open</sequence>
```

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -423,8 +423,13 @@ function updateWidgets(newWidget) {
 	$('.container .col-md-<?=$columnWidth?>').each(function(idx, col) {
 		$('.panel', col).each(function(idx, widget) {
 			var isOpen = $('.panel-body', widget).hasClass('in');
+			var widget_basename = widget.id.split('-')[1];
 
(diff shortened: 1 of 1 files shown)


phil-davis:
In the original version nothing visibly breaks. On loading the dashboard layout, the 'undefined' entries are effectively ignored. So this fix is mainly to avoid people wondering why this crud is in their config.

The widgets that cause this are in RELENG_2_3 also, so this should be back-ported to RELENG_2_3.

If there is another round of building RELENG_2_3_3 then it could be back-ported to RELENG_2_3_3 also.

PR #3546, opened 2017-02-14
Vendor MAC Retention, Restore Logic, Consolidate

NOYB:
Spoof MAC Var Name

Rename 'spoof_mac' var to generic 'mac_addr'.
 a) It may be the vendor MAC or a spoofed MAC.
 b) Update the comment re: not reapplying an already applied MAC.

Vendor MAC Restore Logic

Only use the vendor MAC retention file for restoring the vendor MAC when not booting.
 a) During boot up the current MAC that is obtained from the system is the vendor MAC.
 b) Using this eliminates the inefficient need to open the vendor MAC retention file for every interface during system boot up.

Vendor MAC Retention File Relocate

Relocate the vendor MAC retention file to /var/db directory.
 a) It's more at home here with other network interface stuff.
 b) Friendlier to write cycle sensitive media in a RAM disk enabled system.

Vendor MAC Retention File Consolidate

Use a single file for vendor MAC retention (vendor_mac).
 a) Writes only one file during boot up rather than a file for each interface.
 b) More efficient than numerous tiny files.
 c) Friendlier to write cycle sensitive media in a RAM disk disabled system.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -6005,12 +6005,10 @@ function get_interface_vendor_mac($interface) {
 	global $g;
 
 	$mac = "";
-	$mac_file = "{$g['vardb_path']}/vendor_mac_($interface)";
+	$mac_file = "{$g['vardb_path']}/vendor_mac";
 	if (file_exists($mac_file)) {
-		$mac = file_get_contents($mac_file);
-		if (!is_macaddr($mac)) {
-			$mac = "";
-		}
+		$vendor_mac_arr = json_decode(file_get_contents($mac_file), true);
+		$mac = (is_macaddr($vendor_mac_arr[$interface])) ? $vendor_mac_arr[$interface] : '';
 	}
 
 	return $mac;
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3274,16 +3274,16 @@ function interface_configure($interface = "wan", $reloadall = false, $linkupeven
 
 	$current_mac = get_interface_mac($realhwif);
 	$vendor_mac = get_interface_vendor_mac($realhwif);
-	$spoof_mac = $wancfg['spoofmac'] ?: $vendor_mac;
+	$mac_addr = $wancfg['spoofmac'] ?: $vendor_mac;
 	/*
-	 * Don't try to reapply the spoofed MAC if it's already applied.
+	 * Don't try to reapply the MAC if it's already applied.
 	 * When ifconfig link is used, it cycles the interface down/up, which triggers
-	 * the interface config again, which attempts to spoof the MAC again,
+	 * the interface config again, which attempts to apply the MAC again,
 	 * which cycles the link again...
 	 */
-	if (!empty($spoof_mac) && ($spoof_mac != $current_mac)) {
+	if (!empty($mac_addr) && ($mac_addr != $current_mac)) {
 		mwexec("/sbin/ifconfig " . escapeshellarg($realhwif) .
-			" link " . escapeshellarg($spoof_mac));
+			" link " . escapeshellarg($mac_addr));
 	} elseif ($current_mac == "ff:ff:ff:ff:ff:ff") {
 		/*   this is not a valid mac address.  generate a
 		 *   temporary mac address so the machine can get online.
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3272,8 +3272,13 @@ function interface_configure($interface = "wan", $reloadall = false, $linkupeven
 		interface_wireless_configure($realif, $wancfg, $wancfg['wireless']);
 	}
 
+	/* Get the vendor MAC.  Use source dependent upon whether or not booting. */
 	$current_mac = get_interface_mac($realhwif);
-	$vendor_mac = get_interface_vendor_mac($realhwif);
+	if (platform_booting()) {
+		$vendor_mac = $current_mac;
+	} else {
+		$vendor_mac = get_interface_vendor_mac($realhwif);
+	}
 	$mac_addr = $wancfg['spoofmac'] ?: $vendor_mac;
 	/*
 	 * Don't try to reapply the MAC if it's already applied.
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -6005,7 +6005,7 @@ function get_interface_vendor_mac($interface) {
 	global $g;
 
 	$mac = "";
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -6000,12 +6005,13 @@ function get_interface_vendor_mac($interface) {
...
-	$mac_file = "{$g['cf_conf_path']}/vendor_mac_{$interface}";
+	$mac_file = "{$g['vardb_path']}/vendor_mac";
 	if (file_exists($mac_file)) {
-		$mac = file_get_contents($mac_file);
-		if (!is_macaddr($mac)) {
-			$mac = "";
-		}
+		$vendor_mac_contents = file_get_contents($mac_file);
+		$mac_pattern = '(?:[[:xdigit:]]{2}:){5}[[:xdigit:]]{2}';
+		$pattern = '/^'.$interface.': ('.$mac_pattern.')$/m';
```
rbgarga: If you use '=' to separate interface name from mac address it would be much simpler to read it using explode instead of a regex

src/etc/inc/interfaces.inc
```diff
@@ -3272,18 +3272,23 @@ function interface_configure($interface = "wan", $reloadall = false, $linkupeven
...
 
+	/* Get the vendor MAC.  Use source dependent upon whether or not booting. */
 	$current_mac = get_interface_mac($realhwif);
-	$vendor_mac = get_interface_vendor_mac($realhwif);
-	$spoof_mac = $wancfg['spoofmac'] ?: $vendor_mac;
+	if (platform_booting()) {
+		$vendor_mac = $current_mac;
+	} else {
+		$vendor_mac = get_interface_vendor_mac($realhwif);
+	}
```
rbgarga: Unfortunately you cannot trust interface_configure() will run only once during boot, there are edge cases where it runs more than once and in these cases you would end up with spoofed mac being stored in vendor_mac file.

You should check if there is already a definition in vendor_mac file, and if yes, consider that definition as vendor mac

src/etc/inc/interfaces.inc
NOYB: Been looking at this.  But don't think it is just a simple single operation.  So far what I've come up with is this.  Maybe you have better way?
```
$vendor_mac_lines = explode("\n",$vendor_mac_contents);

foreach ($vendor_mac_lines as $line) {
	$line = explode("=",$line);
	$iface = $line[0];
	$mac = $line[1];
	$vendor_mac[$iface] = $mac;
}

$mac = (is_macaddr($vendor_mac[$interface])) ? $vendor_mac[$interface] : "";
```

src/etc/inc/interfaces.inc
NOYB: Interface configure function doesn't write the vendor MAC file.  Vendor MAC file is written only once, during boot by rc.bootup.  Same as before.  So don't see how it could end up with a spoofed mac being stored in the vendor_mac file.

rbgarga: right, my mistake.

src/etc/inc/interfaces.inc
rbgarga: It could be simpler using file() instead of file_get_contents() like:

```php
$vendor_mac_lines = file("{$g['vardb_path']}/vendor_mac",
    FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
````

Comment (rbgarga) [inline src/etc/inc/interfaces.inc]: Oh, I just thought a second after I submitted previous comment. What about use json_encode() and json_decode(), this way PHP can convert it to an array even easier

Comment (NOYB) [inline src/etc/inc/interfaces.inc]: Changed to use json encode/decode.
Thanks for the pointer.

File: src/etc/rc.bootup [hunk 3]
```diff
@@ -126,10 +126,11 @@ if (mwexec("/bin/kenv -q pfSense.boot 2>/dev/null") != 0) {
 			continue;
 		}
 
-		@file_put_contents("{$g['cf_conf_path']}/vendor_mac_{$if}",
-		    $if_details['macaddr']);
+		$vendor_mac_arr[$if] = $if_details['macaddr'];
 	}
-	unset($ifs, $if);
+	$vendor_mac_contents = json_encode($vendor_mac_arr);
+	@file_put_contents("{$g['vardb_path']}/vendor_mac", $vendor_mac_contents);
```
rbgarga: It's not necessary to set a variable for a single use, you can call json_encode direct on file_put_contents();

src/etc/inc/interfaces.inc
```diff
@@ -6000,12 +6005,11 @@ function get_interface_vendor_mac($interface) {
...
 	$mac = "";
-	$mac_file = "{$g['cf_conf_path']}/vendor_mac_{$interface}";
+	$mac_file = "{$g['vardb_path']}/vendor_mac";
 	if (file_exists($mac_file)) {
-		$mac = file_get_contents($mac_file);
-		if (!is_macaddr($mac)) {
-			$mac = "";
-		}
+		$vendor_mac_contents = file_get_contents($mac_file);
+		$vendor_mac_arr = json_decode($vendor_mac_contents, true);
```
rbgarga: It's not necessary to set a variable for a single use, you can call json_decode direct on file_get_contents();

src/etc/rc.bootup
NOYB: Changed.

src/etc/inc/interfaces.inc
NOYB: Changed.

NOYB:
If we don't unset `$vendor_mac_arr` in rc.bootup, and add a global for it in `interface_configure()`, it would be available for obtaining the interfaces vendor mac during boot up.

```
global $vendor_mac_arr;
if ($vendor_mac_arr) {
	$vendor_mac = (is_macaddr($vendor_mac_arr[$realhwif])) ? $vendor_mac_arr[$realhwif] : '';
} else {
	$vendor_mac = get_interface_vendor_mac($realhwif);
}
```
Would that be better/preferred/beneficial than the if booting `$vendor_mac = $current_mac`.

```
if (platform_booting()) {
	$vendor_mac = $current_mac;
} else {
	$vendor_mac = get_interface_vendor_mac($realhwif);
}
```
If so, could it be boiled down to this?
`
$vendor_mac = (is_macaddr($vendor_mac_arr[$realhwif])) ? $vendor_mac_arr[$realhwif] : get_interface_vendor_mac($realhwif);
`

Perhaps when the API comes about, also adding a global to `get_interface_vendor_mac()` would allow for batch interface config without repeated calls to get the mac from the file.  But maybe that's looking too far down the road.

PR #3547, opened 2017-02-14
Allow multiple copies of a widget on dashboard

phil-davis:
This is a demonstration. It allows any widget to be put on the dashboard multiple times. The multiple copies become system_information-0 system_information-1 interfaces-0 interfaces-1 and so on.
Then the user can filter each widget instance to show just the items that they want. e.g. show a few items in system_information-0 and other items in system_information-1. Or if they have a lot of interfaces, show some in interfaces-0 and others in interfaces-1.
It effectively breaks up the System Information widget by letting the user to display multiple copies and choose which rows to display in each, so the user can break it up how they like.

The code of each widget needs to understand an comply with:
1) It will be widget-xyz-n (where it used to be widget-xyz)
2) If it does timed updating, it needs to update the data in all its instances
3) If it has a filter, it has to make sure to keep the filter lists separate for each instance, and the "All" button to only switch on the items in its own instance.

I have given examples of system_information and interfaces widget changes that are needed. Similar changes are needed in other widgets to make their tool-wrench... stuff work properly when there are multiple of them on the dashboard.

This would let users do other stuff, that may or may not be of any use, like:
- multiple picture widgets with different pictures
- multiple RSS widgets with each on a different RSS feed
...

This looks like it has some merge conflicts already. I don't care about that right now. I just want to get feedback about the idea.


phil-davis:
It's OK - the merge conflict was with a bugfix of mine that got commited while I was doing this new work, easily resolved.

phil-davis:
Note: This is work-in-progress. I need to get to finishing changes in the remaining widgets so they correctly support having multiple instances of a widget on the dashboard at the same time.
But feel free to give comments on the way the various issues here have been solved in each widget. There might be better ways to solve the various cases.

phil-davis:
I will leave this until #3593 has been discussed (and hopefully accepted and merged) because it logically comes first as a relatively small enhancement to the way the widget filter panel currently works, and it will make various conflicts that will need to be resolved with the bigger code changes here.

phil-davis:
There are other things to be done to the underlying widgets first. And refactoring of the way this can all be done. I will leave this open here so that I can find it easily, but there will be a different PR to come once the other widget changes are sorted.

phil-davis:
Closing in favor of #3658

PR #3548, opened 2017-02-16
option to make default deny a reject vs. block

artooro:
I have found a scenario where I would like the default deny rule to use `block return` instead of `block` so that the TCP connection does not continually re-try.
It's currently hard-coded into the system, this pull request makes it an option under System / Advanced / Firewall.

Thanks!


jim-p:
I've thought this over a bit and I can't see any valid use case where this makes sense. You do not want the firewall to return traffic in this way by default. Craft your own rules that reject rather than using the default deny.

PR #3550, opened 2017-02-16
System Info Widget enable All button when disable firmware check is set

phil-davis:
If system firmware disablecheck is set, then the click event for the filter "All" button is also not included in the JS, so the "All" button is not effective.

The setTimeout for systemStatusGetUpdateStatus() can be moved up just below where it is declared and inside the existing "if" that conditionally includes that function - no need for it to be inside events.push

Then the PHP condition around events.push can be removed, so that the "All" button will always be effective.

### DIFF
--- a/src/usr/local/www/widgets/widgets/system_information.widget.php
+++ b/src/usr/local/www/widgets/widgets/system_information.widget.php
@@ -439,6 +439,8 @@ function systemStatusGetUpdateStatus() {
 		}
 	});
 }
+
+setTimeout('systemStatusGetUpdateStatus()', 4000);
 <?php endif; ?>
 
 function updateMeters() {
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: This bug effects RELENG_2_3 also, so I should be back-ported in case there is a 2.3.3-p1 or whatever.

phil-davis:
@jim-p @rbgarga @sbeaver-netgate Anybody there?
I am doing other "feature" things around this code. It would be nice to get this bug-fix merged, that will save me messing about noticing it and getting conflicts in other branches.

sbeaver-netgate:
Sure. Give me a couple of minutes

Steve

phil-davis:
Thanks

sbeaver-netgate:
I'll do so as soon as some repo maintenance work now underway is completed

PR #3553, opened 2017-02-18
Fix NTP widget to show server time (Issue #7245)

doktornotor:
This is a couple of seconds late depending on how much stuff people have on dashboard, but at least displays the server time/TZ instead of the browser time. Plus good riddance wrt those ~250 lines of past century JS.

### DIFF
--- a/src/usr/local/www/widgets/widgets/ntp_status.widget.php
+++ b/src/usr/local/www/widgets/widgets/ntp_status.widget.php
@@ -24,20 +24,7 @@ $nocsrf = true;
 require_once("guiconfig.inc");
 require_once("pfsense-utils.inc");
 require_once("functions.inc");
-
 require_once("/usr/local/www/widgets/include/ntp_status.inc");
-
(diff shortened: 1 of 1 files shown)


dennypage:
Thank you Dok. Was thinking of looking at this over the weekend, but no longer necessary. Have to find something else from my todo list I guess. :)

doktornotor:
I'm kinda disappointed I couldn't fix the one remaining cookie problem with IE5 on W98, though. 🤣

jim-p:
Looks good, definitely not pulling the client time anymore, tested with a VM and my laptop which are a couple seconds apart

ggzengel:
With 2.3.4 I have:
> date
> Fri May  5 17:45:24 UTC 2017

In widget I have:
>Server Time 	19:45:24 UTC
> Sync Source 	85.25.210.112 (stratum 3) 

My local machine has UTC +2

So it should:
Server Time 	19:45:24 GMT +2
-or -
Server Time 	17:45:24 UTC

PR #3554, opened 2017-02-18
Added help descriptions for DNSMadeEasy fields in Dynamic DNS

jeremyrnelson:


pfSense-git-updates:

src/usr/local/www/services_dyndns_edit.php
```diff
@@ -293,6 +293,7 @@ function build_if_list() {
 ));
 
 $group->setHelp('Enter the complete fully qualified domain name. Example: myhost.dyndns.org%1$s' .
+			'DNS Made Easy: Dynamic DNS ID (NOT hostname)' .
```
rbgarga: You should add %1$s to the end of the line, it will be replaced by <br />

src/usr/local/www/services_dyndns_edit.php
```diff
@@ -351,6 +352,7 @@ function build_if_list() {
 	'text',
 	$pconfig['username']
 ))->setHelp('Username is required for all types except Namecheap, FreeDNS and Custom Entries.%1$s' .
+	    		'DNS Made Easy: Dynamic DNS ID' .
```
rbgarga: You should add %1$s to the end of the line, it will be replaced by <br />

src/usr/local/www/services_dyndns_edit.php
```diff
@@ -361,6 +363,7 @@ function build_if_list() {
 	'password',
 	$pconfig['password']
 ))->setHelp('FreeDNS (freedns.afraid.org): Enter the "Authentication Token" provided by FreeDNS.%1$s' .
+	    		'DNS Made Easy: Dynamic DNS Password' .
```
rbgarga: You should add %1$s to the end of the line, it will be replaced by <br />

countingpine:
Hi.
I know this has been merged now, but it's worth saying that the username field is not needed at all.  A URL of this form works: `https://cp.dnsmadeeasy.com/servlet/updateip?id=1234567&password=...&ip=1.2.3.4`

The documentation examples at https://dnsmadeeasy.com/technology/dynamic-dns/ claim `username` should be set to the DNSMadeEasy account name/email, so it's probably unwise to remove support for the field, but it would be nice to downgrade it to an optional one, and not to specify that it should be set to the ID.

So I think perhaps it should recommend that it be set to the account name/email, but also allow it to be left blank.  I've tended to default to a `.` or similar.

jim-p:
This was merged two years ago, so it's a bit late for comments. If you would like to propose a change to the current code, open a new feature request ticket at https://redmine.pfsense.org/

PR #3556, opened 2017-02-19
Fix net.inet.ip.random_id tunable description (Bug #6087)

doktornotor:
https://redmine.pfsense.org/issues/6087


doktornotor:
Is there some problem with this, or???

PR #3558, opened 2017-02-19
Optionally trust CA certificates in GUI for SSL client actions (e.g. updates)

overhacked:
pfSense can be configured to use a proxy server, but if it is a man-in-the-middle SSL proxy, there has been no GUI way to add trusted root certificates for HTTPS client actions like updates, URL alias downloads, etc.

This patch adds a "Trusted" checkbox (default False) to CAs in the GUI that will add the certificate to /etc/ssl/certs. An environment variable is set at boot-time that causes libfetch (eventually libssl) to look for root certificates in that directory.


src/etc/inc/certs.inc
```diff
@@ -281,6 +281,42 @@ function ca_inter_create(& $ca, $keylen, $lifetime, $dn, $caref, $digest_alg = "
...
 }
 
+function ca_trustedcerts_generate() {
+	global $config, $g;
+	if (isset($config['system']['developerspew'])) {
+		$mt = microtime();
+		echo "ca_trustedcerts_generate() being called $mt\n";
+	}
+
+	mwexec("/bin/mkdir -p {$g['etc_path']}/ssl/certs");
```
jim-p: Do not use a shell command for this, we have functions to make directories.

src/etc/inc/certs.inc
```diff
@@ -281,6 +281,42 @@ function ca_inter_create(& $ca, $keylen, $lifetime, $dn, $caref, $digest_alg = "
...
+	/* Remove all certs in the directory (but leave symlinks in place to allow power users to add trusted certs outside the Certificate Manager) */
+	mwexec("/usr/bin/find {$g['etc_path']}/ssl/certs -type f -exec /bin/rm '{}' ';'");
```
jim-p: Do not use a shell command for this, use a function similar to this: https://github.com/pfsense/FreeBSD-ports/blob/devel/net-mgmt/pfSense-pkg-net-snmp/files/usr/local/pkg/net-snmp.inc#L223
If you want to exclude something, then write your files with a unique prefix and then glob using that prefix,

src/etc/inc/certs.inc
```diff
@@ -513,6 +549,12 @@ function cert_get_subject_hash($crt) {
 	return $inf_crt['subject'];
 }
 
+function cert_get_hash($crt) {
+	$str_crt = base64_decode($crt);
```
jim-p: No need to make this variable for a single use, combine the two lines `openssl_x509_parse(base64_decode($crt));`

src/etc/rc.bootup
```diff
@@ -227,6 +229,9 @@ if (!$debugging) {
 /* re-make hosts file after configuring interfaces */
 system_hosts_generate();
 
+/* add configured trusted CA certificates to /etc/ssl/certs */
+ca_trustedcerts_generate();
+
```
NOYB: Why do trusted certs need to be re/generated during boot up?   /etc/ssl/certs is a static dir is it not?  The certs should still be there across reboots.  Only reason I think of is restoring/swapping a config.  But the trusted certs re/generation for that scenario should and could be done by other means.  Such as part of the config restore process.  Or a trusted certs regeneration button in ca manager for those times when manually swapping a config.

src/usr/local/www/system_camanager.php
```diff
@@ -90,6 +90,7 @@
 	$name = $a_ca[$id]['descr'];
 	unset($a_ca[$id]);
 	write_config();
+	ca_trustedcerts_generate();
```
NOYB: Would rather trusted certs be generated/removed on a per cert add/edit basis instead of regenerating all of them every time a change is made.  Even changes that do not make changes to trusted certs settings.

NOYB:
Suggest squash some of these commits to create a cleaner history?  Especially those that are just changes to the changes.

src/usr/local/www/system_camanager.php
```diff
@@ -308,6 +317,7 @@
 
 		if (!$input_errors) {
 			write_config();
+			ca_trustedcerts_generate();
```
NOYB: Would rather trusted certs be generated/removed on a per cert add/edit basis instead of regenerating all of them every time a change is made. Even changes that do not make changes to trusted certs settings.

src/etc/rc.bootup
jim-p: It's safer to do it at boot time to ensure it's consistent. It's not like it's a huge amount of work or writes.

src/usr/local/www/system_camanager.php
```diff
@@ -107,6 +108,7 @@
 	if (!empty($a_ca[$id]['prv'])) {
 		$pconfig['key'] = base64_decode($a_ca[$id]['prv']);
 	}
+	$pconfig['trusted'] = $a_ca[$id]['trusted'];
```
NOYB: Think this is what you really want to do here: `$pconfig['trusted']  = isset($a_ca[$id]['trusted']);`

src/usr/local/www/system_camanager.php
```diff
@@ -115,6 +117,7 @@
 	$pconfig['digest_alg'] = "sha256";
 	$pconfig['lifetime'] = "3650";
 	$pconfig['dn_commonname'] = "internal-ca";
+	$pconfig['trusted'] = false;
```
NOYB: Shouldn't need this as it will not be set by default.

src/usr/local/www/system_camanager.php
```diff
@@ -250,6 +253,12 @@
 
 		$ca['descr'] = $pconfig['descr'];
 
+		if (!isset($pconfig['trusted']) || empty($pconfig['trusted'])) {
+			$ca['trusted'] = false;
+		} else {
+			$ca['trusted'] = $pconfig['trusted'];
+		}
+
```
NOYB: This should be all that it needed here: `$ca['trusted'] = isset($pconfig['trusted']);`

src/usr/local/www/system_camanager.php
```diff
@@ -402,6 +413,7 @@
 				<tr>
 					<td><?=$name?></td>
 					<td><i class="fa fa-<?= (!empty($ca['prv'])) ? "check" : "times" ; ?>"></i></td>
+					<td><i class="fa fa-<?= (isset($ca['trusted']) && $ca['trusted']) ? "check" : "times" ; ?>"></i></td>
```
NOYB: Don't need the `&& $ca['trusted']` here.

src/etc/rc.bootup
NOYB: And completely unnecessary.  It's not like people are restoring or swapping configs in prod every other time they reboot.  Also not friendly to write cycle sensitive media.

NOYB:
Hey Overhacked.  Thanks for doing this.  Really appreciate it.  This has been a thorn in my side for a long time.  I've given quite a few suggestions here.  Most of them stem for what I've been doing for my own config to trust self signed certs.  Please take them as constructive input.  Again thanks.

src/etc/rc.bootup
jim-p: Unless you can check to be sure the state is consistent and update as needed, it is necessary. You don't know what someone might have done, or what might have happened on upgrade, etc. It's not like it's happening every few minutes, the number of writes overall is minimal. Better safe than sorry.

NOYB:
Let me toss another thing your direction to consider.  Removing revoked CA certs.

src/etc/rc.bootup
NOYB: > Unless you can check to be sure the state is consistent and update as needed, it is necessary. 
Which is how it should be done.  Verify what is there against what config says should be there and add/remove accordingly.  Also include a "synchronize" button on the CA page for user to be able to manually initiated the same thing.

NOYB: > Unless you can check to be sure the state is consistent and update as needed, it is necessary. 
Which is how it should be done.  Verify what is there against what config says should be there and add/remove accordingly.  Also include a "synchronize" button on the CA page for user to be able to manually initiated the same thing.

src/etc/inc/certs.inc
```diff
@@ -281,6 +281,42 @@ function ca_inter_create(& $ca, $keylen, $lifetime, $dn, $caref, $digest_alg = "
...
+		echo "ca_trustedcerts_generate() being called $mt\n";
+	}
+
+	mwexec("/bin/mkdir -p {$g['etc_path']}/ssl/certs");
+	/* Remove all certs in the directory (but leave symlinks in place to allow power users to add trusted certs outside the Certificate Manager) */
+	mwexec("/usr/bin/find {$g['etc_path']}/ssl/certs -type f -exec /bin/rm '{}' ';'");
+
+	if (is_array($config['ca'])) {
+		foreach ($config['ca'] as & $ca) {
+			if ($ca['trusted']) {
```
NOYB: And along with the other issets this would be an isset test also. `if (isset($ca['trusted']))`

src/usr/local/www/system_camanager.php
overhacked: Not sure about this one and subsequent comment that suggests deleting `&& $ca['trusted']`. isset() will return TRUE even if the variable is set FALSE, right? Is it *completely* safe to assume that a boolean FALSE never appears in `global $config`?

src/usr/local/www/system_camanager.php
overhacked: I'm not sure it is a good idea to trust that files named `*hashvalue*.0` have the correct certificate as their contents. I could certainly read in the existing file and check if it contains the correct certificate, but that seems like a lot of overhead compared to rewriting the certfiles during change/reboot. Thoughts on security implications of blindly trusting the file contents?

jim-p: There are some more comments on this [up higher](https://github.com/pfsense/pfsense/pull/3558#discussion_r101931396). I'd rather see them rewritten each boot to be absolutely sure. I don't see saving 2-3 writes per reboot as being worth all the extra complexity. You could read the files in and compare, or compute and compare hashes without reading directly, but it's a lot of overhead for minimal benefit.

NOYB: Deleting the entire set of cert files and rewriting them provides no additional safety.

1) If the contained cert does not match the hash name then it will fail for the site.  The certs here are found for the site by their hash name.  They are not all read until a match is found.  That's why they are required to be hash named.

2) Symlinks are not being destroyed.  So anyone who wishes to tamper can do so simply by creating a link to a cert located anywhere on the system they want.

3) Anyone wishing to tamper can simply put a cert of their choosing in the CA bundle (/usr/local/share/certs/ca-root-nss.crt).  It is not purged and reloaded at boot either.

This argument of safety and trust is nonsensical.

No need to read the files in and compare.  Just create an array of file names config says should be there and add/remove accordingly.  Not much work at all really, and far less work than writing files that don't need to be written.  Deleting and rewriting the files is much more work for no benefit.

```
// Get list of trust files for CA certs from config.
foreach($config['ca'] as $ca) {
	$ca_hash = cert_get_hash($ca['crt']);
	$ca_certfiles[] = "{$g['etc_path']}/ssl/certs/{$ca_hash}.0";
}

// Remove cert trust files that have no corresponding CA cert in config.
foreach (glob("{$g['etc_path']}/ssl/certs/*") as $cert_file) {
	if (!in_array($cert_file, $ca_certfiles)) {
		unlink_if_exists($cert_file);
	}
}

// Add missing cert trust files for CA certs in config.  Remove cert trust files of CA certs in config that are not trusted or are revoked.
foreach ($a_ca as $id => $value) {
	ca_trustedcert_add($id);
}
```

src/usr/local/www/system_camanager.php
NOYB: Option set/not set rather that true/false is the way it is done throughout pfSense.  Yes using isset instead of true/false works, is cleaner/clearer and is the consistent boolean option technique in pfSense config.

src/usr/local/www/system_camanager.php
jim-p: Wanting to be sure the firewall has correct contents in the files is hardly "nonsensical". When I say "safe" I mean "Ensure the contents of the files are correct and that all are present", not necessarily "safe" from a security standpoint (which is also important).

If a file is named properly but contains the wrong contents (0-byte file, truncated, overwritten, etc) it would be completely missed unless the contents of the files are checked/verified. There would be no way to correct that unless the file is overwritten or removed and rewritten. The use of the CA would fail but there's no way to correct that, and it wouldn't be obvious to the user that they might have to trigger a forced rewrite by edit/save of the CA if it was fixed to force an update then.

You're still jumping through a lot of hoops to avoid something that isn't even a sizable problem. A few writes per reboot isn't going to kill a disk. We need to be sure the files are in a consistent state with correct contents, making no assumptions and taking no shortcuts.

dennypage: Advantages both ways. Rewriting the file at every boot can be a reasonable prophylactic covering any number of future coding errors. However, in a security audit you generally have to explain the "why" behind each modified file, which is a pain. If the checksum of the generated file is the same each and every time, I would say writing the file at boot is okay. However, if the checksum can change randomly, then I would think the effort to ensure the correctness of the write on demand the code is worth it.

src/etc/inc/certs.inc
overhacked: I agree with this and the previous comment that I should be using pfSense library functions over shell commands. The `mkdir` code is easy to change; I had aped other code in `certs.inc` that did the same thing, so my next pull request may be a few patches to clean up that file.

I will refactor to eliminate the call to `/usr/bin/find` and use `is_link` and `unlink` instead. There's no way to use prefixes in `/etc/ssl/certs`, unfortunately, because OpenSSL requires that the files be named `{cert_hash}.0`, and OpenSSL doesn't support certificates in subdirectories of cert_path, so I can't do that either. Symlinks seemed a reasonable compromise to not completely obliterate the ability of power users and developers to manually place certificates in the directory.

src/etc/inc/certs.inc
overhacked: Will do.

src/usr/local/www/system_camanager.php
overhacked: The file checksum will be constant unless the certificate is updated. Of course, the ctime and mtime will change everytime it is rewritten, so that could trigger an audit. I agree with @NOYB that building an array and policing files that way is probably better than always deleting and recreating. Probably validating contents by checking file size is a reasonable, low-overhead way of checking that the file hasn't been accidentally truncated or overwritten. Obviously that's not security safe, but it does provide some reasonable sanity checking.

NOYB: If that is the reasoning then many other files, including the CA bundle (/usr/local/share/certs/ca-root-nss.crt) should be rewritten from a know good source to insure their contents are correct.  That argument is still nonsensical.  And rewriting them provides no benefit.  The code for checking instead is simple.

If at every turn the thought was it's just a few writes that hardly matter.  Well they add up and before you know it the system takes many minutes to boot due to a bunch of unnecessary disk activity.  All for no benefit.

NOYB: Oh, and by the way it is no more assured that config has the correct content that the file.  Could just as easily be over writing good file with bad.

overhacked:
Re @NOYB commenting above about removing revoked CA certs:

Isn't that something that OpenSSL takes care of when it is properly configured with a CRL? If a user configures one in the GUI, won't or couldn't it already be loaded into the validation via:
```
/* Load CRLs into the `X509_STORE` */

X509_STORE *x509_store = SSL_CTX_get_cert_store(ctx);
X509_STORE_add_crl(x509_store, crl);

/* Enable CRL checking */
X509_VERIFY_PARAM *param = X509_VERIFY_PARAM_new();
X509_VERIFY_PARAM_set_flags(param, X509_V_FLAG_CRL_CHECK);
SSL_CTX_set1_param(ctx, param);
X509_VERIFY_PARAM_free(param);
```
(via https://stackoverflow.com/questions/4389954/does-openssl-automatically-handle-crls-certificate-revocation-lists-now)

src/usr/local/www/system_camanager.php
jim-p: @dennypage SOP on pfSense is to rewrite at boot time to ensure consistency over all. Having to explain each write isn't something that matters here since it's done in numerous other places at boot time to ensure consistency.

@overhacked If you are going to verify, do a full checksum (sha256 or better). Don't just check size. No reason for half measures if you're going through the trouble.

@NOYB Please stop with the "nonsensical" line. It is completely reasonable to want to fully ensure a consistent state at all times. No matter what you believe, it is not "nonsense" to want to be absolutely certain you have good data. If they were rewritten at each filter sync your argument might hold water but it's at boot time or when the settings are saved. It's in line with how the rest of the system operates currently. That may change on the 3.x line but for now, it's what we want. Period.

NOYB:
Haven't study out how pfSense clients will handle that.  But if the cert is not there then it won't be trusted.  If you, or someone, can verify that all pfSense clients, cURL, SMTP, etc. will utilize the CRL then yeah it would matter then.

src/usr/local/www/system_camanager.php
overhacked: So, at boot or resync:

1. Build array of expected cert hashes
1. Unlink existing regular files with unexpected hashes
1. Check file size of expected files against config contents to avoid unnecessary reading of the certs during boot, and rewrite any files with bad sizes
1. Leave symlinks alone unless there is a name collision

When editing a CA:
* Just rewrite the file

When deleting a CA:
* Just unlink the file

NOYB: @jim-p In this case the rewriting provides no benefit.  And other files just as critical are not handled that way to ensure content either.  e.g. the CA bundle.  And recreating form config can just as easily create incorrect content.  There is no more protection of the config contents to be correct than the existing file.

jim-p: Seems OK except for the file size check. If you're going that route you'll have to do a proper checksum comparison, not just size.

jim-p: @NOYB Except that it does offer a benefit. The benefit is you know 100% that the file contents match what is in config.xml. The ultimate reference is config.xml. The contents of config.xml must always be assumed to be correct, and is one of the basic principles of pfSense. If that is wrong then you need to fix what broke config.xml.

A user must be able to take config.xml and restore it to another system and end up with the same consistent system. Though packages offer some exceptions in that area, the base system has to adhere to that principle.

jim-p: @NOYB See above. config.xml is always assumed to be correct and it is always safe to use as a reference.

NOYB: And is more likely to be incorrect than the individual file due to all the activity there.  If we cannot trust these individual cert file contents then why are we trusting the CA bundle of certs content?  Trusting on and not the other is pointless.

jim-p: @NOYB Again, it's about consistency. config.xml is trusted above everything else and it is the LEAST likely thing to be incorrect. The user put what they wanted in config.xml, we need to respect that. The system protects itself against invalid XML, the odds that config.xml is corrupt in a way that is still valid XML are very low, and corruption there is only likely to happen in the case of hardware failure and CAs are the least of your worries then.

I know you don't want to admit you're incorrect here, but you can't use "config.xml might be wrong" as the basis of any valid argument. It's not going to happen, and if it does, you have much worse problems than CAs.

jim-p: ZFS will improve the situation but speaking from experience working here, in the last few years the likelihood of a file in /etc having the wrong contents after a filesystem error is much, much higher than anything else being wrong. In all that time I don't recall ever seen a corrupted config.xml on a system that was still operational. We've seen dozens if not hundreds of systems that needed reinstalled because of filesystem issues that affected files in /etc.

jim-p: We know what we're talking about here. Don't take half measures. Overwrite or verify checksums. No shortcuts.

NOYB: So pfSense needs a complete overhaul to put content of all critical files into the config.xml and rewrite them at boot up to ensure they have the correct (assumed to be correct form config.xml) content.  Including the CA bundle.  That is what the argument you are making amounts to.

The contents of these cert files is no more and no less critical that that of the CA bundle.  If this is necessary and justified for these files to be rewritten from config at every boot up then the same is true for the CA bundle.  It can be corrupted just as easily as these individual cert files.

The reason I can't admit that I'm wrong is because I'm not.  Regardless of how you manage to convince it to be done.  My systems will not be rewriting all the CA's at boot.

jim-p: Not quite what I'm accurately stating (not arguing). The CA bundle is handled via pkg ( ca_root_nss ) so it does not need to be in config.xml. The user has their CAs in config.xml and we need to be sure the CAs written in /etc/ssl/ are the CAs from config.xml. They don't come from other sources.

We need to be absolutely certain that the files match config.xml. Either by overwriting or completely and accurately verifying their contents.

If you want to handle your own system such that you could have an inconsistent state or corrupt files in /etc/ssl/ go ahead, but for code we include, it has to be ensured to be consistent at boot time.

NOYB: The fact that the CA bundle file comes from another source/package makes no difference.  It is just a file and can still just as easily be corrupted by the same problems at would affect these files.  So it should be just as important to ensure the contents of the CA bundle file.  They both serve the same purposes.

overhacked: I patterned my initial code after other features in pfSense that rewrite at boot-time, specifically `php.ini`. I'm going to stick with that approach until there is a future, more comprehensive approach to file integrity in pfSense. I see no need to complicate this code, here and now, by introducing checksumming when **assumed-good** (not *known-good* h/t @NOYB) data in config.xml is available.

jim-p: @NOYB The fact that the bundle is maintained by pkg is significant because pkg could verify that if we wanted, it doesn't need to be handled via the same mechanism. When restoring config.xml from another system, the bundle will already be there from pkg so we do not need to include it in config.xml. Let pkg worry about the bundle, let config.xml and pfSense worry about the contents of the other files. They each get verified by their source.

We'll have to agree to disagree if you can't see the logic behind that.

NOYB: Have never said the certs shouldn't be updated to match a config resotore/swap/move.  Rewriting them all at ever boot to do it is the issue.  Totally unnecessary.  And to require checksuming them and not the CA bundle is illogical.

So the CA bundle could be verified if you wanted it to be.  But for some reason you don't and are willing to trust it not to get corrupted, but not these files.  That's what is not logical.

NOYB: Is php.ini being written from config.xml.  Maybe it's changed by last I looked think it was being written from code with heredoc string, not from the config.

jim-p: Let me reduce it to basics you might get: To blindly assume the files are valid is more dangerous than to be safe and check them and verify or overwrite. The risks of assuming they are good outweigh the little benefit you get by skipping. 

Treating the bundle separately makes sense. It's from pkg, it isn't being changed based on user input like these CAs are. You're trying to lump things together that are clearly separate based on your own arbitrary criteria.

To put it another way, the bundle is to be considered static as a "Base system" file. The CA files are considered volatile because they are generated from user input.

I'm saying verify the CAs because we _can_ and it makes sense to do so here in this code.  It's safer. Touching the bundle does not since this code has nothing to do with the bundle.

The rewards of taking shortcuts are not great enough to counteract the risk.

overhacked:
Just pushed fixes to changes directly requested above, except:

* No handling of CRL's in this code. Should I trust OpenSSL (aka other clients) to check CRLs, or should I check CA certificate validity here?

src/usr/local/www/system_camanager.php
NOYB: First of all that is not the basis of the argument you have been making.  Which is that the file can become corrupted.  Which goes for both equally so if one can not be trusted then neither can the other.

Secondly they are both basically static files.  These certs only get changed if the user changes them.  Just like the CA bundle only gets changed if the package changes it.  In which case they get rewritten.  Otherwise they are static.  Just because it's done through the GUI does not change that or make it any different than a power user adding it manually.

It's not that I don't get it.  It's that the reasoning is inconsistently applied and not logical.

It doesn't matter that this code doesn't touch the CA bundle.  Just like the CA bundle, once it's there it's there and this code doesn't need to touch them again until the user makes a change.  Just like the CA bundle doesn't need to be touched again until the package changes it.

jim-p: See above, re; Agree to disagree then. My logic is sound and I've explained it thoroughly in several ways, consistently.

You can prefer your way, but that is not the way we want pfSense to behave.

NOYB:
So now that you've made those changes.  Thank you.  Here is another thing to look into.

Have only played with this on 2.3 (not 2.4) but are you sure that PHP's curl_exec will look in /etc/ssl/certs for certs?

Maybe the PHP config is different in 2.4 but in 2.3 it does not.  Have to add the CURLOPT_CAPATH option.  Among things that could be affected are URL table downloads and check ip for custom DynDNS and RFC 2136 clients.

Again thanks for doing this.

src/usr/local/www/system_camanager.php
NOYB: Just because it's the way you what pfSense to behave does not make it good or even beneficial.  Microsoft does things because that's they way they want it to be too.  Doesn't necessarily make it better.  Doing that way just because that's the way we want it to be is not sound design.  Any good engineer will tell you that the more you "touch" something increases the probability of introducing an error/failure/etc.  The file is there, it's good.  Leave it alone and let it be.  Just the like CA bundle file.

If customers are willing to accept your inconsistently applied bloatware method of file integrity assurance.  That is between them and you.  But yes I prefer it be left alone once there.  And that's how it will be done on my systems.

If it really is that crittically important to you/pfSense then the rewriting on boot up is still not the "right"/"best" way of ensuring the content is correct, and has much more overhead then doing it in well engineered fashion that could also be applied to any file deemed to be critical.

jim-p: You disagree, I get it. Let it go. No need for insults or name-calling. Your way is not better, it's riskier. You don't know the file is good until you check it to know it's good. You want to allow for the possibility that it's not right -- that's on you, but that isn't how we operate. Overwrite or read/verify checksum, either way is fine since you end up with a file in a known-good state.

Leaving unverified data in place because the filename happens to be correct isn't "well engineered".

NOYB: I agree.  No need for insults or name-calling.

Where did I insult you?  Where did I call you a name?

You certainly hurled some insults at me though.

My way is no more riskier than how the CA bundle is currently being handled.  And like I stated.  Rewriting files at boot up is still not the "right" way to ensure file integrity.  There are better, more reliable methods with far less overhead.

Leaving unverified data in place because the filename happens to be correct isn't "well engineered".

But you and pfSense are okay with that as long is it's other just as critical files.  Yes.  That logic is well reasoned out.  Not.

I have never said they shouldn't be verified.  But if they are going to be then do it "right".  With a well engineer method and consistently apply it to critical files.

jim-p: "inconsistently applied bloatware", for example. Uncalled for. You keep arguing in circles ignoring my explanations, so I'm done.

NOYB: Talking down to someone with statements like; let me boil it down to basic, so you can get it and such.  Yes those kinds of statements that you made are insults.  I wasn't born yesterday.  I know what those statements are about.  They are intended to belittle.

I'm not the one arguing in circles and ignoring your explanations.  I'm debunking them and the reasoning that's being given for them.

overhacked:
Sadly, you're right about `curl_exec()`, @NOYB. `CURLOPT_CAPATH` has to be set for it to look at the certs directory at all, and worse, there is [no php.ini variable](https://secure.php.net/manual/en/curl.configuration.php) to configure it. Going to sleep on this problem and see if I can come up with a solution. I had so hoped that all libssl-using parts of PHP would honor the environment variables.

NOYB:
Fortunately I think there are only a few places the CURLOP_CAPATH is needed.  I'll provide a list of the places I found below.

Look into openssl.capath/cainfo it may be usable for this but I don't know.

Another thing I was thinking is maybe create a PHP var with the same path as being set in the environment variable.  Then CURLOPT_CAPATH where needed could be set from the var instead of hard coded a string everywhere like I did for testing.  It would probably have to be a global var.

Maybe someone knows an even cleaner way.  If so that would be great.

There are some other places than these below where verify host/peer is used but they are for services that should be in the CA bundle already.

```
diff --git a/src/etc/inc/dyndns.class b/src/etc/inc/dyndns.class
index c88fecacdd..1203ac6697 100644
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -674,6 +674,7 @@
 						}
 						if ($this->_curlSslVerifypeer) {
 							curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, TRUE);
+							curl_setopt($ch, CURLOPT_CAPATH, '/etc/ssl/certs');
 						} else {
 							curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, FALSE);
 						}



diff --git a/src/etc/inc/pfsense-utils.inc b/src/etc/inc/pfsense-utils.inc
index 075306f01d..d5da2cf9b2 100644
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -1823,6 +1823,7 @@ function download_file($url, $destination, $verify_ssl = true, $connect_timeout
 	curl_setopt($ch, CURLOPT_URL, $url);
 	curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, $verify_ssl);
 	curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, $verify_ssl);
+	curl_setopt($ch, CURLOPT_CAPATH, '/etc/ssl/certs');
 	curl_setopt($ch, CURLOPT_FILE, $fp);
 	curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, $connect_timeout);
 	curl_setopt($ch, CURLOPT_TIMEOUT, $timeout);
@@ -1878,6 +1879,7 @@ function download_file_with_progress_bar($url, $destination, $verify_ssl = true,
 	curl_setopt($ch, CURLOPT_URL, $url);
 	curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, $verify_ssl);
 	curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, $verify_ssl);
+	curl_setopt($ch, CURLOPT_CAPATH, '/etc/ssl/certs');
 	curl_setopt($ch, CURLOPT_HEADERFUNCTION, 'read_header');
 	curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
 	curl_setopt($ch, CURLOPT_WRITEFUNCTION, $readbody);



diff --git a/src/etc/inc/services.inc b/src/etc/inc/services.inc
index 021fb7e562..6a53378681 100644
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -2035,6 +2035,7 @@ function dyndnsCheckIP($int) {
 		$ip_ch = curl_init($hosttocheck);
 		curl_setopt($ip_ch, CURLOPT_RETURNTRANSFER, 1);
 		curl_setopt($ip_ch, CURLOPT_SSL_VERIFYPEER, $verifysslpeer);
+		curl_setopt($ip_ch, CURLOPT_CAPATH, '/etc/ssl/certs');
 		curl_setopt($ip_ch, CURLOPT_INTERFACE, 'host!' . $ip_address);
 		curl_setopt($ip_ch, CURLOPT_CONNECTTIMEOUT, '30');
 		curl_setopt($ip_ch, CURLOPT_TIMEOUT, 120);

```

Liking what you are doing.  Keep it up.

src/etc/inc/certs.inc
```diff
@@ -281,6 +281,42 @@ function ca_inter_create(& $ca, $keylen, $lifetime, $dn, $caref, $digest_alg = "
...
+function ca_trustedcerts_generate() {
+	global $config, $g;
+	if (isset($config['system']['developerspew'])) {
+		$mt = microtime();
+		echo "ca_trustedcerts_generate() being called $mt\n";
+	}
+
+	safe_mkdir("{$g['etc_path']}/ssl/certs");
+	/* Remove all certs in the directory (but leave symlinks in place to allow power users to add trusted certs outside the Certificate Manager) */
+	array_map('unlink', array_filter('is_file',glob("{$g['etc_path']}/ssl/certs/*")));
```
NOYB: Believe array_filter syntax is incorrect.  The callback function should be the second parameter.
Even so in my testing is_file was true for symlinks too.  So don't think this will do what is expected.

overhacked: Was tired and mimicked `array_map()` syntax. I'll roll this into my next push. Speaking of, regarding your previous comment about squashing, won't a project maintainer be merging all this activity as a squash commit, anyway?

NOYB: The merge should have them all squashed together.  But I know that on mine when I look at a file history it shows all the individual commits.

overhacked:
Most recent commits pushed address this issue in `curl_*()` calls in the pfSense code base. All future code wanting to use the PHP cURL extension should get their cURL handle from `get_curl_handle(…)` in `pfsense-utils.inc` instead of calling `curl_init()` directly.

I deliberately omitted `simplepie.inc` from these changes, as it is an external library.

overhacked:
FYI, I submitted a PR to the PHP project that would address this issue in `php.ini`: https://github.com/php/php-src/pull/2399

overhacked:
Thanks all for the feedback. I'm going to close this PR and submit another. (I will mention commenters above in the new PR.)

I've discovered after doing further testing that the way webConfigurator spawns `pkg`, `curl`, or calls `curl_*` functions varies too widely for any of the above solutions to address. The hardest obstacle to overcome is that nginx fastcgi sanitizes the environment variables. It is possible to set `fastcgi_param` in `/var/etc/nginx-webConfigurator.conf` (generated by `/etc/inc/system.inc`), but those environment variables don't seem to make it all the way through to the `exec()` call that spawns pkg (or curl).

If you have any suggestions about whether it is worth implementing what amounts to kludges for each different way that pfSense spawns TLS-using processes, please chime in. Otherwise, my plan is to use the GUI code I've written and submit a PR that appends to the existing `/etc/ssl/cert.pem` while retaining the original `/usr/local/share/certs/ca-root-nss.crt` to use as the base file every time new CAs are added or removed.

PR #3561, opened 2017-02-19
NTPd GPS Autoset baud rate

jskyboo:
- Adds option to check the baud rate before sending the initialization commands and tries auto correcting the baud rate, after sending the init commands it reloads the configured baud rate.
- Adds option to try to find the correct baud rate automatically and save the configuration.
- Adds option to try to find the correct baud rate automatically every time NTPd is started.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1615,6 +1615,45 @@ function system_timezone_configure() {
 	}
 }
 
+function check_gps_speed($device, $baud) {
+	usleep(1000);
+	// Set timeout to 5s
+	$timeout=microtime(true)+5;
+	if($fp = fopen($device, 'r')) {
+		stream_set_blocking($fp, 0);
+		stream_set_timeout($fp, 5);
+		$contents = "";
+		$cnt = 0;
+		$buffersize = 256;
+		do {
+			$c = fread($fp, $buffersize - $cnt);
+
+			// Wait for data to arive
+			if($c === false || strlen($c) == 0) {
+				usleep(10000);
+				continue;
+			}
+
+			$contents.=$c;
+			$cnt = $cnt + strlen($c);
+		} while($cnt < $buffersize && microtime(true)<$timeout);
+		fclose($fp);
+
+		$nmeasentences = ['RMC', 'GGA', 'GLL', 'ZDA', 'ZDG', 'PGRMF'];
+		foreach ($nmeasentences as $sentence) {
+			if(strpos($contents, $sentence) > 0) {
+				@file_put_contents("/tmp/testread".$baud.$sentence, $contents);
+				return true;
+			}
+		}
+		if (strpos($contents, "0") > 0 && strpos($contents, "`") === false && strpos($contents, "??") === false) {
+			@file_put_contents("/tmp/testread".$baud, $contents);
+			return true;
+		}
+	}
+	return false;
+}
+
 function system_ntp_setup_gps($serialport) {
 	global $config, $g;
 	$gps_device = '/dev/gps0';
@@ -1629,28 +1668,37 @@ function system_ntp_setup_gps($serialport) {
 	@symlink($serialport, $gps_device);
 
 	$gpsbaud = '4800';
-	if (is_array($config['ntpd']) && is_array($config['ntpd']['gps']) && !empty($config['ntpd']['gps']['speed'])) {
-		switch ($config['ntpd']['gps']['speed']) {
-			case '16':
-				$gpsbaud = '9600';
-				break;
-			case '32':
-				$gpsbaud = '19200';
-				break;
-			case '48':
-				$gpsbaud = '38400';
-				break;
-			case '64':
-				$gpsbaud = '57600';
-				break;
-			case '80':
-				$gpsbaud = '115200';
-				break;
-		}
+	$speeds=[0 => '4800', 16 => '9600', 32 => '19200', 48 => '38400', 64 => '57600', 80 => '115200'];
+	$revspeeds=['4800' => 0, '9600' => 16, '19200' => 32, '38400' => 48, '57600' => 64, '115200' => 80];
+	if($config['ntpd']['gps']['speed'] && array_key_exists($config['ntpd']['gps']['speed'], $speeds)) {
+		$gpsbaud = $speeds[$config['ntpd']['gps']['speed']];
 	}
 
 	/* Configure the serial port for raw IO and set the speed */
 	mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
+
+	$autospeed = ($config['ntpd']['gps']['speed'] == 'autoalways' || $config['ntpd']['gps']['speed'] == 'autoset');
+	if($autospeed || ($config['ntpd']['gps']['autobaudinit'] && !check_gps_speed($gps_device, $gpsbaud))) {
+		$found = false;
+		foreach ($speeds as $baud) {
+			mwexec("stty -f {$serialport}.init raw speed {$baud}");
+			@file_put_contents("/tmp/testreadcheck".$baud, 'dffsd');
+			if($found = check_gps_speed($gps_device, $baud)) {
+				if($autospeed) {
+					$saveconfig = ($config['ntpd']['gps']['speed'] == 'autoset');
+					$config['ntpd']['gps']['speed'] = $revspeeds[$baud];
+					if($saveconfig) {
+						write_config(sprintf(gettext('Autoset GPS baud rate to %s'), $baud));
+					}
+				}
+				break;
+			}
+		}
+		if($found === false) {
+			log_error(gettext("Could not find correct GPS baud rate."));
+			return false;
+		}
+	}
 
 	/* Send the following to the GPS port to initialize the GPS */
 	if (is_array($config['ntpd']) && is_array($config['ntpd']['gps']) && !empty($config['ntpd']['gps']['type'])) {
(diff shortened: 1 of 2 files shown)


src/etc/inc/system.inc
```diff
@@ -1629,46 +1666,61 @@ function system_ntp_setup_gps($serialport) {
...
-				break;
-			case '64':
-				$gpsbaud = '57600';
-				break;
-			case '80':
-				$gpsbaud = '115200';
-				break;
-		}
+	$speeds=[0 => '4800', 16 => '9600', 32 => '19200', 48 => '38400', 64 => '57600', 80 => '115200'];
+	$revspeeds=['4800' => 0, '9600' => 16, '19200' => 32, '38400' => 48, '57600' => 64, '115200' => 80];
```
jim-p: There is no need for this to be defined twice manually in different ways. You can use `array_search()` to find a key name by searching for values, or if you prefer to keep the syntax similar, use `$revspeeds = array_flip($speeds);` so you don't have to manually define the list twice.

src/etc/inc/system.inc
```diff
@@ -1629,46 +1666,61 @@ function system_ntp_setup_gps($serialport) {
...
 
+	$autospeed = ($config['ntpd']['gps']['speed'] == 'autoalways' || $config['ntpd']['gps']['speed'] == 'autoset');
+	if($autospeed || ($config['ntpd']['gps']['autobaudinit'] && !check_gps_speed($gps_device))) {
+		$found = false;
+		foreach ($speeds as $baud) {
+			mwexec("stty -f {$serialport}.init raw speed {$baud}");
+			if($found = check_gps_speed($gps_device)) {
+				if($autospeed) {
+					$saveconfig = ($config['ntpd']['gps']['speed'] == 'autoset');
+					$config['ntpd']['gps']['speed'] = $revspeeds[$baud];
```
jim-p: As an additional feature, have you considered also checking the defined init string for a line setting the baud rate and correcting it as well?

src/etc/inc/system.inc
```diff
@@ -1629,46 +1666,61 @@ function system_ntp_setup_gps($serialport) {
...
-	mwexec("cat /tmp/gps.init > {$serialport}");
+	@file_put_contents($gps_device, $gps_init);
+
+	if ($found && $config['ntpd']['gps']['autobaudinit']) {
+		mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
+	}
+
+	/* Remove old /etc/remote entry if it exists */
+	if (intval(`grep -c "^gps0" /etc/remote`) != 0) {
+		mwexec("sed -n '/gps0/!p' /etc/remote > /etc/remote.new");
```
jim-p: You can do this in one step by using the "inplace" parameter for `sed` (`-i ext`). For example `-i ''` will do an inplace replace without saving a backup copy. Also, use the full path to commands, such as `/usr/bin/sed` and `/usr/bin/grep`

src/usr/local/www/services_ntpd_gps.php
```diff
@@ -368,10 +380,19 @@ function build_nmea_list() {
...
+				'It is recommended to configure the GPS to send only one sentence at a baud rate of 4800 or 9600.%1$s' .
+				'Autoset tries to find the correct baud rate of the GPS device and then saves the configuration.%1$s' .
+				'Always Auto tries to find the correct baud rate of the GPS device every time NTPd is started.', '<br /><br />');
+	
+	$section->addInput(new Form_Checkbox(
+		'autobaudinit',
+		null,
+		'Check baud rate before sending init commands (default: unchecked).',
+		$pconfig['autobaudinit']
+		))->setHelp('Before sending the initialization commands, check the baud rate is correct. If it is not correct try to find the correct baud rate automatically, send the initialization commands if the correct rate is found, and then set the baud rate to the configured speed.%1$s' . 'This is useful if the GPS device resets back to a default rate on power loss or when changing the baud rate.', '<br /><br />');
```
jim-p: Change "check the baud rate is correct" to "check the GPS baud rate".

src/etc/inc/system.inc
jskyboo: Writing to a separate file and doing a file move should reduce the possibility for corruption.

src/etc/inc/system.inc
jskyboo: I hadn't thought of that yet. Write the feature request and I'll add it in another PR.

src/etc/inc/system.inc
jim-p: We use `sed -i` in many other places without issue. The only way it would end up corrupted is if the disk ran out of space. If you insist on using two steps, then use `rename()` instead of a shell exec for mv, but I'd much prefer to see it use `sed -i` without the extra unnecessary step.

jskyboo:
Are you waiting on the correcting the init string? Because I think that is a large change that should be merged separately. I think I addressed your other concerns.

jim-p:
No, we've been busy with 2.3.3 release and other tasks, will revisit the open PRs soon.

src/etc/inc/system.inc
```diff
@@ -1721,8 +1788,8 @@ function system_ntp_configure() {
 
 	/* Add PPS configuration */
 	if (is_array($config['ntpd']['pps']) && !empty($config['ntpd']['pps']['port']) &&
-	    file_exists('/dev/'.$config['ntpd']['pps']['port']) &&
-	    system_ntp_setup_pps($config['ntpd']['pps']['port'])) {
+		file_exists('/dev/'.$config['ntpd']['pps']['port']) &&
+		system_ntp_setup_pps($config['ntpd']['pps']['port'])) {
```
rbgarga: You broke style here. Multiline statements should use 4 spaces indent as it was before. Please revert

src/etc/inc/system.inc
```diff
@@ -1760,8 +1827,7 @@ function system_ntp_configure() {
 
 	/* Add GPS configuration */
 	if (is_array($config['ntpd']['gps']) && !empty($config['ntpd']['gps']['port']) &&
-	    file_exists('/dev/'.$config['ntpd']['gps']['port']) &&
-	    system_ntp_setup_gps($config['ntpd']['gps']['port'])) {
+		system_ntp_setup_gps($config['ntpd']['gps']['port'])) {
```
rbgarga: Please revert style breakage

src/etc/inc/system.inc
```diff
@@ -1825,8 +1891,7 @@ function system_ntp_configure() {
 		}
 		$ntpcfg .= "\n";
 	} elseif (is_array($config['ntpd']) && !empty($config['ntpd']['gpsport']) &&
-	    file_exists('/dev/'.$config['ntpd']['gpsport']) &&
-	    system_ntp_setup_gps($config['ntpd']['gpsport'])) {
+		system_ntp_setup_gps($config['ntpd']['gpsport'])) {
```
rbgarga: Please revert style breakage

src/etc/inc/system.inc
```diff
@@ -1842,7 +1907,7 @@ function system_ntp_configure() {
 	/* foreach through ntp servers and write out to ntpd.conf */
 	foreach (explode(' ', $config['system']['timeservers']) as $ts) {
 		if ((substr_compare($ts, $auto_pool_suffix, strlen($ts) - strlen($auto_pool_suffix), strlen($auto_pool_suffix)) === 0)
-		    || substr_count($config['ntpd']['ispool'], $ts)) {
+			|| substr_count($config['ntpd']['ispool'], $ts)) {
```
rbgarga: Please revert style breakage

src/usr/local/www/services_ntpd_gps.php
```diff
@@ -368,10 +380,19 @@ function build_nmea_list() {
...
+				'It is recommended to configure the GPS to send only one sentence at a baud rate of 4800 or 9600.%1$s' .
+				'Autoset tries to find the correct baud rate of the GPS device and then saves the configuration.%1$s' .
+				'Always Auto tries to find the correct baud rate of the GPS device every time NTPd is started.', '<br /><br />');
+	
+	$section->addInput(new Form_Checkbox(
+		'autobaudinit',
+		null,
+		'Check baud rate before sending init commands (default: unchecked).',
+		$pconfig['autobaudinit']
+		))->setHelp('Before sending the initialization commands, check the GPS baud rate. If it is not correct try to find the correct baud rate automatically, send the initialization commands if the correct rate is found, and then set the baud rate to the configured speed.%1$s' . 'This is useful if the GPS device resets back to a default rate on power loss or when changing the baud rate.', '<br /><br />');
```
rbgarga: This is a really long line, please break it into small ones

src/etc/inc/system.inc
```diff
@@ -1629,46 +1672,64 @@ function system_ntp_setup_gps($serialport) {
...
-	mwexec("cat /tmp/gps.init > {$serialport}");
+	@file_put_contents($gps_device, $gps_init);
+
+	if ($found && $config['ntpd']['gps']['autobaudinit']) {
+		mwexec("stty -f {$serialport} raw speed {$gpsbaud}");
+		mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
+	}
+
+	/* Remove old /etc/remote entry if it exists */
+	if (intval(`/usr/bin/grep -c "^gps0" /etc/remote`) != 0) {
```
rbgarga: Do not use backticks. You can trust mwexec return:
```
if (mwexec("/usr/bin/grep -c '^gps0' /etc/remote")) {
```

And personally I would prefer to do this operation using PHP functions instead of shell sessions

sbeaver-netgate:
No action on requested changes

PR #3562, opened 2017-02-21
Add to NTP support for ACTS ref clock

jskyboo:
Add support for driver 18 ACTS ref clock.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1715,7 +1715,7 @@ function system_ntp_setup_acts($serialport) {
 	}
 
 	/* Add /etc/remote entry in case we need to read from the ACTS with tip */
-	if (intval(`grep -c "^{$acts_shortname}" /etc/remote`) == 0) {
+	if (intval(`/usr/bin/grep -c "^{$acts_shortname}" /etc/remote`) == 0) {
 		@file_put_contents("/etc/remote", "{$acts_shortname}:dv={$serialport}:br#{$actsbaud}:pa=none:\n", FILE_APPEND);
 	}
 
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1690,6 +1690,38 @@ function system_ntp_setup_pps($serialport) {
 	return true;
 }
 
+function system_ntp_setup_acts($serialport) {
+	global $config;
+	$acts_device = '/dev/acts0';
+	$serialport = '/dev/'.$serialport;
+
+	if (!file_exists($serialport)) {
+		return false;
+	}
+
+	// Create symlink that ntpd requires
+	unlink_if_exists($acts_device);
+	@symlink($serialport, $acts_device);
+
+	$actsbaud = $config['ntpd']['acts']['speed'];
+	/* Configure the serial port for raw IO and set the speed */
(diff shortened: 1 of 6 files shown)


src/etc/inc/system.inc
```diff
@@ -1690,6 +1690,38 @@ function system_ntp_setup_pps($serialport) {
...
+
+	$actsbaud = $config['ntpd']['acts']['speed'];
+	/* Configure the serial port for raw IO and set the speed */
+	mwexec("stty -f {$serialport} raw speed {$actsbaud}");
+	mwexec("stty -f {$serialport}.init raw speed {$actsbaud}");
+
+	/* Remove old /etc/remote entry if it exists */
+	$acts_shortname = "acts0";
+	if (intval(`/usr/bin/grep -c "^{$acts_shortname}" /etc/remote`) != 0) {
+		mwexec("/usr/bin/sed -i '' '-n '/'{$acts_shortname}'/!p' /etc/remote");
```
jim-p: Those single quotes do not look right and the ! probably needs escaped. Are you certain that command is doing what you intend?

src/etc/inc/system.inc
```diff
@@ -1690,6 +1690,38 @@ function system_ntp_setup_pps($serialport) {
...
+	}
+
+	/* Add /etc/remote entry in case we need to read from the ACTS with tip */
+	if (intval(`/usr/bin/grep -c "^{$acts_shortname}" /etc/remote`) == 0) {
```
jim-p: Rather than running the grep a second time, this should just be an else clause to the if above, since if it fails the `!= 0` test then it must be 0.

src/usr/local/www/services_ntpd_acts.php
```diff
@@ -0,0 +1,323 @@
...
+##|*NAME=Services: NTP ACTS
+##|*DESCR=Allow access to the 'Services: NTP ACTS' page.
+##|*MATCH=services_ntpd_acts.php*
+##|-PRIV
+
+require_once("guiconfig.inc");
+
+if ($_POST) {
+	unset($input_errors);
+
```
jim-p: Input validation here is required, you will need to add code to verify that the input makes sense (proper type, within required ranges, items from a list exist in that list, etc)

src/usr/local/www/services_ntpd_acts.php
```diff
@@ -0,0 +1,323 @@
...
+$tab_array[] = array(gettext("ACTS"), true, "services_ntpd_acts.php");
+display_top_tabs($tab_array);
+
+$form = new Form;
+
+$section = new Form_Section('NTP Serial ACTS Configuration');
+
+$section->addInput(new Form_StaticText(
+	'Notes',
+	'This driver supports the US (NIST and USNO) and European (PTB (Germany), NPL (UK), etc.) modem time services, as well as Spectracom GPS and WWVB receivers connected via a modem. The driver periodically dials a number from a telephone list, receives the timecode data and calculates the local clock correction. It is designed primarily for backup when neither a radio clock nor connectivity to Internet time servers are available. It can also be configured to operate full period.'
```
jim-p: Please split this up into multiple lines at around column 80

jim-p:

Thanks!

sbeaver-netgate:
No activity in almost a year. Please reopen if still desired.

PR #3566, opened 2017-02-21
Do not waste limited menu space (Bug #7238), add User Actions menu (Change password, Logout)

doktornotor:
1/ Shuffle "Gold" under "Help" menu.
2/ While here, I changed the "Logout" button to a dropdown and added a link to "Change Password" for local users.

(It'd be possible to add things like Reboot and Halt there as well - which are currently buried in stupid counter-intuitive places IMNSHO, but that discussion did not lead to anything productive last time I tried, so I skipped that - see https://redmine.pfsense.org/issues/6884). If pfSense guys changed their mind meanwhile, it's simply a matter of adding two lines:

```
$actions_menu[] = array(gettext("Halt System"), "/diag_halt.php");
$actions_menu[] = array(gettext("Reboot"), "/diag_reboot.php");
```

### DIFF
--- a/src/usr/local/www/head.inc
+++ b/src/usr/local/www/head.inc
@@ -398,10 +398,6 @@ $diagnostics_menu[] = array(gettext("Packet Capture"), "/diag_packet_capture.php
 
 $diagnostics_menu = msort(array_merge($diagnostics_menu, return_ext_menu("Diagnostics")), 0);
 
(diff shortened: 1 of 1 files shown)


doktornotor:
Closing. Can revisit in another couple of years, while the Gold thing will still sticking out there like a sore thumb.

PR #3567, opened 2017-02-21
Add syslogd to Status > Services (Feature #4382)

doktornotor:
Sort includes, use full path to pkill and remove openntpd while here.

### DIFF
--- a/src/etc/inc/service-utils.inc
+++ b/src/etc/inc/service-utils.inc
@@ -693,7 +693,7 @@ function service_control_stop($name, $extras) {
 			}
 			break;
 		case 'relayd':
-			mwexec('/bin/pkill relayd');
+			sigkillbyname("relayd", "TERM");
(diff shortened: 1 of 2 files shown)


src/etc/inc/service-utils.inc
```diff
@@ -684,7 +693,17 @@ function service_control_stop($name, $extras) {
 			}
 			break;
 		case 'relayd':
-			mwexec('pkill relayd');
+			mwexec('/bin/pkill relayd');
```
rbgarga: It's better to use sigkillbyname() here

doktornotor:
Done, but I'd rather keep this focused. Apparently, the proper command would be ```/usr/local/sbin/relayctl stop``` but that doesn't seem to work, at least per comments in ```/etc/inc/vslb.inc``` (which also uses the pkill without full path.

PR #3568, opened 2017-02-22
Only start dhcpleases if DHCP server is enabled (Bug #6750)

doktornotor:
As noted on the bug, it doesn't make sense to have a process watch a lease file that will never be populated because dhcpd isn't running. Also, enabling DHCP-related options in forwarder/resolver should only be allowed when we are actually using DHCP server on pfSense.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -570,12 +570,15 @@ function system_hosts_generate() {
 
 function system_dhcpleases_configure() {
 	global $config, $g;
-
+	if (!function_exists('is_dhcp_server_enabled')) {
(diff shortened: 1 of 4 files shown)


rbgarga:
@doktornotor while here, did you check the same for dhcpleases6 / DHCPv6 server ?

doktornotor:
No, my understanding is that DHCPv6 is irrelevant here. These never get registered in the resolver/forwarder -- since the hostname is not supplied. E.g., https://tools.ietf.org/id/draft-ietf-dhc-addr-registration-03.txt

It's just not implemented.

PR #3571, opened 2017-02-23
Fix #7299 Old rules with ICMP types

phil-davis:
This is a minimal patch to just make sure that when there are ICMP types, the 'ipprotocol' gets set to 'inet' if it is not specified in the rule.

It fixes the reported behavior that I managed to reproduce in 2.3.3 - so the patch needs to be back-ported there.


phil-davis:
See #3572 for a more general fix.

phil-davis:
@rbgarga @jim-p  Grischa Zengel has commented on the issue, that after an upgrade a rule like this causes "error loading the rules" and so the rule set is not loaded at all. So it is a big issue for those that come across it.
This, or #3572 or some config upgrade code or something else should be put in place.
I notice that movement on PRs has stopped the last couple of days.

jim-p:
I merged #3572 - Thanks!

PR #3573, opened 2017-02-23
Fix #7300 provide default value for ipprotocol for old rules

phil-davis:
An alternative is to provide upgrade code that adds the 'ipprotocol' = 'inet' tag to old rules, so that there are no longer any old rules around that are missing this field.


phil-davis:
@jim-p This is another part of the "missing ipprotocol" key in old rules that effects the UI. When people edit an old rule like this, it "forgets" the icmp-type(s) that were in the old rule.

PR #3574, opened 2017-02-23
Redmine #7301 Provide word-break opportunity for dynamic DNS host names

phil-davis:
This can later also be extended very easily to places like firewall_aliases.php etc that show full host/domain FQDN names in a column.

And it is easily modified in 1 place if we want to also break on any other characters that might be convenient.

### DIFF
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -439,6 +439,10 @@ function pprint_port($port) {
 	return $pport;
 }
 
+function insert_word_breaks_in_domain_name($domain_name) {
+	return str_replace('.', '.<wbr>', $domain_name);
+}
+
 function firewall_check_for_advanced_options(&$item) {
 	$item_set = "";
 	if ($item['os']) {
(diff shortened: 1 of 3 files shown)


NOYB:
Would prefer the break before dot rather than after so it does like this.

somereallylonghosthame
.somereallylongdomainname
.com

Rather than like this.

somereallylonghosthame.
somereallylongdomainname.
com

Host/domain names just seem more natural as.
host
dot this
dot that
dot com

I know kinda picky isn't it.

phil-davis:
Not picky - I had not thought about it, and what you say is true.

NOYB:
Maybe a generic function that accepts string, break opportunity character, and fore/aft break position.
function word_break_opportunity ($str, $chr, $fore_aft)

phil-davis:
A function like that will be good if we refactor the other places that put word break opportunities in other types of potentially long strings. Then this insert_word_breaks_in_domain_name() would call it, and it would also be called by other new functions like insert_word_breaks_in_alias_name() insert_word_breaks_in_myspecialtypeofstring() and so on, which would specify for different "types" of string, what is the preferred set of wbr-related chars.

phil-davis:
@sbeaver-netgate @rbgarga @jim-p this is another little one that would be easy to review.

PR #3575, opened 2017-02-23
Captive portal: add a captive portal detection bypass feature

plumbeo:
Both Android and iOS include a captive portal detection feature that's supposed to help users notice immediately when they're using a network that requires authenticating on a captive portal. The rationale of this feature is that captive portals assume the user will use a browser and be intercepted and redirected to the login page, but, especially on mobile devices, the browser is just one of many apps running, and it's likely that the user will not launch it at all. If this happens every connection will be blocked, background apps will not be able to connect to the internet and pretty much everything (mail synchronization, social media apps, instant messaging, notifications) will silently stop functioning.

The basic mechanism of a detection feature is simple: the device tries to connect to well-known services that always return well-known results; if the device receives what it's expecting, it assumes the internet connection is working: if it gets anything else it opens a pop-up with a browser that gets duly intercepted and redirected to the captive portal login page, allowing the user to input his credentials and gain access, or to leave the network.

The pop-up window used to be an instance of the system browser, but starting with iOS 7 and Android 5 it relies on a reduced-function browser; in both Android and iOS the CNA/NPD browsers don't support JavaScript and cookies, can't access information stored on the device, including saved password, can't open links on new windows or panes of the system browser, etc. The Android implementation has even more restrictions: the pop-up window doesn't allow special links to open in other apps (for example the Play Store) and as soon as the device detects a working connection (using the method described above) the pop-up window is automatically closed.

While the feature as it is implemented works well enough in most cases, its limitations can still be a problem: for example it's very hard if not impossible to implement authentication via third party services (OAuth, Facebook, etc. etc.) on the limited-function browser, and, on Android, to show the user useful information via a post-authentication landing page (the NPD browser is closed too soon) or directing him to the Play Store to download an app. These issues have been noticed and reported several times (https://code.google.com/p/android/issues/detail?id=173980 for example), but a solution is not forthcoming, so several vendors who provide captive portal features implemented mechanisms to work around the problems, preventing the device from detecting the presence of a captive portal so that the user can login using the full-fledged system browser. Cisco (http://www.cisco.com/c/en/us/td/docs/wireless/controller/7-4/configuration/guides/consolidated/b_cg74_CONSOLIDATED/b_cg74_CONSOLIDATED_chapter_01010101.html), Aruba (https://community.arubanetworks.com/t5/Controller-Based-WLANs/Apple-iOS7-Captive-Network-Assistant-bypass-with-controller/ta-p/181094) and Ruckus (https://support.ruckuswireless.com/answers/000002368), for example, are among these vendors.

This PR implements a similar feature on the captive portal. It's been tested on Android 4.2-7 and iOS 10, and with captive portal bypass enabled the devices never show the pop-up CNA/NPD window. Since the iOS CNA browser is a lot more functional than the Android NPD's one, the bypass feature can be enabled for each OS independently from the other. The behavior of the CNA/NPD feature is little documented if not completely undocumented, so it's possible that future releases will need tweaks, though the basic mechanism hasn't changed in a while.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1338,6 +1338,7 @@ function system_generate_nginx_config($filename,
 	if ($captive_portal !== false) {
 		$cp_interfaces = explode(",", $config['captiveportal'][$captive_portal]['interface']);
 		$cp_hostcheck = "";
+		$cp_bypass = "";
 		foreach ($cp_interfaces as $cpint) {
 			$cpint_ip = get_interface_ip($cpint);
 			if (is_ipaddr($cpint_ip)) {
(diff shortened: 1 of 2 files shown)


sbeaver-netgate:
Change request not addressed in 6 months. Please resubmit if desired.

PR #3577, opened 2017-02-23
Fix #7306 Correctly filter log widget entries by interface description

phil-davis:
conv_log_filter() needs to be passed the interface description "MYLAN", "MYEXTRAIFACE" or whatever rather than the internal interface name "lan", "opt1" etc.

It all works fine when these have not been changed from the defaults (passing "LAN", "OPT1"... is fine). But not when the interface description has been modified.

### DIFF
--- a/src/usr/local/www/widgets/widgets/log.widget.php
+++ b/src/usr/local/www/widgets/widgets/log.widget.php
@@ -71,6 +71,8 @@ if ($_POST) {
 	exit(0);
 }
 
+$iface_descr_arr = get_configured_interface_with_descr();
+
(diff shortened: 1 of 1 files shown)


phil-davis:
@sbeaver-netgate @rbgarga @jim-p review of PRs seems to have come down to a dribble the last week or more. Any chance to look at big fixes and get the fixes in? So people can have the fix and not have to think about the bug any more.

sbeaver-netgate:
The rate at which PRs are received seems to be accelerating still, so we are trying to put some processes and procedures around them - hence the labels, request for review etc. It also works best if one person does the merging. Renato will be getting to these as soon as possible.

PR #3579, opened 2017-02-24
Dyndns dreamhost

frankthetank:
Implemented Dreamhost ISP dynamic DNS updates.

Dreamhost does not allow direct record updates through their API.
Records must be deleted and then added.  Unfortunately, to delete the
record you must first retrieve the existing record to be able to pass
back the old IP address for delete.  

Rather than copy what CloudFlare had done and embed several of the
queries into its own update process I broke out some of the operations
into _lookup_current and _remove(existing_ip) methods.  This will allow
for more consistent logging output if others use these features, too.


pfSense-git-updates:

jim-p:
Please update your fork, there are conflicts.

frankthetank:
Fork updated.

rbgarga:
Can you please rebase your fork one more time, there was a new dyndns added today that created a conflict here. Sorry about that.

frankthetank:
I've noticed that a lot of the input form builders in services_dyndns_edit.php call setHelp() with two strings now.  I cannot get these to work on my local box unless the second parameter is an array().  Am I missing a change from somewhere on my pfsense box?

Leaving the second parameter as a string gives me :
```
/services_dyndns_edit.php: New alert found: PHP ERROR: Type: 4096, File: /usr/local/www/classes/Form/Group.class.php, Line: 89, Message: Argument 2 passed to Form_Group::setHelp() must be of the type array, string given, called in /usr/local/www/services_dyndns_edit.php on line 302 and defined
```

Let me know if I need to change those values back to plain strings rather than arrays.

sbeaver-netgate:
Your local box is on an earlier version of pfSense that used an array as the second element. It was changed to a printf type syntax some months ago.

Steve
--

frankthetank:
@sbeaver-netgate Thanks for that heads up.  I will submit another change shortly that reverts my second parameter arrays in setHelp calls.

src/etc/inc/dyndns.class
```diff
@@ -115,6 +116,8 @@
 	 *  DuiaDNS         - Last Tested: 25 November 2016
 	 *  DuiaDNS IPv6    - Last Tested: 25 November 2016
 	 *  Hover           - Last Tested: 15 February 2017
+	 *  DreamHost       - Not Tested
+	 *  DreamHost IPv6  - Not Tested
```
rbgarga: @frankthetank We do not have access to all DynDNS providers, usually who submits a new provider is responsible to test it and make sure it works properly

frankthetank:
Commit comment says it all - I have implemented the IPv6 for those whose ISP can provide them with IPv6 as it is the same API call just a different record type.  Unfortunately I cannot test it as my ISP does not yet hand out IPv6 addresses.  I verified that the Dreamhost API works the same for both record types by manually calling the API to effect a change.

src/etc/inc/dyndns.class
```diff
@@ -882,6 +921,191 @@
...
+					}
+					if ($this->_dnsPort) {
+						$port = ":" . $this->_dnsPort;
+					}
+					curl_setopt($ch, CURLOPT_URL, $server . $port);
+					curl_setopt($ch, CURLOPT_POSTFIELDS, $post_data);
+					break;
+				default:
+					break;
+				}
```
jim-p: Fix indent (should be one tab stop left of where it is)

src/etc/inc/dyndns.class
```diff
@@ -882,6 +921,191 @@
...
+				if ($this->_dnsService != 'ods') {
```
jim-p: This whole block is indented one level too far

src/etc/inc/dyndns.class
```diff
@@ -882,6 +921,191 @@
...
+					curl_setopt($ch, CURLOPT_HEADER, 1);
+					$response = curl_exec($ch);
```
jim-p: If only dreamhost gets setup for this, wouldn't it be dangerous to execute this for other providers? Perhaps the default case in the switch above should set a flag that causes this block to be skipped as well, for safety.

I know that currently only dreamhost calls this function, but I'd rather be safe and consider the future.

src/etc/inc/dyndns.class
```diff
@@ -882,6 +921,191 @@
...
+					$header = substr($response, 0, $header_size);
+					$data = substr($response, $header_size);
+					$this->_checkStatus($ch, $data, $header);
+					@curl_close($ch);
+				}
+			}
+		
+		/**
+		 * Private Function (added 23 Feb 17)
+		 * Retrieves current DNS records.
```
jim-p: This description should be changed to reflect that it's looking up the address as reported by the provider with an API call rather than a DNS lookup, which is what the statement implies.

src/etc/inc/dyndns.class
```diff
@@ -882,6 +921,191 @@
...
+					if ($this->_dnsPort) {
+						$port = ":" . $this->_dnsPort;
+					}
+					curl_setopt($ch, CURLOPT_URL, $server . $port);
+					curl_setopt($ch, CURLOPT_POSTFIELDS, $post_data);
+					break;
+				default:
+					break;
+			}
+			if ($this->_dnsService != 'ods') {
```
jim-p: As with the previous function, this needs a safety belt to prevent it from running on other providers. Maybe check for an empty CURLOPT_URL, or use a flag variable as I mentioned previously.

Or consider making it a more generic function that can fall back to performing a DNS request if the API call isn't possible.

src/etc/inc/dyndns.class
```diff
@@ -1543,6 +1767,78 @@
...
+						case 'no_record':
+							$status = $status_intro . $error_str . gettext("No record exists.");
+							break;
+						case 'no_type':
+							$status = $status_intro . $error_str . gettext("No type exists.");
+							break;
+						case 'no_value':
+							$status = $status_intro . $error_str . gettext("No value exists.");
+							break;
+						case 'no_such_record ':
```
jim-p: Lots of unnecessary repetition throughout this block. Very similar error messages could also likely be combined if they are effectively the same. Use multiple cases before the code to help with that, such as:

```
case 'no_record':
case 'no_such_record ':
case 'invalid_record':
	$status = $status_intro . $error_str . gettext("No valid record exists.");
	break;
```

src/etc/inc/dyndns.class
frankthetank: The block is actually a copy/paste from the _update() method and the 'ods' conditional can be removed.  To allay your concern with inadvertent calls I've added an internal flag to this method.  Anyone needing to implement a remove operation will have to modify the flag in order for the remove to work.

src/etc/inc/dyndns.class
frankthetank: Is there a way to get the "old" WAN IP address or are we way beyond that at this point?  I was wondering how else I could obtain the previous WAN address rather than having to perform a complete API call.  In the short term I have added a flag in a similar way to above.

src/etc/inc/dyndns.class
frankthetank: I have grouped the no_*/no_such_* values but there is a distinct difference between a record that does not exist and one that is invalid.

PR #3581, opened 2017-02-24
Remove unused ntpWidgetUpdateFromServer() from NTP widget

phil-davis:
I don;t see where this is used at all. Nothing seems to break if I remove it.


doktornotor:
Ah yes, that was some wrapper for the the Win98 garbage.

phil-davis:
@sbeaver-netgate If you are still around, this is another little "junk PR". If it can be merged it will save me possible conflicts when doing other general widget enhancements.

PR #3582, opened 2017-02-24
Fix the pkg_call() and set the timeout to a sane value (Bug #6594)

doktornotor:
What was the idea here?! http://php.net/manual/en/function.stream-select.php
```int stream_select (array &$read, array &$write, array &$except, int $tv_sec [, int $tv_usec = 0 ])```

If ```tv_sec``` is NULL ```stream_select()``` can *block indefinitely*, returning only when an event on one of the watched streams occurs (or if a signal interrupts the system call). So, that original 300 sure like hell is not 300 seconds (as suggested by the comment on line 143). This thing will basically never timeout as is.


rbgarga:
Good catch! Thank you!!!

PR #3583, opened 2017-02-25
Fix #7316 Alias Edit keep pattern, placeholder, title in sync

phil-davis:
1) Add some new rows to what is already displayed, then change the alias type between a host and port type. The pattern for the extra rows is not changed. That is because $counter is a fixed number generated when the page is built with PHP - so the JS only manipulates the first $counter rows initially displayed.
The .each() code fixes this. Note that the special matching is needed, because there are other elements that start with 'address' - e.g. 'address_subnet'

2) Make the title attribute ("hover text") and placeholder attribute also change their text value as the alias type is changed. This is the problem reported in the Redmine issue.

### DIFF
--- a/src/usr/local/www/firewall_aliases_edit.php
+++ b/src/usr/local/www/firewall_aliases_edit.php
@@ -611,6 +611,26 @@ $pattern_str = array(
 	'urltable_ports'	=> '.*'					// Alias Name or URL
 );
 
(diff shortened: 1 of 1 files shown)


phil-davis:
@sbeaver-netgate @rbgarga @jim-p another little fix that is labeled tested and could easily be merged.

PR #3587, opened 2017-02-25
flash_message, commit session when done with it, this prevents blocking other requests.

PiBa-NL:
flash_message, commit session when done with it, this prevents blocking other requests.


PiBa-NL:
Closing this in favor of : https://github.com/pfsense/pfsense/pull/3598 which creates a more 'proper' way to start / commit sessions

PR #3589, opened 2017-02-26
Fix syntax error in services_wol_edit.php

phil-davis:


phil-davis:
@jim-p If you are still around, it would be nice to fix this accidental syntax error, rather than having it keep coming in snaps.

PR #3590, opened 2017-02-26
Add configure link to Status > Queues error message if traffic shaping not configured (Bug #6696)

doktornotor:


src/usr/local/www/status_queues.php
```diff
@@ -112,7 +112,7 @@ class QueueStats {
 include("head.inc");
 
 if (!is_array($config['shaper']['queue']) || count($config['shaper']['queue']) < 1) {
-	print_info_box(gettext("Traffic shaping is not configured."));
+	print_info_box(sprintf(gettext('Traffic shaping is not configured. It can be configured here: %1$s%2$s%3$s.'), '<a href="/firewall_shaper.php">', gettext('Firewall &gt; Traffic Shaper'), '</a>'), 'danger');
```
rbgarga: Second item to be replaced by another gettext() call should be replaced by text IMO. Only HTML tags would be kept as parameters

doktornotor: Hmm, This is very much a verbatim copy of the code from https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/status_upnp.php#L55

doktornotor:
Just wontfix the bug, perhaps. Closing the PR.

PR #3591, opened 2017-02-26
Remove trailing slash from proxy URL (Bug #3445)

doktornotor:


doktornotor:
Uhm.... Apparently no longer needed due to input validation.

PR #3592, opened 2017-02-26
use nameservers from active interface only

peahonen:
Currently there doesn't seem to be a way to only use DNS servers assigned by the active tier ISP in a multi WAN setup.

Please see this forum thread for details
https://forum.pfsense.org/index.php?topic=126017.0

This patch updates the DNS configuration as gateway group state changes and fixes the issue described above. I did not have the opportunity for regression testing on other possible configurations.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -229,6 +229,22 @@ function system_resolvconf_generate($dynupdate = false) {
 
 	return 0;
 }
+function get_dns_glob_pattern() {
+       // names of interfaces to use dns settings from
+        $dns_interfaces = array();
+        $dns_glob_pattern = "*"; // default to using all interfaces
+
+        // get current gateway groups state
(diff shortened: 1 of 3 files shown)


peahonen:
Is there something I can help with to get this merged?

sbeaver-netgate:
Change request not implemented in 6 months. Please resubmit if still desired.

peahonen:
Need straightforward instructions for local ISO image build from source for validating the changes.

rbgarga:
You don't need to build a custom ISO to validate changes. You can install pfSense, then in System -> Package Manager you can install a package called System_Patches, that is a package to help users to get patches applied on current instance.

After that, you can apply a Pull Request patch to your system just adding `.diff` suffix to its URL. for example, to apply this PR you would use the URL `https://github.com/pfsense/pfsense/pull/3592.diff`

PR #3593, opened 2017-02-26
Redmine #7318 None button for widget filters

phil-davis:
1) Handle properly the case when a user hides all the items on a widget. They might as well be allowed to do this, and the code can easily handle having nothing to display. In that case it puts a message in the widget panel to indicate that all items are hidden. That will help people realize the situation and use the settings tool icon to unhide what they want to see.

2) When all items are selected, rename the "All" button to "None". When the button is clicked, unchecked all items. This allows a user to easily uncheck all but 1 or 2 items.

### DIFF
--- a/src/usr/local/www/widgets/widgets/dyn_dns_status.widget.php
+++ b/src/usr/local/www/widgets/widgets/dyn_dns_status.widget.php
@@ -233,13 +233,20 @@ function get_dyndns_service_text($dyndns_type) {
 					<tbody>
 <?php
 				$skipdyndns = explode(",", $user_settings['widgets']['dyn_dns_status']['filter']);
+				$not_all_shown = false;
 				foreach ($all_dyndns as $dyndns):
+					if (in_array(get_dyndnsent_key($dyndns), $skipdyndns)) {
+						$check_box = '';
+						$not_all_shown = true;
+					} else {
+						$check_box = 'checked';
+					}
 ?>
 						<tr>
 							<td><?=get_dyndns_interface_text($dyndns['interface'])?></td>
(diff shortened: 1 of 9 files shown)


phil-davis:
Conflicts resolved - actually it was easy because all the code here was the better code, so in the end I could accept all the code that was here.
Ready for review (do you like the idea and the way I have implemented it) and testing.

phil-davis:
@sbeaver-netgate what do you think about this?
I re-used the "All" button as "None" when all items are already checked. This saves wasting space making another button - which helps stuff to fit if the dashboard has many columns.

As a separate issue, I just noticed that the Traffic Graphs widget has an old-style selector for choosing which interface graphs to display. So I guess that should be updated to the checkbox style, just like the Interfaces and Interface Statistics widgets. I am happy to do that. If you are happy with the "None" button stuff here, then I will do it with the All-None button stuff.

sbeaver-netgate:
Take a look at js/pfSense.js line 169. It seem like if you add the class "checkbox.multi" to the checkbox group, the all/none button is added automagically. This is some Sjon Hortensium code that I have not actually tried, but I remember seeing it use on another page.

phil-davis:
How to make that work? I fiddled a little bit with no success - but it also refers to control-label.
And it looks like it will append an element with an "a" tag to the end of the checkboxes. How can that be made to fit in with the "Save" button also?

sbeaver-netgate:
I don't know, but I will look later today

phil-davis:
I submitted the 1st commit from this PR separately as PR #3652 "Handle widgets having no items selected for display" as that is an independent change/fix which could easily be merged.

phil-davis:
Closing in favor of #3653

PR #3598, opened 2017-02-27
phpsessionmanager, this helps starting and committing the php session…

PiBa-NL:
… preventing other requests from being blocked longer than required.

### DIFF
--- a/src/etc/inc/phpsessionmanager.inc
+++ b/src/etc/inc/phpsessionmanager.inc
@@ -19,8 +19,6 @@
  * limitations under the License.
  */
 
-include_once('notices.inc');
-
 $session_opencounter = 0;
 $session_write = false;
 $session_action_list = array();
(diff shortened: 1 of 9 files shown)


src/etc/inc/phpsessionmanager.inc
```diff
@@ -0,0 +1,80 @@
...
+		if ($session_write) {
+			session_commit();
+			$session_write = false;
+		} else {
+			session_abort();
+		}
+	}
+	if ($session_opencounter < 0) {
+		$session_opencounter = 0;
+		file_notice("sessionmanager", "PHPSESSION closed more often than opened!" . simplestacktrace());
```
jim-p: Is this really so severe a problem that it warrants a notice instead of a log entry?

src/etc/inc/phpsessionmanager.inc
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+function phpsession_cleanupcheck() {
+	global $session_opencounter, $session_action_list;
+	if ($session_opencounter > 0) {
+		file_notice("sessionmanager", "PHPSESSION {$session_opencounter} open sessions left at shutdown script!".print_r($session_action_list, true));
```
jim-p: Is this really so severe a problem that it warrants a notice instead of a log entry?

src/etc/inc/phpsessionmanager.inc
PiBa-NL: Its not really 'severe' like a big problem i guess, but it should be fixed and this way its easy to notice 'something' unexpected happened soon after it happens. I can change it to a log entry should be enough i guess. Though i personally dont often look at systemlogs during development of things. Maybe i should :)

src/etc/inc/phpsessionmanager.inc
PiBa-NL: Same as above.. I will change it to a log entry.

PiBa-NL:
@jim-p are the changes made with the added alright?

PR #3599, opened 2017-02-28
Add config options to DNS Resolver that mitigate a bug w/ Unifi (and maybe others)

luckman212:
This PR adds some configuration options to Unbound (DNS Resolver) that mitigate a bug with Unifi gear failing to reach the controller due to unqualified DNS lookups to 'unifi.' post 2.3.3+/2.4 master.

- adds ability to specify on a host-by-host basis whether to create both qualified and unqualified DNS entries
- adds option to auto-track 'domain' according to System > General - can be nice time saver / safety net when you have a lot of overrides within the default domain
- Unbound overview page displays these as checkmarks/additional columns, domain will be shown in green if it is set to 'track'
- fixes a bug (seemingly caused by commit 13541a81e1173fc02af9af8ab7fe46df2a51007d) that caused an 'host/domain already exists' error to print when trying to edit an existing Host Override
- minimal testing done - "it works for me!"

_references:_
[reddit thread](https://www.reddit.com/r/PFSENSE/comments/5vdjel/cant_ping_hostname_just_fqdn_since_233_upgrade/)
[2.3.3 changes - DNS](https://doc.pfsense.org/index.php/2.3.3_New_Features_and_Changes#DNS_.2F_Resolver_.2F_Forwarder)
[redmine 6064](https://redmine.pfsense.org/issues/6064)

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -345,6 +345,8 @@ function system_hosts_local_entries() {
 
 /* Read host override entries from dnsmasq or unbound */
 function system_hosts_override_entries($dnscfg) {
+	global $config;
+	$syscfg = $config['system'];
 	$hosts = array();
 
 	if (!is_array($dnscfg) ||
@@ -358,11 +360,16 @@ function system_hosts_override_entries($dnscfg) {
 		if ($host['host'] || $host['host'] == "0") {
 			$fqdn .= "{$host['host']}.";
 		}
-		$fqdn .= $host['domain'];
-
+		if ($host['track_system_domain']) {
+			$fqdn .= $syscfg['domain'];
+		} else {
+			$fqdn .= $host['domain'];
+		}
 		$hosts[] = array(
 		    'ipaddr' => $host['ip'],
-		    'fqdn' => $fqdn
+		    'fqdn' => $fqdn,
+		    'name' => $host['host'],
+		    'add_unqualified' => $host['add_unqualified']
 		);
 
 		if (!is_array($host['aliases']) ||
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -564,6 +564,9 @@ function unbound_add_host_entries($cfgsubdir = "") {
 			$added_ptr[$host['ipaddr']] = true;
 		}
 		$unbound_entries .= "local-data: \"{$host['fqdn']} {$type} {$host['ipaddr']}\"\n";
+		if ($host['add_unqualified']) {
+			$unbound_entries .= "local-data: \"{$host['name']} {$type} {$host['ipaddr']}\"\n";
+		}
 	}
 
 	// Write out entries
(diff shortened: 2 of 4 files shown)


doktornotor:
There is no such bug with UniFi. You need to fix your DHCP to pass a search domain to the UniFi gear. That's all. Working on 5 sites with pfSense and UniFi.

luckman212:
Has something changed recently about the way search domain works in DHCP? I usually leave it "blank" which has always caused DHCP to send the domain from System>General as option 15.

The reason I created this patch is because something definitely seems to have changed -- Search Domain was definitely still being handed out, but somehow the Unifi stuff was ignoring it and sending unqualified queries. This might be better discussed on the forums, but this patch is working well for me at 2 locations.

doktornotor:
SSH into the UniFi client and do cat /etc/resolv.conf. Yeah, better discussed in the forums. This has worked for ~2 years with the oldest site we have, across tons of UniFi firmware versions and a quite a bunch of pfSense versions. I always configure things explicitly for the DHCP server, never liked this stuff happening behind the scenes, it's just an endless source of user confusion. (Ditto with the DNS servers.)

luckman212:
Opened a [topic on the forums](https://forum.pfsense.org/index.php?topic=126354.0) to continue discussion until we know what is going on. I will also be working on some Wireshark captures later to dig deeper.

florian-asche:
There is a feature that got removed. The dns lookup will now only work with full domain or configured search-domain on the client. The /var/etc/hosts on pfsense server has changed.

luckman212:
I created a new unified commit 25c163d6c973737beb449864f0a7f5dc6ee7f488 that has the fixes here in one patch so it can be used in System Patches.

I have been doing extensive testing on this issue -- lots of tcpdump'ing and setting up different test scenarios, and I am going to update the [forum topic](https://forum.pfsense.org/index.php?topic=126354.0) shortly with some new (surprising) findings as to why I believe everything in this PR should remain in for now.

jim-p:
Can you rebase your fork? It has conflicts now.

Also, Ubiquity published a firmware update recently that fixed some issues with DHCP search domains, can the problem behavior be reproduced with a unit on current firmware?

doktornotor:
The busybox applet bug is fixed in current stable firmware as far as UBNT is concerned. (Not sure how desirable is this feature otherwise, seems kinda confusing to me.)

luckman212:
@jim-p sure, I will rebase and fix the conflicts. There were originally 2 features + 1 bugfix in here.  The bugfix related to pconfig[$id] was fixed by Phil in https://github.com/pfsense/pfsense/pull/3679, so I will remove that part. Even though Ubnt has fixed their busybox bug, there are still cases where creating the unqualified host entries in DNS could be useful. The other feature (to track the system domain) I think is useful as well. @doktornotor not sure if that was the one you were confused about? It's pretty straightforward, although I could add some additional helptext if you think it would help.

sbeaver-netgate:
Conflicts still unresolved

jim-p:
Hello,

If you could, please associate this PR to an open Redmine issue either by locating an existing issue at https://redmine.pfsense.org or by creating a new issue. Add a link to the Redmine issue that points to this PR, and also add a link on the PR to the Redmine issue.

Thanks!

PR #3600, opened 2017-02-28
fix sftp support

chrcoluk:
Fix sftp support in sshd, this fix can be tested with flashfxp client. https://redmine.pfsense.org/issues/7327#change-31832


jim-p:
We need to test with user privileges that would otherwise lock out or prevent a user from using scp.

With this change the admin user can use sftp which has historically not worked, though that isn't necessarily a bad thing, but we do need to test other scenarios.

jim-p:
See https://redmine.pfsense.org/issues/7327

chrcoluk:
to add, the change is actually unneeded, I am guilty of not reading the documentation, and can confirm when using the right account to login to sftp the sftp server works as is.

PR #3601, opened 2017-03-01
Fix DNS wildcard support for EasyDNS

WarrenSchultz:
DNS wildcard is currently broken for EasyDNS. This patch fixes that issue.

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -450,6 +450,8 @@
 					break;
 				case 'easydns':
 					$needsIP = TRUE;
+					if (isset($this->_dnsWildcard))
+						$this->_dnsWildcard = "ON";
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

WarrenSchultz:
CLA in place.

src/etc/inc/dyndns.class
```diff
@@ -450,6 +450,8 @@
 					break;
 				case 'easydns':
 					$needsIP = TRUE;
+					if (isset($this->_dnsWildcard))
+						$this->_dnsWildcard = "ON";
```
rbgarga: Like other places in the code you should check if it's != "OFF" before assume it should be "ON"

sbeaver-netgate:
No action on requested changes.

PR #3602, opened 2017-03-02
Traffic graphs widget filter

phil-davis:
Refactor the traffic graphs widget to give it the same checkbox UI for selecting which interfaces to show as is used in the other widgets (like interfaces and interface statistics widgets).

1) Make it have the same general code structure as the interfaces widget and gateways widget.
2) Give it the "Show" checkboxes like the other widgets.
3) Use $user_settings to store the settings - this will make the settings be custom for any users who use custom dashboard layout and settings (like all the other widgets)
4) A whole bunch of JS could go, because when changing settings and pressing Save, the dashboard refreshes anyway. So all the special JS that did special saving can go. The standard way of saving works like all the other widgets.
5) Change the way and naming of the various settings so they conform to the array key style used by other widgets. The array of interfaces to show had to be changed anyway, because the "show" checkbox list system records the items that are NOT shown.


phil-davis:
Note1: This code works fine in 2.3.4-DEVELOPMENT and can be back-ported there to make this widget consistent with the others.
Note2: I did not put the All/None button functionality #3593 into this code, because I first would like this code to work and ave a UI just the same as the other widgets in the current master and RELENG_2_3 branches. Then if #3593 is merged, I can easily add that new UI functionality to the traffic graphs widget as a separate commit and PR. That will make it easier to cherry-pick suitable stuff from whatever stage.
Note3: If this is acceptable, then it also needs config upgrade code which I can sort out as an extra commit here.

phil-davis:
@sbeaver-netgate and others, comments please.

phil-davis:
Config upgrade code is in the latest commit above.

phil-davis:
@jdillard @sbeaver-netgate any comments? Is this "a good thing"?

phil-davis:
This code (with some small refactoring) is now in the following chain of PRs:
#3656 Traffic Graphs Widget handle no items selected for display
#3657 Redmine 7318 None button for Traffic Graphs Widget filter

Closing this in favor of the above.

PR #3614, opened 2017-03-05
Remove dummy config_lock() and config_unlock() functions

doktornotor:
Been no-op for ages (https://github.com/pfsense/pfsense/commit/0027de0a544438f146cfc94f005fd6f4ba9f94d7).

Note: Don't merge to RELENG_2_3 until https://github.com/pfsense/pfsense/pull/3613 is done.


jim-p:
I don't see any sign of it used in packages either, looks good to jettison.

PR #3617, opened 2017-03-05
Add missing reasons to write_config() calls for a couple of shell scripts

doktornotor:


phil-davis:

PR #3618, opened 2017-03-05
Add reason to write_config() calls for /etc/inc files

doktornotor:


doktornotor:
@phil-davis - Yeah, definitely. I did just the stuff missing these in /etc.

phil-davis:

PR #3619, opened 2017-03-06
Add reason to write_config() calls for OpenVPN

phil-davis:
Here is an example of what could/needs to be done through the whole of usr/local/www similar to what @doktornotor has done in other pull requests in recent hours.

Is this an acceptable and useful way to go?

If so, a ToDo issue can be raised and people can work on it as they care and have time.


doktornotor:
I think the ToDo has been https://redmine.pfsense.org/issues/204 -- for some 7+ years. 🤣

PR #3624, opened 2017-03-07
Add reason to write_config() calls for Firewall - Aliases/Schedules/Virtual IPs, fix includes

doktornotor:


sbeaver-netgate:
Are include fixes required by the write_config changes? If not, let's separate those please.

doktornotor:
Redone as https://github.com/pfsense/pfsense/pull/3663

PR #3625, opened 2017-03-07
All-None button dyndns widget

phil-davis:
Something like this makes the "All" button change to "None" when all the checkbox items are checked.
It also reacts to the check/uncheck of each item, so it you manually check all the unchecked items, then the All/None button switches behavior to "None".
Similarly if all the items are checked and you uncheck 1 of them, then the All/None button switches behavior to "All".

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -445,6 +445,38 @@ function updateWidgets(newWidget) {
 	$('input[name=sequence]', $('#widgetSequence_form')).val(sequence);
 }
 
+// Determine if all the checkboxes are checked
+function are_all_checked(checkboxes) {
+	var allBoxesChecked = true;
+	$(checkboxes).each(function() {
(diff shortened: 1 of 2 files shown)


phil-davis:
This is a bit different way that #3593 could be done. I made some functions that could be called from any of the widgets, to set the state of the All/None button and to set/clear all the checkboxes etc.
@sbeaver-netgate but probably you have some thoughts about an even smarter way to make this all happen in JS.
Not that this code works better than what is currently in #3593. It reacts to check/uncheck of individual items to decide the All/None button state.

phil-davis:
@sbeaver-netgate any chance to comment on this approach, or what is in #3593 ?
I have working code for #3547 "Allow multiple copies of a widget on dashboard" but it is on top of #3593 and I need to resolve what to do about this "None" button stuff before I can then submit the other.

phil-davis:
Closing in favor of #3653

PR #3626, opened 2017-03-08
Fix #7364 Console assigned VLAN disappears after reboot

phil-davis:
If trigger_initial_wizard is still waiting to happen, then if the user sets up interface assignment/VLAN stuff from the console, we need to let the system know that has been done. Then not try to cross-check any of that at boot time.


phil-davis:
Note: Redmine bug was reported on 2.3.3
I reproduced the bug on 2.4-DEVELOPMENT and the fix is here against master.
This logic should cherry-pick fine back to RELENG_2_3 and RELENG_2_3_3.

PR #3627, opened 2017-03-08
Add option to disable NTP server (Feature #3567)

doktornotor:
Behavior is unchanged unless ntpd is explicitly disabled, so there is no additional configuration required by users.

### DIFF
--- a/src/etc/inc/rrd.inc
+++ b/src/etc/inc/rrd.inc
@@ -795,7 +795,7 @@ function enable_rrd_graphing() {
 		/* End Captive Portal statistics */
 
 		/* NTP, set up the ntpd rrd file */
-		if (isset($config['ntpd']['statsgraph'])) {
+		if (isset($config['ntpd']['statsgraph']) && $config['ntpd']['enable'] != "off") {
 			/* set up the ntpd rrd file */
 			if (!file_exists("$rrddbpath$ntpd")) {
 				$rrdcreate = "$rrdtool create $rrddbpath$ntpd --step $rrdntpdinterval ";
--- a/src/usr/local/www/services_ntpd.php
+++ b/src/usr/local/www/services_ntpd.php
@@ -37,14 +37,7 @@ if (!is_array($config['ntpd'])) {
 }
 
 if (empty($config['ntpd']['interface'])) {
-	if (is_array($config['installedpackages']['openntpd']) && is_array($config['installedpackages']['openntpd']['config']) &&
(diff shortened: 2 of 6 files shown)


src/etc/inc/rrd.inc
```diff
@@ -795,7 +795,7 @@ function enable_rrd_graphing() {
 		/* End Captive Portal statistics */
 
 		/* NTP, set up the ntpd rrd file */
-		if (isset($config['ntpd']['statsgraph'])) {
+		if (isset($config['ntpd']['statsgraph']) && $config['ntpd']['enable'] != "off") {
```
rbgarga: Why not use a boolean type for 'enable' field?

doktornotor: Because pretty much nothing is using boolean? Namely, the reason here was [this code](https://github.com/pfsense/pfsense/blob/master/src/etc/inc/service-utils.inc#L183) (though, after that I've decided to hide the thing from the status page as it seems it's the way it's being done with the rest of "built-in" services. Also see [#5510](https://redmine.pfsense.org/issues/5510)

So yeah, it's be nice to get some consistency here across the code, however this is not the place to do so and there's virtually no consent on what should be used.

src/usr/local/www/services_ntpd.php
```diff
@@ -210,6 +203,13 @@ function build_interface_list() {
 
 $section = new Form_Section('NTP Server Configuration');
 
+$section->addInput(new Form_Select(
+	'enable',
+	'Enable NTP Server',
+	$pconfig['enable'],
+	array('enabled' => 'Enabled', 'off' => 'Disabled')
+));
```
rbgarga: I would like to hear from @sbeaver-netgate about this change. Several screens use a checkbox to enable/disable services

sbeaver-netgate: The Form_Checkbox() class should be used here. There are plenty of examples throughout the codebase.

e.g.: https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/services_snmp.php#L213

doktornotor: A checkbox was not used on purpose, in order to get an _explicit_ value for disabled service -- which can be checked in the rest of the code, without people having to reconfigure anything, without any config upgrade and without any similar nonsense. IOW, so that it would keep working for everyone as it was for years. 

If you with to discuss whether a checkbox is better than dropdown or whether it's important to using false instead of off or disabled, have fun with this. 

P.S. I also am not responsible for the GET -> POST bugs that got spread across entire codebase and that cripple saving configuration.

PR #3628, opened 2017-03-08
Add AJAX to interfaces widget

doktornotor:
Do some whitespace fixes while here.


luckman212:
I like this - had this on my to-do list but you beat me to it. Can I ask one question:

Why the 60-second default refresh? Is this widget particularly CPU-intensive? If I happen to have the dash up, I would rather know within 10 seconds that one of my interfaces got unplugged vs. a whole minute.

PR #3631, opened 2017-03-08
Add missing includes to config.console.inc

doktornotor:

### DIFF
--- a/src/etc/inc/config.console.inc
+++ b/src/etc/inc/config.console.inc
@@ -23,6 +23,11 @@
  * limitations under the License.
  */
 
+require_once("config.inc");
+require_once("globals.inc");
+require_once("interfaces.inc");
+require_once("util.inc");
+
 function set_networking_interfaces_ports() {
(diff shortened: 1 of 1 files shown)


phil-davis:
What is breaking without these?

doktornotor:
@phil-davis - I did not see anything breaking - it would definitely break though if you were to include this file on its own in your own code. I don't think it's a good practice to rely on some magic happening behind the scenes so that in the end you are lucky enough to end up with the required includes being loaded indirectly by some other files.

phil-davis:
True - e.g. in Diag->Command Prompt when executing some PHP that needs it, it is nice if config.console.inc knows how to include everything that it needs itself.

doktornotor:
I filed https://redmine.pfsense.org/issues/7385 about this as it seems it's broken pretty much everywhere. (Fixed a bunch of those in some pending PRs here, but it gets boring and annoying pretty fast.)

PR #3636, opened 2017-03-10
Add option to not create IPv6 entries in /etc/hosts for LAN interfaces

luckman212:
If a LAN interface's IPv6 configuration is set to Track, and the tracked interface loses connectivity, it can cause connections to the firewall that were established via IPv6 hostname to fail. This can happen unintentionally when accessing the system by hostname, since by default both IPv4 and IPv6 entries are added to /etc/hosts. This patch exposes a new `ipv6dontcreatelocaldns` option that works around this issue by preventing those IPv6 records from being created by system_hosts_local_entries().

Set the option in System > Advanced > Networking

**FILES CHANGED**
- /etc/inc/system.inc
- /usr/local/www/system_advanced_network.php

### DIFF
--- a/src/usr/local/www/system_advanced_network.php
+++ b/src/usr/local/www/system_advanced_network.php
@@ -216,9 +216,9 @@ $section->addInput(new Form_Checkbox(
 	'Do not generate local IPv6 DNS entries for LAN interfaces',
 	$pconfig['ipv6dontcreatelocaldns']
 ))->setHelp('If a LAN interface\'s IPv6 configuration is set to Track, and the tracked interface loses connectivity, '.
-    'it can cause connections to this firewall that were established via IPv6 hostname to fail. This can happen '.
-	'unintentionally when accessing the system by hostname, since by default both IPv4 and IPv6 entries are added '.
-	'to /etc/hosts. Enabling this option prevents those IPv6 records from being created.');
+    'it can cause connections to this firewall that were established via hostname to fail. This can happen '.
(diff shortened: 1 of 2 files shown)


src/usr/local/www/system_advanced_network.php
```diff
@@ -203,6 +210,16 @@
...
 
+$section->addInput(new Form_Checkbox(
+	'ipv6dontcreatelocaldns',
+	'IPv6 DNS entry',
+	'Do not generate local IPv6 DNS entries for LAN interfaces',
+	$pconfig['ipv6dontcreatelocaldns']
+))->setHelp('If a LAN interface\'s IPv6 configuration is set to Track, and the tracked interface loses connectivity, '.
+    'it can cause connections to this firewall that were established via IPv6 hostname to fail. This can happen '.
+	'unintentionally when accessing the system by hostname, since by default both IPv4 and IPv6 entries are added '.
+	'to /etc/hosts. Enabling this option prevents those IPv6 records from being created.');
```
jim-p: I would not mention `/etc/hosts` here -- That isn't necessarily where they're going or where they're pulled from (unbound doesn't use it). Say "entries are added for resolution by DNS" or some variation of that.

luckman212: I'll fix that, thanks for reviewing.

luckman212:
@jim-p are we gtg now on this PR?

jim-p:
Looks OK to me

PR #3640, opened 2017-03-10
Redmine # 5112: Captive Portal - Use User Manager as Authentication source

Augustin-FL:
The captive portal currently only supports local user database, RADIUS or voucher authentication.

I added support for LDAP Authentication (using the User Manager, exactly like OpenVPN does).


Augustin-FL:
indeed. it's done

Augustin-FL:
@rbgarga , @luckman212 , @jdillard , could this PR be merged ?

jippi:
bump :)

jim-p:
Redmine Issue: https://redmine.pfsense.org/issues/5112

Augustin-FL:
Hi @jim-p  , 

I did a main refactoring of the code as you asked. Captiveportal now use a server from the user manager as source.

final

Here are the main changes i did :

- I updated the 6 functions dealing with RADIUS authentication in the captiveportal to make them use servers from server manager : 
 
```
radius() - need RADIUS auth
portal_allow() - need RADIUS accounting start
captiveportal_prune_old()  - need RADIUS accounting start & stop, need RADIUS auth
captiveportal_send_server_accounting() - need RADIUS accounting on & off
captiveportal_disconnect() - need RADIUS accounting stop
captiveportal_radius_stop_all() - need RADIUS() accounting stop
```

All theses functions now rely on user manager servers. This is the most significant part of this Pull Request.

- I added way to log in through an LDAP server from the user manager. I had to copy LDAP authentication code from auth.inc to captiveportal.inc, instead of just using the code of auth.inc.
Yeah, that's bad i know. The reason why i did this is the following problem : ```ldap_backed()``` and ```radius_backed()```  only return TRUE or FALSE, without returning the reason why ("could not contact server"? "invalid credential"?). 

Captiveportal some others services of pfSense (VPN, IPsec) need to know the reason why the auth didn't work, because they need to display an error message. 
Maybe theses functions should be modified so that they return an array and not just a boolean?

Anyway, tell me if i'm wrong...but i believe that this should be done in another Pull Request, shouldn't it? 


- I also updated the format of the SQLite database which contain users, since they was references on RADIUS servers inside it. I mention it here because *this may break all existing captiveportals during update process*.
- As you saw on the above picture, I have rewritten many help text, I changed the order of inputs etc...on the captiveportal configuration page. 


I also fixed few bugs : 
1. When re-configuring a captiveportal, users went disconnected (because captiveportal's rules were flushed) but they was still considered as connected (because they was not removed from the SQLite database). i fixed this by flushing the SQLite db, and added a warning.
2. I fixed a (very strange) parse error bug induced very long time ago by 856e58a

Only Regression : You cannot have multiple RADIUS server. Do you want me to add this feature back ?

jim-p:
If you look at how it works now, all of the entries under "RADIUS options" in the portal config are shared between the servers. The only thing that changes is the address, port, and shared secret. They are specific to the portal, not each individual RADIUS server.

Augustin-FL:
> all of the entries under "RADIUS options" in the portal config are shared between the servers.

Okay....But I am also facing a more deep problem : RADIUS attributes are not used in the same way in OpenVPN, IPsec and CaptivePortal when making an Authentication.

It was very confusing for me so i made a table of most general RADIUS attributes pfSense currently deal with (excluding Radius attributes used for credentials, such as User-Name/User-Password/CHAP-Password/etc...And Vendor-Specific Attributes) :

Attribute given while making an Access-request | Captive Portal (using functions from ```radius_authentication.php```) | OpenVPN (using functions from ```auth.inc```) | IPSec (using functions from ```auth.inc```)
------------ | -------------|------------- | -------------
NAS-Identifier | / _(Support for it has been coded, but its code is currently not executed...If the code was executed, it would send the string given by radiusnasid, or pfSense Hostname)_ | "OpenVPN"  | pfSense's Hostname
NAS-IP-Address | / | The WAN interface IP address or 0.0.0.0 | The WAN interface IP address or 0.0.0.0
Calling-Station-Id | If "default" NAS type set : captiveportal's user MAC address (or MAC address of the last Router encountered by the user); ```If "cisco" NAS type set : captiveportal's user IP address``` | / | /
Called-Station-Id | If "default" NAS type Set : An IP address given by radiussrcip_attribute or IP address of WAN; ```If "cisco" NAS type Set : the captiveportal's user MAC Address (or MAC address of the last Router encountered by the user)``` |  The String corresponding to (NAS-IP-Address):(NAS-Port) | /
Framed-IP-Address | Captiveportal's user IP address | / | /
NAS-Port | The IP/MAC Firewall Rule that will be used to allow the user after successful auth | The TCP/UDP port OpenVPN is listening to | /
NAS-Port-Type | Ethernet" | "Virtual" | /
Service-Type | "Login" | / | /

In order to move completely RADIUS captiveportal login functions to the user manager, i need firstly to update the way user manager use RADIUS Attributes, to make it standard. 

I suggest to make these updates :
- Create an User-Manager setting for NAS-IP-Address, which would be like radiussrcip_attribute in captiveportal. This setting would be stored in the User Manager (so it would be common to all authentications) and would let the choice of wich IP should be used. I plan to make something like this : 
image
- Set the "Service-Type" Attribute globally to "Login"
- Set NAS-Identifier for Captive Portal to "CaptivePortal", instead of letting user the choice of it, as it is done by OpenVPN (maybe IPsec Auth should also be updated?)
- Remove the "cisco" mode, since it probably should not exist. I made some research in cisco documentation, this mode seems to not correspond to any RADIUS behavior. This mode have been added 12 years ago by 856e58a at m0n0wall time, when radius Auth was just being set up and was very buggy...i believe it should be removed.
- Remove Called-Station-Id for captiveportal, as it is giving the same information as Nas-IP-Address...And it should not be used for this purpose. [According to RFC3580](https://mrncciew.com/2013/07/22/called-calling-station-id/), Called-Station-Id should be used to store the MAC address of any Switch/bridge between user's device and pfSense's captiveportal (such as Wi-Fi Access point).

Do you agree with me ?

jim-p:
Seems reasonable but you might want to hold off on doing too much more just yet. We are in the middle of converting over to PHP 7.2 and there will have to be some other internal RADIUS changes since the current RADIUS code isn't compatible as-is, it's based on a very ancient version of the PEAR RADIUS code. We're still discussing internally what will happen, but it will almost certainly change what you'll have to do as well.

Augustin-FL:
hi @jim-p 
I did the changes you asked me. CaptivePortal now rely on a multi-server list, and i completly moved the Authentication to the user manager. 

Would it be possible to have a review on my code ?

sbeaver-netgate:
Please rebase or otherwise resolve conflicts

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+	1 = failed user auth
+	2 = failed user auth with custom server reply recieved
+	3 = successful auth
+	*/
+	$authlevel = 0;
+
+	/* Getting authentication servers from captiveportal configuration */
+	$auth_servers = array();
+	
+	if($cpcfg['auth_method'] === 'none') {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+		$auth_servers[] = array('type' => 'none');
+	} else {
+		$fullauthservers = explode(",", $cpcfg['auth_server']);
+		foreach($fullauthservers as $authserver) {
```
jim-p: Needs a space after `foreach`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+			if(strpos($authserver, ' - ') !== false) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+				$authserver = explode(' - ', $authserver);
+				array_shift($authserver);
+				$authserver = implode(' - ', $authserver);
+				
+				if(auth_get_authserver($authserver) !== null) {
+					$auth_servers[] = auth_get_authserver($authserver);
+				}
```
jim-p: `else` should be on the same line as this closing brace.

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+		$auth_servers[] = array('type' => 'none');
+	} else {
+		$fullauthservers = explode(",", $cpcfg['auth_server']);
+		foreach($fullauthservers as $authserver) {
+			if(strpos($authserver, ' - ') !== false) {
+				$authserver = explode(' - ', $authserver);
+				array_shift($authserver);
+				$authserver = implode(' - ', $authserver);
+				
+				if(auth_get_authserver($authserver) !== null) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+			}
+		}
 	}
 
-	$radiusservers = captiveportal_get_radius_servers();
-
-	if (is_null($radiusctx)) {
-		$radiusctx = 'first';
+	/* Unable to find the any authentication server config - shouldn't happen! - bail out */
+	if(count($auth_servers) === 0) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+		log_error("Zone: {$cpzone} - No valid server could be used for authentication.");
+		$reply_message = gettext("Internal Error");
 	}
```
jim-p: `else` should be on the same line as this closing brace.

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
-			$clientmac,
-			$username,
-			$password,
-			$auth_list,
-			$pipeno,
-			$radiusctx);
-	} else {
-		captiveportal_free_dn_ruleno($pipeno);
+	else {
+		foreach($auth_servers as $authcfg) {
```
jim-p: Needs a space after `foreach`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+			if($authlevel < 3) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+				$radmac_error = false;
+				$attributes = array("nas_identifier" => "CaptivePortal",
+					"nas_port_type" => RADIUS_ETHERNET,
+					"nas_port" => $pipeno,
+					"framed_ip" => $clientip);
+				if(mac_format($clientmac) !== null) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					$attributes["calling_station_id"] = mac_format($clientmac);
+				}
+				
+				$result = null;
+				$status = null;
+				$msg = null;
+				
+				/* Radius MAC authentication */
+				if($cpcfg['auth_method'] === 'radmac' && $clientmac) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					if($authcfg['type'] === 'radius') {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					} else {
+						/* Trying to perform a Radius MAC authentication on a non-radius server - shouldn't happen! - bail out */
+						$msg = gettext("Internal Error");
+						log_error("Zone: {$cpzone} - Trying to perform RADIUS MAC authentication on a non-RADIUS server !");
+						$radmac_error = true;
+						$result = null;
+					}
+				}
+				
+				if(!$radmac_error) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					if($authcfg['type'] === 'none') {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+						$result = true;
+					} else {
+						$result = authenticate_user($login, $password, $authcfg, $attributes);
+					}
+					
+					if(!empty($attributes['error_message'])) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+						$msg = $attributes['error_message'];
+					}
+					
+					if ($authcfg['type'] == 'Local Auth' && $result && isset($cpcfg['localauth_priv'])) {
+						if(!userHasPrivilege(getUserEntry($login), "user-services-captiveportal-login")) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+							$result = false;
+							$msg = gettext("Access Denied");
+						}
+					}
+					
+					if(empty($status)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					
+					if ($authcfg['type'] == 'Local Auth' && $result && isset($cpcfg['localauth_priv'])) {
+						if(!userHasPrivilege(getUserEntry($login), "user-services-captiveportal-login")) {
+							$result = false;
+							$msg = gettext("Access Denied");
+						}
+					}
+					
+					if(empty($status)) {	
+						if($result === true) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+							$status = "ACCEPT";
+						} else if($result === null) {
```
jim-p: Use `elseif`, needs a space after `elseif`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+							$status = "ERROR";
+						} else {
+							$status = "FAILURE";
+						}
+					}
+					
+					if($cpcfg['auth_method'] === 'radmac' && $login == mac_format($clientmac) || $authcfg['type'] === 'none' && empty($login)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+						$login = "unauthenticated";
+					} 
+					// We determine a flag
+					if($result === true) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+						$val = 3;
+					}
```
jim-p: Change to `elseif`, move up to the same line as the closing brace, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					else if($result === false && !empty($attributes['reply_message'])) {
+						$val = 2;
+						$msg = $attributes['reply_message'];
+					}
```
jim-p: Change to `elseif`, move up to the same line as the closing brace, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					else if($result === false) {
+						$val = 1;
+					}
```
jim-p: Change to `elseif`, move up to the same line as the closing brace, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -1584,46 +1493,158 @@ function captiveportal_syslog($message) {
...
+					else if($result === null) {
+						$val = 0;
+					}
+				}
+				
+				if($val >= $auth_val) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+	$cpcfg = $config['captiveportal'][$cpzone];
+	$acctcfg = auth_get_authserver($cpcfg['radacct_server']);
+	
+	if (!isset($cpcfg['radacct_enable']) || empty($acctcfg)) {
+		return null;
+	}
+	
+	if ($type === 'on') {
+		$racct = new Auth_RADIUS_Acct_On;
+	} else if($type === 'off') {
```
jim-p: Change to `elseif`, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
 		$racct = new Auth_RADIUS_Acct_Off;
+	} else if($type === 'start' ) {
```
jim-p: Change to `elseif`, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct = new Auth_RADIUS_Acct_Start;
+		if(!is_int($start_time)) {
+			$start_time = time();
+		}
+	} else if($type === 'stop') {
```
jim-p: Change to `elseif`, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct = new Auth_RADIUS_Acct_Stop;
+		if(!is_int($stop_time)) {
+			$stop_time = time();
+		}
+	} else if ($type === 'update') {
```
jim-p: Change to `elseif`, add space after `elseif`.

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		return null;
+	}
+	
+	if ($type === 'on') {
+		$racct = new Auth_RADIUS_Acct_On;
+	} else if($type === 'off') {
 		$racct = new Auth_RADIUS_Acct_Off;
+	} else if($type === 'start' ) {
+		$racct = new Auth_RADIUS_Acct_Start;
+		if(!is_int($start_time)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+			$start_time = time();
+		}
+	} else if($type === 'stop') {
+		$racct = new Auth_RADIUS_Acct_Stop;
+		if(!is_int($stop_time)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+			$stop_time = time();
+		}
+	} else if ($type === 'update') {
+        $racct = new Auth_RADIUS_Acct_Update;
+		if(!is_int($stop_time)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
 	}
-	foreach ($radiusservers['first'] as $radsrv) {
-		// Add a new server to our instance
-		$racct->addServer($radsrv['ipaddr'], $radsrv['acctport'], $radsrv['key']);
+
+	$racct->addServer($acctcfg['host'], $acctcfg['radius_acct_port'],
+		$acctcfg['radius_secret'], $acctcfg['radius_timeout']);
+	
+	$racct->authentic = RADIUS_AUTH_RADIUS;
+	if(!empty($username)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+	if (PEAR::isError($racct->start())) {
+		captiveportal_syslog('RADIUS ACCOUNTING FAILED : '.$racct->getError());
 		$racct->close();
-		return $retvalue;
-	}
+		return null;
+	}
+	
+	$nasip = $acctcfg['radius_nasip_attribute'];
+	if(!is_ipaddr($nasip)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$nasip = get_interface_ip($nasip);	
+		if (!is_ipaddr($nasip)) {
+			$nasip = get_interface_ip();//We use WAN interface IP as fallback for NAS-IP-Address
+		}
+	}
+	$nasmac = get_interface_mac(find_ip_interface($nasip));
+	$racct->putAttribute(RADIUS_NAS_IP_ADDRESS, $nasip, "addr");
+	$racct->putAttribute(RADIUS_NAS_IDENTIFIER, 'CaptivePortal');
+	
+	if(is_int($ruleno)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct->putAttribute(RADIUS_NAS_PORT_TYPE, RADIUS_ETHERNET);
+		$racct->putAttribute(RADIUS_NAS_PORT, intval($ruleno), 'integer');
+	}
+	
+	if(!empty($sessionid)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct->putAttribute(RADIUS_ACCT_SESSION_ID, $sessionid);
+	}
+
+
+	if(!empty($clientip) && is_ipaddr($clientip)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct->putAttribute(RADIUS_FRAMED_IP_ADDRESS, $clientip, "addr");
+	}
+	if(!empty($clientmac)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct->putAttribute(RADIUS_CALLING_STATION_ID, mac_format($clientmac));
+	}
+	if(!empty($nasmac)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		$racct->putAttribute(RADIUS_CALLED_STATION_ID, mac_format($nasmac).':'.gethostname());
+	}
+	
+	// Accounting request Stop and Update : send the current data volume
+	if(($type === 'stop' || $type === 'update') && is_int($start_time)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
+		define("RADIUS_ACCT_OUTPUT_GIGAWORDS", "53");
+		$racct->putAttribute(RADIUS_ACCT_OUTPUT_GIGAWORDS, intval($volume['output_gigawords']), "integer");
+		$racct->putAttribute(RADIUS_ACCT_INPUT_GIGAWORDS, intval($volume['input_gigawords']), "integer");
+		// Set session_time
+		$racct->session_time = $session_time;
+	
+	}
+	
+	if ($type === 'stop') {
+		if(empty($term_cause)) {
```
jim-p: Needs a space after `if`

src/etc/inc/captiveportal.inc
```diff
@@ -2593,49 +2565,168 @@ function captiveportal_blocked_mac($mac) {
...
 	if (PEAR::isError($result)) {
-		$retvalue['acct_val'] = 1;
-		$retvalue['error'] = $result->getMessage();
-	} else if ($result === true) {
-		$retvalue['acct_val'] = 5 ;
-	} else {
-		$retvalue['acct_val'] = 1 ;
+		 captiveportal_syslog('RADIUS ACCOUNTING FAILED : '.$racct->getError());
+		 $result = null;
+	} else if ($result !== true) {
```
jim-p: Change to `elseif`.

src/usr/local/captiveportal/index.php
```diff
@@ -79,6 +77,11 @@
 	include("{$g['varetc_path']}/captiveportal-{$cpzone}-logout.html");
 	ob_flush();
 	return;
+} else if (!empty($cpsession)) {
```
jim-p: Change to `elseif`

src/usr/local/captiveportal/index.php
```diff
@@ -192,65 +178,55 @@
...
 
-} else if ($_POST['accept'] && $radius_enable) {
-	if (($_POST['auth_user'] && isset($_POST['auth_pass'])) || ($_POST['auth_user2'] && isset($_POST['auth_pass2']))) {
-		if (!empty($_POST['auth_user'])) {
-			$user = $_POST['auth_user'];
-			$paswd = $_POST['auth_pass'];
-		} else if (!empty($_POST['auth_user2'])) {
-			$user = $_POST['auth_user2'];
-			$paswd = $_POST['auth_pass2'];
+} else if ($_POST['accept'] || $cpcfg['auth_method'] === 'radmac') {
```
jim-p: Change to `elseif`

src/usr/local/captiveportal/index.php
```diff
@@ -192,65 +178,55 @@
...
+	
+	$pipeno = captiveportal_get_next_dn_ruleno();
+	/* if the pool is empty, return appropriate message and exit */
+	if (is_null($pipeno)) {
+		$replymsg = gettext("System reached maximum login capacity");
+		if($cpcfg['auth_method'] === 'radmac') {
```
jim-p: Needs a space after `if`

src/usr/local/captiveportal/index.php
```diff
@@ -192,65 +178,55 @@
...
+			portal_reply_page($redirurl, "error", $replymsg);
 		}
-		$auth_list = radius($user, $paswd, $clientip, $clientmac, "USER LOGIN", $radiusctx);
+		log_error("Zone: {$cpzone} - WARNING!  Captive portal has reached maximum login capacity");
+		
+	}
+	
+	$auth_result = captiveportal_authenticate_user($_POST['auth_user'], $_POST['auth_pass'], $clientmac, $clientip, $pipeno);
+	
+	if($auth_result['result']) {
```
jim-p: Needs a space after `if`

src/usr/local/captiveportal/index.php
```diff
@@ -192,65 +178,55 @@
...
-			captiveportal_logportalauth($user, $clientmac, $clientip, "FAILURE", $auth_list['reply_message']);
-			portal_reply_page($redirurl, $type, $auth_list['reply_message'] ? $auth_list['reply_message'] : $errormsg);
-		}
-	} else {
-		if (!empty($_POST['auth_user'])) {
-			$user = $_POST['auth_user'];
-		} else if (!empty($_POST['auth_user2'])) {
-			$user = $_POST['auth_user2'];
+		
+		if($auth_result['login_message']) {
```
jim-p: Needs a space after `if`

src/usr/local/captiveportal/index.php
```diff
@@ -192,65 +178,55 @@
...
 		}
-
-		if ($loginok) {
-			captiveportal_logportalauth($_POST['auth_user'], $clientmac, $clientip, "LOGIN");
-			portal_allow($clientip, $clientmac, $_POST['auth_user']);
+		
+		captiveportal_logportalauth($_POST['auth_pass'], $clientmac, $clientip, $auth_result['login_status'], $replymsg);
+
+		/*Radius MAC authentication. $_POST['auth_user'] has been updated by captiveportal_authenticate_user() */
+		if($cpcfg['auth_method'] === 'radmac' && $type !== 'redir') {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -151,9 +152,10 @@
 	$pconfig['freelogins_updatetimeouts'] = isset($a_cp[$cpzone]['freelogins_updatetimeouts']);
 	$pconfig['enable'] = isset($a_cp[$cpzone]['enable']);
 	$pconfig['auth_method'] = $a_cp[$cpzone]['auth_method'];
+	$pconfig['auth_server'] = explode(",",$a_cp[$cpzone]['auth_server']);
```
jim-p: Needs spaces between parameters

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
-	}
-
 	if ($_POST['maxproc'] && (!is_numeric($_POST['maxproc']) || ($_POST['maxproc'] < 4) || ($_POST['maxproc'] > 100))) {
 		$input_errors[] = gettext("The maximum number of concurrent connections per client IP address may not be larger than the global maximum.");
 	}
-
-	if (trim($_POST['radiusnasid']) !== "" && !preg_match("/^[\x21-\x7e]{3,253}$/i", trim($_POST['radiusnasid']))) {
-		$input_errors[] = gettext("The NAS-Identifier must be 3-253 characters long and should only contain ASCII characters.");
+	
+	if($_POST['auth_method']) {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
+		if($_POST['auth_method'] !== 'none' && empty($_POST['auth_server'])) {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
+			$input_errors[] = gettext("You need to select at least one authentication server.");
+		}
+		/* If RADMAC auth method is selected : carefully check that the selected server is a RADIUS one */
+		if($_POST['auth_method'] === 'radmac') {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
+			foreach($_POST['auth_server'] as $server) {
```
jim-p: Needs a space after `foreach`

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
+				
+				$realauthserver = explode(' - ', $server);
+				array_shift($realauthserver);
+				$realauthserver = implode(' - ', $realauthserver);
+				$realauthserver = auth_get_authserver($realauthserver);
+				
+				if($realauthserver === null || $realauthserver['type'] !== 'radius') {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
+					$input_errors[] = gettext("RADIUS MAC Authentication can only be performed on a RADIUS server.");
+				}
+			}
+			
+			if(isset($_POST['nomacfilter'])) {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -296,48 +274,39 @@
...
+				$input_errors[] = gettext("RADIUS MAC Authentication cannot be used if MAC filtering is disabled");
+			}
+		}
+	}
+	
+	if(isset($_POST['radacct_enable']) && empty(auth_get_authserver($_POST['radacct_server']))) {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -520,6 +422,21 @@ function build_cert_list() {
...
 
+
+function build_authserver_list() {
+
+	$authlist = auth_get_authserver_list();
+	$options = array();
+
+	/* auth types are used by javascript */ 
+	foreach ($authlist as $i => $auth) {
+		if($auth['type'] != 'radius' || $auth['type'] == 'radius' && !empty($auth['radius_auth_port'])) {
```
jim-p: Needs a space after `if`

src/usr/local/www/services_captiveportal.php
```diff
@@ -705,210 +627,144 @@ function build_cert_list() {
...
-	'radiuskey3',
-	null,
-	'text',
-	$pconfig['radiuskey3']
-));
+$options = array();
 
-$section->add($group);
+foreach (auth_get_authserver_list() as $i => $auth) {
+	if($auth['type'] == 'radius' && !empty($auth['radius_acct_port']))
```
jim-p: Needs a space after `if`, fix brace style.

src/etc/inc/captiveportal.inc
```diff
@@ -184,15 +185,14 @@ function captiveportal_configure_zone($cpcfg) {
...
-			unlink_if_exists("{$g['vardb_path']}/captiveportal{$cpzone}.db");
 		} else {
 			captiveportal_syslog("Reconfiguring captive portal({$cpcfg['zone']}).");
 		}
 
-		/* init ipfw rules */
+		/* (re)init ipfw rules. Cause all users to disconnect */
+		captiveportal_radius_stop_all();
 		captiveportal_init_rules(true);
+		unlink_if_exists("{$g['vardb_path']}/captiveportal{$cpzone}.db");
```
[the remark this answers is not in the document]
Augustin-FL: For the main CP zone : Because currently when settings are saved, captiveportal rules are re-appplied unconditionally to the network interface, meaning **all ipfw rules are unconditionally flushed**.

This is a big problem when editing captive portal settings while some users are connected : When saving the settings, **users go technically disconnected** and are redirected to the login page (because ipfw rules are flushed),  but they are still considered as connected and **are unable to log-in again**(because they are still present in the sqlite database).

I solved this issue by flushing SQLite DB when rules are re-generated, and i added a warning "there is connected users, they will get disconnected. Do you want to continue ?" on the GUI. 

For the RADIUS database : because this database now obsolete, captiveportal don't use it anymore.

Augustin-FL:
(please do not merge yet. Conflicts resolved but untested)

src/etc/inc/captiveportal.inc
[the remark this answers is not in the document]
jim-p: > all ipfw rules are unconditionally flushed.

That would be considered a separate issue and would be under its own redmine issue and PR. Though it's tempting to try rolling everything into one, everything should be as narrowly scoped as possible.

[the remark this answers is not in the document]
Augustin-FL: @jim-p No problem. I'll remove this fix and submit it in another PR+Redmine issue.

Should i remove the [GUI Warning](https://github.com/pfsense/pfsense/pull/3640/files#diff-7e49d642280675092542de8bd5454f3cR1202) as well?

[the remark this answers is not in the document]
jim-p: Yes, take that out as well and combine it with the other issue. Ideally we shouldn't clear out the users/rules/etc when saving portal settings so updating settings is less disruptive, but if we can't make that work then yes it should be consistent in the other direction.

Augustin-FL:
Hello,
@rbgarga i think this is a work in progress but to let you informed : 
```
Fatal error: Call to undefined method Auth_RADIUS_MSCHAPv1::raiseError() in /usr/local/share/pear/Auth/RADIUS.php on line 431 PHP ERROR: Type: 1, File: /usr/local/share/pear/Auth/RADIUS.php, Line: 431, Message: Call to undefined method Auth_RADIUS_MSCHAPv1::raiseError()
```
Look like there is some trouble with [this line](https://github.com/pfsense/FreeBSD-ports/commit/0c7f7304b2cfc7dc1bd5570f47198fbc79e37a6e#diff-1d03251a6672978e64594664254f5aadR66) ?

Also, i noticed that you [copied some code about "standard attributes"](https://github.com/pfsense/FreeBSD-ports/blob/0c7f7304b2cfc7dc1bd5570f47198fbc79e37a6e/net/pear-Auth_RADIUS/files/patch-Auth_RADIUS.php#L16-L57) from old ```radius.inc``` to the new ```patch-Auth_RADIUS.php```. This code is a little bit weird and make no real sense as [RADIUS attributes are not used in the same way](https://github.com/pfsense/pfsense/pull/3640#issuecomment-385279841) in OpenVPN, IPsec and CaptivePortal when performing an Authentication. 

I tried to standardize the way attributes are used in this pull request, but outside of the RADIUS class ([here](https://github.com/Augustin-FL/pfsense/blob/f17effa1ee7fe8bcdead47bd75ab35f2c85bc7ff/src/etc/inc/auth.inc#L1625-L1661)). My plan was to remove the content of [putStandardAttributes](https://github.com/pfsense/pfsense/blob/29a272f7361689c87dd7ad9fc1c903e843a1c593/src/etc/inc/radius.inc#L296-L326) function in the old ```radius.inc```, and to use attributes in a standard/common way.

If you plan to merge my pull request and you agree with this, would it be possible to delete lines 19 to 57 in ```patch-Auth_RADIUS.php```? I think it would be better to handle standard attributes outside RADIUS class (so that no need to create functions like ```getNasIp``` and to use ``function_exists()``` anymore )?

Augustin-FL:
- [x] Ready for merge 
- [x] strongly tested 
- [x] Port patch submitted. Please merge it at the same time as this PR
- [x] Resolves issues : 
   - Provide Captiveportal LDAP Authentication - https://redmine.pfsense.org/issues/5112
   - Provide MAC instead of IP in Called-Station-Id and Calling-Station ID Attribute when using RADIUS auth : https://redmine.pfsense.org/issues/4294
   - Use provided user name if there is one when using ```Authentication: None``` : https://redmine.pfsense.org/issues/8361
   - Use a standardized NAS-Identifier when using RADIUS auth : https://redmine.pfsense.org/issues/3686
   - Accounting updates are not sent in a particular situation : https://redmine.pfsense.org/issues/8655

src/usr/local/www/services_captiveportal.php
```diff
@@ -705,210 +627,143 @@ function build_cert_list() {
...
+	$pconfig['radiussession_timeout']
+))->setHelp('When enabled, clients will be disconnected after the amount of time retrieved from the RADIUS Session-Timeout attribute.');
 
-$group->add(new Form_Input(
-	'radiusport2',
-	null,
-	'number',
-	$pconfig['radiusport2']
-))->setHelp('RADIUS port. Leave blank for default (1812)');
+-$section->addInput(new Form_Checkbox(
```
Augustin-FL: I need to update this line as well

Augustin-FL:
>  I don't see any upgrade code, which will be required to migrate settings from the old auth style to the new style.

I agree. Following your message, am working on an ```update_config.inc``` function in a [separate branch](https://github.com/Augustin-FL/pfsense/tree/bump-config). i'll add it here when it will be ready.

Augustin-FL:
@jim-p Ready for merge or for another review.

Augustin-FL:
> Thus there should be a second Authentication server multi-select for the "secondary authentication source", which should be pre-selected with what the user had previously chosen for the secondary authentication primary/secondary servers.

Hum....I would suggest instead the following feature : Captiveportal should remember the last server used to authenticate each user, and try this server firstly when re-authenticating.

This would avoid having two multi-select list (which might be confusing). Do you think it is a nice idea ?

jim-p:
That is not the point of the feature. It is not for a fallback or retry. The point is to have a completely separate set of login fields on the login page

auth_user + auth_pass => Primary Authentication Source
auth_user2 + auth_pass2 => Secondary Authentication Source

https://github.com/pfsense/pfsense/blob/master/src/usr/local/captiveportal/index.php#L195

Augustin-FL:
Ok, i understand now.

I'll update this PR but it will require me some time, as i will probably need to change many things to make it work.

Augustin-FL:
Hello @jim-p 
- I updated this PR as asked, it does handle radius context now
- The PHP warning issue is also fixed

Augustin-FL:
I apologize about the commit noises.
It should be good now, ready for another review or for merge.

Augustin-FL:
Yes, sure ! I completly agree, 66 commits is too much....it would be better to squash. 

@rbgarga how do you want to proceed? do you want to ```git merge --squash``` on your side? or do you want me to rebase on my side?

rbgarga:
@Augustin-FL it would be nice if you rebase it and do a forced push. We commit it first on an internal repo using scripts, not trivial as github's squash option :)

Augustin-FL:
@rbgarga it's done :)

jim-p:
Looks OK now with changes. I did hit an issue while testing but I backed out the patch and confirmed it was unrelated. Just a different/new CP regression on 2.4.4 (See https://redmine.pfsense.org/issues/8783)

At least from my basic tests here it all works. Auth servers imported OK and worked after upgrade. I could auth from RADIUS or LDAP. Primary and secondary auth worked.

Should be good enough to merge for wider testing.

PR #3641, opened 2017-03-11
Work around syntax error on filtering SKIP ICMP type (Bug #7372)

doktornotor:


phil-davis:
I did this a bit differently in #3642 (using more lines of code). It would be nice not to have to worry if "skip" happens to be a sub-string of some other icmp-type key word. And a function replace_element_in_list() might be handy for other things to call that have these kind of delimiter-separated lists.

doktornotor:
Shrug, it don't think there's any other skip anywhere near ICMP.

PR #3642, opened 2017-03-12
Fix #7372 Cannot filter ICMP Type SKIP

phil-davis:
Replace 'skip' with '39', the ICMP type code. pf handles the numbered code OK, but barfs when fed the pf keyword "skip".


phil-davis:
This should cherry-pick back to 2.3 easily.

PR #3643, opened 2017-03-13
Redmine #7381 Disable detail in alias popup

phil-davis:
Give the option for users to choose if they want to display alias details in the popups on Firewall Rules, NAT etc. This works like the other user preferences - there is an overall system setting in General Setup webConfigurator section, and the preference can be set per-user in User Manager/User Settings.


phil-davis:
The places that need to be changed for this illustrate that there is more refactoring that could be done - e.g. system.php system_usermanager.php and system_user_settings.php all have similar code for loading and storing the user preference settings. And there could be a list of what the keyword names of the preference settings are, which could be used everywhere to load/store them (e.g. in auth.inc etc) so that there are not long lists of "field names" being repeatedly set/stored...
But actually that applies to the whole code base, validation logic embedded in front-end UI display code, similar tests repeated in different places that have to be modified in-sync... - so I'm not about to refactor the whole of pfSense!

phil-davis:
and as a separate issue, the webConfigurator preferences section of General Setup is getting longer. Perhaps some day that should be made into a separate page/menu item System->Preferences.

phil-davis:
If this is the solution that is wanted, then it should cherry-pick back to 2.3.* easily.

PR #3645, opened 2017-03-13
Fix comment typo in firewall_rules.php

phil-davis:
while I notice it


phil-davis:
May as well cherry-pick this back to 2.3.* also just to keep the 2 branches as similar as reasonable.

PR #3647, opened 2017-03-13
Add reason to write_config() call for Firewall - Rules, fix includes

doktornotor:


sbeaver-netgate:
Are the "Fix broken includes"  commits required by the write_config updates? If not, let's separate them please

doktornotor:
Redone as https://github.com/pfsense/pfsense/pull/3663

PR #3650, opened 2017-03-17
No need to check for HTML in NAT descr

phil-davis:
This loop checking for htmlentities in any of the data has been in firewall_nat_edit.php etc for years. It prevents people from using some chars in the description. e.g. see forum:
https://forum.pfsense.org/index.php?topic=127350.0

This generic loop checking htmlentities is not in firewall_rules_edit.php - you can put strings in the description there like:
```
<b>bold "stuff" here</b>
```
that looks like you are trying to make it render bold in the Description and nothing bad happens. All those "special" characters are treated just as ordinary characters and are seen in the firewall rules table etc.

Other pages, like NAT 1to1 and NAT Out have slightly different code in their "htmlentities loop" which rubs out any ">" chars. Why is that done differently there?

This seems to be a classic case of breaking the DRY (Do not Repeat Yourself) rule (which is broken all the time in pfSense code) - whatever generic validation of input data is needed should be done in a common routine that can be called by every page.

The "fix" here is just a kludge to avoid doing the htmlentities check for the "descr" field. Actually somebody needs to decide what "htmlentities" protection is really needed, then implement it consistently across all pages.

### DIFF
--- a/src/usr/local/www/firewall_nat_1to1_edit.php
+++ b/src/usr/local/www/firewall_nat_1to1_edit.php
@@ -104,6 +104,10 @@ if ($_POST['save']) {
 	 *	cannot think he is slick and perform a XSS attack on the unwilling
 	 */
 	foreach ($_POST as $key => $value) {
+		if ($key == 'descr') {
+			continue;
+		}
+
 		$temp = str_replace(">", "", $value);
 		$newpost = htmlentities($temp);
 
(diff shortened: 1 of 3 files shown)


phil-davis:
This needs a design decision and then implementing a consistent "fix". Do not commit as-is.
The decided design would be expected to easily port back to 2.3.*

phil-davis:
Well this does work-around the direct issue that I managed to produce. But it does not resolve the general "policy" of what is this htmlentities() stuff really intended to achieve, and how should it be done consistently across GUI pages.

PR #3654, opened 2017-03-19
Traffic Graphs Widget whitespace

phil-davis:
No functional change, just code formatting ready for future changes.

PR #3655, opened 2017-03-19
Traffic graph widget filter checkboxes

phil-davis:
This refactors the Traffic Graphs Widget so that it has checkboxes for selecting the graphs to display, and so that the settings are stored in the "standard" format that is used by other widgets.


phil-davis:
This sits on top of #3654 , which could be merged first.

phil-davis:

PR #3652, opened 2017-03-19
Handle widgets having no items selected for display

phil-davis:
1) The existing code is a bit "surprising". If the user unchecks all items and presses save, then it behaves as if all items are checked (i.e. everything displays).

2) If there were some items checked for display (e.g. some WoL entries or...) and those items have since been deleted from the config, then in practice there will be no entries that get displayed.

For both cases it seems useful/reasonable that the widget displays the fact that there is nothing selected for display. Without doing this, users are likely to wonder why the widget is empty.

PR #3656, opened 2017-03-19
 Traffic Graphs Widget handle no items selected for display

phil-davis:
This is the equivalent of #3652 (which implements this logic for the other widgets).
A separate PR is needed here, because the Traffic Graphs Widget first needs to have a code base that works in a similar way to the other widgets. That code base is provided by the following other PRs that are waiting to be reviewed/merged:

Those need to be merged first in order, then this one will show just a few lines of diff.


phil-davis:

PR #3653, opened 2017-03-19
Redmine #7318 None button for widget filters

phil-davis:
When all items are selected, rename the "All" button to "None". When the button is clicked, unchecked all items. This allows a user to easily uncheck all but 1 or 2 items.

PR #3657, opened 2017-03-19
Redmine #7318 None button for Traffic Graphs Widget filter

phil-davis:
This implements the "None" button functionality on the Traffic Graphs widget, in the same way as proposed in #3653 
This is a separate PR because it needs all the underlying code in the Traffic Graphs Widget to work in a similar way to the other widgets. That is in the following other PRs which are waiting for review/merging:
#3656 Traffic Graphs Widget handle no items selected for display

Those need to be merged first in order, then this one will show just a few lines of diff.

### DIFF
--- a/src/etc/inc/globals.inc
+++ b/src/etc/inc/globals.inc
@@ -37,6 +37,7 @@ define('DMYPWD', "********");
 
 global $g;
 $g = array(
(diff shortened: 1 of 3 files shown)


phil-davis:

PR #3659, opened 2017-03-20
pkg_call_plugins put includefile and supported plugins in the normal …

PiBa-NL:
…config.xml so there is no need to parse the package xml for them. this improves performance significantly for several pages like such as 'ipsec overview' and 'openvpn server edit page' which use certificates and gatewaygroups which acquire some information from plugins.


rbgarga:

rbgarga:
It can cause a problem when system files are updated but package don't, so these items won't be registered in config.xml. I'll merge it but keep it 2.4 only. When people migrate from < 2.4 all packages will be reinstalled for sure

PR #3661, opened 2017-03-21
Gateways Widget backport to RELENG_2_3

phil-davis:
These Gateways widget changes to implement the "show" selection checkboxes did not get backported to RELENG_2_3 (not sure why they got missed). Other widgets have these checkboxes in RELENG_2_3 and even also in RELENG_2_3_3 released code.

Backporting this will make the Gateways Widget have the same "rev level" as other widgets in RELENG_2_3.

Then I can think about backporting the All/None button changes - those backport easily once Gateways widget is up-to-date here, and it would be nice that 2.3.4-DEVELOPMENT would have the "All/None" button thing.


phil-davis:
This could then be taken back to RELENG_2_3_3 to make the Gateways widget have consistent behavior to the other widgets in 2.3.3

PR #3662, opened 2017-03-21
Upgrade widget settings for multiple copies of widgets

phil-davis:
Upgrades any old widget preferences to the new format in the config.
Part of implementation of https://redmine.pfsense.org/issues/7182

### DIFF
--- a/src/etc/inc/upgrade_config.inc
+++ b/src/etc/inc/upgrade_config.inc
@@ -5192,4 +5192,105 @@ function upgrade_162_to_163() {
 		unset($config['widgets']['trafficgraphs']['shown']);
 	}
 }
+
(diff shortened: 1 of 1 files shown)


phil-davis:
@sbeaver-netgate I tried this with a config from a 2.3.4-development system that had various widget preferences in the old formats for both system-wide and user-specific dashboard. I believe it should deal with any dashboard widget preferences coming from old configs.

phil-davis:
The last remaining item on this is to make the Traffic Graphs widget be able to have multiple copies o the dashboard at once. That will require a bit more code changes, because Traffic Graphs widget JS has quite a few global data objects that it uses between functions that now need to be per-widget-copy.

PR #3664, opened 2017-03-21
Gateways Widget backport to RELENG_2_3_3

phil-davis:
This is the same code cherry-picked as in #3661 but now applied to RELENG_2_3_3. It makes the Gateways widget have the same checkbox show/hide behavior as the other widgets, as released in 2.3.3.

I don't know what the policy is about backporting into 2.3.3 - this is not strictly a fix, but it does "fix" a UI inconsistency, and I think it would be useful to have [if|when] a 2.3.3-p2 is needed.


sbeaver-netgate:
As we get closer to 2.4, I am being more strict about "bug fixes only" for 2.3, but this one is worth sneaking in :)

PR #3665, opened 2017-03-21
System Info Widget enable All button when disable firmware check is set

phil-davis:
If system firmware disablecheck is set, then the click event for the filter "All" button is also not included in the JS, so the "All" button is not effective.

The setTimeout for systemStatusGetUpdateStatus() can be moved up just below where it is declared and inside the existing "if" that conditionally includes that function - no need for it to be inside events.push

Then the PHP condition around events.push can be removed, so that the "All" button will always be effective.


phil-davis:
This fix did not seem to get back-ported.

PR #3667, opened 2017-03-22
Bump config rev for widget preference setting changes

phil-davis:
This should have been bumped along with https://github.com/pfsense/pfsense/commit/071d8a610047fe1994c9fa28e339fe93f4498c73


phil-davis:
@sbeaver-netgate are you around? This is needed to make sure the upgrade code is actually run.

PR #3668, opened 2017-03-23
Fix display of BIOS only on Sytem Info widget

phil-davis:
If I filter the System Info Widget to display only the BIOS item, it thinks that no rows are displayed and show the message "All System Information items are hidden." - see attached screen shot.

Make it know that an item has been selected for display.
bios-info-only

### DIFF
--- a/src/usr/local/www/widgets/widgets/system_information.widget.php
+++ b/src/usr/local/www/widgets/widgets/system_information.widget.php
@@ -161,6 +161,7 @@ $rows_displayed = false;
 <?php
 	endif;
 	if (!in_array('bios', $skipsysinfoitems)):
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: There is a little other "feature" for the items that can be selected for display, but information may not be available on the system concerned. e.g. select just the Temperature or Hardware Crypto item on system without temperature support or without crypto support. The row for the item does not show, but neither does the message "All System Information items are hidden." - because actually not all items are hidden, some are selected but they have no data to show.
Nothing breaks, you just get an empty System Information widget, which does not hurt anybody.

jim-p:
Good catch, thanks! It'll be merged momentarily.

PR #3669, opened 2017-03-24
Redmine #7422 remove bogus z at end of words

phil-davis:
These were my bad - was testing some translation stuff, putting an extra 'z' in places so the translation would fail. But I accidentally left it in the commits/PRs.
bridge edit: https://github.com/pfsense/pfsense/commit/398821c4aa05e3ae88731b8d9ce26e8d889d9fb8
openvpn client: https://github.com/pfsense/pfsense/commit/3e142087ca38917200fd01e2d8b604c7a5135b2b

Found them by a bit of regex
```
find /usr/local/www -type f -exec grep -H '\wz\W' {} \; > /tmp/zwww.txt
```
which finds words ending in "z". There are lots of OK ones (Hz, tgz and so on), but these couple are dodgy.


phil-davis:
As far as I can see, these are only in master - the changes are not back in RELENG_2_3* so no back-porting required.

phil-davis:
@sbeaver-netgate another bit of UI crud for when you wake up. Sorry!

PR #3670, opened 2017-03-26
Redmine #7428 Handle empty port alias

phil-davis:
pf rules do not cope with having a port alias that is empty. In this case, leave the rule out of the rule set and file_notice() to the dashboard etc, since a rule that matches a NULL list of ports would not match anything any way.

I did not want to write another "if is_port() or is_port_range()" test - so I refactored that to have some is_portorrange() ... functions available and condense all the occurrences of that test.

I did not want to repeat lines that set up $aliasnesting and $aliasaddrnesting for a call to filter_generate_nested_alias() - those are used internally by the function recursion, which should not require the first caller to set them up. So I refactored that also.

Then the 3rd commit does the enhance checking to avoid using an empty alias in a rule.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1124,6 +1124,11 @@ function is_portrange($portrange) {
 	return (count($ports) == 2 && is_port($ports[0]) && is_port($ports[1]));
 }
 
+/* returns true if $port is a valid TCP/UDP port number or range ("<port>:<port>") */
+function is_portorrange($port) {
+	return (is_port($port) || is_portrange($port));
+}
+
 /* returns true if $port is a valid port number or an alias thereof */
 function is_portoralias($port) {
 	global $config;
@@ -1140,6 +1145,11 @@ function is_portoralias($port) {
 	} else {
 		return is_port($port);
 	}
+}
+
+/* returns true if $port is a valid TCP/UDP port number or range ("<port>:<port>") or an alias thereof */
+function is_portorrangeoralias($port) {
+	return (is_portoralias($port) || is_portrange($port));
 }
 
 /* create ranges of sequential port numbers (200:215) and remove duplicates */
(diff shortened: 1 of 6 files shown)


phil-davis:
Agree - I did a bit of refactoring to prevent continuing to duplicate bits of code/logic in more places. And since the ruleset generation is right at the core of pfSense, it is good to make sure it does not break in some other way.

src/etc/inc/util.inc
```diff
@@ -1142,6 +1147,11 @@ function is_portoralias($port) {
 	}
 }
 
+/* returns true if $port is a valid TCP/UDP port number or range ("<port>:<port>") or an alias thereof */
+function is_portorrangeoralias($port) {
+	return (is_portoralias($port) || is_portrange($port));
```
rbgarga: I guess you meant is_portorrange() instead of is_portrange() here, no?
I'd also suggest to reverse this clause since is_portorrange() is cheaper than is_portoralias()

phil-davis: In the 2nd call it only needs to check if it is a "portrange", because the 1st part has already verified if it is a "portoralias". So this will return true if it is any of those 3 things.
The order of evaluation came directly from its use in firewall_nat_out_edit.php
I guess there could be a function is_portalias() that separates out the "is_alias" and "is_port" stuff in "is_port_or_alias". Then we can have:
```
function is_portorrangeoralias($port) {
	return (is_port($port) || is_portrange($port) || is_portalias($port));
}
```

phil-davis: @rbgarga Before I refactor more stuff here, and have more commits that may or may not be what you are thinking of, please confirm what you would like.
1) I could put the "_" in function names is_port_or_alias is_port_or_range is_port_or_range_or_alias ... and change all uses of those throughout the system.
2) I could refactor the existing is_portoralias() to split out the code block that effectively does is_portalias() and then make is_portoralias() be something like:
```
    return (is_port || is_port_alias);
```
What takes your fancy?

phil-davis:
See the code in #3677  for this stuff with optional refactoring. If you like it you could either just merge that, or merge this first and then #3677

PR #3673, opened 2017-03-26
get_stats, optimizations. dont call unused functions, call get_mbuf 1x instead of twice

PiBa-NL:
get_stats, optimizations.
-dont call unused interface statistics functions
-call get_mbuf only once (its netstat call is cpu intensive.)

### DIFF
--- a/src/usr/local/www/includes/functions.inc.php
+++ b/src/usr/local/www/includes/functions.inc.php
@@ -33,13 +33,9 @@ function get_stats() {
 	$stats['states'] = get_pfstate();
 	$stats['temp'] = get_temp();
 	$stats['datetime'] = update_date_time();
-	$stats['interfacestatistics'] = 0;//get_interfacestats();//unused
-	$stats['interfacestatus'] = 0;//get_interfacestatus();//unused
 	$stats['cpufreq'] = get_cpufreq();
 	$stats['load_average'] = get_load_average();
-	get_mbuf($mbuf, $mbufpercent);
-	$stats['mbuf'] = $mbuf;
-	$stats['mbufpercent'] = $mbufpercent;
+	get_mbuf($stats['mbuf'], $stats['mbufpercent']);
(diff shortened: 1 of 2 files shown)


src/usr/local/www/includes/functions.inc.php
```diff
@@ -33,12 +33,13 @@ function get_stats() {
 	$stats['states'] = get_pfstate();
 	$stats['temp'] = get_temp();
 	$stats['datetime'] = update_date_time();
-	$stats['interfacestatistics'] = get_interfacestats();
-	$stats['interfacestatus'] = get_interfacestatus();
+	$stats['interfacestatistics'] = 0;//get_interfacestats();//unused
+	$stats['interfacestatus'] = 0;//get_interfacestatus();//unused
```
rbgarga: If above items are not used anywhere, why not just remove them?

src/usr/local/www/includes/functions.inc.php
```diff
@@ -33,12 +33,13 @@ function get_stats() {
...
 	$stats['cpufreq'] = get_cpufreq();
 	$stats['load_average'] = get_load_average();
-	$stats['mbuf'] = get_mbuf();
-	$stats['mbufpercent'] = get_mbuf(true);
+	get_mbuf($mbuf, $mbufpercent);
```
jim-p: Why use the interim variables here? Just use the targets directly rather than using single-use variables. Unless that breaks in some way.

PiBa-NL: Was thinking it would be slightly more 'readable' but changed it now, i'm okay with it either way.

src/usr/local/www/includes/functions.inc.php
PiBa-NL: Removed now, including the functions they call, and changed the index numbers that changed by removing them from the array.

PR #3675, opened 2017-03-26
ipsec webgui, move building the $iflabels array outside if the loop

PiBa-NL:


src/usr/local/www/vpn_ipsec.php
```diff
@@ -289,25 +308,6 @@
 						<td>
 <?php
 			if ($ph1ent['interface']) {
```
rbgarga: While you are changing it please also check if `$iflabels[$ph1ent['interface']]` is defined

PiBa-NL:
@rbgarga how can i remove that 'changes requested' label? It should have been back on 'review pending' for over a week..

src/usr/local/www/vpn_ipsec.php
PiBa-NL: done

PR #3677, opened 2017-03-27
Handle empty port alias refactor

phil-davis:
This sits on top of #3670 and refactors to:
1) Add "_" characters to various is_port_or* functions.
2) Adds an is_portalias function, refactors is_port_or_alias to use it, refactors is_port_or_range_or_alias to call each of its 3 "parts" to get the answer.


rbgarga:
Note: There are 2 ports that uses is_portoralias(). haproxy and haproxy-devel

PR #3678, opened 2017-03-28
Redmine #7435 Fix edit IP address in host override

phil-davis:
It would "forget" $id that was passed in in the initial request. Then things would go wrong through the validation checks because $_POST if "id" would be empty (although the $_GET has the id from the URL line).

Same problem for Forwarder and Resolver (don't you love copied code - I can fix 2 bugs in only a little longer than fixing 1 - sarcasm).

I made different input error messages that will tell the user in a bit more detail what the problem is.

### DIFF
--- a/src/usr/local/www/services_captiveportal_vouchers_edit.php
+++ b/src/usr/local/www/services_captiveportal_vouchers_edit.php
@@ -69,6 +69,7 @@ if (isset($id) && $a_roll[$id]) {
 	$pconfig['count'] = $a_roll[$id]['count'];
 	$pconfig['minutes'] = $a_roll[$id]['minutes'];
 	$pconfig['descr'] = $a_roll[$id]['descr'];
+	$pconfig['id'] = $id;
 }
 
 $maxnumber = (1<<$config['voucher'][$cpzone]['rollbits']) -1;	// Highest Roll#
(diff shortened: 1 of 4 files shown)


phil-davis:
Note: In RELENG_2_3 the codechecks both $_GET and $_POST of "id". Things work out there, because actually $_GET "id" has an id in it. $_POST "id" is empty, but the code has a check for that, and leaves the value from $_GET.

In master it condenses the checks into a single check of $_REQUEST id. $_REQUEST gets assembled by PHP from $_GET values then overlayed with $_POST values. So the code in master sees the "" value of $_POST "id".

So this kind of bug is brought out by the changes in master to not use $_GET directly.

Actually the fixes here are applicable to RELENG_2_3 (and I guess RELENG_2_3_3) to make things be working "correctly" underneath, and it will give users the enhanced input error messages. So I would back-port this.

phil-davis:
This is a classic case for having (automated) test cases that can try all sorts of add/edit/delete sequences and find these sort of underlying bugs that "pop out" when other changes are made.

phil-davis:
Forwarder Domain Override - Edit - edit an entry, press Save, you get a duplicate (new) entry.
Captive Portal Voucher Roll - Edit - press Save, you get an input error message about the roll already exists. Press Save again and it saves.
Both of these are manifestations of this "id" get/post thing also.
Neither cause a user-visible problem in RELENG_2_3

phil-davis:
After finding these other cases, I realized that the way it was done elsewhere was slightly different.
#3679 has code that conforms to the rest of the system.
Closing this one in favor of #3679

PR #3679, opened 2017-03-28
Redmine #7435 Fix edit problems due to POST id code

phil-davis:
This code would "forget" the proper value of $id and so there were effects when editing existing entries:
1) Captive Portal Vouchers, edit a roll, press save. You get told that the roll already exists.
2) DNS Forwarder, edit a host override, press save. You get told that the host override already exists.
3) DNS Resolver, edit a host override, press save. You get told that the host override already exists.
4) DNS Forwarder, edit a domain override, press save. You get a duplicate entry.

For when the user edits the IP address in a host override, I made the input error messages a bit more informative if they change the IP address V4<->V6 and there is already an entry for the new address family.

### DIFF
--- a/src/usr/local/www/services_captiveportal_vouchers_edit.php
+++ b/src/usr/local/www/services_captiveportal_vouchers_edit.php
@@ -212,7 +212,7 @@ if (isset($id) && $a_roll[$id]) {
 		'id',
 		null,
 		'hidden',
-		$pconfig['id']
+		$id
 	));
 }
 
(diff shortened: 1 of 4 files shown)


phil-davis:
These are all the places that had the string:
```
$pconfig['id']
```
in a bad way

phil-davis:
These problems all are apparent in 2.4 master due to the changes related to $_GET, $_POST, $_REQUEST. But actually the problem has been latent in the code in 2.3.* so the fixes could (should?) be back-ported there also.

PR #3683, opened 2017-03-29
add ability to specify multiple IPs in dhcprejectfrom

luckman212:
This PR adds the ability to ignore DHCP offers from multiple servers.
- Forum thread: https://forum.pfsense.org/index.php?topic=124046.msg705100#msg705100
- related dhclient source:
https://github.com/pfsense/FreeBSD-src/blob/devel/sbin/dhclient/clparse.c#L945

**changed files:**
/usr/local/www/interfaces.php
/etc/inc/interfaces.inc

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -38,6 +38,27 @@ function interfaces_bring_up($interface) {
 }
 
 /*
+ * Validate comma-separated list of IPv4 addresses
+ */
+function validate_ipv4_list($value) {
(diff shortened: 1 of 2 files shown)


luckman212:
@jim-p Sorry about that miss! I had removed the is_ipaddrv4() from my local interfaces.inc at some point during testing which is why I didn't catch this. Updated now.

jim-p:
Looks good now, the value makes it into the config and dhclient is happy with the directive contents. Thanks!

PR #3684, opened 2017-03-29
Add hint to web interface that DynDNS will automatically try to fetch…

bonki:
… the public IP in case the selected interface's IP is private.

See https://github.com/pfsense/pfsense/blob/master/src/etc/inc/dyndns.class#L1684.
This doesn't seem to be documented?


bonki:
I've just signed the CLA, is there a delay before the status correctly shows on Github?

PR #3685, opened 2017-03-30
Add user option to sort interface names

phil-davis:
This has been talked abut a few times in the forum and other places.
https://forum.pfsense.org/index.php?topic=128043.0

Add the option instead of just talking about it.


phil-davis:
Note: The System->General Setup, webConfigurator section is getting quite long now. Let me know if you think it should be separated out into a separate GUI page, and I can do that after this is merged.
(e.g. System->webConfigurator Settings, webConfigurator Preferences, webGui Settings, webGUI Preferences or just "Preferences"?)

PR #3688, opened 2017-04-02
Add local directive to nsupdate file. Fixes #7446.

derelict-pf:


jim-p:
PR was merged but it didn't get marked as merged. See https://redmine.pfsense.org/issues/7446

PR #3690, opened 2017-04-09
phpsessionmanager, csrf should use phpsession_end before exit to avoid syslog warning about a open session

PiBa-NL:
phpsessionmanager, csrf should use phpsession_end before exit to avoid syslog warning about a open session


rbgarga:
Manually merged, thanks!

PR #3692, opened 2017-04-15
Add ability to ping from ARP listing page

sergiocarlosmorales:
Relates to: https://redmine.pfsense.org/issues/7442

### DIFF
--- a/src/usr/local/www/diag_arp.php
+++ b/src/usr/local/www/diag_arp.php
@@ -30,10 +30,8 @@
 ##|*MATCH=diag_arp.php*
 ##|-PRIV
 
-@ini_set('zlib.output_compression', 0);
-@ini_set('implicit_flush', 1);
-
 require_once("guiconfig.inc");
+require_once("diag_ping_shared.php");
 
 // delete arp entry
 if (isset($_POST['deleteentry'])) {
@@ -275,19 +273,7 @@ if ($savemsg) {
 }
 ?>
 
-<!-- On modern hardware the table will load so fast you may never see this! -->
-<div id="loading">
-	<?= gettext(" Loading, please wait...")?>
-</div>
-
 <?php
-
-// Flush buffers out to client so that they see Loading, please wait....
-for ($i = 0; $i < ob_get_level(); $i++) {
-	ob_end_flush();
-}
-
-ob_implicit_flush(1);
 
 // Resolve hostnames and replace Z_ with "".  The intention
 // is to sort the list by hostnames, alpha and then the non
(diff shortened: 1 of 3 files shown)


src/usr/local/www/diag_arp.php
```diff
@@ -30,10 +30,8 @@
 ##|*MATCH=diag_arp.php*
 ##|-PRIV
 
-@ini_set('zlib.output_compression', 0);
```
sergiocarlosmorales: These removals are intentional.
They were added 8 years ago to output a "loading" dialog to make the page load seem speedy. Unfortunately this messes up with the CSRF magic output handler which was added years later.

Comments near this "loading" dialog (line 278) mention that on modern hardware this should not be a problem. 8 years later I don't think this is a concern anymore.

If page load time is still a concern here, a proper solution can be implemented like loading the actual ARP table via an async AJAX call, not by playing around with output handlers.

src/usr/local/www/diag_ping.php
```diff
@@ -50,8 +49,8 @@
 	$reqdfieldsn = array(gettext("Host"), gettext("Count"));
 	do_input_validation($_REQUEST, $reqdfields, $reqdfieldsn, $input_errors);
 
-	if (($_REQUEST['count'] < 1) || ($_REQUEST['count'] > MAX_COUNT)) {
-		$input_errors[] = sprintf(gettext("Count must be between 1 and %s"), MAX_COUNT);
+	if (($_REQUEST['count'] < 1) || ($_REQUEST['count'] > MAX_PING_COUNT)) {
```
sergiocarlosmorales: In the context of the ping page, MAX_ & DEFAULT_COUNT constants may be explicit enough. But since now they can be included from another page, renamed them to clearly state they are related to ping actions and potentially avoid collisions.

src/usr/local/www/diag_ping_shared.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/*
+ * diag_ping.php
```
rbgarga: Filename is wrong

src/usr/local/www/diag_ping_shared.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2004-2017 Rubicon Communications, LLC (Netgate)
+ * Copyright (c) 2003-2005 Bob Zoller (bob@kludgebox.com)
```
rbgarga: This file only contain bootstrap forms, I don't believe there are any content added by Bob here

src/usr/local/www/diag_ping_shared.php
```diff
@@ -0,0 +1,76 @@
...
+ * All rights reserved.
+ *
+ * originally based on m0n0wall (http://m0n0.ch/wall)
+ * Copyright (c) 2003-2004 Manuel Kasper <mk@neon1.net>.
+ * All rights reserved.
```
rbgarga: I don't think there are m0n0wall code here, only bootstrap forms

src/usr/local/www/diag_arp.php
rbgarga: I agree

rbgarga:
I liked @jim-p 's idea

sergiocarlosmorales:
Alright, will close this PR and explore @jim-p's suggested approach in a separate one.
Thanks for the feedback guys.

PR #3694, opened 2017-04-16
Fix #7394 rename alias correctly on edit

phil-davis:
If you change an alias name, but have other input errors on the initial attempt to save, then fix the input errors and save again, the alias name did not get updated in Rules/NAT/etc where it was used.

Refactor update_alias_names_upon_change also (which I was looking at before seeing this bug), to get rid of all that repeated code. Where it is used in diag_backup when restoring a m0n0wall config, it does not hurt to try to rename any port aliases etc also, in case somehow there were port aliases in an old m0n0wall config.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -2078,6 +2078,37 @@ function update_progress_bar($percent, $first_time) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


phil-davis:
Note: Redmine 7473 is a duplicate of 7394. This PR fixes both.

PR #3696, opened 2017-04-16
Add ability to delete NDP entry

sergiocarlosmorales:


src/usr/local/www/diag_ndp.php
```diff
@@ -33,10 +33,28 @@
...
-
+define('NDP_BINARY_PATH', '/usr/sbin/ndp');
 require_once("guiconfig.inc");
 
-exec("/usr/sbin/ndp -na", $rawdata);
+// Delete ndp entry.
+if (isset($_POST['deleteentry'])) {
+	$ip = $_POST['deleteentry'];
+	if (is_ipaddrv6($ip)) {
+		$commandReturnValue = mwexec(NDP_BINARY_PATH . " -d " . $ip, true);
```
jim-p: The IPv6 address value must be validated and escaped. If it's not a valid IPv6 address, the command should not be run at all. And even when it is, the value must be escaped using `escapeshellarg()`.

sergiocarlosmorales:
@jim-p Thanks for your feedback, the command is only executed if the IP is deemed valid by is_ipaddrv6() (which rejects emptry strings, null and other dummy values).

I've added escaping to the parameter when sent to the command, let me know if there is additional validation to be added.

jim-p:
Looks OK now, tested/works.

PR #3700, opened 2017-04-17
Fix #7474 Save interface and gateway together

phil-davis:
This refactors a bunch of code to take the gateway validation and saving code out into functions that can be called from both system_gateways_edit.php and interfaces.php
This allows interfaces.php to gather the interface parameters and any gateway parameters for a gateway(s) that the user wants to add, then validate both the interface and gateway parameters.
If there are validation errors with either, then the page is redisplayed showing the errors.
If both interface and gateway pass validation then they are both committed.

This ensures that the added gateway falls in the interface subnet, even when the users changes/adds the interface IP at the same time as adding the gateway.


phil-davis:
Note1: A bit of cleanup can be done to remove some 'isAjax' tests that are no longer relevant (there is no longer cross-posting from interfaces.php into system_gateways_edit.php)
Note2: If there are input errors, the gateway parameters to add are lost from the refreshed screen - I will have a look at how best to get back the POSTed values into the refreshed page, so the user does not have to re-enter them just because they made some dumb mistake in some other field.

src/etc/inc/gwlb.inc
```diff
@@ -864,6 +869,15 @@ function return_gateways_array($disabled = false, $localhost = false, $inactive
 		$gateways_arr['Null4'] = $gwlo4;
 		$gateways_arr['Null6'] = $gwlo6;
 	}
+
+	if ($integer_index) {
+		$gateways_arr_temp = array();
```
jim-p: This whole block could be reduced to: `$gateways_arr = array_values($gateways_arr);`

Or maybe just use that where it's needed rather than adding it to the function?

phil-davis: Yes, I looked at that and thought so. But then I got worried about "was there some reason it was coded like this?" and so I just refactored-out the repeated code into the function, so that at least it only appeared once. (it wasn't my objective at the time to "fixup" this)
The subsequent use of $gateways_arr assumes that the index into $gateways_arr of gateways that do exist in the config is the same index they have in the config. The way that return_gateways_array() builds its list means that the array entries should be added in the same order as in the config.
That order must not be messed up by re-indexing the keys.
I will do a few tests/checks to make sure that array_values() will not cause a problem.

phil-davis: It is not so fussy as I thought - system_gateways and system_gateways_edit use the same array returned by return_gateways_array() to index the 'id' parameter to know which gateway to edit. As long as return_gateways_array() is consistent, all is good anyway.
system_gateways_edit finds the "real" gateway index to update in the config by:
```
$realid = $a_gateways[$id]['attribute'];
```
This looks not as critical as I had thought it might be.

phil-davis:
Using array_values() works fine.

PR #3702, opened 2017-04-20
added tlsauth keydir options to openvpn client

vpiserchia:


pfSense-git-updates:

src/etc/inc/openvpn.inc
```diff
@@ -447,6 +447,16 @@ function openvpn_get_digestlist() {
 	return $digests;
 }
 
+function openvpn_get_keydirlist() {
+	$keydirs = array(
+		'-1' => gettext('Use default keydir direction'),
```
jim-p: This value should probably be unset/empty `''` because otherwise upgraded configurations will never match any of these cases.

src/etc/inc/openvpn.inc
```diff
@@ -1096,8 +1106,32 @@ function openvpn_reconfigure($mode, $settings) {
 					$tls_directive = "tls-auth";
 					if ($mode == "server") {
 						$tlsopt = 0;
+						//switch($settings['tlsauth_keydir']){
```
jim-p: Why add all of this commented out? The setting should also be valid for servers, provided that it is set properly compared to the client (one side set to "1", the other "0", or both set to omit). It would need code on the server page, of course.

src/etc/inc/openvpn.inc
```diff
@@ -1096,8 +1106,32 @@ function openvpn_reconfigure($mode, $settings) {
...
+								//$tlsopt = 1;
+								//break;
+							//case '2':
+								//$tlsopt = '';
+								//break;
+						//}
 					} else {
 						$tlsopt = 1;
+						switch($settings['tlsauth_keydir']){
+							case '-1':
```
jim-p: See previous note, this should be empty/unset

src/etc/inc/openvpn.inc
```diff
@@ -1096,8 +1106,32 @@ function openvpn_reconfigure($mode, $settings) {
...
+							case '1':
+								$tlsopt = 1;
+								break;
+							case '0':
+								$tlsopt = 0;
+								break;
+							case '2':
+								$tlsopt = '';
+								break;
+						}
```
jim-p: Switch needs a default case, which could also be used to apply the previous/existing default behavior.

src/usr/local/www/vpn_openvpn_client.php
```diff
@@ -150,6 +151,8 @@
 		} else {
 			$pconfig['shared_key'] = base64_decode($a_client[$id]['shared_key']);
 		}
+		// OpenVPN Defaults to -1 if unset
+		$pconfig['tlsauth_keydir'] = !(empty($a_client[$id]['tlsauth_keydir'])) ? $a_client[$id]['tlsauth_keydir'] : -1;
```
jim-p: `empty()` will return true when the value is `0` which will be a problem since `0` is one of the valid settings for this variable.

src/usr/local/www/vpn_openvpn_client.php
```diff
@@ -638,6 +642,13 @@
 	    'This key is used to sign control channel packets with an HMAC signature for authentication when establishing the tunnel. ', '<br/>');
 
 	$section->addInput(new Form_Select(
+		'tlsauth_keydir',
+		'*TLS keydir direction',
+		$pconfig['tlsauth_keydir'],
+		openvpn_get_keydirlist()
+	))->setHelp('Leave this unchecked if you are not sure');
```
jim-p: It's not a checkbox so the note doesn't make sense here. Perhaps something like this instead:

"The TLS Key Direction must be set to complementary values on the client and server. For example, if the server is set to 0, the client must be set to 1. Both may be set to omit the direction, in which case the TLS Key will be used bidirectionally."

src/usr/local/www/vpn_openvpn_client.php
```diff
@@ -638,6 +642,13 @@
 	    'This key is used to sign control channel packets with an HMAC signature for authentication when establishing the tunnel. ', '<br/>');
 
 	$section->addInput(new Form_Select(
+		'tlsauth_keydir',
+		'*TLS keydir direction',
```
jim-p: Use "TLS Key Direction" here, "keydir direction" reads awkwardly.

src/etc/inc/openvpn.inc
jim-p: The description for this setting should be "Use default direction".

src/etc/inc/openvpn.inc
```diff
@@ -447,6 +447,16 @@ function openvpn_get_digestlist() {
 	return $digests;
 }
 
+function openvpn_get_keydirlist() {
+	$keydirs = array(
+		'-1' => gettext('Use default keydir direction'),
+		'2' => gettext('No keydir direction'),
```
jim-p: The description for this setting should be "Both Directions". This option should also be last, not one of the first options.

src/etc/inc/openvpn.inc
```diff
@@ -447,6 +447,16 @@ function openvpn_get_digestlist() {
...
+		'0'  => gettext('Use keydir direction 0'),
```
jim-p: The description for this setting should be "Direction 0".

src/etc/inc/openvpn.inc
```diff
@@ -447,6 +447,16 @@ function openvpn_get_digestlist() {
...
+		'1'  => gettext('Use keydir direction 1'),
```
jim-p: The description for this setting should be "Direction 1".

src/etc/inc/openvpn.inc
vpiserchia: an other is to remove this and for the client use the direction 1 as default (as it is now)
Eventually for the server it will be defaulted to 0. So we can remove this completely ?

vpiserchia: see above

src/etc/inc/openvpn.inc
vpiserchia: +1

src/etc/inc/openvpn.inc
vpiserchia: +1

src/etc/inc/openvpn.inc
vpiserchia: +1 in case of client this is the default as keeps backwards compatibility ?

src/etc/inc/openvpn.inc
vpiserchia: yes they are there for the server side, but the server page is ongoing. Also there should be a way to correlate client/server choosen parameter to always use the right one.
Of course this is not possible for those clients you do not have control the corresponding server' otherwise one can do the correlation.

jim-p: I'd prefer to see the server page implemented completely before this is imported rather than doing it halfway.

We can't automatically correlate client/server, that will have to be left to the end user to handle. I wouldn't worry about that case at all. So long as the defaults preserve the existing behavior it should be safe.

src/etc/inc/openvpn.inc
jim-p: I'd rather have an option to preserve the default behavior rather than assuming it in the GUI. It is safer than forcing everything to have an explicit option. We can't risk breaking any existing configurations.

src/etc/inc/openvpn.inc
jim-p: See above, I'd keep the default option and leave these are simply "0" and "1". We may default them a certain way but the other way is also valid so long as it is set in a complementary way, so I wouldn't change the text to indicate anything other than 0/1 since it is ultimate up to the user.

rbgarga:
@vpiserchia Before this pull request can be accepted you must first sign a CLA as described at https://www.pfsense.org/about-pfsense/#cla. Please read for more details

sbeaver-netgate:
Requested changes not implemented in 7 months. PLease resubmit if still desired.

PR #3703, opened 2017-04-21
Split System Widget in default dashboard

phil-davis:
This is a suggested split of the system widget on the default dashboard. The name/version... are in column 1. The "performance" items are in column2. Users can customize this anyway, but this might be a useful starting point to give on the default dashboard.
Open for discussion. See forum https://forum.pfsense.org/index.php?topic=129040.0

### DIFF
--- a/src/conf.default/config.xml
+++ b/src/conf.default/config.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <pfsense>
-	<version>15.4</version>
+	<version>16.4</version>
 	<lastchange/>
 	<system>
 		<optimization>normal</optimization>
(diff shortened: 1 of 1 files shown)


rbgarga:
Just a note, before change config version from 15.4 to 16.4 you need to verify if all changes made since 15.5 are applied in this config because config upgrade functions will not be called

phil-davis:
I just checked, all the things converted from 15.4 up to 16.3 are not in the default config, so nothing to change for that. 16.4 converts the dashboard widget layout settings to the "new multi-widget on the dashboard" capability, which happens to be exactly the config section that is updated here anyway, effectively converting it. So that is why the default config version needs to be bumped to 16.4.

So the config version change looks fine to me.

phil-davis:
After #3707 it will also be possible to give each copy of the System Information widget a custom title, e.g. "System Information" and "System Performance" or...

phil-davis:
PR for "Allow customize of dashboard widget names" is now #3710 
This "Split System Widget in default dashboard" can wait until #3710 is done, then take advantage of it.

rbgarga:
#3710 is now merged @phil-davis. Are you going to rework this change?

phil-davis:
Closing in favor of #3727

PR #3704, opened 2017-04-23
Add expiration field to NDP listing

sergiocarlosmorales:
Now that we have the ability to delete an NDP entry, add a field to show when they will expire on their own.

### DIFF
--- a/src/usr/local/www/diag_ndp.php
+++ b/src/usr/local/www/diag_ndp.php
@@ -65,9 +65,15 @@ foreach ($ifdescrs as $key =>$interface) {
 	$hwif[$config['interfaces'][$key]['if']] = $interface;
 }
 
-/* Array ( [0] => Neighbor [1] => Linklayer [2] => Address
-[3] => Netif [4] => Expire [5] => S
-[6] => Flags ) */
+/*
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_ndp.php
```diff
@@ -65,9 +65,15 @@
 	$hwif[$config['interfaces'][$key]['if']] = $interface;
 }
 
-/* Array ( [0] => Neighbor [1] => Linklayer [2] => Address
```
sergiocarlosmorales: This key map was incorrect for keys 1 & 2.

src/usr/local/www/diag_ndp.php
```diff
@@ -128,10 +135,11 @@ function _getHostName($mac, $ip) {
 	<table class="table table-striped table-condensed table-hover sortable-theme-bootstrap" data-sortable>
 		<thead>
 			<tr>
-				<th><?= gettext("IPv6 address"); ?></th>
```
sergiocarlosmorales: Spacing of this block was inconsistent with the rest of the page (spaces after ?= & trailing '; ')

PR #3705, opened 2017-04-25
Fix #7493 only check interface IP if static

phil-davis:
"ipaddr" is only saved to the interface config if "staticv4" type is selected.
"ipaddrv6" is only saved to the interface config if "staticv6" type is selected.
So these fields need only be checked for validity when "staticv4"/"staticv6" type is selected.

The report in the Redmine issue has a more complex situation in which this bug is apparent, but actually you can easily see the problem:
1) Edit an interface with a static IP
2) Change the IP Address field to have an invalid IP address in that field
2) Change the "type" to "none" or "DHCP" (or any other "type")
3) Press Save

An error is given saying that the IP address must be valid.
Actually that is a "dumb" error to report, because the invalid IP address string is about to be discarded anyhow.


phil-davis:
Note: This is an issue in all versions. Could be cherry-picked to RELENG_2_3 and RELENG_2_3_3

PR #3707, opened 2017-04-25
Allow customize of dashboard widget names

phil-davis:
Forum: https://forum.pfsense.org/index.php?topic=129040.msg713499#msg713499

For all widgets that already have a "wrench tool" (i.e. they already have some custom settings or filter), this adds a setting to specify the widget title. So if you have a widget on the dashboard twice (presumably with different settings/filter selections), you will have the ability to give each widget a custom title.

### DIFF
--- a/src/etc/inc/xmlparse.inc
+++ b/src/etc/inc/xmlparse.inc
@@ -248,7 +248,8 @@ function dump_xml_config_sub($arr, $indent) {
 						    (substr($ent, 0, 9) == "ldap_bind") ||
 						    (substr($ent, 0, 11) == "ldap_basedn") ||
 						    (substr($ent, 0, 18) == "ldap_authcn") ||
-						    (substr($ent, 0, 19) == "ldap_extended_query")) {
+						    (substr($ent, 0, 19) == "ldap_extended_query") ||
+							($ent == "customwidgettitle")) {
 							$xmlconfig .= "<$ent><![CDATA[" . htmlentities($cval) . "]]></$ent>\n";
 						} else {
 							$xmlconfig .= "<$ent>" . htmlentities($cval) . "</$ent>\n";
@@ -280,6 +281,7 @@ function dump_xml_config_sub($arr, $indent) {
 				    (substr($ent, 0, 11) == "ldap_basedn") ||
 				    (substr($ent, 0, 18) == "ldap_authcn") ||
 				    (substr($ent, 0, 19) == "ldap_extended_query") ||
+				    ($ent == "customwidgettitle") ||
 				    (substr($ent, 0, 5) == "text")) {
 					$xmlconfig .= "<$ent><![CDATA[" . htmlentities($val) . "]]></$ent>\n";
 				} else {
--- a/src/etc/inc/xmlparse.inc
+++ b/src/etc/inc/xmlparse.inc
@@ -248,8 +248,7 @@ function dump_xml_config_sub($arr, $indent) {
 						    (substr($ent, 0, 9) == "ldap_bind") ||
 						    (substr($ent, 0, 11) == "ldap_basedn") ||
 						    (substr($ent, 0, 18) == "ldap_authcn") ||
-						    (substr($ent, 0, 19) == "ldap_extended_query") ||
-							($ent == "customwidgettitle")) {
+						    (substr($ent, 0, 19) == "ldap_extended_query")) {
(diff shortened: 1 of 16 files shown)


phil-davis:
Annoyed at myself for leaving that in there - fixed.

src/usr/local/www/guiconfig.inc
```diff
@@ -577,6 +577,25 @@ function genhtmltitle($title, $links=true) {
 	return $bc;
 }
 
+function gen_customwidgettitle_div($widgettitle) {
+	$divstr = '<div class="form-group">';
+	$divstr .= '  <label for="customwidgettitle" class="col-sm-4 control-label">' . gettext('Widget title'). '</label>';
+	$divstr .= '  <div class="col-sm-4">';
+	$divstr .= '    <input type="text" name="customwidgettitle" id="customwidgettitle" value="'. $widgettitle . '" class="form-control" />';
```
jim-p: Needs encoding (like `htmlspecialchars()`) around input sent back to the user in this way. You have that on the actual title output but this field value needs it for the form, too.

src/usr/local/www/guiconfig.inc
```diff
@@ -577,6 +577,25 @@ function genhtmltitle($title, $links=true) {
...
+	$divstr .= '  </div>';
+	$divstr .= '</div>';
+
+	return $divstr;
+}
+
+function set_customwidgettitle(& $user_settings) {
+	if ($_POST['customwidgettitle']) {
+		$user_settings['widgets'][$_POST['widgetkey']]['customwidgettitle'] = trim($_POST['customwidgettitle']);
```
jim-p: Consider limiting what they can set as the title, perhaps just letters/numbers/basic punctuation. Either filter/strip it out or generate an input validation failure for it. If the field doesn't get CDATA protected in XML, then special/international characters could also cause problems.

phil-davis: For users in other languages they will want to be able to use a range of Unicode chars/symbols... so I made the field be CDATA encoded.
See #3708 for some different issues with the CDATA encoding field name matching.

src/usr/local/www/guiconfig.inc
phil-davis: If I do that, then it keeps "amplifying" the special chars on each edit.
e.g. If I put in some title text that makes an attempt to put the widget title in italics:
```
abc <i>zxc नेपाल </i>
```
It correctly escapes the attempted HTML tags, and the title is rendered literally as I typed it, showing the "<" "i" ">" as literal chars - good.
Then when I try to edit the title again, it fills the input area with:
```
abc &lt;i&gt;zxc नेपाल &lt;/i&gt;
```
and if I save without touching anything, I get that literal string as the new widget title.
Then edit/save again "expands/escapes" another round:
```
abc &amp;lt;i&amp;gt;zxc नेपाल &amp;lt;/i&amp;gt;
```
and so on ad-infinitum.

jim-p: Testing it, it seems to be OK without the change. Strange though as that has been a problem in the past. Must be getting filtered somewhere along the way already.

src/etc/inc/xmlparse.inc
```diff
@@ -248,7 +248,8 @@ function dump_xml_config_sub($arr, $indent) {
 						    (substr($ent, 0, 9) == "ldap_bind") ||
 						    (substr($ent, 0, 11) == "ldap_basedn") ||
 						    (substr($ent, 0, 18) == "ldap_authcn") ||
-						    (substr($ent, 0, 19) == "ldap_extended_query")) {
+						    (substr($ent, 0, 19) == "ldap_extended_query") ||
+							($ent == "customwidgettitle")) {
```
jim-p: Rather than adding a new key, why not use "descr" for this field name in the config, which already gets protected?

phil-davis: Now that I am aware of this stuff embedded in the XML parser, that seems a good idea. The "descr" key will be a sub-key of each individual widget anyway, so the "descr" key name will seem reasonable for being put in the "widget title".

phil-davis:
All comments addressed. Seems good to go. I can make a new PR with a single commit if you like - saves having all these little commits end up merged in master.

phil-davis:
See #3710 for a PR for this with just a single commit.

PR #3708, opened 2017-04-27
Fix comparisons for CDATA tags in config

phil-davis:
Some length numbers here do not match the strings they are comparing with. That looks very odd.
Note that:
```
substr($ent, 0, 5) == "text"
```
will return true when $ent is "text". So actually this "works". But it returns false if $ent is "text1" "texta" etc.
So it depends what the intention was.

If it was intended to match just "text" then I could be just:
```
$ent == "text"
```
If it was intended to match "text" "text1" "texta" etc, then it needs to be fixed like I suggest here.

### DIFF
--- a/src/etc/inc/xmlparse.inc
+++ b/src/etc/inc/xmlparse.inc
@@ -247,7 +247,7 @@ function dump_xml_config_sub($arr, $indent) {
 						    (substr($ent, 0, 9) == "ldap_attr") ||
 						    (substr($ent, 0, 9) == "ldap_bind") ||
 						    (substr($ent, 0, 11) == "ldap_basedn") ||
-						    (substr($ent, 0, 18) == "ldap_authcn") ||
+						    (substr($ent, 0, 11) == "ldap_authcn") ||
 						    (substr($ent, 0, 19) == "ldap_extended_query")) {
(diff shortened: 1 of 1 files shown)


phil-davis:
@jim-p stuff I noticed while looking at how the config save works out when to put CDATA.

Most of the substr() comparisons here seem odd/unneeded. e.g. ldap_authcn is only ever used as-is, there are no uses of ldap_authcn1 ldap_authcnzzz etc. So a simple test ($ent == "ldap_authcn") would suffice. So what is going on there?

jim-p:
That's a good question. Appears to just be a typo as far as I can tell in either case.

phil-davis:
Note: making this change will mean that fields with names like "textbox" "textarea" "text1" etc will start to e CDATA escaped in config.xml.
I searched the code, and strings like "textarea" and "texbox" appear as "data/input types" - which is fine - but not as application field names. So that seems fine - there will be no real change to behaviour.
Similarly there are no "ldap_authcn*" field names - just "ldap_authcn" appears in the code.

jim-p:
That should be safe. The only danger would be if something used to be CDATA escaped and then wasn't. Fields will just have the CDATA protection added the next time the config is written out.

PR #3709, opened 2017-04-27
Another wat to match tags for CDATA

phil-davis:
I suspect that we only want to match exact field names like "ldap_authcn" and so on anyway, so this code could remove the substr() stuff.
I left the "descr" and "detail" clauses, because maybe it is intended to match "descr1" "detail01" and stuff like that also.


phil-davis:
See #3708 for minimal fix to substr() counts.
This is a possible alternate way to go to refactor and fix.

jim-p:
I believe the original idea was to protect any strings starting with these names, rather than the names exactly. To reduce the length of the list overall. Though I'm not sure how much we leverage that in practice, I'd rather not risk breaking something that has been working. Fixing the individual tests seems like the better fix to me.

phil-davis:
Yes, I agree. Making a minimal change here for now is the safest thing to do. Then refactor/improve later.

PR #3711, opened 2017-04-27
ipsec, allow configuration of multiple ike phase1 encryption ciphers

PiBa-NL:
ipsec, allow configuration of multiple ike phase1 encryption ciphers (algo/bits/hash/dh)

this is useful for mobile users that need to connect with different operating systems. This way there is no need to find a single commonly supported weaker cipher.


rbgarga:

PiBa-NL:
@rbgarga, ive rebased the config upgrade version numbers

PiBa-NL:
Rebased.

rbgarga:
@PiBa-NL can you please rebase it one more time?

jim-p:
The IPsec portions look OK at a glance, will test it more once it gets rebased. The GUI additions will need some review from @sbeaver-netgate

sbeaver-netgate:
I will review after the conflicts are resolved.

PiBa-NL:
@rbgarga @jim-p @sbeaver-netgate , Rebased it, lemme know if something needs changing/fixing.

sbeaver-netgate:
Clicked the "Add" button twice and added two new algorithm sections, but I could then only delete the second added section. The first added agm displays a delete button, but it doesn't do anything.

PiBa-NL:
@sbeaver-netgate there should always be at least 1 algorithm section, so probably the delete button should get 'hidden' once there is only 1 left i havnt implemented that yet.. But it is imho a small visual detail that can be fixed later assuming the pull-request is 'technically approved' so to say.?

sbeaver-netgate:
I had two selections on screen, and could not delete the seconds one.

PiBa-NL:
hmm i cant seem to reproduce that..

jim-p:
It seems to pass some basic tests here. I was able to add multiple P1 encryption sets to a tunnel. They show in the summary view, they are in the config. Clients could still connect (Win 10 and OS X). Editing the tunnel I could remove any entry I tried (first, last, middle). Looking good so far :+1:

sbeaver-netgate:
Yeah - Sorry. That was my fault. There is not much visual distinction between each encryption section, and the authentication section above. - Perhaps others will make the same mistake. I'm trying to think of a way to improve that and also fix having the delete button hidden when only one is left. (Because I spent so much time on that with all the other "repeatable" blocks in pfSense.

Would this display be an easier way to do it? (The diff is agains current master, not your version)
pfsense_localdomain_-_vpn__ipsec__tunnels__edit_phase_1

PiBa-NL:
Adding a section 'title' should be easy also with the current list implementation. But i must agree that your mock-up looks good. Will have to look if there was some technical difficulty that prevented me from using that. One nice thing about the new 'list' component would be that it supports more control's in their original layout. The repeatable group is limited in its ability to have more controls afaik.?. but in this case 4 is enough. So i'm not sure anymore why i started implementing a complete separate solution..

Ill take a look at using that instead, will take a few days though..

jim-p:

Thanks!

jim-p:
Redmine issue: https://redmine.pfsense.org/issues/8186

PiBa-NL:
Changes implemented from the sample code (the usable parts).

If you thinks that is not the case, can you point me to what i missed?

sbeaver-netgate:
Looks very nice. Thank you. Sorry for the confusion, but for some reason you update was given a date from April so we did not realize it was new :(

I needed to bump the config version umbers since another update was made prior to this one. It is now 17.5 - I renamed the update function accordingly.

PiBa-NL:
Okay i understand, thanks :)

PR #3712, opened 2017-04-29
Remove bonus end td tag in interface statistics widget

phil-davis:


phil-davis:
@jim-p @sbeaver-netgate this is a little bit of rubbish that could easily be merged and backported before you guys make 2.3.4

PR #3714, opened 2017-04-29
Fix interface down in interface statistics widget

phil-davis:
If an interface is down, the widget would write the interface description in a column heading but then omit the stats items. So later columns (to the right) would have the headings and stats items out of line.
This also fixes the performance problem that get_interface_info($ifdescr) was being called for every row of stats data for every interface. i.e. it was called 7 times for each interface when 1 time is enough.


phil-davis:
@jim-p @sbeaver-netgate this should also backport OK to 2.3.*

PR #3715, opened 2017-04-29
Fix #7501 Make Interface Stats widget layout more flexible

phil-davis:
1) Let the user filter the interface statistics table by statistic (as well as by interface), so they can choose to show just a subset of statistics if they wish.
2) Add an option to display the table with 1 row per interface, rather than the current default of 1 column per interface.


phil-davis:
Note: it would be best to first review and merge:
#3712 - a minor "fix" to existing code that can be immediately and easily back-ported to 2.3.* branches
#3714 - fix the case when some interface is down (the column data would not line up with the headings). This has more lines of change, but should also backport easily to 2.3.*

Then this functional enhancement can be looked at on top of all that.

There will be conflicts when those above 2 changes are merged. The code here has those fixes already, so actually resolving the "conflicts" will come down to just confirming that the existing code here is what is wanted.

rbgarga:
@phil-davis please rebase

phil-davis:
I think that should be fine now.

rbgarga:
@phil-davis not yet, it's conflicting

phil-davis:
I need to test this now - some conflicts were non-obvious.

phil-davis:
All combinations work for me.
Hopefully this is the last set of changes it will conflict with.

stilez:
Works for me (after manually updating index.php and the functions "gen_customwidgettitle_div()" and "set_customwidgettitle()" which have respectively been updated since 2.3.4 and added to guiconfig.inc since 2.3.4). Looks good, solves redmine issue

phil-davis:
Yes, to backport to the 2.3.* series a few dependent things need to be checked as to what of the widget title editing changes could be easily backported, or whether to just remove that code from this widget when back-porting it.

PR #3716, opened 2017-04-29
Set default language in General Setup to en_US

phil-davis:
Since https://github.com/pfsense/pfsense/commit/fdcde31b4a910c4e058513c1e3f68a62e722da6b added German to the top of the get_locale_list() array, if you start with a default system and go to System->General Setup (make some changes to other fields if you like) and press Save, you end up in German because that is what gets selected in the dropdown when the config does not yet have a language in it.
The code assumed that the language of the default system is the language at the top of the list, which is no longer true.

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -47,6 +47,11 @@ if (!isset($config['system']['webgui']['dashboardcolumns'])) {
 	$config['system']['webgui']['dashboardcolumns'] = 2;
 }
 
+// set default language if unset
+if (!isset($config['system']['language'])) {
+	$config['system']['language'] = 'en_US';
+}
+
 $dnsgw_counter = 1;
 
 while (isset($config["system"]["dns{$dnsgw_counter}gw"])) {


phil-davis:
Note: This could also be fixed by:
1) Edit get_locale_list() to keep en_Us at the top; or
2) Put the language tag into the default config, and ad a config upgrade step to set the language to en-SU if it is not already set; or
3) Abstract this higher up somewhere, have something that specifies the "build language" and is used as the default language in all places if nothing else is specified. Then "en_US" would only be specified in 1 place.

jim-p:
I'm leaning toward setting the default in /etc/inc/globals.inc, and then using that global default to pre-fill the variable as you've done in this PR. The value in globals.inc could be changed/overridden by a separate build if needed.

phil-davis:
@jim-p I am happy if you do something like that and close this.

jim-p:
Turns out it was already defined in globals.inc. I merged this and then adjusted the default setting to pull from there. Thanks!

PR #3717, opened 2017-05-01
Integrate various cloudflare changes

phil-davis:
I suspect this is what is needed to properly integrate with the change that added the "proxied" option for cloudflare.

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -498,6 +498,7 @@ events.push(function() {
 				hideCheckbox('proxied', false);
 				hideInput('zoneid', true);
 				hideInput('ttl', true);
+				break;
 			default:
 				hideGroupInput('domainname', true);
 				hideInput('resultmatch', true);
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -469,8 +469,6 @@ events.push(function() {
 				hideInput('ttl', false);
 				break;
 			case "namecheap":
-			case "cloudflare-v6":
-			case "cloudflare":
 			case "gratisdns":
 			case "hover":
 				hideGroupInput('domainname', false);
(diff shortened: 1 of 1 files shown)


phil-davis:
If it works, it needs to be cherry-picked back to RELENG_2_3 etc, wherever the "duplicated" cloudflare cases in the switch statement are found.

CarlGill:
@phil-davis   This change would definitely work.

What needs to be confirmed though is if it is needed.

I Tested the 2.3.3 against cloudflare and it all seemed to work.

I will test when I get back to my home lab whether we can completely strip out the proxied feature all together.  If we can, then there is no reason to go down this rabbit hole.

phil-davis:
@CarlGill any home lab results?

rbgarga:
@CarlGill did you have the chance to test it?

CarlGill:
@phil-davis @rbgarga   Sorry, I have not had a chance yet.  I should Have time Tonight or tomorrow assuming I finish at the office in a reasonable hour.

crkochan:
I have tested these changes on 2.3.4-release on an SG-2220. 

I arrived at the same changes on my own in a branch I was working on before realizing this PR existed. Oops.

Originally, I thought that there also needed to be a break statement at the end of the 'cloudflare' block on line 500, but the interface seems to work without it.

The fix restores the missing UI checkbox element, the use of which is properly reflected when checking the results via the CloudFlare portal.

crkochan:
Tested on 2.3.4-release on an SG-2220.

Restores hidden CloudFlare Proxy checkbox.

Ran multiple updates updates with proxy features enabled and disabled. The CloudFlare portal reflected the proper proxy status each time.

One question, does the 'cloudflare' block need a `break;` at the end of it? The 'custom-v6', 'route53', and 'hover' blocks all have them.

phil-davis:
The break there was missing all along! I have added it. That should make both the domainname input and proxied checkbox appear.

PR #3718, opened 2017-05-02
Typos in pkg management code comments

phil-davis:
While looking at this code, may as well fix the typos in comments.


phil-davis:
@jim-p @sbeaver-netgate this is another little bit of rubbish that could easily be merged and backported before you guys make 2.3.4

PR #3720, opened 2017-05-03
Fix #7508 stop write_config after reset_factory_defaults

phil-davis:
On shutdown, some things try to write_config to (possibly) save some state. e.g. Captive Portal vouchers, and maybe some packages might try to do write_config() in their stop function.

If we are in the process of doing reset to factory defaults, then that write_config writes over the top of the default config that is waiting to be implemented. So do not do that.


phil-davis:
@jim-p @sbeaver-netgate I don't like doing these sort of global passing things, but it works here. I got a bit verbose in the comments, but it needs at least some explanation about why this flag is used.

PR #3727, opened 2017-05-11
Split System Widget in default dashboard

phil-davis:
This is a suggested split of the system widget on the default dashboard. The name/version... are in column 1. The "performance" items are in column2, with widget title "System Performance". Users can customize this anyway, but this might be a useful starting point to give on the default dashboard.
Open for discussion. See forum https://forum.pfsense.org/index.php?topic=129040.0

### DIFF
--- a/src/conf.default/config.xml
+++ b/src/conf.default/config.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <pfsense>
-	<version>15.4</version>
+	<version>16.5</version>
 	<lastchange/>
 	<system>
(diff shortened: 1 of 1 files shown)


phil-davis:
Yes, I just did this to point out that it is now possible without any real code changes - just modifying the default config. So if someone wants to split the System Information widget - either in the default config, or on their own dashboard - it is this easy.
If splitting it on the default dashboard is not required, then feel free to close this. It is just that there had been a few mentions of maybe doing that.

PR #3728, opened 2017-05-11
auth_check lowers cpu usage for checking if the user has permission

PiBa-NL:
auth_check lowers cpu usage for checking if the user has permission for the page requested when used in place of guiconfig, especially useful for frequent requests like those made by stats and traffic widgets

### DIFF
--- a/src/etc/inc/auth_check.inc
+++ b/src/etc/inc/auth_check.inc
@@ -1,0 +1,79 @@
+<?php
+/*
+ * auth_check.inc
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2017 Rubicon Communications, LLC (Netgate)
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


src/etc/inc/auth_check.inc
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// this function is a duplicate from cmp_page_matches() in priv.inc
+// however unconditionally including priv.inc takes significant more time/cpu
```
rbgarga: Why not move the function here instead of duplicating it and make sure priv.inc to require this file?

rbgarga: I don't like the idea of duplicating code around, it can create headaches in the future when one function is changed and other doesn't

PR #3729, opened 2017-05-15
Remove stray debug-code (/tmp/script)

znerol:


znerol:
This line is there since the initial commit 5b4ee05e5

PR #3731, opened 2017-05-18
Removed htmlspecialchars when loading leases from config

ylecuyer:
Removed htmlspecialchars when loading leases from config.

Fixes a bug introduced in pfsense/pfsense@49a6769 where description and hostname would be aplied htmlspecialchars twice when loaded from the config.


pfSense-git-updates:

ylecuyer:
Done

PR #3740, opened 2017-05-24
If user-entered PHP CLI errors out, display error info to help

stilez:
If the user enters PHP in the command-line page, and it errors out, it's usually a typo or something minor. But the user is left with an error that references a /tmp file which doesn't exist at the point they read the message, a line number that's incorrect, and their input which doesn't display line numbers anyhow.

This patch fixes this, so that a user who wants to enter php command-line in the GUI and gets an error, can quickly identify the exact line causing the error. 

If an error occurs (detected by either non-zero return value or output matching an error message referencing the temp script file), then an short DIV is shown above the usual PHP output for "error location". It gives the correct error line number, and shows the code with line numbers and syntax highlighting (using php's native "-s" option), pre-scrolled to the error line.


stilez:
Test this PR by entering code with unmatched {  }, missing semicolon,bad syntax, require_once with nonexistent files/bad paths, etc.

src/usr/local/www/diag_command.php
```diff
@@ -255,39 +255,86 @@ function Reset_onClick( form ) {
...
+			$chars = strlen(count($syntax_output)) + 1;
+			for ($lineno = 1; $lineno < count($syntax_output) - $lineno_correction; $lineno++) {
+				if ($lineno == $errline) {
+					$lineno_prefix = "&gt;&gt;&gt;" . str_repeat('&nbsp;', $chars - strlen($lineno));
+				} else {
+					$lineno_prefix = str_repeat('&nbsp;', ($chars + 3) - strlen($lineno));
+				}
+				$html .= "<span style='color:black;backgroundcolor:lightgrey'><tt>" . $lineno_prefix . $lineno . ":</tt></span>&nbsp;&nbsp;{$syntax_output[$lineno + $lineno_correction - 1]}<br/>\n";
+			}
+			echo sprintf(gettext('Line %s appears to have generated an error, and has been highlighted. The full response is below.<br/>' .
```
sbeaver-netgate: Changing this to a "print_info_box()" call makes for a better display I think:

`			print_info_box(sprintf(gettext('Line %s appears to have generated an error, and has been highlighted. The full response is below.<br/>' .
				'Note that the line number in the full PHP response will be %s lines too large.'), $errline, $lineno_correction), 'danger');`

src/usr/local/www/diag_command.php
```diff
@@ -255,39 +255,86 @@ function Reset_onClick( form ) {
...
+				'Note that the line number in the full PHP response will be %s lines too large.'), $errline, $lineno_correction);
+			echo "<div style='margin:20px'><b>" . gettext("Error locator:") . "</b>\n";
+			echo "<div id='errdiv' style='height:7em; width:60%; overflow:auto; white-space: nowrap; border:darkgrey solid 1px; margin-top: 20px'>\n";
+			echo $html . "\n</div></div>\n";
+		}
```
sbeaver-netgate: We have converted almost all of the "echo" statements to "print()" in pfSense. Would be good to do the same here please

src/usr/local/www/diag_command.php
sbeaver-netgate: Let's make the '<br />' a positional parameter too. We are getting serious about making the translator's jobs less of a pain.

stilez:
If someone knows a better way to scroll the div to the right position (or to an anchor in the contents) can they add it here so I can use it. Hard to know what works best across browsers and with different zoom settings. If its difficult, don't worry though.

stilez:
Updated the code - 

1) convert echo to print, and add infobox, as requested in PR comments
2) If no identifiable line number, say so
3) EVAL gives an error message format that wasn't picked up, it is now. Test case - enter as the code:  eval("X");
4) Warn that EVAL gives a "spurious" "error at line 1" as well, the real error is in the (...). See above test case for this.
5) Simplify translators' job a bit
6) Try to reduce errors in user code giving NOTICES (see PR comments). Define a temp dir in globals.inc for user code in diag_command.php only (in /tmp/user_code) so we can accurately avoid creating notices if the error was from such a file, without excluding anything else. Then modify file_notice() sdo it doesn't file the notice if the error happened from such a file.

There is a risk of issues happening from the last step - for example, user code is used to update something; but somehow hits an error that should be reported (disk space or missing system file), then this won't be reported as a notice. Is this an issue, and if so, can we at least exclude some user-code errors if not all, from being added to notices? (One option - also require error text to match some common issue in user code, like "syntax error"?)

PR #3741, opened 2017-05-27
Convert bit-based error types to human readable E_* codes when reporting errors

stilez:
error_get_last() includes a bit-based error type which corresponds to E_WARNING, E_DEPRECATED etc. Report the text tags rather than the numeric code if possible, as it makes more sense on reading notices.


PiBa-NL:
Any reason not to write the array values as they are?:
```
	$PHP_error_types = array (
		   1 => 'E_ERROR',           2 => 'E_WARNING',           4 => 'E_PARSE',            8 => 'E_NOTICE',
		  16 => 'E_CORE_ERROR',     32 => 'E_CORE_WARNING',     64 => 'E_COMPILE_ERROR',  256 => 'E_USER_ERROR',
		 512 => 'E_USER_WARNING', 1024 => 'E_USER_NOTICE',    2048 => 'E_STRICT',        4096 => 'E_RECOVERABLE_ERROR',
		8192 => 'E_DEPRECATED',  16384 => 'E_USER_DEPRECATED'
	);
```
Looks like it would be faster even if only by a millisecond. As globals is included almost anywhere there is no point to doing the string parsing and eval dynamic code every time from php code.
imho anyhow, @jim-p what do you think.?.

jim-p:
If you're generating it from a static list anyhow, may as well do the array method rather than a foreach/eval loop.

If you don't want to hardcode anything and don't care about speed, you could parse it out of get_defined_constants(true) by getting any E_* variables from the 'core' section of the returned list.

stilez:
Makes sense. I did it that way because of the old principle that codes intended for human use tend to be static, those for internal use (their numeric equivalents) are more prone to change. Probably not relevant here. Worth changing to fully static definition?

sbeaver-netgate:
Please consider using the array as PiBa-NL suggests, and resolve conflicts

jim-p:

Thanks!

sbeaver-netgate:
Please rebase to resolve conflict. We can then merge this PR.

sbeaver-netgate:
Closing due to unresolved conflicts. Please resubmit if you are able to rebase.

PR #3742, opened 2017-05-27
Remove duplicate code in get_uptime()

stilez:
We already have a function for seconds -> (d) hms when displaying time intervals, so use it. (Also ensures consistent presentation of time periods, and also avoids the need for gettext() on the words used.)

For niceness, minor enhancement also to pfsense-utils.inc that allows time intervals to be displayed a little nicer

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -1442,8 +1442,11 @@ function is_pppoe_server_enabled() {
 	return $pppoeenable;
 }
 
-/* Optional arg forces hh:mm:ss without days */
-function convert_seconds_to_dhms($sec, $showhoursonly = false) {
+/* Optional args:
+ * $showhoursonly forces hh:mm:ss without days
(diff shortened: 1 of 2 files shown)


src/etc/inc/pfsense-utils.inc
```diff
@@ -1453,7 +1456,7 @@ function convert_seconds_to_dhms($sec, $showhoursonly = false) {
 					(int)(($sec % 3600)/60),
 					$sec % 60
 				);
-	return ($d > 0 ? $d . 'd ' : '') . sprintf('%02d:%02d:%02d', $h, $m, $s);
+	return ($d > 0 ? $d . 'd ' : '') . sprintf(($showHMS ? '%02dh %02dm %02ds' : '%02d:%02d:%02d'), $h, $m, $s);
```
rbgarga: For these one big line statements are a bit confusing and hard to read. Do you mind to rework it using variables and whatever is necessary to reduce these ternary operators?

sbeaver-netgate:
Requested change not addressed in 4 months. Please resubmit if still desired

PR #3743, opened 2017-05-27
 system_identify_specific_platform() shouldn't return "pfSense" when $g['platform'] contains product name

stilez:
I've checked all uses of this function. In every case, the calling code is looking for the hardware platform, not the firmware name. Unfortunately if it's not a specific recognised hardware, this function falls back on $g['platform'] and returns that, which by default contains the product name. 

This PR ensures that if `system_identify_specific_platform()` is going to return the product name, it doesn't, to ensure the returned string is always as expected a platform name.

It also ensures that a usable string is always returned.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -2466,7 +2466,8 @@ function system_identify_specific_platform() {
 	}
 	unset($dmesg_boot);
 
-	if (strlen($g['platform']) > 0) {
+	if (strlen($g['platform']) > 0 && $g['platform'] != $g['product_name']) {
+		// if $g['platform'] contains the software/firmware name rather than a hardware name, don't use it.
 		return array('name' => $g['platform'], 'descr' => $g['platform']);
 	} else {
 		return array('name' => 'Unknown platform', 'descr' => 'Unknown platform');
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -2470,7 +2470,7 @@ function system_identify_specific_platform() {
 		// if $g['platform'] contains the software/firmware name rather than a hardware name, don't use it.
 		return array('name' => $g['platform'], 'descr' => $g['platform']);
 	} else {
-		return array('name' => 'Unknown platform', 'descr' => 'Unknown platform');
+		return array('name' => 'Generic/unknown', 'descr' => 'Generic/unknown');
 	}
 }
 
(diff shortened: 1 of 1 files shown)


stilez:
The problem is not a big deal , if the question needed an answer, see below.  

"Unidentified hardware" would also be fine. My wording came from looking at what would seem sensible to a user. - my logic was a bit like this:  When one usually thinks of "hardware platform" as distinct from the individual components, one is referring to the actual brand of the box or server - a WRAP, an ALIX, a PowerEdge, or whatever.  While the major hardware components are usually able to be named and often branded, the **platform** may not be, and often doesn't have a formal model/name.  I figured that many/most users probably run white box pfsense where the platform itself doesn't have a "name".  How would such users expect their hardware platform to be described on the system widget (the main place this info will be used for non-name boxes)? "Generic x86 platform" seemed a decent answer - "unidentified" sounded like an error, posing the question whether it _should_ be identified. Generic/unknown implies something more reassuring, that it's a generic x86 build, or built using off the shelf components, but doesn't imply as "unidentified" might, that there is a value here that _should_ be identifiable. So I went with "Generic/unknown" because the boxes we can't explicitly identify will almost always be generic x86/embedded boxes, or unknown brand boxes, and it's accurate/informative and didn't make it sound like there was info missing.

As to reason, if it matters - these strings are only used in 2 places - code that needs to check if the hardware is a specific type, in order to take it into account when generating GUI or processing user responses (eg offering or hiding specific options), and code that wants to inform the user as to their system generally (system widget etc).  In neither of these cases is it useful to report the hw platform as "pfsense" - in the first case it's never useful but it's iunvisible. In the second case presumably the user would appreciate hardware platform and system software platform not being mixed up, with "pfsense" being given for both. Logically in all cases, it's cleaner to have one route that clearly delivers a hardware (not system/OS) platform info, or says it's not known, and another that delivers a software/firmware/"running system" platform info, and that they are clearly different and shouldn't be seen as conflated.  Also if the two are confused, it adds confusion in output (if the user code wants to know specifically the hardware to report it, how can it be sure that's the info it is actually printing to GUI, and that it's not printing "pfsense" for both hardware and software, if the back-end is itself allowed to be muddled?)  As this was easy to do in the code and makes the output cleaner with less not more confusion, I did it, and it's neater to do so I think.  

tl;dr, let me know what string you prefer :)

jim-p:
You have a good point that "Unknown" and "Unidentified" both could be mistaken as errors, but "Generic" also has negative implications. We could maybe go for something like "Common Hardware" or "Commodity Hardware".

In the end though I'm not sure any change is warranted. The location where the text is shown to the user is not specifically labeled as a description of the hardware. And at the end of the day, seeing "pfSense" in the system description is most likely less confusing than any other text. The old "Platform" section of the system info widget is gone on 2.4 so there isn't any redundancy there in these cases.

stilez:
This PR can be closed when #3944 is committed; it dup's some of that, and is behind it.

sbeaver-netgate:
No longer required

PR #3746, opened 2017-05-28
Allow user to specify smaller traffic graphs on dashboard widget

stilez:
A dashboard should usually aim to be visible to the eye, not needing scrolling etc too much. But even with extra columns, the traffic graphs are quite large, so this adds a "chartsize" option to the widget to set a smaller size.

It works well for me, with one CSS related caveat which I can't fix, and need someone who knows the widget system better to put it right ( @phil-davis ? :) )

The sole issue is the scaling and positioning of the graphs. At the moment I can't figure how to do this "natively", by scaling the svg's themselves, or their viewport, or a better way. So I'm solving it using CSS zoom which is well supported on all browsers for many years. It gives almost the right results but also a lot of trailing white space, which I can't figure how to remove.

But as the code's essentially there with the exception of CSS/appearance, can I ask someone who knows the codebase and widgets better than me, to figure out the required final change to make this good, and I'll push it to my repo when that's done. Everything else is good. Makes the dashboard a lot easier to get an overview of the system's activity.

### DIFF
--- a/src/usr/local/www/widgets/widgets/traffic_graphs.widget.php
+++ b/src/usr/local/www/widgets/widgets/traffic_graphs.widget.php
@@ -57,8 +57,12 @@ if ($_POST) {
 		$user_settings["widgets"]["traffic_graphs"]["backgroundupdate"] = $_POST["backgroundupdate"];
 	}
 
-	if (isset($_POST["size"])) {
+	if (isset($_POST["size"]) && ($_POST["size"] == 1 || $_POST["size"] == 8)) {
 		$user_settings["widgets"]["traffic_graphs"]["size"] = $_POST["size"];
+	}
+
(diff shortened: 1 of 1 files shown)


jdillard:
I'm not a fan of adding options when they aren't absolutely necessary, as they tend to add a ton of technical debt within themselves and especially when compounded on existing/future options.

I agree that traffic graphs should be smaller, but think we should just make them smaller across the board. A simple CSS change to pfSense.css from a height of 250px to 150px:

```
.traffic-widget-chart, .traffic-widget-chart svg {
    padding: 0;
    height: 150px;
}
```

produces this:

screen shot 2017-11-29 at 1 11 53 pm

what do you think?

grandrivers:
I would NOT want min any smaller than currently are !

PR #3748, opened 2017-05-29
Rework on "system_usermanager_settings" and "system_authservers"

stilez:
I refactored these two pages, although the underlying code isn't changed at all.

The session timeout in user manager settings hasn't got much to do with individual users or their permissions / authentication, it's a global timeout setting far closer to other admin security measures controlling security of logins on WebGUI sessions which are without exception all under **system -> advanced admin"**. Long standing "urgh". The two pages have 3 other functions split between them which are very closely related - (1) view a list of auth servers, (2) select an auth server, and (3) edit auth servers. It seems very unhelpful to have (1) + (3) on the same page (via add/edit buttons) and (2) as a solo control on a different page, so I've put the GUI auth selection dropdown on the same page as the list of auth servers, which works nicely and makes logical sense.  That left nothing on usermanager_settings, which seems reasonable - it only had 2 controls, one related to admin access security and the other to selecting an auth server, neither was related to the user manager itself. Finally for code simplicity, I split **system_authservers.php** into two files, one for the table display and the other for entry editing, creating a new `system_authservers_edit.php` page from the existing code, because that function really is quite long and complex - basically it's exactly how it's done on other pages with a substantial edit page (such as firewall_rules.php + firewall_rules_edit.php).

**tl;dr summary:**

- `usermanager_settings.php` removed. There _weren't_ any settings for the user manager itself, it only had 2 controls, and both were very closely tied to other pages (one much more related to admin security but unhelpfully, not kept with other admin security functions, and the other related to how users authenticate rather than any settings of the usermanager). The controls moved with their code intact, one moved to `system->advanced admin`, the other to `system->user manager -> authentication servers`. They look a lot more at home there, and much easier for GUI users.  

- `system_authservers.php` was very substantial because of all the auth settings for editing an auth server. In that regard it's closer to firewall rules, which have a separate settings page. It basically forked at `head.inc` to output both the main and edit pages. I split it into 2 pages the same way as firewall rules pages and others - a main page (containing the table and "add" button", and an edit page.  This also let me be fairly sure that it wouldn't would cause issues, and kept the PR much simpler, by making it a direct copypaste with only minor changes.

- I then added the `GUI login auth server selector` code to be above the table on authservers.

- I deliberately avoided code change to the underlying code for ease of review.  It's basically copy-paste + minor glue. The only changes are as follows:  (1) Minor UI wording improvements on authserver.gui form, to be clearer what the selector is for, see screenshot, (2) Minor enhancements to validity checking on $_POST/$_REQUEST, (3) simplifying the args for the new `authservers_edit.php`:  it doesn't need `?act=new/edit` in its links now, and can be simplified like firewall_rules_edit.php: the presence of a valid `?id=` signifies edit, anything else signifies new. So I simplified that aspect and $act is set based on `id`.  (4) consequential edits to`help.php` + `priv.defs.inc` (because of the new page and removed page), and other user manager subpages (to remove the usermanager_settings tab),

I don't use LDAP or other auth servers myself which limits checking, but I haven't modified the code knowingly other than peripheral as described, and the redone code looks good here from what I can see. Much more intuitive and simpler.

Screenshot below.


stilez:
**Screenshot:**

tmp

stilez:
**Screenshot:**

tmp

stilez:
@jim-p Thanks for the review. Will do as requested shortly.

stilez:
Working on this. Not yet ready for review. Will comment here when it's done.

jim-p:

Thanks!

rbgarga:
No changes since last June. Please re-open if you still want to work on that

PR #3749, opened 2017-05-31
Provide backend functions to get safe versions of release info, and HTML from a pfsense doc page (redmine feature 5074)

stilez:
Provide functions to get pfSense release info for GUI use. 
See @jim-p comment on [redmine 5074](https://redmine.pfsense.org/issues/5074)

It would be very useful when upgrading, to have easy access to the changelog from `system -> system update`.

But there's no easy way from the GUI to view the changelog for the current release, or, if a release was skipped, the skipped releases up to the current release.  We could just add a crude link, but the user might need more than one version's changes (if they skipped a release) and it's nicer anyway to present it in-GUI.

To make this possible, we first need to get the pfSense release info in usable form. This PR adds 3 underlying functions needed.  That way when there's a new version, `system -> system update` can make use of this info and provide an easy way to view relevant changelogs for the proposed update, in future. The functions added are:

1. `get_release_list()` - returns a list of releases taken directly from the doc page _"Versions of pfSense and FreeBSD"_, including all table data (dates, status, doc pages of the release notes etc)
2. `get_sanitised_doc_page()` - returns a sanitised HTML version of a single doc page, such as the release notes for an update version.  All attribs (except a single valid href) and formatting/CSS are removed to make it easy to match the GUI's existing style/theme or add needed CSS.
3. `update_all_release_details()` - maintain a tiny cache ( ~ 20 pages x 2KB) for the bare release notes of known versions. Updated when user enters `system -> system update`. Cache is kept in `/tmp/doc_pages_cache/PAGE@TIMESTAMP` (controlled by `$g['doc_pages_cache']`).   As these doc pages are mostly static and the code checks last modified time, once cached we rarely need to update, so it should be extremely fast (immediate) after the 1st run.  Also as these pages are small and relatively few, and rarely added to, the cache won't grow large, nor will old pages become irrelevant easily, so no need for complex caching code. Pretty much just check for updates and not much more.  Stops the page from taking time to grab multiple static doc pages when we've almost certainly grabbed them previously and can reuse.

These aren't used yet, but adding them so that progress can be made on redmine 5074.

**DATA SANITISING NOTE  -** There is careful attention to sanitising , so that if there were a problem at the docs website end it doesn't create a vulnerability from reusing doc content in the router context. But as sanitising is easily done wrong, extra attention to check that it really is robust (and fix if it isn't) would be good.

### DIFF
--- a/src/etc/inc/globals.inc
+++ b/src/etc/inc/globals.inc
@@ -85,7 +85,8 @@ $g = array(
 	"pkg_prefix" => "pfSense-pkg-",
 	"default_timezone" => "Etc/UTC",
 	"language" => "en_US",
-	"default_config_backup_count" => 30
+	"default_config_backup_count" => 30,
+	'doc_root_url' => 'doc.pfsense.org/index.php/'
 );
 
 /* IP TOS flags */
--- a/src/etc/inc/globals.inc
+++ b/src/etc/inc/globals.inc
@@ -86,7 +86,8 @@ $g = array(
 	"default_timezone" => "Etc/UTC",
 	"language" => "en_US",
 	"default_config_backup_count" => 30,
-	'doc_root_url' => 'doc.pfsense.org/index.php/'
+	'doc_root_url' => 'doc.pfsense.org/index.php/',
+	'doc_api_url' => 'doc.pfsense.org/api.php?'
 );
 
 /* IP TOS flags */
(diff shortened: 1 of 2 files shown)


stilez:
Editing this to clean it up - will repost in a bit. Closed for now.

PR #3754, opened 2017-06-12
Pattern matching correction in carp status

derelict-pf:
Match whitespace after $vhid to avoid partial matches. Fixes #7638

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -723,7 +723,7 @@ function get_carp_interface_status($carpid) {
 
 	$vhid = $vip['vhid'];
 	$carp_query = '';
-	$_gb = exec("/sbin/ifconfig $interface | /usr/bin/grep carp: | /usr/bin/grep \"vhid $vhid\"", $carp_query);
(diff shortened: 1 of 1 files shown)


src/etc/inc/pfsense-utils.inc
```diff
@@ -723,7 +723,7 @@ function get_carp_interface_status($carpid) {
 
 	$vhid = $vip['vhid'];
 	$carp_query = '';
-	$_gb = exec("/sbin/ifconfig $interface | /usr/bin/grep carp: | /usr/bin/grep \"vhid $vhid\"", $carp_query);
+	$_gb = exec("/sbin/ifconfig $interface | /usr/bin/grep carp: | /usr/bin/grep \"vhid $vhid \"", $carp_query);
```
jim-p: As long as we're fixing it, combine those greps and fix the variable refs so it's simpler, such as:

`exec("/sbin/ifconfig {$interface} | /usr/bin/grep \"carp:.* vhid {$vhid} \"", $carp_query);`

jim-p:
Since @derelict-pf is out for most of this week I committed the changed version directly rather than waiting on him to get back and update the PR, so we can get this into 2.3.4-p1 if it happens that quickly.

PR #3757, opened 2017-06-13
system_gateways, default gateway switching based on a set 'priority'

PiBa-NL:


jim-p:
This is interesting but still seems like a half-measure to fix https://redmine.pfsense.org/issues/4507. Having to visit each of the gateways and set a priority seems counterintuitive and difficult for the user.

A better way would be to allow a single gateway group to be marked as default when gateway switching is enabled. Groups already have a familiar GUI interface for tiers and functions for finding current active gateways. That way the current behavior could be preserved if there was no default group.

PiBa-NL:
fyi i'm working on the proposed changes.. Will hopefully have some progress in a week or two..

jim-p:
@PiBa-NL Thanks!

PiBa-NL:
Closing this one, as a different/better implementation is in progress https://github.com/pfsense/pfsense/pull/3781

PR #3758, opened 2017-06-14
changed how openvpn calls php for authentication

dcnoye:
Currently openvpn is using the executable fcgicli to do authentication, which, from what I can tell uses php-fpm via socket connection. I've changed this to use system php instead of the service ph p-fpm. 

This will prevent individuals from being locked of a vpn connection if php-fpm would crash.

### DIFF
--- a/src/etc/inc/openvpn.auth-user.php
+++ b/src/etc/inc/openvpn.auth-user.php
@@ -32,6 +32,8 @@ require_once("radius.inc");
 require_once("auth.inc");
 require_once("interfaces.inc");
 
+parse_str(implode('&', array_slice($argv, 1)), $_GET);
+
 /**
  * Get the NAS-Identifier
  *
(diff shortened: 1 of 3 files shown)


jim-p:
I'm not sure this would be an overall positive change. PHP-FPM is much better from a performance and resource usage perspective, especially with large numbers of users and/or a busy system.

It would be better to focus on why PHP-FPM is failing and fix that instead.

dcnoye:
The binary that calls/utilizes php-fpm is actually only used in a couple place within pfsense.

 ( https://github.com/pfsense/pfsense/search?utf8=%E2%9C%93&q=fcgicli&type= )

The pull request actually follows more in suit with the rest of the framework. I do however agree that the reason for php-fpm crashing should be investigated and fixed, this will be an overall more robust solution.

rbgarga:
I agree with @jim-p here. I don't think this is the right way to go

PR #3760, opened 2017-06-18
loginpage, move css out of the folder used for enumerating themes, and remove duplicated code.

PiBa-NL:
loginpage, move css out of the folder used for enumerating themes, and remove duplicated code.


sbeaver-netgate:
The login page is an early experiment. Thanks for your work, but I don't want to accept a pull request against the page until it is closer to being finished.

PR #3761, opened 2017-06-20
Add Option to use static IPV6 over v4 parent ( PPPoE )

marjohn56:
A new option when setting a v6 static on the WAN to allow the connection to use the parent V4 interfaces i.e. PPPoE.

Some ISP's allow the use of static V6 addresses and PD over a PPPoE link. This patch adds a new option under WAN settings for IPv6 static address, the option to use the parent v4 ( pppoe) interface.  The change was needed as although it is possible to set up the static and it works just using the given address, the gateway and monitoring did not work. it is essential to know what the gateway address is when using this option, as you'll need to manually add the gateway. In my case this is found by using dhcp in the first instance to get that information, and then adding a new gateway etc.

### DIFF
--- a/src/usr/local/www/interfaces.php
+++ b/src/usr/local/www/interfaces.php
@@ -1903,7 +1903,7 @@ $section->addInput(new Form_IpAddress(
 $section->addInput(new Form_Checkbox(
 	'ipv6usev4iface',
 	'Use IPv4 connectivity as parent interface',
-	'IPV6 will use  the IPv4 connectivity link (PPPoE)',
+	'IPv6 will use the IPv4 connectivity link (PPPoE)',
 	$pconfig['ipv6usev4iface']
 ));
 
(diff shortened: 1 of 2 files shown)


src/usr/local/www/interfaces.php
```diff
@@ -1895,6 +1900,13 @@ function build_gatewayv6_list() {
 	'V6'
 ))->addMask('subnetv6', $pconfig['subnetv6'], 128);
 
+$section->addInput(new Form_Checkbox(
+	'ipv6usev4iface',
+	'Use IPv4 connectivity as parent interface',
+	'IPV6 will use  the IPv4 connectivity link (PPPoE)',
```
jim-p: Change "IPV6" to "IPv6", and there are two spaces between "use" and "the" instead of just one.

Otherwise seems OK.

PR #3763, opened 2017-06-24
gateway monitoring, give apinger some time to properly 'initialize' before using its results

PiBa-NL:
gateway monitoring, give apinger some time to properly 'initialize' before using its results

this avoids needles "gateway added/removed from routing group" emails, and possible other events that might trigger when a gateway fails or comes back online

(This sometimes happens when actually nothing changed regarding the status gateways themselves..)

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -323,26 +323,44 @@ function get_dpinger_status($gwname, $detailed = false) {
 		log_error("dpinger: status socket {$proc['socket']} not found");
 		return false;
 	}
+	$timeoutcounter = 0;
+	while (true) {
+		$fp = stream_socket_client("unix://{$proc['socket']}", $errno, $errstr, 10);
+		if (!$fp) {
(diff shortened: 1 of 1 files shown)


luckman212:
Nice patch @PiBa-NL. Am I right in thinking that this helps reduce the "mail bomb" issue from [redmine #4031](https://redmine.pfsense.org/issues/4031)? I looked at the code, and one question I had was on the `usleep(10000)` and `timeoutcounter`.  It appears that the loop will break out after 3 seconds (.000001x10000x300=3s). Is this the intended timing? Is that enough time for dpinger to stabilize its stats and start returning sane values?

dennypage:
FWIW, a simpler approach would be to simply wait until dpinger reports something other than 0 0 0. This will only happen on startup before the first probe is resolved.

dennypage:
BTW, the maximum wait time for 0 0 0 to clear really should be equal to the loss time provided to dpinger rather than a hard coded value.

PR #3764, opened 2017-06-25
Issue #7659 foreach warning on adding gateway

phil-davis:
To make this happen:
1) Start with a system with just default DHCP on WAN - the system generates gateways at run ti,e for that and ```$config['gateways']['gateway_item']``` is empty.
2) Edit some interface, set it to Static IPV4, put in some IPv4 address, click "Add Gateway" and give it some gateway IP.
3) Save and apply
4) Go to the dashboard - the "crash" message is given.

This fix adds protection around the foreach() to not bother going there if there are no existing gateway items to process.

Signed-off-by: Phil Davis <phil@jankaritech.com>


rbgarga:
Already cherry-picked from #3766

rbgarga:
Ops, fat finger here. Reopening

PR #3768, opened 2017-06-25
notices, background delivery of growl messages and combined mail messages from a queue

PiBa-NL:
notices, background delivery of growl messages and combined mail messages from a queue so that in case of failure the timeout period for connecting does not impact the functionality of the calling scripts themselves

1st message is delivered directly after it is queued.
2nd and following messages that are send within a 10 second window are combined into 1 email
when no emails were send for 1 minute, then the monitor script ends and a new 1st message can be delivered directly after it gets queued

### DIFF
--- a/src/etc/inc/notices.inc
+++ b/src/etc/inc/notices.inc
@@ -215,6 +215,97 @@ function are_notices_pending($category = "all") {
 	return false;
 }
 
+function notices_sendqueue() {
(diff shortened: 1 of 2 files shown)


PiBa-NL:
Should improve/fix : https://redmine.pfsense.org/issues/4031 as well.

PR #3770, opened 2017-06-29
Add strongswan rekeymargin attribute to vpn ipsec phase1 view and activate RADIUS accounting for mobile ipsec

hamnur:


hamnur:
Bug: RADIUS accounting is not activated, although it is selected in the auth server view. 

Bugfix: in strongswan.conf "accounting= yes" has to be set

		eap-radius {
			class_group = yes
			eap_start = no
			**accounting = yes**
                ...

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -238,6 +240,10 @@
 		$input_errors[] = gettext("The P1 lifetime must be an integer.");
 	}
 
+	if (($pconfig['rekeymargin'] && !is_numericint($pconfig['rekeymargin']))) {
+	  $input_errors[] = gettext("Rekey Margin must be an integer.");
+  }
```
rbgarga: Fix indent of this block

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -830,6 +837,13 @@ function build_eal_list() {
 	$pconfig['rekey_enable']
 ));
 
+$section->addInput(new Form_Input(
+	'rekeymargin',
+	'*Rekey Margin (Seconds)',
```
jim-p: This doesn't appear to be required, so you can remove the `*`.

PR #3773, opened 2017-07-03
ipsec, prevent simultaneous/repeated calling of vpn_ipsec_configure() by /etc/rc.newipsecdns

PiBa-NL:
ipsec, prevent simultaneous/repeated calling of vpn_ipsec_configure() by /etc/rc.newipsecdns
also exit the script without parsing a config if not needed.

p.s.
I think this would decrease the amount of ipsec restarts when events needing such restarts occur.. One might however argue that the vpn_ipsec_configure() itself also performs locking which would make all sleeping and locking in the rc.newipsecdns unneeded.?

### DIFF
--- a/src/etc/rc.newipsecdns
+++ b/src/etc/rc.newipsecdns
@@ -25,8 +25,21 @@
  * limitations under the License.
  */
 
-/* parse the configuration and include all functions used below */
 require_once("util.inc");
+require_once("globals.inc");
+
+/* make sure to wait until the boot scripts have finished */
+if (file_exists("{$g['varrun_path']}/booting")) {
+	return;
+}
+
+$ipseclck_pending = try_lock('ipsecdns_pending', 0);
+if (!$ipseclck_pending) {
+	/* if a vpn_ipsec_configure() is still pending no need to stack up another one */
+	return;
+}
+
+/* parse the configuration and include all functions used below */
 require_once("config.inc");
 require_once("gwlb.inc");
 require_once("functions.inc");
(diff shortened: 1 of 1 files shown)


src/etc/rc.newipsecdns
```diff
@@ -36,21 +49,20 @@ require_once("auth.inc");
...
+sleep(3);
+log_error("IPSEC: One or more IPsec tunnel endpoints has changed its IP. Refreshing.");
+
+/* make sure we have the latest configuration changes loaded. */
+$config = parse_config();
 
 vpn_ipsec_configure();
 
 unlock($ipseclck);
-?>
```
jim-p: Why remove this line?

PiBa-NL: why not? afaik its not needed to have, and it avoids possible 'output' send by white space that might be there after it.. i can undo it if you want..

src/etc/rc.newipsecdns
```diff
@@ -36,21 +49,20 @@ require_once("auth.inc");
...
-if (file_exists("{$g['varrun_path']}/booting")) {
-	return;
-}
-
-if (ipsec_enabled()) {
-	sleep(15);
-	log_error("IPSEC: One or more IPsec tunnel endpoints has changed its IP. Refreshing.");
-} else {
+if (!ipsec_enabled()) {
+	unlock($ipseclck_pending);
```
rbgarga: IMO it would be better to check earlier if IPsec is enabled before try_lock() call and if it's not it doesn't need to try to lock

PiBa-NL: In the structure it would indeed be somewhat more logical, however calling try_lock is a lot 'cheaper' if it turns out that it can just 'return' than parsing the configuration into memory by the include_once calls below it.

If you want, i can move it around.

PR #3778, opened 2017-07-09
Fix #7682 system_authserver delete UI bugs RELENG_2_3

phil-davis:
After deleting an entry, regenerate the local array of auth servers based on the order in the config. That way the displayed list has indexes that correctly reflect the config, and so the next edit/delete actions will index correctly.


jim-p:
I committed and cherry-picked the simplified version I mentioned on https://github.com/pfsense/pfsense/pull/3780, thanks!

PR #3780, opened 2017-07-09
Fix #7682 system_authserver delete UI bugs

phil-davis:
After deleting an entry, regenerate the local array of auth servers based on the order in the config. That way the displayed list has indexes that correctly reflect the config, and so the next edit/delete actions will index correctly.

### DIFF
--- a/src/usr/local/www/system_authservers.php
+++ b/src/usr/local/www/system_authservers.php
@@ -32,6 +32,15 @@ require_once("guiconfig.inc");
 require_once("auth.inc");
 require_once("pfsense-utils.inc");
 
+function gen_auth_server_array() {
+	$a_servers = auth_get_authserver_list();
+	foreach ($a_servers as $servers) {
+		$auth_server_array[] = $servers;
+	}
+
+		return $auth_server_array;
+}
+
 // Have we been called to populate the "Select a container" modal?
 if ($_REQUEST['ajax']) {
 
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_authservers.php
```diff
@@ -103,11 +112,7 @@
 	$config['system']['authserver'] = array();
 }
 
-$a_servers = auth_get_authserver_list();
-
-foreach ($a_servers as $servers) {
-	$a_server[] = $servers;
-}
+$a_server = gen_auth_server_array();
```
jim-p: Looking at that original code, I'm not sure why it was done that way. That code, and the new function, could be replaced by `$a_server = array_values(auth_get_authserver_list());` and it would be a lot simpler.

src/usr/local/www/system_authservers.php
```diff
@@ -132,11 +137,11 @@
...
 	}
 
-	/* Remove server from temp list used later on this page. */
-	unset($a_server[$_POST['id']]);
-
 	$savemsg = sprintf(gettext("Authentication Server %s deleted."), htmlspecialchars($serverdeleted));
 	write_config($savemsg);
+
+	/* Recreate the locally-used list to close up the deleted entry. */
+	$a_server = gen_auth_server_array();
```
jim-p: Using `$a_server = array_values($a_server);` here would fix the key IDs, or you could use the same code as the previous comment to recreate it.

phil-davis:
I was too lazy to think about it, just put the existing inline code that generates $a_server into a function and call it where needed. Feel free to make it nicer if you like, and close this.

jim-p:
OK, I committed and cherry-picked the simplified version. Thanks for tracking this down!

PR #3782, opened 2017-07-12
nat, portforwards should not make up a new destination information when unkown

PiBa-NL:
nat, portforwards should not make up a new destination information when a configured dhcp interface does not currently have an address.

fixes: https://forum.pfsense.org/index.php?topic=127585.msg733528#msg733528

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2246,10 +2246,6 @@ function filter_nat_rules_generate() {
 			$srcaddr = trim($srcaddr);
 			$dstaddr = trim($dstaddr);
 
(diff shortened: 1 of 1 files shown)


vicomte:
This is a dumb question, but wouldn't this affect any default setup of 2.3 since the auto nat rules all make reference to 'WAN Address'?

PiBa-NL:
@vicomte does it? if you suspect a problem please verify the situation you think might go wrong. But afaik normally the address to act on should always be configured. If configuration doesnt match the active situation pfSense imho should not alter rules to 'something different'..

PR #3791, opened 2017-07-27
dhcp6c Advanced Config prefix interface

marjohn56:
Currently, when using dhcp6c advanced configuration the prefix interface is WAN, this is not very useful!

The changes here allow the user to enter the interface name, that's the REAL interface name, RE1 RE0 etc.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -4407,10 +4407,11 @@ function DHCP6_Config_File_Advanced($interface, $wancfg, $wanif) {
 			}
 			$id_assoc_statement_prefix .= ";";
 		}
-
+		
+		$realif = get_real_interface($wancfg['adv_dhcp6_prefix_selected_interface']);
 		if (is_numeric($wancfg['adv_dhcp6_prefix_interface_statement_sla_id'])) {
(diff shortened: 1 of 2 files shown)


jim-p:
Forcing the user to enter a physical interface is the wrong way to handle that kind of setting. It should be a drop-down list of interfaces and then the code can figure it out from there. Keep it abstracted.

marjohn56:
It's in the advanced config, only advanced users should be there. :)

jim-p:
Doesn't matter, it still is not something we can reasonably expect any user in the GUI to put in by hand. Especially not when the GUI can easily provide a selection list and figure out what to use.

marjohn56:
OK, the Mk II commit used a drop-down list.

marjohn56:
@rbgarga - I have some additions ready for dhcp6c, these allow raw options to be sent, hence the reason I found this issue. 

The upstream FreeBSD dhcp6c and the pfSense dhcp6c are not the same, I think there have been some changes upstream that are not reflected. Do you want me to PR to the upstream or the pfSense version?

rbgarga:
@marjohn56 can you please send me an email with more information about it and we can discuss the best approach? renato at netgate dot com is my email. Thanks!

PR #3792, opened 2017-07-30
status_queues, provide 'realtime' statistics

PiBa-NL:
status_queues, provide 'realtime' statistics
-retrieve 'current' numbers from pfSense

not using qstats provides the following advantages:
-no long lag which requires 18 updates to get 90% accurate values on screen
-showing queue's in 2 browsers does not show half the value

### DIFF
--- a/src/usr/local/www/getqueuestats.php
+++ b/src/usr/local/www/getqueuestats.php
@@ -1,0 +1,40 @@
+<?php
+/*
+ * getqueuestats.php
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2017 Rubicon Communications, LLC (Netgate)
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+##|+PRIV
+##|*IDENT=page-getqueuestats
+##|*NAME=AJAX: Get Queue Stats
+##|*DESCR=Allow access to the 'AJAX: Get Stats' page.
+##|*MATCH=getqueuestats.php*
+##|-PRIV
+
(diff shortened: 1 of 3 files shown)


src/usr/local/www/getqueuestats.php
```diff
@@ -0,0 +1,119 @@
...
+
+//required for convert_real_interface_to_friendly_interface_name() to get altq interface name..
+include_once("interfaces.inc");
+
+function get_queue_stats() {
+	// do to sysutils\qstats not providing accurate stats with 'currently' valid numbers 
+	// in its current implementation this php function does the job for now..
+	$result = array();
+	$result['timestamp'] = gettimeofday(true);
+	$r = exec("pfctl -s queue -v", $output, $ret);
```
jim-p: Should use the full path for `/sbin/pfctl`

src/usr/local/www/getqueuestats.php
```diff
@@ -0,0 +1,119 @@
...
+header("Expires: " . gmdate("D, j M Y H:i:s", time()) . " GMT");
+header("Cache-Control: no-cache, no-store, must-revalidate"); // HTTP/1.1
+header("Pragma: no-cache"); // HTTP/1.0
+
+require_once("auth_check.inc");
+
+//required for convert_real_interface_to_friendly_interface_name() to get altq interface name..
+include_once("interfaces.inc");
+
+function get_queue_stats() {
```
jim-p: Since this function is used in multiple files, is there any particular reason it needs to be here instead of in a library file like `/etc/inc/shaper.inc`?

src/usr/local/www/status_queues.php
```diff
@@ -136,13 +85,125 @@ function getqueueactivity() {
...
+		}
+		timestampprevious = timestamp;   
 	}
 
 	$(document).ready(function() {
 		setTimeout(getqueueactivity, 150);
 	});
 });
+
+function formatSpeedBits(speed) {
```
jim-p: This should probably be in `/etc/inc/util.inc` next to `format_bytes()` and `format_number()`.

PiBa-NL: It seems difficult to move.. <?=gettext("Kbps"); ?> doesnt work in pfSenseHelper.js as thats not parsed through php.. so maybe this is the best spot after all. even though it does duplicate javascript code thats also in graph.php. not sure what would be the proper way do de-duplicate this code..

PiBa-NL:
@jim-p @rbgarga @sbeaver-netgate , i have rebased this PR, can you guy's review and pull it please? thanks.

jim-p:

Thanks!

jim-p:
Redmine issue: https://redmine.pfsense.org/issues/8185

PR #3798, opened 2017-08-10
Update system_update_settings.php

IknowJoseph:
Tidied the Dashboard update settings text

### DIFF
--- a/src/usr/local/www/system_update_settings.php
+++ b/src/usr/local/www/system_update_settings.php
@@ -158,7 +158,7 @@ $section = new Form_Section('Updates');
 $section->addInput(new Form_Checkbox(
 	'disablecheck',
 	'Dashboard check',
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_update_settings.php
```diff
@@ -158,7 +158,7 @@ function get_repo_name($path) {
 $section->addInput(new Form_Checkbox(
 	'disablecheck',
 	'Dashboard check',
-	'Disable the automatic dashboard auto-update check',
+	'Disable the Dashboard auto-update check',
```
rbgarga: Dashboard shouldn't be capitalized IMO

IknowJoseph:
Thanks for the feedback, I went with D as I thought "Dashboard" was a
proper noun relating to that component. Happy to use whatever the
convention is, however.

PR #3802, opened 2017-08-18
PPP down script change

svenauhagen:
Hi,

in the PPP down script the line

/etc/rc.kill_states ${IF} ${LOCAL_IP}

is killing cellular modems when the script is executed.
The cuaXX port is blocked after this command and the mpd deamon can not talk to the modem anymore.
Only a complete modem reset (power down/power up) brings it back up.

Therefore I added an if statement to stop the command on PPP modems.

Please let me know if you can merge that in.

Best and thanks
Sven

Voleatech

### DIFF
--- a/src/usr/local/sbin/ppp-linkdown
+++ b/src/usr/local/sbin/ppp-linkdown
@@ -27,7 +27,7 @@ if [ -f /tmp/${IF}up ] && [ -f /conf/${IF}.log ]; then
 	/usr/local/sbin/ppp-log-uptime.sh $seconds ${IF} &
 fi
 
-if [$IF !=~ "ppp[0-9]*"]
+if [${IF} !=~ "ppp[0-9]*"]
 then
 	/etc/rc.kill_states ${IF} ${LOCAL_IP}
 fi
--- a/src/usr/local/sbin/ppp-linkdown
+++ b/src/usr/local/sbin/ppp-linkdown
@@ -27,7 +27,7 @@ if [ -f /tmp/${IF}up ] && [ -f /conf/${IF}.log ]; then
 	/usr/local/sbin/ppp-log-uptime.sh $seconds ${IF} &
 fi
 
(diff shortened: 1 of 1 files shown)


src/usr/local/sbin/ppp-linkdown
```diff
@@ -27,7 +27,9 @@ if [ -f /tmp/${IF}up ] && [ -f /conf/${IF}.log ]; then
 	/usr/local/sbin/ppp-log-uptime.sh $seconds ${IF} &
 fi
 
-/etc/rc.kill_states ${IF} ${LOCAL_IP}
+if [ ${IF} !=~ "ppp[0-9]*" ]; then
```
rbgarga: `test` (aka `[`) doesn't  support this operator `!=~` so this will not work on FreeBSD's `/bin/sh`

svenauhagen:
Hi,

I did not know that.
How about testing with egrep?

`if $(echo "${IF}" | egrep -v "ppp[0-9]+" > /dev/null)`

would that befine?
I did a quick test and that worked in sh.

Best
Sven

rbgarga:
Yeah, it's better. Just remove the >/dev/null and $() and use -q grep parameter:
`if echo "${IF}" | egrep -qv "ppp[0-9]+"`

src/usr/local/sbin/ppp-linkdown
```diff
@@ -27,7 +27,9 @@ if [ -f /tmp/${IF}up ] && [ -f /conf/${IF}.log ]; then
 	/usr/local/sbin/ppp-log-uptime.sh $seconds ${IF} &
 fi
 
-/etc/rc.kill_states ${IF} ${LOCAL_IP}
+if echo "${IF}" | egrep -qv "ppp[0-9]+"; then
```
jim-p: Please use the full path for `/usr/bin/egrep`, otherwise this seems ok.

svenauhagen:
Hi,

do you have any more change requests for this?

Best
Sven

PR #3803, opened 2017-08-19
Fix IPsec widget when mobile server is defined (Bug #7784)

doktornotor:


sbeaver-netgate:
Thanks Doktor. I merged your change manually and have confirmation that it was effective :)

doktornotor:
Thanks, 👍

PR #3805, opened 2017-08-26
Add reason to write_config() calls - interfaces*.php

doktornotor:


doktornotor:
@sullrich Probably, if someone understands the code better than me or has more time to dig into it. Nice to see you're still around. 😉 

P.S. Is the [bounty](https://redmine.pfsense.org/issues/204#note-1) still in place? 🤣

doktornotor:

PR #3812, opened 2017-08-30
Update head.inc

IknowJoseph:
Added rel='noopener' to _blank links because I don't trust the web. Imagining a situation where a pfsense admin is tricked into entering their credentials into a malicious third party site (I have not tested if this is possible, however).

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

### DIFF
--- a/src/usr/local/www/head.inc
+++ b/src/usr/local/www/head.inc
@@ -509,7 +509,7 @@ if (are_notices_pending()) {
 					['name' => 'Status',	     'menu' => $status_menu,	  'href' => null],
 					['name' => 'Diagnostics',    'menu' => $diagnostics_menu, 'href' => null],
 //					['name' => 'Gold',		     'menu' => $gold_menu,		  'href' => '_blank'],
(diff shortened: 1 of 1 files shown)


doktornotor:
_Imagining a situation where a pfsense admin is tricked into entering their credentials into a malicious third party site_ 

Yeah, then they already have the credentials. noopener or no noopener. WTF is this hype about?

IknowJoseph:
So I update an external help page (or other) to include:

window.opener.location = 'https://fakewebsite/pfsense/login.html';

user then returns to their pfsense admin tab and enters their user/name
password into my page. It's not a likely scenario, I admit, but it's not
hard to protect against either.

doktornotor:
This is just absurd. If someone thinks there's a security issue here, maybe get the browsers append that stuff automatically.

`$ grep -Rni _blank /usr/local/www | wc -l`
`     166`

sbeaver-netgate:
Not required

PR #3813, opened 2017-08-30
Remvoed dnsstuff.com link

IknowJoseph:
http://www.dnsstuff.com links were not working. I replaced the whois link and removed the "IP tools" as I did not know what it did.


doktornotor:
It works just fine here. And not understanding what something does is not a good reason to remove links to useful tools.

🤦‍♂️

IknowJoseph:
Apologies. At the time of the change it was producing an error. Looks fine
now.

PR #3817, opened 2017-08-30
Unbound Serve expired

marjohn56:
Serve expired – Records stay in cache after TTL expires, with a TTL value of 0, if a new lookup is requested the cached record will be served for maximum performance, but at same time the resolver will ask for a new value from upstream to refresh the value and TTL.


chrcoluk:
can confirm this has no issues, testing without problems.

PR #3826, opened 2017-09-21
Unbound - allow snoop from localhost

doktornotor:
dig +trace fails without this, which is super annoying for debugging/diagnostics/benchmarking or whatever similar purposes. Allowing both recursive and non-recursive queries should be of no security concern as long as it's localhost-only.

https://redmine.pfsense.org/issues/7884


jim-p:
Looks good and has the intended effect

PR #3828, opened 2017-09-24
Build sysutils/devcpu-data for CPU microcode updates

doktornotor:
Part of https://redmine.pfsense.org/issues/7890

Not exactly sure whether this is enough to get [cpuctl(4)](https://www.freebsd.org/cgi/man.cgi?query=cpuctl&sektion=4&manpath=FreeBSD+11.1-RELEASE) support in kernel. Kindly advise if additional bits are required in https://github.com/pfsense/FreeBSD-src/blob/devel-11/sys/amd64/conf/pfSense or elsewhere.


doktornotor:

PR #3830, opened 2017-09-24
pppoe, allow configuring pppoe on a carp interface so its only active on the master

PiBa-NL:
pppoe, allow configuring pppoe on a carp interface so its only active on the master

https://redmine.pfsense.org/issues/8184

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1764,6 +1764,27 @@ function handle_pppoe_reset($post_array) {
 	}
 }
 
+function restart_ppp_interfaces_using_interfaces($triggerinterfaces) {
+	global $config;
+	$ppp_list = array();
+	if (is_array($config['ppps']['ppp']) && count($config['ppps']['ppp'])) {
+		foreach ($config['ppps']['ppp'] as $pppid => $ppp) {
+			$ports = explode(",", $ppp['ports']);
+			foreach($ports as $port) {
+				foreach($triggerinterfaces as $vip) {
+					if ($port == "_vip{$vip['uniqid']}") {
+						$if = convert_real_interface_to_friendly_interface_name($ppp['if']);
+						$ppp_list[$if] = 1;
+					}
+				}
+			}
+		}
+	}
+	foreach($ppp_list as $pppif => $dummy) {
+		interface_ppps_configure($pppif);
+	}
+}
+
 /*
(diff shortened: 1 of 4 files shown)


rbgarga:

PiBa-NL:
@jim-p 
broken connectivity on 1 box is indeed expected, there are probably workarounds possible around that though it gets complicated fast. kinda similar though where you would have wan-interfaces on a private ip while having a carp-ip for the public ip's to reduce overhead there (i used to run past years with 5 carp-ips on the our isp /29 subnet.. and private addresses on the interfaces). My goal is to have 2 pfSense boxes where if 1 fails or needs maintenance it can be turned off without much extra effort internal routing between vlans continues, and though it does break connectivity to the internet for a minute for the ppp to connect back.. then does restore connectivity.

In my current/new situation we have got 1 static pppoe IP and a routed /29 subnet from our new ISP, sofar this change seems to work for us.. not sure if this falls under your 'dynamic wan' criteria? but i don't have a better idea on how to have 2 pfSense boxes 'share' the single pppoe connection. they cant both actively connect as only 1 will receive the traffic.
i hope it is oke for merging like this.

@rbgarga 
commit is rebased.

jim-p:
@PiBa-NL In that instance, you'd have the CPE perform PPPoE and expose the /29 directly on its back end (firewall WAN subnet), otherwise you get into a weird mess of broken connectivity and other complications. We are probably better off having this support in the tree, but I wouldn't encourage anyone to use it because of all the downsides. If they have no other choice, perhaps.

PiBa-NL:
@jim-p, i'm not sure about your CPE proposal.. our fiber-provider installed a https://genexis.eu/product/fibertwist/ in the building and they 'manage' that device remotely, they provides a vlan for our internet-provider. The internet-provider then provides the internet service over the vlan. And we need to plug in a rj45 cable to and use pppoe over that. I suppose it would probably be possible to put a gateway device like small draytek in between to do the pppoe connection. But it seemed 'cleaner' to do it straight from pfSense itself. Maybe indeed it shouldn't be encourage'd to do it like this, but it seems like a nice possibility to have.. Now the pfSense box has 9 usable public IP's instead of just 5.

jim-p:

Thanks!

mikes-gh:
What happened with this.  I am guessing its a common senario?

PR #3832, opened 2017-09-26
Fix for dhcpleases dupes

luckman212:
Same as f0f56a7f4ea9b6636102c380d2d210983a08c996 just retargeted at master

Fixes duplicate blocks of dhcp lease entries getting added to /etc/hosts. Further discussion: https://forum.pfsense.org/index.php?topic=131554.0

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -616,25 +616,26 @@ function system_dhcpleases_configure() {
 			$unbound_conf = "";
 		}
 
-		if (isvalidpid($pidfile)) {
-			/* Make sure dhcpleases is using correct unbound or dnsmasq */
-			$_gb = exec("/bin/pgrep -F {$pidfile} -f {$dns_pid}", $output, $retval);
-			if (intval($retval) == 0) {
-				sigkillbypid($pidfile, "HUP");
-				return;
-			} else {
-				sigkillbypid($pidfile, "TERM");
-			}
-		}
+		/* best to never send a HUP to dhcpleases, it is bugged */
+		// if (isvalidpid($pidfile)) {
+		// /* Make sure dhcpleases is using correct unbound or dnsmasq */
+		//	$_gb = exec("/bin/pgrep -F {$pidfile} -f {$dns_pid}", $output, $retval);
+		//	if (intval($retval) == 0) {
+		//		sigkillbypid($pidfile, "HUP");
+		//		return;
+		//	} else {
+		//		sigkillbypid($pidfile, "TERM");
+		//	}
+		// }
 
 		/* To ensure we do not start multiple instances of dhcpleases, perform some clean-up first. */
-		if (is_process_running("dhcpleases")) {
-			sigkillbyname('dhcpleases', "TERM");
-		}
+		mwexec("/bin/pkill -x dhcpleases", true, true, false);
 		@unlink($pidfile);
+		usleep(250000); // quarter-second pause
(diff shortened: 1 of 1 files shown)


src/etc/inc/system.inc
```diff
@@ -616,25 +616,26 @@ function system_dhcpleases_configure() {
...
+		// if (isvalidpid($pidfile)) {
+		// /* Make sure dhcpleases is using correct unbound or dnsmasq */
+		//	$_gb = exec("/bin/pgrep -F {$pidfile} -f {$dns_pid}", $output, $retval);
+		//	if (intval($retval) == 0) {
+		//		sigkillbypid($pidfile, "HUP");
+		//		return;
+		//	} else {
+		//		sigkillbypid($pidfile, "TERM");
+		//	}
+		// }
```
rbgarga: If the code is useless, just remove it instead of commenting it out

luckman212: Yes, true. I will submit an update to remove this.

luckman212: The `pkill -x` was to ensure that all instances were killed to avoid multiple instances (which is something I observed happening occasionally while I was debugging). But, if you prefer to use existing functions then I suppose `sigkillbyname('dhcpleases', "TERM")` does the same thing.

My thinking was that just killing it was a little faster than first calling `is_process_running()` and then `sigkillbyname()` since in the end we are just trying to kill all instances, regardless of whether 0, 1 or 10 of these procs exist.

src/etc/inc/system.inc
```diff
@@ -616,25 +616,26 @@ function system_dhcpleases_configure() {
...
 
 		/* To ensure we do not start multiple instances of dhcpleases, perform some clean-up first. */
-		if (is_process_running("dhcpleases")) {
-			sigkillbyname('dhcpleases', "TERM");
-		}
+		mwexec("/bin/pkill -x dhcpleases", true, true, false);
```
jim-p: There isn't much to be gained by going this route. sigkillbyname() does its own check if a process is running first but it's a fairly fast pgrep followed by `killall`, which is roughly the same as `pkill -x`.

The only change needed is to remove the router is_process_running() since sigkillbyname() already does that. And though sigkillbyname does use mwexec itself, we prefer to use functions for those where possible so that if we find a better way to implement that action, we don't need to hunt down all of the shell execs.

src/etc/inc/system.inc
```diff
@@ -616,25 +616,26 @@ function system_dhcpleases_configure() {
...
 		@unlink($pidfile);
+		usleep(250000); // quarter-second pause
 		mwexec("/usr/local/sbin/dhcpleases -l {$g['dhcpd_chroot_path']}/var/db/dhcpd.leases -d {$config['system']['domain']} -p {$g['varrun_path']}/{$dns_pid} {$unbound_conf} -h {$g['etc_path']}/hosts");
-	} elseif (isvalidpid($pidfile)) {
-		sigkillbypid($pidfile, "TERM");
+		
+	} else {
+		mwexec("/bin/pkill -x dhcpleases", true, true, false);
```
jim-p: Same as above, use `sigkillbyname('dhcpleases', "TERM");` here.

jim-p:

Thanks!

sbeaver-netgate:
Please advise if you expect to be able to complete the requested changes.

sbeaver-netgate:
Please resubmit when you are able to get to these changes

PR #3836, opened 2017-09-27
Improve flexibility of DHCPD config

colinligertwood:
- New checkbox "Virtual IPs" to allow dhcpd to use ipalias subnets
- Allow creation of static mappings on ipalias subnets

I've minimally tested this. It's mostly for comment at this point. 

This is a very minimal effort to allow creation of dhcpd static host configurations on ipalias subnets. It is enabled by checking the "Virtual IPs" checkbox on the DHCPD service configuration view. Once enabled, the user will be able to create static mappings with addresses that match any subnet bound to the interface, including ipalias address subnets.

In order to facilitate this change, once enabled, the dhcpd.conf is modified to include all subnets of a given interface within a "shared-network" block named for the interface the subnets live on. No options are allowed within ipalias subnets.

This is as far as I'd like to go with this feature, but it could be expanded to allow for "Additional Pools" to be created within the ipalias subnets. I did some work towards this, but have stopped as it becomes very difficult to validate the new pools and all of their options without a major refactor if services_dhcp.php.

Know issues:
- Once "Virtual IPs" is disabled, the out of subnet static mappings remain
- dhcpd.conf indentation is inconsistent

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -688,6 +688,18 @@ EOPP;
 		$subnet = gen_subnet($ifcfgip, $ifcfgsn);
 		$subnetmask = gen_subnet_mask($ifcfgsn);
 
+		$vipaliases = Array();
+		if (is_array($config['virtualip']['vip'])) {
+			foreach ($config['virtualip']['vip'] as $vipent) {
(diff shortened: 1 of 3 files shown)


rbgarga:
Please re-send it against branch `master`. We don't do direct commits on stable branches

PR #3837, opened 2017-09-27
Improve flexibility of DHCPD configuration

colinligertwood:
- New checkbox "Virtual IPs" to allow dhcpd to use ipalias subnets
- Allow creation of static mappings on ipalias subnets

I've minimally tested this. It's mostly for comment at this point. 

This is a very minimal effort to allow creation of dhcpd static host configurations on ipalias subnets. It is enabled by checking the "Virtual IPs" checkbox on the DHCPD service configuration view. Once enabled, the user will be able to create static mappings with addresses that match any subnet bound to the interface, including ipalias address subnets.

In order to facilitate this change, once enabled, the dhcpd.conf is modified to include all subnets of a given interface within a "shared-network" block named for the interface the subnets live on. No options are allowed within ipalias subnets.

This is as far as I'd like to go with this feature, but it could be expanded to allow for "Additional Pools" to be created within the ipalias subnets. I did some work towards this, but have stopped as it becomes very difficult to validate the new pools and all of their options without a major refactor if services_dhcp.php.

Know issues:
- Once "Virtual IPs" is disabled, the out of subnet static mappings remain
- dhcpd.conf indentation is inconsistent

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1049,6 +1049,9 @@ EOD;
 }
 
 EOD;
+		if (isset($config['dhcpd'][$dhcpif]['enable_aliases'])) {
+			$dhcpdconf .= "}\n";
+		}
 
 		/* add static mappings */
 		if (is_array($dhcpifconf['staticmap'])) {
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -651,6 +651,18 @@ EOPP;
 		$subnet = gen_subnet($ifcfgip, $ifcfgsn);
 		$subnetmask = gen_subnet_mask($ifcfgsn);
 
+		$vipaliases = Array();
+		if (is_array($config['virtualip']['vip'])) {
+			foreach ($config['virtualip']['vip'] as $vipent) {
+				if ($vipent['mode'] != 'ipalias') continue;
+				if ($vipent['interface'] != $dhcpif) continue;
+				$vipalias = Array();
+				$vipalias['subnet'] = $vipent['subnet'];
+				$vipalias['subnet_bits'] = $vipent['subnet_bits'];
+				array_push($vipaliases, $vipalias);
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


colinligertwood:
This is a pull request to master. It's untested.

src/etc/inc/services.inc
```diff
@@ -651,6 +651,18 @@ EOPP;
 		$subnet = gen_subnet($ifcfgip, $ifcfgsn);
 		$subnetmask = gen_subnet_mask($ifcfgsn);
 
+		$vipaliases = Array();
+		if (is_array($config['virtualip']['vip'])) {
+			foreach ($config['virtualip']['vip'] as $vipent) {
+				if ($vipent['mode'] != 'ipalias') continue;
```
jim-p: Technically speaking, CARP VIPs should be viable for this as well.

src/etc/inc/services.inc
```diff
@@ -763,7 +775,15 @@ EOPP;
...
 			}
 		}
-
+		if (isset($config['dhcpd'][$dhcpif]['enable_aliases'])) {
+			$dhcpdconf .= "shared-network {$dhcpif} {\n";
+			foreach ($vipaliases as $vipalias) {
+				$vipsubnet = gen_subnet($vipalias['subnet'], $vipalias['subnet_bits']);
+				$vipsubnetbits = $vipalias['subnet_bits'];
+				$vipsubnetmask = gen_subnet_mask($vipalias['subnet_bits']);
+				$dhcpdconf .= "subnet {$vipsubnet} netmask {$vipsubnetmask} {}\n";
```
jim-p: You also need to consider failover here. Though it is only being used for static addresses in this setup, it should probably still have the failover options present. Needs testing, though.

jim-p:

Thanks!

sbeaver-netgate:
No action since November 2017. Please resubmit if you are able to accommodate the requested changes.

PR #3841, opened 2017-10-03
[RELENG_2_3_4] Honor login page color

doktornotor:
The System - General Setup - webConfigurator - Login page color setting is ignored without this.


doktornotor:

PR #3843, opened 2017-10-04
Added Granularity for Firewall Logging Settings

kfiducia:
Allows snort and virusprot table logs to be enabled independently from "default"
If "Log Default Blocks" is enabled the functionality is unchanged. If not set, a user can log snort/virusprot table blocks without all of the default blocks.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -3129,6 +3129,12 @@ function filter_rules_generate() {
 	if (isset($config['syslog']['nologdefaultpass'])) {
 		$log['pass'] = "log";
 	}
+	if (isset($config['syslog']['logsnortblock']) || !isset($config['syslog']['nologdefaultblock'])) {
+		$log['snortblock'] = "log";
+	}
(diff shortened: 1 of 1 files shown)


doktornotor:
Somehow missing the settings...

kfiducia:
The settings _could_ be added to the user interface, if that what you're getting at. I intended this as something that users can add, for now, through the command prompt page. A much easier method than editing the filter.inc file and keeping track of it through system updates in order to see when snort blocks something. I figured I would wait to see if its something that would be accepted before spending a bunch of time building the UI and testing it. "Missing" settings shouldn't change current functionality if users choose not to use the option. 

Adding the settings is pretty simple:
$config['syslog']['logsnortblock'] = ''; write_config('Logging snort blocks.');

Does that make sense?

kfiducia:
I'll add the GUI portion.

jim-p:

Thanks!

sbeaver-netgate:
No activity in 3+ months. Please reopen if still desired

PR #3844, opened 2017-10-04
Don't create unwanted gateways for OpenVPN interfaces

luckman212:
This patch adds a feature to the OpenVPN client & server config pages that allows specifying what types of gateways are automatically created for _assigned_ interfaces. This helps to keep the System>Routing page shorter and easier to read, especially when a user has many OpenVPN connections defined.

Tested on 2.4.0RC and 2.4.1 snaps.

_further discussion on forum_
https://forum.pfsense.org/index.php?topic=137317.0

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -699,15 +699,28 @@ function return_gateways_array($disabled = false, $localhost = false, $inactive 
 			default:
 				$tunnelif = substr($ifcfg['if'], 0, 3);
 				if (substr($ifcfg['if'], 0, 4) == "ovpn") {
-					// if current iface is an ovpn server endpoint then check its type, skip tap only
-					if (substr($ifcfg['if'], 4, 1) == 's') {
-						$ovpnid = substr($ifcfg['if'], 5);
-						if (is_array($config['openvpn']['openvpn-server'])) {
-							foreach ($config['openvpn']['openvpn-server'] as & $ovpnserverconf) {
-								if ($ovpnserverconf['vpnid'] == $ovpnid) {
-									if ($ovpnserverconf['dev_mode'] == "tap") {
-										continue 3;
-									}
+					switch (substr($ifcfg['if'], 4, 1)) {
+						case "c":
+							$ovpntype = "openvpn-client";
+							break;
+						case "s":
+							$ovpntype = "openvpn-server";
+							break;
+						default:
+							// unknown ovpn type
+							continue 2;
+					}
+					$ovpnid = substr($ifcfg['if'], 5);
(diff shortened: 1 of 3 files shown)


src/etc/inc/gwlb.inc
```diff
@@ -699,15 +699,28 @@ function return_gateways_array($disabled = false, $localhost = false, $inactive
...
 				if (substr($ifcfg['if'], 0, 4) == "ovpn") {
-					// if current iface is an ovpn server endpoint then check its type, skip tap only
-					if (substr($ifcfg['if'], 4, 1) == 's') {
-						$ovpnid = substr($ifcfg['if'], 5);
-						if (is_array($config['openvpn']['openvpn-server'])) {
-							foreach ($config['openvpn']['openvpn-server'] as & $ovpnserverconf) {
-								if ($ovpnserverconf['vpnid'] == $ovpnid) {
-									if ($ovpnserverconf['dev_mode'] == "tap") {
-										continue 3;
-									}
```
rbgarga: Above code is only dealing with `openvpn-server` items, and below it's also considering `openvpn-client`. Could you please elaborate a bit more why are you changing that behavior?

rbgarga: Above code is only dealing with openvpn-server items, and below it's also considering openvpn-client. Could you please elaborate a bit more why are you changing that behavior?

luckman212: The patch was created to eliminate unwanted gateways from OpenVPN interfaces (client *or* server). The old code only considered servers so both V4+V6 client gateways were _always_ created, which was a bit messy. The added code considers both scenarios. The first block is for V4 and the later block handles V6.

luckman212:
@jim-p It would be trivial to add a "none" option if you think that is useful...

src/usr/local/www/vpn_openvpn_client.php
```diff
@@ -899,6 +905,37 @@
...
+
+	$group->add(new Form_Checkbox(
+		'create_gw',
+		null,
+		'IPv6 only',
+		($pconfig['create_gw'] == "v6only"),
+		'v6only'
+	))->displayAsRadio();
+
+	$group->setHelp('If you assign a virtual interface to this OpenVPN client, ' .
```
jim-p: The more I consider this, I do think a 'none' option should be added as long as the code is being changed. It looks like it should be fairly easy to accommodate.

jim-p: But since this was already merged, that can be done in a separate PR.

luckman212: Will do a PR for that @jim-p - thanks

PR #3845, opened 2017-10-08
Unbound - add a checkbox to optionally serve expired responses (Feature #7814)

doktornotor:
See https://redmine.pfsense.org/issues/7814


marjohn56:
I already did this Doc, #3817 :)

doktornotor:
Yeah I redid it because I disliked your variable's naming. 👅 

On a more serious note, someone really needs to go through the piling PRs and merge them.

marjohn56:
:)

I think we must be patient, getting 2.4 out as it stands must take priority. 

However, after that it's going to get busy here..

PR #3849, opened 2017-10-13
Interfaces is the same word on english and spanish.

cmmora:
"Interfases" is not correctly on spanish


rbgarga:
We cannot make changes direct to .po files. Zanata is the colaboration tool we use to take care of it. You can see instructions about how to contribute at https://www.netgate.com/blog/pfsense-software-translations-with-zanata.html

Thanks

PR #3854, opened 2017-10-20
Remove leftover CLA bits, use consistent uppercase

doktornotor:


doktornotor:

PR #3855, opened 2017-10-22
Don't leave sshlockout_pf process orphans when restarting syslogd (Bug #7984)

doktornotor:
Also remove redundant checks while here, since they are already done by sigkillbypid().

### DIFF
--- a/src/etc/inc/service-utils.inc
+++ b/src/etc/inc/service-utils.inc
@@ -700,14 +700,12 @@ function service_control_stop($name, $extras) {
 			sigkillbyname("relayd", "TERM");
 			break;
 		case 'syslogd':
-			if (isvalidpid("{$g['varrun_path']}/syslog.pid")) {
-				sigkillbypid("{$g['varrun_path']}/syslog.pid", "TERM");
-				usleep(100000);
-			}
-			if (isvalidpid("{$g['varrun_path']}/syslog.pid")) {
-				sigkillbypid("{$g['varrun_path']}/syslog.pid", "KILL");
-				usleep(100000);
-			}
+			sigkillbypid("{$g['varrun_path']}/syslog.pid", "TERM");
+			usleep(100000);
+			sigkillbypid("{$g['varrun_path']}/syslog.pid", "KILL");
+			usleep(100000);
+			// Bug #7984: do not leave sshlockout_pf process orphans
+			sigkillbyname("sshlockout_pf", "TERM");
 			break;
 		default:
 			stop_service($name);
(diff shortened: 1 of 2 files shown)


src/etc/inc/service-utils.inc
```diff
@@ -700,14 +700,12 @@ function service_control_stop($name, $extras) {
...
-				usleep(100000);
-			}
-			if (isvalidpid("{$g['varrun_path']}/syslog.pid")) {
-				sigkillbypid("{$g['varrun_path']}/syslog.pid", "KILL");
-				usleep(100000);
-			}
+			sigkillbypid("{$g['varrun_path']}/syslog.pid", "TERM");
+			usleep(100000);
+			sigkillbypid("{$g['varrun_path']}/syslog.pid", "KILL");
+			usleep(100000);
```
PiBa-NL: If TERM was successful, and process stopped running no need kill and sleep again? And if for some reason its not running at all, no need to try the TERM+sleep either.? So why make the sleep unconditional? If waiting isn't needed i would prefer to avoid it where possible.

doktornotor: It won't kill anything if not running. It does exact same thing like before, without a bunch of duplicate checks. Plus 0.1s delay which of course is a huge deal.

doktornotor:
Removed the usleep() calls after SIGKILL from both files since they were equally useless before and now. Waiting for the outcome won't terminate the sshlockout_pf process regardless of sleeping or not.

PiBa-NL:
I agree it wont matter for sshlockout_pf. i do wonder why they were made put there in the first place.. Is it guaranteed that the process immediately gone and OS sockets cleaned up when the SIGKILL is executed? So a new syslogd can startup without any issue.?

It still seems the first sleep is sill useless if syslogd wasn't running at that time the original isvalidpid() check was not 'redundant' for that part imho, even though most of the time syslogd would be running, during bootup it will add at 100ms of doing nothing.. though small it can be better with little effort. actually it seems you put effort in to make it worse by removing good working code, that is what surprises me.
Probably only adding the `sigkillbyname("sshlockout_pf", "TERM");` would have been enough to 'fix' my reported issue. But was wondering if that is the proper fix or syslogd maybe should keep track if its children or something different..

Nonetheless thanks for the PR. Ill leave it at that, and see what the official devs think about it.

doktornotor:
Well, IMHO doing `isvalidpid()` which in turn calls `sigkillbyname()` which in turn does `isvalidpid()` yet again is just braindead. The entire overhead is probably very much close to those 100ms that noone cares about anyway. YMMV.

doktornotor:

PR #3856, opened 2017-10-22
Add option to disable HSTS for nginx (Bug #6650)

doktornotor:
As far as webconfigurator is concerned, current default behavior remains unchanged (HSTS enabled with HTTPS) if the toggle is unset.

If HSTS is desired for CP, I can add a toggle there as well. IMNSHO, it's pretty much never desired since it makes testing or changing your mind very much impossible as you have no control over the CP clients in general.

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -425,7 +425,7 @@ function captiveportal_init_webgui_zone($cpcfg) {
 		}
 		system_generate_nginx_config("{$g['varetc_path']}/nginx-{$cpzone}-CaptivePortal-SSL.conf",
 			$crt, $key, $ca, "nginx-{$cpzone}-CaptivePortal-SSL.pid", $listenporthttps, "/usr/local/captiveportal",
-			"cert-{$cpzone}-portal.pem", "ca-{$cpzone}-portal.pem", $cpzone);
+			"cert-{$cpzone}-portal.pem", "ca-{$cpzone}-portal.pem", $cpzone, false);
 	}
 
 	/* generate nginx configuration */
@@ -436,7 +436,7 @@ function captiveportal_init_webgui_zone($cpcfg) {
 	}
 	system_generate_nginx_config("{$g['varetc_path']}/nginx-{$cpzone}-CaptivePortal.conf",
 		"", "", "", "nginx-{$cpzone}-CaptivePortal.pid", $listenporthttp, "/usr/local/captiveportal",
-		"", "", $cpzone);
+		"", "", $cpzone, false);
 
 	@unlink("{$g['varrun']}/nginx-{$cpzone}-CaptivePortal.pid");
 	/* attempt to start nginx */
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1335,12 +1335,13 @@ function system_webgui_start() {
 			$portarg = "443";
 		}
 		$ca = ca_chain($cert);
+		$hsts = isset($config['system']['webgui']['disablehsts']) ? false : true;
 	}
 
 	/* generate nginx configuration */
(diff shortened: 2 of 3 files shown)


NOYB:
> If HSTS is desired for CP, I can add a toggle there as well. IMNSHO, it's pretty much never desired since it makes testing or changing your mind very much impossible as you have no control over the CP clients in general.

Once a client has received the HSTS cookie you're still out of luck to change your mind.  Until client clears that cookie the associated host is going to https.  Regardless of any changes made to the server.
As for not meeting testing needs, customize the test environment.

For HSTS enabled clients, the proper solution is to use https services, or use a different host name for accessing non-https services.

HSTS is performing it's job very well.  Better to be inconvenienced to setup services properly than to disable security features and some day find out have been going along with stuff inadvertently in the clear.

There are appropriate viable solutions to this perceived HSTS issue without this feature.
```
if (security.value < convenience.value) { fail(); }
```

miken32:
The “proper solution” is to allow users to customize the firewall as they see fit, which this patch does, and not to impose an onerous security system on users without any choice in the matter.

NOYB:
HSTS enabled clients using a different host name than the router for non-https services is hardly onerous.  Plus it actually does solve the issue of changing your mind.  Where as this feature does not.  Just change host name.

Users can customize the firewall all they want.  It's open source.  Mine is highly customized with features I like and find useful.

This patch does not solve the stated problem.  It merely alters and shifts it to some degree in some circumstances.

This feature to disable HSTS for sake of convenience when there are easy, appropriate, viable capabilities to set up services appropriately such that they are not affected by the routers HSTS, is misguided as a solution.

A service that is undesirably affected by the router GUI access HSTS is a poorly designed service.  Use a different host name.  Not the router host name.

doktornotor:
There is a number of issues caused by hard-coded HSTS mentioned on the linked bug, which include captive portals, undesirable issues with pfSense packages such as HAproxy (you cannot enable HSTS globally there because you end up sending the header twice) or bandwidthd (does not support HTTPS at all, it's completely broken out-of-the-box with workarounds involving complex HAproxy setup.)

Use a different hostname doesn't cut it. The hardcoded HSTS with unconfigurable max-age of 1 year is just broken design, sorry.

jim-p:
Given the special way that HSTS gets handled by browsers, it can be too much of a good thing in certain cases. Enough cases that I think the option is worth including.

NOYB:
What is broken is those packages.  Accommodating broken packages prolongs their broken state and all but ensures they will never be fixed.  They should not be relying on the router hostname if they cannot handle how it may be configured.  Allowing them to dictate and prevent the use of HSTS by the router's GUI is broken design.  Properly designed they should not be dependent on this nor require that it be disabled.  They are broken!  Sorry!

Ah ain't that just terrible that security feature would require complex setup.  Such a shame.  Any network engineer worth their salt can handle it.

If using a different hostname that the client has not associated with an HSTS cookie doesn't cut it then HSTS is not the issue.

But apparently the tradition of brokenness accommodation and convenience at expense of security is continuing, alive and well.

jim-p:
No, we merely acknowledge that we do not live in a perfect world. Outside of the lab, there are real requirements that can't always be that strictly enforced. The good thing is that you don't have to touch a thing to keep the same behavior it has now. The option doesn't hurt you, so just ignore it and move on.

NOYB:
Actually, in a way it does potentially hurt me, and others.  Maybe I, and others, would like to make use of some of those broken packages if they were fixed rather than requiring HSTS to be disabled for the router's GUI.  Accommodating them in their broken state practically ensures they will never be fixed.  Yes that is an impact to those who would like to use them, but don't want to disable HSTS on the router's GUI to do so.

This is a like a bacteria laced band-aid that will ensure the wound doesn't heal.

doktornotor:

PR #3857, opened 2017-10-22
pf, retry pfctl -f rule loading when pf is 'busy'

PiBa-NL:
pf, retry pfctl -f rule loading when pf is 'busy', don't try and fail to force -d -e as that would also fail at this point in time..

Some background info..  https://forum.pfsense.org/index.php?topic=138123.msg757793#msg757793

It seems to be related to this issue reported on FreeBSD 10.4 where also a 'retry' is advised as the proper thing to do. https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=223093

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -422,11 +422,29 @@ function filter_configure_sync($delete_states_if_needed = true) {
 		$mt = microtime();
 		echo "pfctl being called at $mt\n";
 	}
-	unset($rules_loading, $rules_error);
-	$_grbg = exec("/sbin/pfctl -o basic -f {$g['tmp_path']}/rules.debug 2>&1", $rules_error, $rules_loading);
-	if (isset($config['system']['developerspew'])) {
-		$mt = microtime();
-		echo "pfctl done at $mt\n";
+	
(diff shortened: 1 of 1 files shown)


src/etc/inc/filter.inc
```diff
@@ -422,11 +422,31 @@ function filter_configure_sync($delete_states_if_needed = true) {
...
 	}
-	unset($rules_loading, $rules_error);
-	$_grbg = exec("/sbin/pfctl -o basic -f {$g['tmp_path']}/rules.debug 2>&1", $rules_error, $rules_loading);
-	if (isset($config['system']['developerspew'])) {
-		$mt = microtime();
-		echo "pfctl done at $mt\n";
+	
+	$pfctl_try = 0;
+	while ($pfctl_try < 10) {
+		$pfctl_try++;
```
jim-p: It has the same net effect, but using `for` is preferred for a loop with a set number of iterations. It would combine these three lines into one and be easier to read:
`for ($pfctl_try = 0; $pfctl_try < 10; $pfctl_try++) {`

PiBa-NL:
@jim-p, new commit pushed with a `for` loop as requested, indeed looks better :)

PR #3827, opened 2017-09-23
Allow specific gateways to be excluded from default gateway switching

luckman212:
_Rebase/update of my old PR #3609 -- trying again!_

This commit adds a feature to allow you to optionally prevent gateways from being elected as default during Gateway Switching operations in a multi-WAN environment when the "Enable default gateway switching" option is enabled. Some possible use cases:

- non-internet-facing gateways
- metered connections
- VPN gateways

This only controls the default gateway of the firewall itself; it does not change the behavior of Policy Based Routing.

**Changes/new features**

-adds a `no_defgw_switch` flag to gateway config that will cause it to be excluded from the default gateway selection process
-indicate this setting (as an icon) in System > Routing > Gateways
-updated help text for Default gateway switching in System > Advanced > Misc to better describe the usage of that feature
-adds some debug logging for gw change events
-setting a gateway as default on the edit page automatically toggles/hides this option, as having both enabled would not be sane

_Files changed:_
/etc/inc/gwlb.inc
/usr/local/www/system_advanced_misc.php
/usr/local/www/system_gateways.php
/usr/local/www/system_gateways_edit.php

PR #3781, opened 2017-07-11
Gateways, allow for configuring a gatewaygroup as the default gateway.

PiBa-NL:
Gateways, allow for configuring a gatewaygroup as the default gateway.
-Avoid changing routes by just visiting a webgui page.
-Avoid change some unneeded events when nothing changed.
  
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8187
- [x] Ready for review

PR #3861, opened 2017-11-01
Allow gateways to be excluded from default gw switching

luckman212:
Unified default gw switching + notifications patch
supersedes PR#3827 and PR#3609, targets 2.4.2+/master


jim-p:
Looking at the comments on #3827, I don't think we actually need this now. After 2.4.2 we will probably pull in #3781 which would make this unnecessary. @rbgarga ?

rbgarga:
true, let me close this one then

PR #3862, opened 2017-11-01
Set VLAN Priority on dhcp6c packets

marjohn56:
Add VLAN tag to outgoing dhcp6c packets. Option setting in WAN dhcp6c section.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -3100,7 +3100,7 @@ function filter_generate_user_rule($rule) {
 }
 
 function filter_rules_generate() {
-	global $config, $g, $FilterIflist, $time_based_rules, $GatewaysList, $tracker;
+	global $config, $g, $FilterIflist, $time_based_rules, $GatewaysList, $tracker, $vlanprio_values;
 
 	$fix_rule_label = 'fix_rule_label';
 	$increment_tracker = 'filter_rule_tracker';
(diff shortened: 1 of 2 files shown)


src/etc/inc/filter.inc
```diff
@@ -3100,8 +3100,9 @@ function filter_generate_user_rule($rule) {
 }
 
 function filter_rules_generate() {
-	global $config, $g, $FilterIflist, $time_based_rules, $GatewaysList, $tracker;
+	global $config, $g, $FilterIflist, $time_based_rules, $GatewaysList, $tracker, $vlanprio_values;
 
+	$wancfg = $config['interfaces']['wan'];
```
rbgarga: Why is it using `'wan'` interface hardcoded here?

marjohn56: Don't know... copy paste finger trouble perhaps.

It's Removed.

marjohn56: I had a muppet moment... try the update.

marjohn56:
Removed fixed 'wan' and used the existing routine to get the info required.

kwillers:
Finally back in France where this functionality is needed with Orange FTTP service 

Works like a charm

marjohn56:
Good stuff.

We still need to get the modified dhcp6c client sorted too, that will finally end the problems with Orange France.

kwillers:
Indeed if we can get the modified dhcp6c client into the main build along with the modified dhclient client then Orange fttp users in France will able to upgrade via the GUI and not need manually replace the files each time

The two modified clients are very stable as my firewall has been up without issues for 3 months

marjohn56:
I have issues a PR for dhcp6c changes on hrs-allbsd/wide-dhcpv6

Add the ability to use RAW parameters in the config PR #8

jim-p:

Thanks!

marjohn56:
Redmine 8200

marjohn56:
Redmine #8200

jim-p:
Redmine Issue: https://redmine.pfsense.org/issues/8200

PR #3863, opened 2017-11-02
routing, support use of recursive network aliases in static routes

PiBa-NL:


src/etc/inc/util.inc
```diff
@@ -2460,21 +2486,8 @@ function get_staticroutes($returnsubnetsonly = false, $returnhostnames = false,
...
-					if (is_ipaddrv4($net)) {
-						$net .= "/32";
-					} else if (is_ipaddrv6($net)) {
-						$net .= "/128";
-					} else if ($returnhostnames === false || !is_fqdn($net)) {
-						continue;
-					}
-				}
+			$alias_subnets = alias_to_subnets_recursive($route['network'], $returnhostnames);
+			foreach ($alias_subnets as $net) {
```
jim-p: Instead of creating a single use variable there, just call that function directly in the foreach:

`foreach (alias_to_subnets_recursive($route['network'], $returnhostnames) as $net) {`

The original code had the same issue but it's better to fix it than repeat the error.

PiBa-NL: That one is done.

PR #3818, opened 2017-09-06
Adding Route 53 v6 

chewrocca:
Adding option to use Route 53 v6 for Dynamic DNS services.

PR #3869, opened 2017-11-17
Updated PR to add Route 53 IPv6 dynamic DNS support

jsl:
This PR adds a commit to the work in #3818 to make it mergeable with upstream changes.

I have submitted a signed CLA.


jim-p:
#3818 was already merged, can you fix this PR so it's only the difference between that PR and what you want to see, or is it all OK now? If it's OK in the repo now, this can be closed.

PR #3870, opened 2017-11-18
dhcp6c lock files not removed after unclean shutdown when using "Do not wait for an RA" on WAN interface

jtl999:
https://redmine.pfsense.org/issues/8106


marjohn56:
It does not matter whether it is using 'Do not wait for RA' or not. The lock is set in either mode and should be deleted if it exists during boot before the interfaces are brought up.

jtl999:
Very true

PR #3874, opened 2017-11-24
Ensure only one instance of system_dhcpleases_configure runs concurrently

NewEraCracker:
I wonder if this is also a possible fix for dhcpleases dupes.

Please test.
Thanks in advance.

Possible alternative to PR: #3832


jim-p:
Closing in favor of #3832

PR #3877, opened 2017-11-24
Enable use of auth script for getting dynamic credentials on openvpn

ssbarnea:
Patch for enabling the ability to specify auth scripts that do generate user/password for the openvpn client, enabling OTP users to configure persistent VPNs.

https://redmine.pfsense.org/issues/8122

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1074,6 +1074,11 @@ function openvpn_reconfigure($mode, $settings) {
 			if (!($settings['auth_user'] && $settings['auth_pass'])) {
 				$userpass .= "\n";
 			}
+
+			// if auth_script is not empty, we use its output instead of user/pass
+			if ($settings['auth_script']) {
+				$userpass = shell_exec($settings['auth_script']);
+			}
 			file_put_contents($up_file, $userpass);
 		}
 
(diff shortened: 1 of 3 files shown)


jim-p:

Thanks!

ssbarnea:
@jim-p If you can provide some feedback on this PR and eventually the linked bug, it would be highly appreciated.

ssbarnea:
@jim-p Please suggest alternatives to this. As far as I know, running a script is the only possible to enable OTP authentication.

Now what would be maybe more important from the security point of view is to find a way to contain this, as in running this code as a user with minimal permissions.

Here is my current script:
```
echo -e "johndoe\n5bprefix`/usr/local/bin/oathtool --totp -b ABCDEF`"
```

As you can imagine different users may have different ways to obtain or generate OTP credentials, so trying to put these into fields would now work for them. Imagine that maybe someone may want to use `curl` to get the credentials using a HTTP request.

Any ideas?

jim-p:
At the moment I do not have any alternative suggestions to offer. Whatever is added must be self-contained in the configuration.

sbeaver-netgate:
Please resubmlit if you are able to accommodate the requested changes.

ssbarnea:
@sbeaver-netgate @jim-p I think that you created a chicken and the egg kind of issue here, and here are few facts:

1. user needs to be able to write script for obtaining OTP credentials
2. it is up to the user to assure the executable exists
3. the export/import config issue is still valid for an endless number of issue, for example how about crontab entries that are defined by user? Obviously that if user moves config from one device to another which does not have the tools used installed, the script will fail.
4. I do not see a case where import/export would break a router, even if this is performed on older version that does not support this feature.
5. the requirement for self-contained doesn't hold water because you have no control over what user writes in the script.

Yep, an `echo -e "username\npassword"` script would be self contained. Does this help?

jim-p:
1-2 fail because they require the user to maintain something outside of config.xml.
3. Cron entries are stored in config.xml, so that is not a valid example.
4. Having a directive in the OpenVPN config that calls a script which doesn't exist is an error
5. It isn't what they write that is the problem, it's a concern, but the problem is that the script content is not included in the configuration itself. That is the self-contained issue.

The moment the user _is required_ to touch something outside of config.xml to restore, you have broken the entire model on which the self-contained configuration was designed.

PR #3878, opened 2017-11-26
Increase FPM process availability in high ram systems

marjohn56:
To reduce chance of nginx gateway error when interacting with FPM backend, this patch does the following on systems with 1 gig or more of ram:

1. Starts up extra FOM server processes at startup
2. Allows more to stay running on standby
3. Increases automatic shutdown time from 5 seconds to one hour.

Thanks to Chris Collins for this one. 

Redmine #8125


rbgarga:
@marjohn56 please rebase it

jim-p:
This looked OK, but two unrelated files had permissions changed. If you resubmit it, make sure not to include those permission changes in the PR.

marjohn56:
That's why I closed this PR and issued a new one. My PC Github Desktop did a nasty and messed up several things/

marjohn56:
New PR #3881

PR #3879, opened 2017-11-29
Widget refresh fix; minor widget UI tweaks

scizzr:
executewidget
- Renamed function to executewidgets for grammar's sake
- Moved ajaxspecs.length check out of the loop - if ajaxspecs.length isn't more than 0 at page load, there's no reason to keep checking it. can always move the check back into executewidgets function if there ever arises a case where widgets can be inserted into ajaxspecs dynamically.
- Changed to a setInterval instead of a self-calling setTimeout - code is cleaner that way
- Fixed logic in widget loop - previously, the setTimeout loop was only incrementing ajaxidx every time it was executed, which means that the function is only checking one single widget every second. If you have 6 widgets registered (my case, for example), that means that the loop only comes back around to a given widget (log.widget.php for example) every 6th run. So if you have your interval set to 1, it actually only refreshes every 6 seconds; if the interval is set to 2, it refreshes every 12 seconds. Changed ajaxidx to a for loop, checking each widget every setInterval execution.

make_ajax_call
- Improved logic flow - should be doing your failure checks, then continuing if no failures are detected
- Cleaned up unnecessary variable ajaxtimeout - don't need to capture this in a variable if it's not checked elsewhere
- Cleaned up unnecessary variable ajaxmutex - you shouldn't use a global widget lock; see next point for replacement
- Added a per-widget lock, which expires regardless of success or failure

### DIFF
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -608,8 +608,8 @@ function gen_customwidgetinterval_div($widgetinterval) {
 }
 
 function set_customwidgetinterval(& $user_settings) {
-	if ($_POST['interval']) {
-		$user_settings['widgets'][$_POST['widgetkey']]['interval'] = trim($_POST['interval']);
+	if ($_POST['interval'] && is_numeric($_POST['interval'])) {
+		$user_settings['widgets'][$_POST['widgetkey']]['interval'] = int_val($_POST['interval']);
 	} else {
 		unset($user_settings['widgets'][$_POST['widgetkey']]['interval']);
 	}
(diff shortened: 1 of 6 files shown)


jim-p:

Thanks!

rbgarga:
@scizzr do you still plan to work on this? If yes please see @sbeaver-netgate's comment or just close it

NobleKangaroo:
Closing PR; will resubmit it when I have more time to work out the minutiae of everything.

scizzr:
Closing PR; will re-submit at a later date.

PR #3885, opened 2017-12-04
Fix #2345 by setting ptr-domain, ddnsdomainkeyname and ddnsdomainkey

JoeriCapens:


JoeriCapens:
Wrong bug #, should be #6319

PR #3887, opened 2017-12-07
Add DHCP Dynamic DNS key algorithm choice. Implements #6621

JoeriCapens:
This adds a choice for the key algorithm used by DHCP to push DNS updates. Implements [#6621](https://redmine.pfsense.org/issues/6621)

Since this requires a new "ddnsdomainkeyalgorithm" in the config, I'm not sure how to make sure this doesn't break an upgrade. Should the default (hmac-md5) be used in function dhcpdkey($dhcpifconf) when $dhcpifconf['ddnsdomainkeyalgorithm'] is empty? Or should this default be added to the config file during an upgrade? If so, how?

Any other feedback is more than welcome!

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -1101,8 +1101,7 @@ $section->addInput(new Form_Input(
 	'DDNS Domain',
 	'text',
 	$pconfig['ddnsdomain']
-))->setHelp('Leave blank to disable dynamic DNS registration.%1$s' .
-			'Enter the dynamic DNS domain which will be used to register client names in the DNS server.', '<br />');
+))->setHelp('Enter the dynamic DNS domain which will be used to register client names in the DNS server.');
 
 $section->addInput(new Form_Checkbox(
 	'ddnsforcehostname',
@@ -1144,7 +1143,7 @@ $section->addInput(new Form_Input(
 	'DNS Domain key secret',
 	'text',
 	$pconfig['ddnsdomainkey']
-))->setHelp('Dynamic DNS domain key secret (HMAC-MD5) which will be used to register client names in the DNS server.');
+))->setHelp('Dynamic DNS domain key secret which will be used to register client names in the DNS server.');
 
 // Advanced MAC
 $btnadv = new Form_Button(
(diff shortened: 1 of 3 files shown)


JoeriCapens:
It's probably best to use hmac-md5 as a default when the value is not specified in the general config file. This makes an upgrade go smooth and it allows to use an old config (i.e. from a backup) without problems. I've added commit "Set default key algorithm to hmac-md5" for this.

src/usr/local/www/services_dhcp.php
```diff
@@ -313,15 +314,14 @@ function validate_partial_mac_list($maclist) {
 	if ($_POST['maxtime'] && (!is_numeric($_POST['maxtime']) || ($_POST['maxtime'] < 60) || ($_POST['maxtime'] <= $_POST['deftime']))) {
 		$input_errors[] = gettext("The maximum lease time must be at least 60 seconds and higher than the default lease time.");
 	}
-	if (($_POST['ddnsdomain'] && !is_domain($_POST['ddnsdomain']))) {
+	if ($_POST['ddnsupdate'] && !is_domain($_POST['ddnsdomain'])) {
```
jim-p: If you are changing the test to key off that checkbox, then the descriptive text under the DDNS Domain field can be changed to reflect that. Currently it says "Leave blank to disable dynamic DNS registration" which isn't true any longer.

JoeriCapens: Thanks! Changed the help text for both DHCPv4 and v6. I also changed the help text for the DHCPv4 key secret which still mentioned HMAC-MD5.

There are quite some differences in wording between the DHCPv4 and v6 pages ("Primary domain name server..." vs. "Enter the primary domain name server...", "DNS Domain key" vs. "DDNS Domain Key name", etc.) but I won't change those here. I'll create another branch and pull request later with proposals to make the wording consistent.

yamlCase: This change now seems to (hopefully inadvertently) enforce the requirement of a A valid domain key name, algorithm and secret when none is desired.  The particular use-case is running the 'bind' pfsense package in a sandboxed enviornment.  If the desire is to enforce secrets, I can certainly open a feature request, however as it is I am unable to make configuration changes to services_dhcp.php without this error presenting itself.

I am prepared to be properly scolded or ignored if this is the wrong place to raise the issue.  Thanks!

jim-p: This is a PR that was merged nearly a year ago, so this is definitely not the right place. Start a thread on the forum at https://forum.netgate.com and describe your use case and what isn't working like you expect. If there is a bug or something that could be handled better, it can eventually find its way into a redmine entry, but it should be discussed on the forum first.

PR #3890, opened 2017-12-09
Fix #6319 again by adding missing dns-servers

JoeriCapens:
Not sure how I missed this but the change in pull request #3886 is wrong. The $reversezone['dns-servers'] array needs to be set for dhcpdzones() to create the correct reverse zone definition in dhcpdv6.conf


jim-p:
Redmine Issue: https://redmine.pfsense.org/issues/6319

PR #3892, opened 2017-12-17
Fixes #7413: Some DHCPv6 leases are not displayed in the GUI

al-right:
. Better handling/parsing of ISC dhcpv6 leases file and removal of unnecessary properties.
. Experimental Pools/failover handling in the parsing of DHCPv6.
. Cleanup on the Status DHCVPv6 Leases page.
. Correction re. Mac address sent to 'Services' - 'Wake-on-LAN' - 'Edit' page when clicking the + button for 'Add WOL mapping' on the Status DHCPv6 Leases page.
. IPv6 matcher/parser and unit testing of it.
. Small debugging tool for inspecting lease file content.

Corresponds to this commit
https://github.com/al-right/pfSense-dhcpv6-gui-leases-patch/tree/5501616d5b9fc7bcad855b0dadd9d2fef8b4fa43

Please read parser/README.md and https://redmine.pfsense.org/issues/7413#note-11 for more notes about the changes.

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/7413
- [ ] Ready for review


src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+			$inner_debug = function ($rerun, $var, $params) use (&$inner_debug) {
+			        static $has_utf8, $index_mark;
+			        if (!isset ($has_utf8)) {
+			          // For testing try: setLocale(LC_CTYPE, 'C');
+			          $has_utf8 = preg_match ('/utf-?8$/i', setLocale(LC_CTYPE, 0));
+			          if ($has_utf8) {
+			            $index_mark['last'] = index_mark['utf8']['last'];
+			            $index_mark['normal'] = index_mark['utf8']['normal'];
+			          }
+			          else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+			          }
+			        }
+
+				static $the_choice;
+				if (!isset($the_choice)) {
+					$the_choice = function ($is_last, $spaces, $key, $value) use (&$the_choice, &$inner_debug, &$index_mark) {
+						if ($is_last) {
+							$selected_char = $index_mark['last'];
+						}
+						else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+					};
+				}
+
+				static $spaces_func;
+				if (!isset($spaces_func)) {
+					$spaces_func = function ($is_last, $string, &$spaces) use (&$index_mark) {
+						if ($is_last) {
+							$selected_char = $index_mark['last'];
+						}
+						else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+					if (is_array ($var)) {
+						// Print out identification that indicates an array
+						echo "[ ]\n";
+						$dec_counter = count ($var);
+						foreach ($var as $key => $value) {
+							$dec_counter-=1;
+							$the_choice ( ! $dec_counter, '', $key, $value);
+						}
+					}
+					else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+						if (isset ($var)) {
+  							echo "$var\n";
+						}
+						else {
+							echo "undefined@7\n";
+						}
+					}
+				} // Or are we given a $var who is not an array
+				else if ( ! is_array ($var)) {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+
+					// If yes, print it
+					if (setnlen ($var, 0)) {
+						echo "$var\n";
+					}
+					else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+							// handle if the value is an array
+							$the_choice ( ! $dec_counter, $spaces, $pkey, $value);
+						}
+						else {
+							// parameter not an array var
+							$the_choice ( ! $dec_counter, $spaces, $pkey, $value);
+						}
+					}
+				}
+				else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+
+								// key2 does not need to be printed as well
+								// since it is our lookup table for $var.
+								// Here we are namely only interested in the values.
+								if (is_array ($value2)) {
+									$inner_debug (1, $var, $value2); // And here is an exception to our NOTICE above! :)
+									// no problems with spacing as we are traversing params for
+									// indexes.
+								}
+								else {
```
jim-p: Brace style

src/usr/local/www/parser/tools_for_debugging_n_testing.inc
```diff
@@ -0,0 +1,237 @@
...
+									// use value2 (indirect from $params) as the look up index
+									// in $var and use the_choice to return the tree of return
+									// values.
+									$the_choice ( ! $dec_counter, '', $value2, (setnlen ($var, $value2) ? $var[$value2] : 'undefined@9'));
+								}
+							}
+						}
+						else {
```
jim-p: Brace style

smos:
It works! I can see the leases. Kind of handy with the neigh impossible hoops that MacOS makes you jump through.

smos:
Is the "Add static mapping" button missing? It doesn't appear for any lease at this moment. The correlation between the mac address is a lot better, I can now refer some of the static v4 leases to the v6 leases. Which helps.

al-right:
Hi all

smos: Please tell me more about 'Is the "Add static mapping" button missing?'
It is the blue + (plus) with the white background.
I see the button on pfSense 2.4.2 (stable). I guess you might be running a nightly build?
I have not made any amends to this.

Currently I am amending the requested changes and I hope to be done with
that in the comming weekend (if nothing gets in my way - that does happen)
The list:
* Making requested changes to follow coding guide lines re. input from jim-p.
* Making assigning of values and the checking more streamlined so that there is less text when assigning, and the checking is done in a similar way (but hidden away in function calls). E.g. if a value has already been set then that may be interesting to know, because this may be an indication that something is wrong and therefore helpful when debugging with the command line version of the status dhcpv6 leases 'page' that I made, because...
* I have identified a problem re. when a device shares its device DUID between several of its interfaces (which is quite okay for a device to do). Windows 10 behaves this way. Each interface has a unique IAID, but share the DUID. I saw this in a real life example on our local network and checked the setup of that computer (WiFi+cable, no bridging).
So the problem in code is that we map the IP address with the DUID:
$mappings[$entry['duid']] = $entry['ip'];
, where in fact we need to map the IP to both the DUID+IAID.
But it is fairly easy to fix.
Normally Windows 10 is not used as a router, because $mappings is only used for mapping IA-NA to IA-PD (meaning showing the route to the prefix), but maybe other onlink routers have more interfaces (connected to the same pfsense box) that is shown on the common Status DHCPv6 Leases page and therefore it seems fair/makes sense to fix the above and also to avoid setting the same $mappings[$entry['duid']] e.g. twice with different values.

In the example above it seems like both leases are active, but only one interface on link.
But it seems logical, because the computer can choose which interface should have the default route, but maybe it may confuse some. In the example the dhcpv6 lease updates of the interfaces happened almost at the same time typically just with about 10 seconds in difference.
On our network we use both DHCPv6+stateless autoconf.
The computers I guess mostly choose an address from the stateless autoconf (typically the random privacy extended one.)
This also means (I guess all of you know) that we may have active leases to interfaces, but those addresses are not showing online, because they are never used even though the interfaces are on link (using another address.)

It would be powerful to be able to combine/link mac addresses from the NDP table with the info from the DHCPv6 leases page so that all registered addresses (stateless as well as dhcpv6 requested addresses) were linked to one mac, but I do not see how to accomplish that - anyway out of scope for this patch. The DHCPv6 lease page is kind of giving half the info in a typical scenario with DHCPv6+stateless autoconf.

One could also argue that maybe wake on lan could be added to the NPD (and ARP) table(s).
But this is also out of scope for this patch.

al-right:
I have updated the PR with yet another patch.
Please see the commit log message for notes,
changes and things to consider for the future.

Applying the newest patch with System Patches
won't work. I guess it is because of the files
that moved.
Please check out the patch and e.g. scp the
files to their target locations when testing.

Thanks

jim-p:
Looks good to me so far. I had a number of leases on my edge firewall that were not showing before, and now they all show up with this patch applied.

The diff didn't want to apply at first because the copyright years are changed in this PR (when they shouldn't be) and they were updated in base as well. Github doesn't seem to care, merging would probably ignore it, but it threw off manual patching.

src/etc/inc/parser_ipv6.inc
```diff
@@ -0,0 +1,242 @@
+<?php
+/*
+ * tools_for_debugging.inc
```
rbgarga: filename is wrong here

al-right:
Thanks for testing guys!
Please just say if you need any further amends!

PR #3868, opened 2017-11-12
interfaces_assign.php UI speedup as requested on forum bounty board

loonylion:
Improves significantly and hopefully fixes bug#6400

Refactored interfaces_assign.php to to benefit people with large numbers of VLANs, as requested on the forum at https://forum.pfsense.org/index.php?topic=137391.0. Also contains a minor speedup for interfaces_vlan.php. Modified functions are contained in interfaces_fast.inc. Profiling code is still present but commented out, as is replaced code.

PR #3894, opened 2018-01-03
fix bugs 

loonylion:
Fixes:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8223
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8222
- [x] Ready for review

Fixes bugs causing previous pull request (https://github.com/pfsense/pfsense/pull/3868) to be reverted


jim-p:
Since the old PR was reverted, a new PR should include both the old changes and the new changes so it can be brought back in as a single piece

loonylion:
How do I achieve that?

PR #3895, opened 2018-01-06
openvpn, fix the ifconfig ip1 ip2 when subnet /31 is used

PiBa-NL:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8261
- [x] Ready for review


PiBa-NL:
What i was wondering, does that spot also need a check for  =ipv4 or (=ipv6 and >= 126) .?

jim-p:
No, there isn't an IPv6 point-to-point mode, it always behaves like tap/subnet topology

PR #3903, opened 2018-01-22
filter.inc, make filter_expand_alias_array() return consistent results between first and second call.

PiBa-NL:
filter.inc, make filter_expand_alias_array() return consistent results between first and second call.

Currently the 1st php call to: filter_expand_alias_array('mixed_ip_dns_alias') does not return the fixed IP's but a second call to filter_expand_alias_array('mixed_ip_dns_alias') does.. This is at least 'unexpected'.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8290
- [x] Ready for review

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -693,7 +693,6 @@ function filter_generate_nested_alias_recurse($name, $alias, &$aliasnesting, &$a
 				$filterdns["{$address}{$name}"] = "pf {$address} {$name}\n";
 			}
 		}
-		$finallist = '';
 	}
 
 	return $finallist;


rbgarga:
@PiBa-NL do you mind to share instructions about how to reproduce it?

PiBa-NL:
For a reproduction, create a alias the_alias with 2 entries:  1.2.3.4/32, domain.tld

Then run the diagnostics/php command:
```
include_once('filter.inc');
print_r(filter_expand_alias_array('the_alias'));
print_r(filter_expand_alias_array('the_alias'));
```

Result
```
Array
(
    [0] =>
)
Array
(
    [0] => 1.2.3.4/32
)
```
Expected result would be to be the same twice..

Other actual scenario that happens with haproxy package is written in the redmine ticket: https://redmine.pfsense.org/issues/8290

rbgarga:
@PiBa-NL Can you try this patch and let me know how it goes? https://gist.github.com/rbgarga/b74db181dd9a4b8ef58ff8d960426d6a

rbgarga:
Just remember to revert your original patch before...

rbgarga:
I've pushed a different fix. Lets keep track of tests in redmine.

PR #3905, opened 2018-01-29
Fix dyndns update with gateway group

pulcov:
If dynamic dns interface is a gateway group, DNS entry is update only when primary route is up again. When primary route down, nothing happen.
This patch fix this issue.


sbeaver-netgate:
Testing completed by sd738

jim-p:
This was missing a redmine entry, I added one for it at https://redmine.pfsense.org/issues/8333

PR #3908, opened 2018-02-26
Resolves #7905 - Asynchronous Authentication Plugin

pdemonaco:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/7905
- [x] Ready for review
- [x] Plugin port contributed to FreeBSD upstream

This correction replaces the call to auth-user-pass-verify with a plugin that has identical symantics and implements the deferred auth hook. The source of the plugin can be found in my other repository [here](https://github.com/pdemonaco/auth-script-openvpn/tree/pass-script-arguments).

Note that this has been tested with the following OpenVPN authentication configurations:
- [x] Remote Access (SSL/TLS + User Auth) - Local Database
- [x] Remote Access (User Auth) - External RADIUS + Duo MFA
- [x] Remote Access (SSL/TLS) - Obviously this isn't impacted.

I'm running this in production at my company as of this afternoon on 2.4.2-RELEASE. Note that the openvpn.inc file of that build was manually modified to match this change as the current master has significant other changes.

Please note that before this change can be merged some mechanism to build and install the actual plugin shared library must be added. I'm currently thinking it would make sense to create a BSD port dedicated to this purpose but I have a few questions before I go down that road.

1. How can a port be added to the standard install? 
2. Should I attempt to contribute the port to the upstream FreeBSD repo or can I add it directly to the pfsense fork?

Alternatively, it might make sense to add the plugin as a patch to the openvpn port. 

Any feedback would be appreciated.


jim-p:
About the port, the best course of action would be to submit it to FreeBSD directly for inclusion in the ports tree. We prefer to keep as much as possible upstreamed. If that doesn't work out, however, you can submit it as a pull request to our https://github.com/pfsense/FreeBSD-ports/ repository directly.

pdemonaco:
Thanks Jim, I'll get that process started. 

After taking a closer look at the structure of the core pfsense repository I think I'll need to add the package to the [poudriere_bulk](https://github.com/pfsense/pfsense/blob/master/tools/conf/pfPorts/poudriere_bulk) list so it's included in the base build. Is this correct?

rbgarga:
@pdemonaco You can add it as a dependency of security/pfSense port and it will be built.

I can help you to get it added into FreeBSD ports tree before we merge this change. Let me know when you submit the bugzilla ticket.

pdemonaco:
@rbgarga I'll let you know as soon as I have it ready - it may take me a bit as I haven't created a port for FreeBSD before and I'm not very familiar with the package manager.

I'm assuming you're referring to the FreeBSD bugzilla?

rbgarga:
Yes

pdemonaco:
@rbgarga I've submitted my patch as [bug 226492](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=226492)

I've also redeployed the plugin on my production environment using the library as compiled by the port on my build system.

rbgarga:
Great! I noted another committer already took it so lets wait until it's added to the tree and then we cherry-pick it to pfSense repository

pdemonaco:
Sounds good - I'll update this when it's commited!

pdemonaco:
@rbgarga Do you have any thoughts on the inclusion or exclusion of the **SSP_CFLAGS=    stack-protector-strong** flag in the port Makefile? The committer who picked up my bug report asked about it and I haven't heard back in a few days.

pdemonaco:
@rbgarga Is there anything we can do to speed up the review process on the FreeBSD side? I'd like to get this closed out before 2.4.4 if possible so I don't have to continue manually integrating my fix on each update.

rbgarga:
Since another FreeBSD developer already took the ticket I cannot interfere. We need to wait. It will not take too long and it will be plenty time before 2.4.4

pdemonaco:
Fair enough. Is there anything I could do to make the port more likely to be approved based on the content of comment log for the FreeBSD ticket?

pdemonaco:
@rbgarga the FreeBSD port has been committed!

I'm assuming there are a couple more changes I'll need to make in this pull request to add a requirement for the inclusion of that port. Can you point me in the right direction?

rbgarga:
Port has been added as a dependency of main pfSense meta-pkg on devel branch, it will be available on 2.4.4 snapshots soon. After that we can go ahead and start testing this change

pdemonaco:
Great! Let me know if there's anything I can do to help!

PR #3912, opened 2018-03-05
Add support for interface groups in nat rules.

mhalden:
This pr adds support for interface groups for nat. This is useful when having multiple equal cost wan interfaces as you won't have to duplicate all your nat rules.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/1933
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/firewall_nat_1to1_edit.php
+++ b/src/usr/local/www/firewall_nat_1to1_edit.php
@@ -429,29 +429,6 @@ $section->addInput(new Form_Checkbox(
 	$pconfig['nobinat']
 ))->setHelp('Excludes the address from a later, more general, rule.');
 
-$iflist = get_configured_interface_with_descr(true);
-
-foreach ($iflist as $if => $ifdesc) {
-	if (have_ruleint_access($if)) {
-		$interfaces[$if] = $ifdesc;
-	}
-}
-
-if ($config['l2tp']['mode'] == "server") {
-	if (have_ruleint_access("l2tp")) {
-		$interfaces['l2tp'] = gettext("L2TP VPN");
-	}
-}
-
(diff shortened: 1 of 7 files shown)


src/usr/local/www/firewall_nat_1to1_edit.php
```diff
@@ -461,7 +461,7 @@ function dsttype_selected() {
 	'interface',
 	'*Interface',
 	$pconfig['interface'],
-	$interfaces
```
jim-p: You should be able to remove the code that builds `$interfaces`, or replace it with `$interfaces = build_interface_list();` and then leave this part as `$interfaces`, depending on what else uses $interfaces.

src/usr/local/www/firewall_nat_edit.php
```diff
@@ -702,7 +702,7 @@ function dsttype_selected() {
 	'interface',
 	'*Interface',
 	$pconfig['interface'],
-	$interfaces
```
jim-p: See previous note about the code that builds `$interfaces`.

src/usr/local/www/firewall_nat_out_edit.php
```diff
@@ -504,7 +504,7 @@ function build_target_list() {
 	'interface',
 	'*Interface',
 	$pconfig['interface'],
-	$interfaces
```
jim-p: See previous note about the code that builds `$interfaces`.

mhalden:
Thank you for the feedback.

I have removed the code related to $interfaces (that I somehow managed to miss) as suggested. The function was renamed to avoid a naming collision which would break IPsec phase 1 and there is a fix for an UI bug which would make the interface name be empty in the list of nat-rules.

Considering I got the function originally from firewall_rules_edit.php, I changed that and firewall_rules.php (which used yet another way to get the same list) to use the same function.

PR #3914, opened 2018-03-09
igmp, Add option to disable the igmp service (fix item name..)

PiBa-NL:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8356
- [x] Ready for review


jim-p:
Already pushed fixes for these separately, but thanks!

PR #3916, opened 2018-03-18
Change CloudFlare to Cloudflare

robjarsen:
This commit also includes misc. whitespace removal on the affected files.

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -326,7 +326,7 @@ $section->addInput(new Form_Checkbox(
 	'Cloudflare Proxy',
 	'Enable Proxy',
 	$pconfig['proxied']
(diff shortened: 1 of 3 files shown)


src/usr/local/www/services_dyndns_edit.php
```diff
@@ -323,12 +323,12 @@ function build_if_list() {
 
 $section->addInput(new Form_Checkbox(
 	'proxied',
-	'CloudFlare Proxy',
+	'Cloudflare Proxy',
 	'Enable Proxy',
 	$pconfig['proxied']
-))->setHelp('Note: This enables CloudFlares Virtual DNS proxy.  When Enabled it will route all traffic '.
+))->setHelp('Note: This enables Cloudflares Virtual DNS proxy.  When Enabled it will route all traffic '.
```
jim-p: This should probably be "... enables the Cloudflare Virtual DNS proxy", to avoid the awkward use of their name in a possessive.

PR #3918, opened 2018-03-27
Patch for specifying router preference for the default ipv6 route pushed by radvd.conf

tkrajca:
This patch simply adds AdvRoutePreference to the default route pushed by radvd.conf so that router preference is honored.

This fixes https://redmine.pfsense.org/issues/6237 and https://redmine.pfsense.org/issues/7204.

The issue is that on systems with multiple RAs with different priorities (master and backup, for example), the default route/gateway might be taken from the first received RA because the explicit default ::/0 route doesn't specify priority, i.e. defaults to medium for both/all RAs.

Alternative fix would be to exclude the route ::/0 {} directive altogether as the RA advertiser is taken as default gateway by default.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -232,6 +232,17 @@ function services_radvd_configure($blacklist = array()) {
 			}
 		}
 		$radvdconf .= "\troute ::/0 {\n";
+		switch ($dhcpv6ifconf['rapriority']) {
+			case "low":
+				$radvdconf .= "\t\tAdvRoutePreference low;\n";
+				break;
+			case "high":
+				$radvdconf .= "\t\tAdvRoutePreference high;\n";
+				break;
+			default:
+				$radvdconf .= "\t\tAdvRoutePreference medium;\n";
+				break;
+		}
 		$radvdconf .= "\t\tRemoveRoute on;\n";
 		$radvdconf .= "\t};\n";
 


jim-p:
If the user sets the priority in the GUI, that already sets the default priority for advertised routes (`AdvDefaultPreference`), which should be carried through to this automatically. Are you sure this change has any effect beyond what the GUI already does?

Seems like a GUI option to disable that block entirely would be useful, though.

tkrajca:
Thanks @jim-p, 

> If the user sets the priority in the GUI, that already sets the default priority for advertised routes (AdvDefaultPreference), which should be carried through to this automatically

That is unfortunately not the case (I run pfSense 2.4.2). I would also assume that radvd would automatically default to the **AdvDefaultPreference** for the explicit routes but unfortunately it defaults to **medium** (if not set). tcpdump confirms that the RAs are advertised with **medium** preference if unset.

It would definitely be useful to be able to disable the entire block from the GUI but I was not sure what the implications of that would be, I don't quite understand why the block is there at all when not having it there should have the same effect. I did git blame, the code block was put there in 

commit 2a5960b07bb2a40aeaf81423b0c52d5629e87e49
Author: Luiz Otavio O Souza <luiz@netgate.com>
Date:   Mon Feb 8 19:22:40 2016 -0600

but there is no reference to the radvd config change in the commit message.

tkrajca:
Hmm, I could track it further to commits dc131dfe2d2 and 8859c0a69b4, my guess is that it might have been put there for the **RemoveRoute on** option so that the route gets uninstalled as soon as the corresponding RA goes offline.

jim-p:
Ah, ok. This should be good given how that behaves then. A GUI option for the other part could be a separate PR.

tkrajca:
Thanks for quick feedback @jim-p, it actually took me about two years to work out what the issue was :) I first suspected it could be linux kernel and some misconfiguration on our side as it worked flawlessly on CentOS 6 but CentOS 7 would always take the first RA regardless of priority. It turns out CentOS 6 doesn't understand the route ::/0 {} RA option so it happily ignored it and looked at the AdvDefaultPreference.

PR #3920, opened 2018-03-29
interfaces, aliases on lo0 should get configured at bootup

PiBa-NL:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8393
- [x] Ready for review


loos-br:
@PiBa-NL thanks for the quick fix!

I committed a slightly different patch as I noted that my last commit was wrong.

The checks that I added already exists in interface_ipalias_configure(), so I just copied them to interface_carp_configure(), which allowed me to reduce some duplicate code.

PR #3924, opened 2018-03-31
filter vip usage, ipv6 vips cause invalid rules because a empty item …

PiBa-NL:
…gets added to the vips list for a interface

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8408
- [x] Ready for review

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -1152,6 +1152,7 @@ function filter_generate_optcfg_array() {
 					if (!is_array($oic['vips'])) {
 						$oic['vips'] = array();
 					}
+					$oic['vips'][$vipidx]['mode'] = $vip['mode'];
 					$oic['vips'][$vipidx]['ip'] = $vip['subnet'];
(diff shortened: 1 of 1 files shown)


PiBa-NL:
Well it solves the cause of the problem (in this case), and the mode wasn't being set for vips6 array.. Not sure where its used but probably something got fixed there as well by this change. Ill add a commit to do some actual sanity checking though.

PiBa-NL:
Actually looking a bit closer, there seems no need for sanity checking as the check for a valid ipv4 / ipv6 ip is already done when building the 'vips' array.:  https://github.com/pfsense/pfsense/blob/2d113b128f270302cc5380669f257e8bd1cb3a15/src/etc/inc/filter.inc#L1159 already checks for valid ipv4 ip's and on 1169 for a valid v6 one.? So imho there is no need to check again ?

jim-p:
That may be OK then, I'm just trying to avoid any possible way it could end up with empty/invalid data in the line.

PR #3927, opened 2018-04-02
Redmine #8418: Enable OCSP Stapling

peterberbec:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8418
- [ ] Ready for review

Here's a patch to enable OCSP stapling and add an option in the "System - advanced - admin access" to enable and disable this option.


src/etc/inc/system.inc
```diff
@@ -1504,6 +1504,12 @@ EOD;
 		$nginx_config .= "\t\tadd_header X-Content-Type-Options nosniff;\n";
 		$nginx_config .= "\t\tssl_session_tickets off;\n";
 		$nginx_config .= "\t\tssl_dhparam /etc/dh-parameters.4096;\n";
+		if ($config['system']['webgui']['webgui-ocsp-staple'] !== false) {
+			$nginx_config .= "\t\tssl_stapling on;\n";
+			$nginx_config .= "\t\tssl_stapling_verify on;\n";
+			$nginx_config .= "\t\tresolver 8.8.8.8 1.1.1.1 valid=300s;\n";
```
jim-p: The resolver addresses should not be hardcoded, and definitely shouldn't default to 8.8.8.8 and 1.1.1.1. There should be configured DNS servers to use in the system general settings. Or failing that, 127.0.0.1 so it will use the DNS Resolver or Forwarder. Basically, use the same logic that is used to create resolv.conf.

jim-p:
I'm not sure if it's something we can parse out with the available functions but it might be nicer to either:
* Automatically enable this if the user chooses a certificate with must-staple for the GUI
or
* Present an error if the GUI certificate requires it, but the option is disabled.

src/etc/inc/system.inc
peterberbec: Sounds smart. I'll hunt around for that logic and change this.

peterberbec:
1. nginx throws a warning if the stapling config is present and the cert doesn't have the chain. Is there a way to check the nginx log?

2. nginx does not throw any error if the cert requires ocsp, but the browser does. Problem is, once this error comes up, you cannot get to the webgui while browsing using a url with domain, you'd have to go via IP.

Is there a way to tell, from an uploaded certificate if OCSP stapling is required?

jim-p:
That's what I'm not 100% sure on. We have functions to retrieve some certificate properties but I don't have a cert with must-staple on it to check against, to see if it comes up in any of the property lists.

peterberbec:
If you tell me how you do the test, I can check with my certificate.

peterberbec:
OK I grabbed the code used to generate resolv.conf and put it into the section that generates the nameservers for the resolver line.

Should this get broken out as a seperate function, now that it is being used twice? 
`function get_dns_servers()
// do all teh stuff
return $dns_server_array;`

src/etc/inc/system.inc
peterberbec: Added.

jim-p:
Ideally it would be best to have that in a separate function rather than duplicating the code.

As for the stapling property, basically I'd need to see something like the output of this

    echo var_export(openssl_x509_parse("your cert string"), true);

peterberbec:
`array ( 'name' => '/CN=berb.ec',
'subject' => array ( 'CN' => 'berb.ec', ),
'hash' => '09366148',
'issuer' => array ( 'C' => 'US', 'O' => 'Let\'s Encrypt', 'CN' => 'Let\'s Encrypt Authority X3', ),
'version' => 2,
'serialNumber' => '355220098167100636559376717253396255280753',
'validFrom' => '180331070058Z',
'validTo' => '180629070058Z',
'validFrom_time_t' => 1522479658,
'validTo_time_t' => 1530255658,
'signatureTypeSN' => 'RSA-SHA256',
'signatureTypeLN' => 'sha256WithRSAEncryption',
'signatureTypeNID' => 668,
'purposes' => array (
	1 => array ( 0 => true, 1 => false, 2 => 'sslclient', ),
	2 => array ( 0 => true, 1 => false, 2 => 'sslserver', ),
	3 => array ( 0 => false, 1 => false, 2 => 'nssslserver', ),
	4 => array ( 0 => false, 1 => false, 2 => 'smimesign', ),
	5 => array ( 0 => false, 1 => false, 2 => 'smimeencrypt', ),
	6 => array ( 0 => false, 1 => false, 2 => 'crlsign', ),
	7 => array ( 0 => true, 1 => true, 2 => 'any', ),
	8 => array ( 0 => true, 1 => false, 2 => 'ocsphelper', ),
	9 => array ( 0 => false, 1 => false, 2 => 'timestampsign', ),
	),
'extensions' => array (
	'keyUsage' => 'Digital Signature', 'extendedKeyUsage' => 
	'TLS Web Server Authentication, TLS Web Client Authentication',
	'basicConstraints' => 'CA:FALSE',
	'subjectKeyIdentifier' => '8E:80:81:97:BD:90:17:1B:CD:0C:3C:63:D8:F6:4B:23:17:39:6B:5A',
	'authorityKeyIdentifier' => 'keyid:A8:4A:6A:63:04:7D:DD:BA:E6:D1:39:B7:A6:45:65:EF:F3:A8:EC:A1 ',
	'authorityInfoAccess' => 'OCSP - URI:http://ocsp.int-x3.letsencrypt.org CA Issuers - URI:http://cert.int-x3.letsencrypt.org/ ',
	'subjectAltName' => 'DNS:*.berb.ec, DNS:*.berbec.com, DNS:berb.ec, DNS:berbec.com', 
	'1.3.6.1.5.5.7.1.24' => '0',
	'certificatePolicies' => 'Policy: 2.23.140.1.2.1 Policy: 1.3.6.1.4.1.44947.1.1.1 CPS: http://cps.letsencrypt.org User Notice: Explicit Text: This Certificate may only be relied upon by Relying Parties and only in accordance with the Certificate Policy found at https://letsencrypt.org/repository/ ',
	'ct_precert_scts' => 'Signed Certificate Timestamp: Version : v1(0) Log ID : 29:3C:51:96:54:C8:39:65:BA:AA:50:FC:58:07:D4:B7: 6F:BF:58:7A:29:72:DC:A4:C3:0C:F4:E5:45:47:F4:78 Timestamp : Mar 31 08:00:58.613 2018 GMT Extensions: none Signature : ecdsa-with-SHA256 30:44:02:20:46:A9:9E:D6:DC:51:5F:81:A4:05:42:28: 38:C3:45:EA:2D:E7:E3:DE:ED:EB:37:DD:8D:DD:02:91: 9F:11:64:29:02:20:25:AC:2D:0E:FD:7E:6B:6E:19:93: 64:A9:B1:63:E4:67:E4:03:66:D7:47:95:73:42:C4:90: 04:B9:65:E7:76:1C Signed Certificate Timestamp: Version : v1(0) Log ID : DB:74:AF:EE:CB:29:EC:B1:FE:CA:3E:71:6D:2C:E5:B9: AA:BB:36:F7:84:71:83:C7:5D:9D:4F:37:B6:1F:BF:64 Timestamp : Mar 31 08:00:58.639 2018 GMT Extensions: none Signature : ecdsa-with-SHA256 30:44:02:20:4F:82:33:5D:8B:48:3E:6E:E5:0E:75:E9: 22:C8:DE:50:4F:AE:F1:DB:CB:A5:C5:50:F6:A2:40:A0: 51:D6:6E:95:02:20:2A:6A:41:51:5E:EA:82:54:38:67: 55:46:1C:96:71:7B:33:0B:9B:34:6C:4C:AF:5B:A0:EC: 29:FE:F0:48:DF:46', ),
)`

jim-p:
OK, thanks. It's extension OID `1.3.6.1.5.5.7.1.24` so that means we can find it. I'll commit a function to locate it and have it display in the certificate properties list in the GUI shortly, then you can call `cert_get_oscpstaple(<encoded cert string>)` to check. True means it must staple, false means it doesn't.

jim-p:
OK, `cert_get_oscpstaple()` is there in 00e54150ea33d7a5b7e8f28923210dedb915ee43

peterberbec:
Made a get_dns_nameservers function. I'm going to have to get this on my pfsense box so I can debug it. Coding on github is possible, but very likely to introduce errors.

peterberbec:
It's OCSP, btw.

Now that the function exists, should I add logic to hide the option entirely if the cert isn't must-staple, or should this not be an option at all anymore?

jim-p:
d'oh. I typed it right in the GUI text but somehow did that wrong twice in the function. Fixed now.

I'm torn on the GUI option. I'm leaning toward just enabling it automatically as needed. Maybe it might be good to keep the GUI option to force it on, in case the detection isn't perfect. If we do that, be sure to note in the option's help text that it will be handled automatically in most cases.

peterberbec:
Added your function to my fork so if you decide to pull, it'll be there.

peterberbec:
I am thinking calling this an **option** is a misnomer. If you have a must staple cert, not having this in the nginx configuration leaves you with a broken site.

Could we hide it in System -> Advanced -> System Tuneables? `net.ssl.ocsp`. 0 for force disable, 1 for autodetect, 2 for force disable?

jim-p:
I agree that it's required for certificates with the property set. My fear is that, despite the auto-detection, someone might need it when that doesn't work as expected. It could be called "Force OCSP Stapling"  or similar in that case, to reflect that it's overriding a default behavior. Forcing it off isn't possible and could break, but if the autodetection doesn't kick in, that could help someone in a bind.

The tunables are FreeBSD sysctl OIDs, so we can't just make up new ones. 

You could remove the GUI option but keep the configuration paramter check, then someone could manually edit it into their config.xml if need be.

peterberbec:
Sounds smart to me. 

Here's what I see my work as now:

1. Change the option to "Force OCSP Stapling"
2. Hide the option if the installed and used cert returns true from `cert_get_ocspstaple()`
3. Enable the stapling if the return value of `cert_get_ocspstaple()` is true.

What's the best way to find the certificate used for the pfsense website? Is there `cert_get_inuse_for_webConfigurator()`?

peterberbec:
Actually, this needs to be written into the cert import process, as well, does it not?

jim-p:
> What's the best way to find the certificate used for the pfsense website? Is there `cert_get_inuse_for_webConfigurator()`?

https://github.com/pfsense/pfsense/blob/master/src/etc/inc/system.inc#L1328

> Actually, this needs to be written into the cert import process, as well, does it not?

No, the OCSP Must Staple property is a part of the certificate data, it does not need to be set or maintained outside of the certificate.

peterberbec:
Sorry, what I meant was this needs to be tested for when a cert is assigned to be used for the webconfigurer, and this flag set then.

jim-p:
> Sorry, what I meant was this needs to be tested for when a cert is assigned to be used for the webconfigurer, and this flag set then.

If the code is setting it up automatically there isn't a need to test when it's selected, only when generating the nginx configuration. There wouldn't be a flag to set, really. The only time an option would need to be set is manually by the user if they need stapling but it didn't detect it automatically. And since we can't detect that case automatically, there isn't anything we can do to check/set for that.

peterberbec:
For testing on my box, there seems to be an guide to updating pfsense from git [here](https://doc.pfsense.org/index.php/Updating_pfSense_code_between_snapshots)

If I change it to point to my repo, by the last command of 
[code]rm -rf /root/pfsense/
cd /root/pfsense/pfSenseGITREPO/pfSenseGITREPO; git remote set-url origin git://github.com/pfsense/pfsense.git[/code]

Should this give me my git version on my box? Sorry, not the best at github.

peterberbec:
I rebased my fork to take your updates into account

README.md
```diff
@@ -1,5 +1,7 @@
 # pfSense
 
+peterberbec mods to support OCSP must-staple
```
jim-p: You'll have to remove these changes to README.md, it's trying to merge them into the pfSense repo, they aren't just local to you.

src/etc/inc/system.inc
```diff
@@ -1365,6 +1343,49 @@ function system_webgui_start() {
...
+   		(empty($config['dnsmasq']['port']) || $config['dnsmasq']['port'] == "53") &&
+	      	(empty($config['dnsmasq']['interface']) ||
+	       	in_array("lo0", explode(",", $config['dnsmasq']['interface'])))) ||
+	     	((isset($config['unbound']['enable'])) &&
+	      	(empty($config['unbound']['port']) || $config['unbound']['port'] == "53") &&
+	      	(empty($config['unbound']['active_interface']) ||
+	       	in_array("lo0", explode(",", $config['unbound']['active_interface'])) ||
+	       	in_array("all", explode(",", $config['unbound']['active_interface']), true)))) &&
+	     	(!isset($config['system']['dnslocalhost']))) {
+			$dns_nameservers[$i++] = "127.0.0.1";
```
jim-p: Is there a reason you're indexing these manually vs using `$dns_nameservers[]` which will automatically add to the end of the array?

src/etc/inc/system.inc
```diff
@@ -1365,6 +1343,49 @@ function system_webgui_start() {
 	return $res;
 }
 
+function get_dns_nameservers() {
+	global $config;
+	
+	$dns_nameserver_number = 0;
```
jim-p: This variable is declared but never used.

src/etc/inc/system.inc
```diff
@@ -1365,6 +1343,49 @@ function system_webgui_start() {
...
+		}
+	}
+	if (is_array($syscfg['dnsserver'])) {
+		foreach ($syscfg['dnsserver'] as $sys_dnsserver) {
+			if ($sys_dnsserver && (!in_array($sys_dnsserver, $ns))) {
+				$dns_nameservers[$i++] = "$sys_dnsserver";
+			}
+		}
+	}
+	return $dns_nameservers;
```
jim-p: As long as we're collecting these in an array, for good measure this should probably do `return array_unique($dns_nameservers);`, to avoid using the same server more than once.

src/etc/inc/system.inc
```diff
@@ -1504,6 +1525,17 @@ EOD;
 		$nginx_config .= "\t\tadd_header X-Content-Type-Options nosniff;\n";
 		$nginx_config .= "\t\tssl_session_tickets off;\n";
 		$nginx_config .= "\t\tssl_dhparam /etc/dh-parameters.4096;\n";
+		if (($config['system']['webgui']['ocsp-staple'] == true) ||
+		    cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) {
+			$nginx_config .= "\t\tssl_stapling on;\n";
+			$nginx_config .= "\t\tssl_stapling_verify on;\n";
+			$nginx_config .= "\t\tresolver ";
+			foreach(get_dns_nameservers() as $dns_ns) {
```
jim-p: This could be optimized further, instead of doing a foreach, replace that block with:

    $nginx_config .= implode(" ", get_dns_nameservers()) . " valid=300s;\n";

src/usr/local/www/system_certmanager.php
```diff
@@ -1223,7 +1223,6 @@ function list_csrs() {
 						if (cert_get_ocspstaple($cert['crt'])) {
 							$certextinfo .= '<b>' . gettext("OCSP: ") . '</b> ';
 							$certextinfo .= gettext("Must Staple");
-						}
```
jim-p: Why did this brace get deleted?

README.md
peterberbec: That was to ensure I was looking at the right repo locally, i'l delete it obviously!

src/etc/inc/system.inc
peterberbec: I'm a c programmer at heart. I was originally going to use a for loop and I don't know about php's variable declarations greatly. I learned about foreach a little later. deleting.

src/etc/inc/system.inc
peterberbec: good idea. changed

src/usr/local/www/system_certmanager.php
peterberbec: When I rebased my repo on the main one, It created a duplicate of the code block. I hit 'd5d' in vi instead of 'd4d'. fixed

src/etc/inc/system.inc
peterberbec: because I'm a stupid C programmer. fixing now

peterberbec:
Cleaned up stuff at your direction. Mostly C vs PHP stuff. I wasn't aware of quite a few features (`$array[] = $x` automatically assigns to next free member is great!)

src/usr/local/www/system_advanced_admin.php
```diff
@@ -556,6 +580,7 @@
 
 	hideInput('ssl-certref', $('input[name=webguiproto]:checked').val() == 'http');
 	hideCheckbox('webgui-hsts', $('input[name=webguiproto]:checked').val() == 'http');
+	hideCheckbx('ocsp-staple', $(cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) == true);
```
jim-p: One more thing I noticed: This is JavaScript here, not PHP, so it should probably be something closer to:

    hideCheckbx('ocsp-staple', <?= cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) ? "true" : "false"; ?>);

peterberbec: and I spelled hideCheckbox wrong!

peterberbec: Forgive ignorance, but is it:

`<?= cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) ? "true" : "false"; ?>`

or 

`<?php cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) ? "true" : "false"; ?>`

jim-p: You can use either method so long as you handle it properly.
<?= $foo ?> performs an implied echo/print of the returned value, which makes cases like this easier.
<?php ... ?> requires you to do your own echo, so as written in your example it would output nothing, but if you put `echo ` before the function call it would have the same net effect.

peterberbec: ok. i pushed the `<?=` version already.

peterberbec: and for my own enlightenment, the other option would be:

```<?php echo cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) ? "true" : "false"; ?>```

or would it need to be

```<?php cert_get_ocspstaple(lookup_cert($config['system']['webgui']['ssl-certref'])) ? echo "true" : echo "false"; ?>```

Comment (jim-p) [reply src/usr/local/www/system_advanced_admin.php]: Either of those should have the same effect in the end.

**Comment by peterberbec** [comment]
@jim-p Is the method I laid out for updating my router to test the right way to do it?

**Comment by jim-p** [comment]
You can do that, but it's easier to do a regular gitsync and then apply the PR as a patch using the System Patches package. Add `.diff` after the PR url and that's what you feed into the System Patches package.

https://patch-diff.githubusercontent.com/raw/pfsense/pfsense/pull/3927.diff

Just remember to revert the patch before re-fetching to apply new changes.

**Comment by peterberbec** [comment]
OK I'll try that tonight. My only pfsense box is in use, so I have to wait until late to play. Is there a guitar method to gitsync?

**Comment by peterberbec** [comment]
OK gitsync worked, but applying the patch from https://patch-diff.githubusercontent.com/raw/pfsense/pfsense/pull/3927.diff did not function correctly. Any tips?

```
/usr/bin/patch --directory=/ -t -p1 -i /var/patches/5ac3f7a6d5809.patch --check --forward 

Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/etc/inc/system.inc b/src/etc/inc/system.inc
|index 11136bd952..d295660cd6 100644
|--- a/src/etc/inc/system.inc
|+++ b/src/etc/inc/system.inc
--------------------------
No file to patch.  Skipping...
Hunk #1 ignored at 122.
Hunk #2 ignored at 134.
Hunk #3 ignored at 141.
Hunk #4 ignored at 1343.
Hunk #5 ignored at 1524.
5 out of 5 hunks ignored while patching src/etc/inc/system.inc
Hmm...  The next patch looks like a unified diff to me...
The text leading up to this was:
--------------------------
|diff --git a/src/usr/local/www/system_advanced_admin.php b/src/usr/local/www/system_advanced_admin.php
|index dee707182c..5c636ed4a7 100644
|--- a/src/usr/local/www/system_advanced_admin.php
|+++ b/src/usr/local/www/system_advanced_admin.php
--------------------------
No file to patch.  Skipping...
Hunk #1 ignored at 42.
Hunk #2 ignored at 159.
Hunk #3 ignored at 411.
Hunk #4 ignored at 580.
4 out of 4 hunks ignored while patching src/usr/local/www/system_advanced_admin.php
done
```

**Comment by peterberbec** [comment]
OK got it up. doing some debugging now.

There seems to be an error in `cert_get_ocspstaple`:

```
[03-Apr-2018 18:40:31 EST5EDT] PHP Warning:  base64_decode() expects parameter 1 to be string, array given in /etc/inc/certs.inc on line 665
```

That line is:

```
663 function cert_get_ocspstaple($str_crt, $decode = true) {
664 	if ($decode) {
665		$str_crt = base64_decode($str_crt);
666	}
667	$crt_details = openssl_x509_parse($str_crt);
668	if (!empty($crt_details['extensions']['1.3.6.1.5.5.7.1.24'])) {
669		return true;
670	}
671	return false;
672 }
```

**Comment by peterberbec** [comment]
I've made that and a few other changes. I'm going to rebase my system from the main git branch, apply this diff, and see if everything works. I haven't figured out how to push to my own git from the pfsense box, so I'm copying what I do on the router over to another machine.

It's seems to be working properly now. Setting the config in nginx, hiding the check box as it should!

**Comment by peterberbec** [comment]
Caught something on reboot. I think it's a scope issue, i'm not quite sure.

**Comment by jim-p** [comment]
It might be that the NS list is empty at some point. Safest thing to do is add an `is_array()` check before running `foreach` and skip the `foreach` if it's false, or set it to an empty array so the `foreach` will succeed either way.

**Comment by peterberbec** [comment]
Oh, so it's not scope? I just wasn't sure if the $ns array existed outside the foreach loop if initialized inside of it. I moved it outside of the loop, but will also add some testing.

**Comment by jim-p** [comment]
Well it is, it should be declared before the loop, but it's still safest to do an is_array check before a foreach.

**Comment by peterberbec** [comment]
OK I threw an `is_array` in there

**Comment by peterberbec** [comment]
OK I woke up in the middle of the night, redid a gitsync and applied the new github.com diff patch. It booted without errors now.

**Comment by jim-p** [comment]
Looks OK to me now, only part I'm concerned about is the changes to the name server list generation. We just need to beat it up a bit and make sure it still behaves identically to the older code.

**Comment by peterberbec** [comment]
Is there a pfsense fuzziness library? :)

**Comment by peterberbec** [comment]
Apologies, I was trying to add something to my personal box and didn't think it would affect this pull. I removed the files.

**Comment by sdj38** [comment]
I was testing this PR and noticed something that may be amiss. If you are using the default webConfigurator as the SSL certificate. If you check the OCSP must staple box and save it. It will then say it is reloading in 20 seconds... After waiting 20 seconds it will show up unchecked.

**Comment by peterberbec** [comment]
So force-stapling isn't working if the cert doesn't support stapling?

**Comment by sdj38** [comment]
I just wasn't sure if it should uncheck the box afterward without giving an indication or information as to why it got unchecked. It will say it is applied successfully but after it reloads it will no longer be checked.

**Comment by sbeaver-netgate** [comment]
It would seem like an error message "The selected certificate does not support stapling" or something like that is required here.

**Comment by peterberbec** [comment]
It should force it on. I wonder, if in the chain of events, it enables
stapling; checks the cert; notices it doesn't support this and then
disables the option.

This setting is for edge cases where we can't detect the OCSP property
correctly and it needs to be enabled. I'm not sure it has any other use.

I'm mobile only now. I'll look at the code more. Is there a way to detect
if the option is properly set? Does pfsense save the configuration
somewhere I can check?

**Comment by sdj38** [comment]
It looks like if you change these lines it will stay checked/unchecked after reloading
```
//$pconfig['ocsp-staple'] = $config['system']['webgui']['ocsp-staple'];
$pconfig['ocsp-staple'] = isset($config['system']['webgui']['ocsp-staple’]);

        else {
			if ($config['system']['webgui']['ocsp-staple'] == true) {
				$restart_webgui = true;
			}

                          //$config['system']['webgui']['ocsp-staple'] = false;
			unset($config['system']['webgui']['ocsp-staple’]); 
		}
```
I hope that helps.

PR #3930, opened 2018-04-02
webgui, top menu scroll bar for many items, and absolute position for smaller screens to keep menu usable

PiBa-NL:
webgui, top menu scroll bar for many items, and absolute position for smaller screens to keep menu usable

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8419
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/css/pfSense.css
+++ b/src/usr/local/www/css/pfSense.css
@@ -857,6 +857,14 @@ table[data-sortable].sortable-theme-bootstrap thead th {
     }
 }
 
+@media (min-width:991px){
+	.navbar-fixed-top .navbar-nav .open .dropdown-menu {
+		/* show a scrollbar in the menu when the menu is higher than the viewport */
+		max-height: calc(100vh - 50px);
+		overflow-y: auto;
+	}
+}
+
 @media (max-width: 991px) {
     /* change top navbar from horizontal to vertical */
     .navbar-header {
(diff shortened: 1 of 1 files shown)


PiBa-NL:
I don't have an iPhone.. but on Android with a Chrome browser and fixed top menu as it used to be that menu cannot be used as it doesn't show half of the menu options.... So that the menu is not fixed anymore is better in that case, i have intentionally broken that 'fix top menu' for such smaller screens..

The scrollbar should work for screens wider than 990px at least on a regular W10 computer.

So imho its still a improvement.?

sbeaver-netgate:
The intention was that the user select "Scroll with screen" on small screens, "Fixed" on large screens. There are always more ways of addressing a problem though.

While this PR may, or may not be an improvement, I don't think we should render an existing control ineffective in this way. If this mode of operation is desired (and I'm not sure it is) then at least the Setup->General->web UI control needs to be changed to reflect the new operation.

I wonder if a better solution would be to have the control automatically force "scroll with screen" mode on small screens?

The scroll bar you added does not seem to work on iOS devices.

PiBa-NL:
Okay, i can remove the "@media (max-width: 991px) {" part as that changes settings 'undesirably'. (I do wonder how you intend to use the menu though if you cant reach some of the buttons.. Or change the setting each time when sitting behind a different device?)

Would the part for "@media (min-width:991px){" part either improve or leave behavior the same for the other devices? Not sure if youve seen the expected effect.?: https://snag.gy/AuYMxm.jpg  also make sure to hard-refresh the browser cache..

sbeaver-netgate:
I have seen that effect, but it does not appear to work on iOS devices. Of course I have been careful to clear the cache and/or use a private browser window in each case.

PiBa-NL:
Ok, so no 'improvement' there.. is it the same or worse as before.?

i.e. Would the >990 part be 'merge-able' ? If so i can change the commit for that. Or should i make a 3rd choice for the user to pick this specific new behavior.?

sbeaver-netgate:
I have just spoken to some others (with more web UI experience than me) and the consensus seems to be that if we want to retain the two menu display options, the best approach is to change the control so that it becomes a "Fixed top menu on large screens" checkbox. If the user is on a tablet or phone, it should automatically switch to "scrolls with page".

On the other hand, it was suggested that perhaps the "fixed" option is of no real use and we should just do away with it.

sbeaver-netgate:
Closing this PR for now. Wile it offers an improvement in some areas, in others it adds confusion. We may also ditch the "fixed to top of page" option before long.

PR #3934, opened 2018-04-05
outbound-nat, make sure rules generated with outbound-nat rules are valid

PiBa-NL:
outbound-nat, make sure rules generated with outbound-nat rules are valid also on interfaces with mixed ipv4/ipv6 address environments

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8437
- [x] Ready for review

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -1617,8 +1617,8 @@ function filter_generate_reflection_proxy($rule, $nordr, $rdr_ifs, $srcaddr, $ds
 function filter_nat_rules_automatic_tonathosts($with_descr = false) {
 	global $config, $FilterIflist, $GatewaysList;
 
-	$tonathosts = array("127.0.0.0/8");
-	$descriptions = array(gettext("localhost"));
+	$tonathosts = array("127.0.0.0/8", "::1/128");
+	$descriptions = array(gettext("localhost"), gettext("localhost"));
 
 	// Add any enabled static routes
 	foreach (get_staticroutes(false, false, true) as $route) {
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -1905,6 +1905,10 @@ function filter_nat_rules_generate_if ($if, $descr = "", $ipprotocol = "", $src 
 	if (!empty($descr)) {
 		$descr = " # " . $descr;
 	}
+	
+	if (empty($if_friendly) || empty($src) || empty($dst)) {
+		return "# validation of rule properties failed to verify the interface/source/destination" . " in nat rule: {$descr}\n";
+	}
 	/* Put all the pieces together */
 	$natrules = "";
 	if ($nonat) {
@@ -1912,10 +1916,18 @@ function filter_nat_rules_generate_if ($if, $descr = "", $ipprotocol = "", $src 
 		$natrules .= "no nat on \${$if_friendly} {$inet}{$protocol} from {$src} to {$dst}{$descr}\n";
 	} else {
 		if ($inet == "" || $inet == 'inet') {
-			$natrules .= "nat on \${$if_friendly} inet{$protocol} from {$src} to {$dst} -> {$tgt4}{$tgtport} {$poolopts}{$staticnatport_txt}{$descr}\n";
+			if (empty($tgt4)) {
+				$natrules .= "# validation of rule properties failed to verify the IPv4 target" . " in nat rule: {$descr}\n";
+			} else {
+				$natrules .= "nat on \${$if_friendly} inet{$protocol} from {$src} to {$dst} -> {$tgt4}{$tgtport} {$poolopts}{$staticnatport_txt}{$descr}\n";
+			}
 		} 
 		if ($inet == "" || $inet == 'inet6') {
-			$natrules .= "nat on \${$if_friendly} inet6{$protocol} from {$src} to {$dst} -> {$tgt6}{$tgtport} {$poolopts}{$staticnatport_txt}{$descr}\n";
+			if (empty($tgt6)) {
+				$natrules .= "# validation of rule properties failed to verify the IPv6 target" . " in nat rule: {$descr}\n";
+			} else {
+				$natrules .= "nat on \${$if_friendly} inet6{$protocol} from {$src} to {$dst} -> {$tgt6}{$tgtport} {$poolopts}{$staticnatport_txt}{$descr}\n";
+			}
 		}
 	}
 	return $natrules;
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -1786,26 +1786,72 @@ function filter_nat_rules_outbound_automatic($src) {
 }
 
 /* Generate a 'nat on' or 'no nat on' rule for given interface */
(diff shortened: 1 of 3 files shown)


jim-p:
I'm very tempted to reject IPv6 input in the outbound NAT rules rather than add all this. I'm not entirely convinced we should be accommodating even more NAT with IPv6, when the goal is less.

PiBa-NL:
Well i can make a smaller uglier version.. at first i didnt have the protocol selectionbox for the user.. And it worked decently.. However to have a rule apply to any<>any traffic which i think is nice to have when having lots of internal v4 subnets.., it didnt know to autodetect to only make a ipv4 rule. When 'normal' IPv6 is present. So added the gui selection box to give the user full control..

And yes IPv6 is 'meant' to be used without natting, however ISP's make a mess of things all the time, and its nice to have ways to work around funky ISP configurations.. Imho it IPv6 natting would be nice to have.. Anyhow the possibility to write 'wrong' rules is worse than writing working but 'undesirable' rules.. If you want me to 'hide' part of this stuff and perhaps make them advanced settings or something i'm open to do that.

src/etc/inc/filter.inc
```diff
@@ -1789,17 +1835,13 @@ function filter_nat_rules_generate_if ($if, $src = "any", $srcport = "", $dst =
...
-	}
+	
+
+	$tgtport = "";
 	/* Add the hard set source port (useful for ISAKMP) */
 	if ($natport != "") {
-		$tgt .= " port {$natport}";
-	}
+		$tgtport .= " port {$natport}";
+	}
```
jim-p: Little whitespace error here

jim-p:
I didn't have anything specific in mind, but we've had several instances lately of unexpected values breaking the ruleset because of assumptions made about what had already been validated, plus a fun race condition that popped up (See https://redmine.pfsense.org/issues/8604 ), so I want to be absolutely sure that when we're forming rules that every component will be valid so the rules don't get tossed out.

PiBa-NL:
@jim-p , added a commit to 'validate' some required properties are present.. quite a few parameters used are optional so cant be checked easily without duplicating the code that 'should' fill them.. so not sure if this way is sufficient.. can you take a look.? thanks in advance.

jim-p:
That looks like it should be good enough, thanks!

src/etc/inc/filter.inc
```diff
@@ -1757,26 +1757,72 @@ function filter_nat_rules_outbound_automatic($src) {
...
+			} else {
+				$tgt4 = "{$natip}/32";
+			}
 		} else {
-			$tgt = "{$natip}/32";
+			$ifip = get_interface_ip($if);
+			if (is_ipaddrv4($ifip)) {
+				$tgt4 = "{$ifip}/32";
+			} else {
+				$tgt4 = "(" . $FilterIflist[$if]['if'] . ")";
```
jim-p: This needs an extra safety belt to NOT generate the rule if the interface has no IPv4 address for this family. Similar to the check I added in 6f8e648f5c88e04166539ab27872b13dfd587cb8 -- Otherwise the ruleset will fail to load. It's more important when it comes to IPv6 where it's more likely to be empty. I'll add another note on the other place that needs fixed.

src/etc/inc/filter.inc
```diff
@@ -1757,26 +1757,72 @@ function filter_nat_rules_outbound_automatic($src) {
...
+			} else {
+				$tgt6 = "{$natip}/128";
+			}
 		} else {
-			$tgt = "(" . $FilterIflist[$if]['if'] . ")";
+			$ifip = get_interface_ipv6($if);
+			if (is_ipaddrv6($ifip)) {
+				$tgt6 = "{$ifip}/128";
+			} else {
+				$tgt6 = "(" . $FilterIflist[$if]['if'] . ")";
```
jim-p: This needs an extra safety belt to NOT generate the rule if the interface has no IPv6 address for this family. Similar to the check I added in 6f8e648f5c88e04166539ab27872b13dfd587cb8 -- Otherwise the ruleset will fail to load.

jim-p:
I'm not sure why the errors were not caught in the other places we discussed, but somehow it's still ending up with ruleset errors when an interface doesn't have a matching address.

src/etc/inc/filter.inc
jim-p: Yeah that's a bit confusing to me as well, the only thing I can think of is that up to now it's fairly difficult for an interface to be enabled and _not_ have an IPv4 address, but it's trivial for an interface to not have an IPv6 address. I'm surprised this hasn't caused problems in the past, but at the same time, it would be tough to paint it into that corner. It's also possible some change in FreeBSD/pf has made the problem more apparent.

PiBa-NL: ok then ill add a 'safety belt' and remove the $FilterIflist[$if]['if'] as that would no-longer be 'reachable code'

jim-p: OK, so long as it ends up skipping the entire rule when the interface doesn't have an address on it.

PiBa-NL:
@jim-p ive changed the ' $FilterIflist[$if]['if'] ' items to a empty "", and for a 'safety belt' there was already a check for "if (empty($tgt4)) {" in place, but filling it with the interface was thought to be safe, so it wasn't caught by that check, now when setting tgt4 and tgt6 to a empty string it will be caught and not write the rule..

So should be safe to go now.?

jim-p:
Looks like it should be good, will need to run some more tests tomorrow to confirm. One of our employees was able to replicate the case where it was empty.

jim-p:
OK, after some digging it turns out the empty interface address was **not** actually the cause of that error, so we can probably revert that last change and let the `(<interface>)` syntax work.

The problem is that the `tonatsubnets` table does not always contain both IPv4 and IPv6 addresses, so pf fails to expand the rule properly and throws an error.

It would appear a better fix is to change src/etc/inc/filter.inc#L1620 to:

    $tonathosts = array("127.0.0.0/8", "::1/128");

So that pf always has a way to expand IPv6 addresses in those rules. We're still testing this theory though.

PiBa-NL:
@jim-p i guess the theory was tested successfully? ive added a commit for the $tonathosts as requested

jim-p:
We couldn't make it break with that in `tonatsubnets`, so it looks like you can revert 8517667d7038c0835f7d41ff5b1e6e293880cbc1 as well since that check is unnecessary -- it was `tonatsubnets` breaking the rule and not the target.

PiBa-NL:
https://github.com/pfsense/pfsense/commit/8517667d7038c0835f7d41ff5b1e6e293880cbc1 was added after your comment about safetychecks: https://github.com/pfsense/pfsense/pull/3934#issuecomment-401931135

The 3rd commit i added later (removing the "(" . $FilterIflist[$if]['if'] . ")" ) for the "filter load errors" you mentioned later is already undone, and is no longer present in the commits in this PR. It is replaced by the commit that adds IPV6 to the $tonatsubnets

Are you sure the other 'safety belt' can be removed as well? I don't think it hurts anything and yes it doesn't help anything either that i 'know' of.. But i don't know everything ;)

jim-p:
OK, I was looking at the commits it shows on the PR and it didn't add any note in the timeline that it was removed. I think, aside from the tonatsubnets bit, the extra safety belts I kept going on about are not needed since it wasn't actually failing in the way I thought it was failing. The pf error wasn't clear about which part was to blame and I (wrongly) blamed the incorrect part of the rule. :-)

Looking over the whole diff again instead of just the commits I do think what's there now is probably good enough to move forward with, thanks!

PR #3936, opened 2018-04-13
8122: enable use of auth script v3

ssbarnea:

- [ ] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1108,6 +1108,11 @@ function openvpn_reconfigure($mode, $settings) {
 			if (!($settings['auth_user'] && $settings['auth_pass'])) {
 				$userpass .= "\n";
 			}
+
+			// if auth_script is not empty, we use its output instead of user/pass
+			if ($settings['auth_script']) {
+				$userpass = shell_exec($settings['auth_script']);
+			}
 			file_put_contents($up_file, $userpass);
 		}
 
--- a/src/usr/local/www/vpn_openvpn_client.php
+++ b/src/usr/local/www/vpn_openvpn_client.php
@@ -101,7 +101,7 @@ if ($act == "new") {
 }
 
 global $simplefields;
-$simplefields = array('auth_user', 'auth_pass');
+$simplefields = array('auth_user', 'auth_pass', 'auth_script');
 
 if ($act == "edit") {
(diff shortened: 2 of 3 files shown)


jim-p:

Thanks!

src/usr/local/www/vpn_openvpn_client.php
```diff
@@ -634,6 +634,13 @@
 		$pconfig['auth_pass']
 	))->setHelp('Leave empty when no password is needed');
 
+	$section->addInput(new Form_Textarea(
+		'auth_script',
+		'Auth Script',
+		$pconfig['auth_script']
+	))->setHelp('Script that would output credentials as "user<newline>password".' .
+	    'When defined, this will be executed and the results used instead of username and password. Allows use of OTP authentication.', '<br/>');
```
jim-p: If this is intended to be a full path to a script on the filesystem, that should be mentioned in the help text.

Input validation must also ensure this is a path to a script file alone and not a string of other commands or parameters, otherwise you're handing the user a way to exec anything they want in the shell.

This should be more clear about it being executed when the settings are saved and not every time the client connects. Given that limitation I'm not sure how well it will effectively work with OTP since if the client gets disconnected you would have to re-save the settings or otherwise trigger an openvpn config rewrite to get a new token in the file.

src/etc/inc/openvpn.inc
```diff
@@ -1108,6 +1108,11 @@ function openvpn_reconfigure($mode, $settings) {
 			if (!($settings['auth_user'] && $settings['auth_pass'])) {
 				$userpass .= "\n";
 			}
+
+			// if auth_script is not empty, we use its output instead of user/pass
+			if ($settings['auth_script']) {
+				$userpass = shell_exec($settings['auth_script']);
```
jim-p: This needs a test to check that the script exists before it is executed, otherwise you are blindly sending whatever the user entered to run as a shell command.

jim-p:
I thought this seemed familiar.

Old PR with more discussion, and similar points: https://github.com/pfsense/pfsense/pull/3877
Redmine issue: https://redmine.pfsense.org/issues/8122

This new PR doesn't appear to address any of the specific concerns raised in the old PR.

sbeaver-netgate:
No action in > 30 days. Please resubmit if you are able to address the requested changes

PR #3939, opened 2018-04-21
Improve/add help messages warning about bad security choices

seanm:


seanm:
This is another attempt after this: https://github.com/pfsense/pfsense/pull/3524

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -595,7 +595,7 @@ function idinfo_to_pconfig($prefix, & $idinfo, & $pconfig) {
 	'*Protocol',
 	$pconfig['proto'],
 	$p2_protos
-))->setHelp('ESP is encryption, AH is authentication only.');
+))->setHelp('ESP (Encapsulating Security Payload) is encryption, AH (Authentication Header) is authentication only.');
```
jim-p: If you're writing both out, it's better to write out the full name and then put the abbreviation in parenthesis, e.g. "Encapsulating Security Payload (ESP)"

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -851,6 +851,7 @@ function build_eal_list() {
 	$section->add($group);
 	$counter += 1;
 }
+$section->setHelp('Note: Blowfish, 3DES, CAST128, MD5, SHA1, and DH groups 1, 2, 22, 23, and 24 provide weak security and should not be used.');
```
jim-p: I prefer "should be avoided" rather than "should not be used" here. The latter makes it sound like we're lazy and haven't removed them when we know better, rather than the option still being there because some people need them because they are burdened with third-party compatibility or other vendor/business needs.

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -632,7 +632,7 @@ function idinfo_to_pconfig($prefix, & $idinfo, & $pconfig) {
 
 
 	if ($i == $rows) {
-		$group->setHelp('Use 3DES for best compatibility or for a hardware crypto accelerator card. Blowfish is usually the fastest in software encryption.');
+		$group->setHelp('Note: Blowfish, 3DES, and CAST128 provide weak security and should not be used.');
```
jim-p: See above, re: "avoided".

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -650,6 +650,8 @@ function idinfo_to_pconfig($prefix, & $idinfo, & $pconfig) {
 		(empty($pconfig['halgos']) ? '' : in_array($algo, $pconfig['halgos'])),
 		$algo
 	))->addClass('multi')->setAttribute('id');
+
+	$group->setHelp('Note: MD5 and SHA1 provide weak security and should not be used.');
```
jim-p: See above, re: "avoided".

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -661,14 +663,18 @@ function idinfo_to_pconfig($prefix, & $idinfo, & $pconfig) {
 	'PFS key group',
 	$pconfig['pfsgroup'],
 	$sm ? $p2_pfskeygroups:array()
-))->setHelp($sm ? '':'Set globally in mobile client options');
+));
+
+$helpstr = $sm ? '':'Set globally in mobile client options. ';
+$helpstr .= 'Note: groups 1, 2, 22, 23, and 24 provide weak security and should not be used.';
```
jim-p: G in Groups should be capitalized here.
Also, see above, re: "avoided".

seanm:
OK fixed.

BTW: please check syntax carefully.  I don't know PHP, I just hacked this up, and didn't actually test it!

PR #3941, opened 2018-05-05
Add dummynet AQM and scheduler configuration support to pfSense Limit…

mattund:
Add dummynet AQM and scheduler configuration support to pfSense Limiters through the GUI.  Only shaper.inc was changed.

Screenshot: https://i.imgur.com/N36gpXF.png
Related discussion: https://forum.pfsense.org/index.php?topic=126637

**Highlights**

- FQ_CODEL instead of altq method (FairQ + single CoDel queue)
- Shape Multi-WAN + Multi-LAN configurations in one queue per direction per WAN
- Shape ixgbe, LAGG, among others not supported via altq (in my experience, anyway)
- Load default params from sysctl, or set your own for PIE, FQ_PIE, RED, GRED, CoDel, and FQ_CoDel (target ms, flows, etc.)
- Set human-readable names for parameters for ease of use ("tudpate" becomes "Interval (ms)")
- Descriptive help for schedulers+AQMs to make sense of options
- Future development: add/change AQM/Scheduler arrays/models to drop in/change out features, parameters, etc.

**Usage**

1. Create a Limiter and configure basic BW/schedule options
2. Assign a scheduler and/or AQM to the limiter
3. Create a queue on the limiter
4. Assign an AQM, if applicable, to the queue
5. Make a floating rule, dir out, on the iface desired to shape and assign pipes' queues as usual

**Concept & Changes**

I have added a new section ("Queue") to both Limiters and their child queues.  Limiters have applicable several options:
- AQM Algorithm
- AQM Parameters (if algorithm has any)
- Scheduler
- Scheduler Algorithm (if scheduler has any)
- Queue Length (moved from Advanced Options)
- ECN

NOTE: Limiters by default have an in-configurable (?) FIFO scheduler attached to the underlying pipe.  So, you should _not_ attach traffic  to a limiter if you expect to schedule it with a different scheduler than FIFO.  For this, you have to attach it to a child queue.  See the related forum thread for details on this.

Limiter children have all of the same options, except do not define a scheduler.  This is _inherited_ from their configured parent's scheduler, and shared among the respective limiter's queues:
- AQM Algorithm
- AQM Parameters (if algorithm has any)
- Queue Length (moved from Advanced Options)
- ECN

Generated ipfw commands (rules.limiter):
```
pipe # config [bw...] {pipe->aqm->parameter_format}
sched # config pipe # {pipe->scheduler->parameter_format}
queue n config sched # {queue->aqm->parameter_format}
queue n+1 config sched # {queue->aqm->parameter_format}
```

- [x] Redmine Issue: https://redmine.pfsense.org/issues/6620
- [x] Ready for review


sbeaver-netgate:
All PHP code needs to be PHP 7 compliant since support for 5.6 ends this year.

mattund:
@sbeaver-netgate,

Thanks -- good catch.  I based my shaper.inc on an earlier point in the branch.  I've revised and made the requested changes.

Manevolent:
There is an issue with the patch where I write out `noecn` when the scheduler/AQM doesn't support that param.  Fixing now.

mattund:
Just noticed a problem with this patch where if you select a scheduler or AQM which doesn't support ECN it still writes out `noecn` so I'm going to write a quick fix for that

mattund:
Also,

```
5.2.1.  Interval

   The _interval_ parameter has the same semantics as CoDel and is used
   to ensure that the minimum sojourn time of packets in a queue used as
   an estimator by the CoDel control algorithm is a relatively up-to-
   date value.  That is, CoDel only reacts to delay experienced in the
   last epoch of length interval.  It SHOULD be set to be on the order
   of the worst-case RTT through the bottleneck to give end-points
   sufficient time to react.

   The default interval value is 100 ms.
```

My sysctl calls assume a unit in milliseconds.  I need to factor for this.  Will make another change

mattund:
Hey all,

Changes made as requested last week.  Any chance this could get reviewed soon?

sbeaver-netgate:
Nothing further needed. Just requires testing

grandrivers:
what is status on this is there an easy way to help test this?

src/etc/inc/shaper.inc
```diff
@@ -4095,6 +4507,16 @@ class dnqueue_class extends dummynet_class {
 
 	function validate_input($data, &$input_errors) {
 		parent::validate_input($data, $input_errors);
+		
+		
+		/* Limiter patch */
+		$selectedAqm = getAQMs()[$data['aqm']];
+		if (!$selectedAqm)
+			$input_errors[] = gettext("Selected scheduler not recognized: " . $data['sched'] . '.');
```
jim-p: See previous comment about printing this value, and also the message says scheduler but it looks like it should be AQM

jim-p:
@grandrivers You can test it by using the System Patches package, add `.diff` to the end of the PR URL and feed that to the package, set a path strip value of 2. So you'd use https://github.com/pfsense/pfsense/pull/3941.diff for this PR.

mattund:
1. I rewound src/etc/inc/config.inc back to you guys' base.  It was some funny EOL stuff that happened.
2. Unwrapped gettext()
3. Agreed.  Sanitized.
4. Unwrapped gettext()
5. Took out input_errors item
6. Took out input_errors item
7. I like the idea of this.  I would love to add heavier validation; unless I just affix for datatype I would need to know the constraints on which the parameters live for the AQM/scheds.  Maybe that's documented.  I can revisit this.
8. Took out input_errors item

Thanks for taking a look, Jim!  I appreciate the feedback.

blackyellowred:
on 2.4.4 devel applied using url: https://github.com/pfsense/pfsense/pull/3941.diff

This is what I see:

fq_codel

blackyellowred:
Another issue: whenever firewall rules are reloaded (by pfblockerng in this example) there is a kernel error: `config_aqm Unable to configure flowset, flowset busy!`

codel_fq

mattund:
@blackyellowred 

This was my fault; fixed now :)

blackyellowred:
Thanks its fixed now

jsbsmd:
Updated shaper.inc and rebooted. Comes up with php warning warning error.

Crash report begins.  Anonymous machine information:

amd64
11.2-RELEASE
FreeBSD 11.2-RELEASE #18 7ea1caafea9(RELENG_2_4_4): Tue Jun 26 19:56:12 EDT 2018     root@buildbot3:/builder/ce-master/tmp/obj/builder/ce-master/tmp/FreeBSD-src/sys/pfSense

Crash report details:

PHP Errors:
[28-Jun-2018 12:00:00 America/Toronto] PHP Warning:  Invalid argument supplied for foreach() in /etc/inc/shaper.inc on line 3937
[28-Jun-2018 12:00:00 America/Toronto] PHP Warning:  Invalid argument supplied for foreach() in /etc/inc/shaper.inc on line 3937
[28-Jun-2018 12:00:02 America/Toronto] PHP Warning:  Invalid argument supplied for foreach() in /etc/inc/shaper.inc on line 3937
[28-Jun-2018 12:00:02 America/Toronto] PHP Warning:  Invalid argument supplied for foreach() in /etc/inc/shaper.inc on line 3937

No FreeBSD crash data found.

mattund:
Hey @jsbsmd it's because I set the default scheduler to "FIFO" when it's defined as "fifo" in code, but PHP is case sensitive.  I'm making a patch right now for everyone.

jsbsmd:
works perfect now. thanks.

src/etc/inc/shaper.inc
```diff
@@ -3824,11 +3824,7 @@ class dnpipe_class extends dummynet_class {
 		
 		/* Limiter patch */
 		$selectedScheduler = getSchedulers()[$data['sched']];
-		if (!$selectedScheduler)
-			$input_errors[] = gettext("Selected scheduler not recognized: " . $data['sched'] . '.');
```
jim-p: This should still be tested -- Keep the test but do not put the variable in the error message.

src/etc/inc/shaper.inc
```diff
@@ -3824,11 +3824,7 @@ class dnpipe_class extends dummynet_class {
...
 		$selectedAqm = getAQMs()[$data['aqm']];
-		if (!$selectedAqm)
-			$input_errors[] = gettext("Selected scheduler not recognized: " . $data['sched'] . '.');
```
jim-p: This should still be tested -- Keep the test but do not put the variable in the error message.

src/etc/inc/shaper.inc
```diff
@@ -4511,8 +4507,6 @@ class dnqueue_class extends dummynet_class {
 		
 		/* Limiter patch */
 		$selectedAqm = getAQMs()[$data['aqm']];
-		if (!$selectedAqm)
-			$input_errors[] = gettext("Selected scheduler not recognized: " . $data['sched'] . '.');
```
jim-p: This should still be tested -- Keep the test but do not put the variable in the error message.

mattund:
@jim-p, thanks again for taking a look -- I've made the requested changes.

dtaht:
I just wanted to applaud y'all for getting fq_codel in there. Got any benchmarks? (rrul test from flent.org is good)

Manevolent:
Firstly, my baseline RTT and dev to the test server:
```
128 packets transmitted, 128 received, 0% packet loss, time 127200ms
rtt min/avg/max/mdev = 65.799/68.391/75.375/1.455 ms
```

Before: https://i.imgur.com/MyQGroP.png

After: https://i.imgur.com/w3F6duK.png

dtaht:
That's quite an improvement. Welcome to the debloated internet club. :) I'm so glad pfsense got this code running and y'all can have *way* better internet.

We have flent servers all over the world, if you would like to try a closer one (my guess is you are on the east coast?) - try flent-newark.bufferbloat.net.

It would be interesting to get a docsis-pie result also (just shape the download), but at first eyeball I don't think pie is on. (one way to tell might be: flent -H flent-newark.bufferbloat.net tcp_upload) - pie ping latency on the upload would peak at ~20ms)

"This is, of course, my home connection. I'm testing from a Debian VM hosted on a 2x X5650 server using an I350-T as the adapter in 2xGigE active load-balance LACP with the Modem. The pfSense VM is virtualized via Hyper-V on the same server, and I'm not using SR-IOV or vf's with the NIC. The Debian VM communicates with pfSense over a vSwitch on that host."

As for this I can't help but laugh at how geeky we all are. Can you fix your parent's internet with something less extreme? Your neighbors? :) It's kind of weirdly wonderful that with all those layers you can still have reasonably accurate clock resolution.

dtaht:
I'm also curious as to what happens with w/wo fq_codel, applied without a shaper, to the ethernet interface, over a 1gige or 100mbit local, native link. Linux regulates the local ring buffer with BQL (see https://lwn.net/Articles/454390/ ) which is necessary as they went hog-wild with offloads.

victorhooi:
@Manevolent Curious - how did you create those charts in https://github.com/pfsense/pfsense/pull/3941#issuecomment-413206004 above?

victorhooi:
Oh - Flent looks very awesome!

Just trying to find articles about it now:

https://blog.tohojo.dk/2017/04/the-story-of-flent-the-flexible-network-tester.html

iperf/netperf is for testing within the LAN however - so I'm curious what people are using as the remote endpoint to test for bufferbloat?

Or do you setup your own netperf/iperf system on a cloud server somewhere nearby and try that?

dtaht:
we have 6-7 public flent servers in the cloud at any given time.

flent-fremont.bufferbloat.net (west coast)
flent-newark.bufferbloat.net (east coast)
flent-london.bufferbloat.net (british coast)

And the rest are in europe. We generally hope that folk will setup netserver and irtt ( https://github.com/heistp/irtt ) on their own clouds/servers... these are not good much past 5-600Mbit.

if anyone wants to donate a server, that's nice too. :)

PR #3944, opened 2018-06-01
Fix and simplify IPv6 NAT

overhacked:
HEADS UP, this bumps the config version because of deprecating
the NAT "IPV6" protocol type. Sorry if that's not OK to do as a contributor;
please let me know the best way to submit a PR that changes the schema.

It doesn't appear that this feature was working via the GUI.
The generated pf rule contained a "proto ipv6" that wasn't
matching any packets. It was also possible to create a config
via the GUI that would cause pf to bail, complaning about an
address family mismatch between source and target.

I've eliminated the "IPV6" protocol choice, which is the
equivalent of "any", just with v6 addresses. The GUI save
code checks to make sure that Source agrees with both Destination
and Target in terms of address family. If any of those fields
are configured as aliases, the address family checking is
skipped. Most other places, pfSense doesn't do that kind of
"deep" verification, and just lets the alerts thrown to the
admin at filter reload tell the story.

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8437
- [ ] Ready for review


jim-p:
Bumping the configuration rev is OK, the only danger is that if something else bumps it before your PR is fully merged, you may have to fix the PR to bump it again to avoid a conflict.

src/etc/inc/filter.inc
```diff
@@ -1793,10 +1793,18 @@ function filter_nat_rules_generate_if ($if, $src = "any", $srcport = "", $dst =
 			$tgt = "{$natip}/32";
 		}
 	} else {
-		$natip = get_interface_ip($if);
-		if (is_ipaddr($natip)) {
-			$tgt = "{$natip}/32";
+		if (is_subnetv6($src) || is_subnetv6($dst)) {
```
jim-p: Should they not both have to be IPv6? `&&` seems safer than `||` here.

src/etc/inc/filter.inc
```diff
@@ -1793,10 +1793,18 @@ function filter_nat_rules_generate_if ($if, $src = "any", $srcport = "", $dst =
...
+			$natip = get_interface_ipv6($if);
+			if (is_ipaddrv6($natip)) {
+				$tgt = "{$natip}/128";
+			}
 		} else {
```
jim-p: This should probably test for IPv4, in case it's a broken IPv6 scenario. Then add another `else` to deal with a case that can't be resolved properly, so that it never makes a broken rule.

sbeaver-netgate:
No action in two months. Please resubmit if you can addresses the issues noted.

PR #3946, opened 2018-06-06
Update to DNSimple APIv2 endpoint

martinisoft:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8071
- [x] Ready for review

This PR updates the dnsimple dyndns integration by updating to the new v2 API endpoint. The v1 API is no longer available. I re-purposed the username to account ID and password to the API bearer token. I can assist with any documentation updates to make it clearer to the user what is required for each field since it won't be username and password anymore.

Let me know if you need help getting setup with an account to test this integration with the new API as I work for DNSimple.

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -364,6 +364,7 @@ $section->addInput(new Form_Input(
 ))->setHelp('Username is required for all types except Namecheap, FreeDNS and Custom Entries.%1$s' .
 	                'Azure: Enter your Azure AD application ID%1$s' .
 			'DNS Made Easy: Dynamic DNS ID%1$s' .
+			'DNSimple: User account ID (In the URL after the '/a/')%1$s' .
 			'Route 53: Enter the Access Key ID.%1$s' .
 			'GleSYS: Enter the API user.%1$s' .
 			'Dreamhost: Enter a value to appear in the DNS record comment.%1$s' .
@@ -378,6 +379,7 @@ $section->addPassword(new Form_Input(
 ))->setHelp('FreeDNS (freedns.afraid.org): Enter the "Authentication Token" provided by FreeDNS.%1$s' .
 	                'Azure: client secret of the AD application%1$s' .
 			'DNS Made Easy: Dynamic DNS Password%1$s' .
+			'DNSimple: User account token%1$s' .
 			'DigitalOcean: Enter API token%1$s' .
 			'Route 53: Enter the Secret Access Key.%1$s' .
 			'GleSYS: Enter the API key.%1$s' .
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_dyndns_edit.php
```diff
@@ -364,6 +364,7 @@ function build_if_list() {
 ))->setHelp('Username is required for all types except Namecheap, FreeDNS and Custom Entries.%1$s' .
 	                'Azure: Enter your Azure AD application ID%1$s' .
 			'DNS Made Easy: Dynamic DNS ID%1$s' .
+			'DNSimple: User account ID (In the URL after the '/a/')%1$s' .
```
jim-p: Does that actually work like you show? It has unescaped single quotes inside single quotes looks like it would throw an error and/or not format that string properly.

sbeaver-netgate: That syntax error leads me to think that the code was not tested sufficiently. Please perform a regression test of your changes and post hte results here.

martinisoft: > That syntax error leads me to think that the code was not tested sufficiently. Please perform a regression test of your changes and post hte results here.

Any pointers on where I can find instructions running regression tests? I have since corrected the issue with escaped quotes in bad3542. The other half of this code is currently running on my local install of pfSense.

martinisoft:
I've corrected the changes so I'd like to see if this can be reviewed and potentially merged. We've just fully removed the code for the v1 API so this will come in the nick of time.

rbgarga:
@martinisoft the only pending here is testing. A testing account would be awesome!

martinisoft:
> @martinisoft the only pending here is testing. A testing account would be awesome!

I'd be happy to, but I don't see any test setup instructions within the project. Can you point me to them?

emanuelstanciu:
@martinisoft Hi Aaron, I searched around for this v2 implementation and this is the only thread I could find with even remotely relevant information for a new issue I'm seeing.
I configured the Dynamic DNS with DNSimple and filled all the relevant fields. The DNS is correctly being updated on the domain, however the Cached IP stays at 0.0.0.0 and is red.

In the syslog I see:

```
Apr 15 12:14:39 | php-fpm |   | /services_dyndns_edit.php: phpDynDNS (<domain>): (Unknown Response)
-- | -- | -- | --
Apr 15 12:14:39 | php-fpm |   | /services_dyndns_edit.php: phpDynDNS (<domain>): PAYLOAD: {"data":{"id":record_id,"zone_id":"domain","parent_id":null,"name":"","content":"ip_addr","ttl":60,"priority":null,"type":"A","regions":["global"],"system_record":false,"created_at":"2019-02-21T08:34:00Z","updated_at":"2019-04-15T11:13:28Z"}}
```

So, it seems to be the response that's now changed and interface/cached ip would not be updated and seen as a successful update.
Do you know what this can be? Thanks in advance for your help.

martinisoft:
Hi @emanuelstanciu 

I'm not sure what the issue is in the response or what it's looking for. Some of the options like the `$needsIP` thing aren't well documented so getting this change in was a lot of just fixing their curl commands to work with our new API structure and we return a response with the change that was implemented.

PR #3947, opened 2018-06-09
routing, fix setting the default-route when the configured default gateway is a dynamic pppoe gateway

PiBa-NL:
fix setting the default-route when the configured default gateway is a dynamic pppoe gateway. It doesnt have a gateway-status when it hasn't connected yet.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8561
- [x] Ready for review

### DIFF
--- a/src/etc/rc.newwanip
+++ b/src/etc/rc.newwanip
@@ -183,7 +183,12 @@ if (!empty($gre)) {
 	}
 }
 
-if (platform_booting()) {
+if (platform_booting() && in_array(substr($interface_real, 0, 3), array("ppp", "ppt", "l2t"))) {
+	// unlike dhcp interfaces which wait until they get an ip, a ppp connection lets the boot continue while 
+	// trying to aquire a ip address so to avoid a race condition where it would be possible that the default
+	// route would not be set, this script must continue to use the new assigned ip even while booting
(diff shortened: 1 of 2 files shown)


PiBa-NL:
This one is 'ready', as it fixes the problem that no default-route is set for pppoe during bootup when one is configured.

Ill add a different pull-request for the feature/bug where default-gateway setting 'none' is not really none. (discussed that change with jimp, but i think these fixes should be pulled asap for pppoe people..)

src/etc/rc.newwanip
```diff
@@ -183,7 +183,12 @@ if (!empty($gre)) {
 	}
 }
 
-if (platform_booting()) {
+if (platform_booting() && strpos($interface_real, "ppp") !== 0) {
```
jim-p: This could also affect PPTP WAN and L2TP WANs, the same as PPPoE and PPP, so you should check for those as well. Maybe something like this:
```
in_array(substr($interface_real, 0, 3), array("ppp", "ppt", "l2t"))
```
Or a `preg_match()` with some variation of `"/^ppp|^l2tp|^pptp/"`

PiBa-NL:
new commit pushed over the previous one. to apply the fix to all of array("ppp", "ppt", "l2t")

PR #3904, opened 2018-01-23
IPsec mobile clients with different (virtual) IP addresses by (EAP) identity

Hobby-Student:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8292
- [ ] Ready for review

this is the first mod for extending mobile connection. this enables multiple mobile clients with different IPsec options handled by firewall rules.
this is a strongswan default feature, it's just not configurable by the WebUI of pfSense.

please also have a look at the forum post: https://forum.pfsense.org/index.php?topic=142560.0

PR #3949, opened 2018-06-20
[Fix for #3904] IPsec mobile clients with different (virtual) IP addresses by (EAP) identity

Hobby-Student:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8292
- [x] Ready for review

there were 2 failures in #3904 which are now fixed. Also added support for individual DNS per user (strongswan config option).

sorry for the wrong brackets.

### DIFF
--- a/src/usr/local/www/vpn_ipsec_keys_edit.php
+++ b/src/usr/local/www/vpn_ipsec_keys_edit.php
@@ -185,20 +185,20 @@ $section->addInput(new Form_Input(
 
 $section->addInput(new Form_Select(
 	'ident_type',
-	'*Identifier type',
+	'Identifier type',
 	$pconfig['ident_type'],
 	build_ipsecid_list()
 ))->setWidth(4)->setHelp('Optional: specify identifier type for strongswan');
 
 $section->addInput(new Form_IpAddress(
 	'pool_address',
-	'*Virtual Address Pool',
+	'Virtual Address Pool',
 	$pconfig['pool_address']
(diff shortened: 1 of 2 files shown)


src/usr/local/www/vpn_ipsec_keys_edit.php
```diff
@@ -186,9 +192,15 @@ function build_ipsecid_list() {
 
 $section->addInput(new Form_IpAddress(
 	'pool_address',
-	'*IPv4 address',
+	'*Virtual Address Pool',
```
jim-p: Why the name change? If it must be an IPv4 Address, the original name seems fitting.

Also if it's optional we can remove that `*` from before the name since that denotes a required option. While you are here, the option just above, "Identifier Type", can also have the `*` removed since it is optional.

Hobby-Student: Virtual Address Pool seems to fit better, because you can choose the subnet yourself. For me, IPv4 Address means one single host...

for the rest, you are absolutely right. will change it.

jim-p: But it still only assigns the client a single address with the chosen subnet mask, right?
I suppose the description works either way.

Hobby-Student: yes, every client will get its own address based on the subnet. It's the same like the Virtual Address Pool in "VPN/IPsec/Mobile Clients". Now the description is consistent over different (IPsec) settings pages.

PR #3950, opened 2018-06-20
Fix System information dashboard to check all swap on system. Bug #8587

marcelloc:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8587
- [ ] Ready for review

### DIFF
--- a/src/usr/local/www/includes/functions.inc.php
+++ b/src/usr/local/www/includes/functions.inc.php
@@ -177,7 +177,7 @@ function disk_usage($slice = '/') {
 }
 
 function swap_usage() {
-	exec("/usr/sbin/swapinfo", $swap_info);
+	exec("/usr/sbin/swapinfo | /usr/bin/grep Total", $swap_info);
 	$swap_used = "";
 	foreach ($swap_info as $line) {
 		if (preg_match('/(\d+)%$/', $line, $matches)) {
--- a/src/usr/local/www/includes/functions.inc.php
+++ b/src/usr/local/www/includes/functions.inc.php
@@ -177,7 +177,7 @@ function disk_usage($slice = '/') {
 }
 
 function swap_usage() {
-	exec("/usr/sbin/swapinfo | /usr/bin/grep Total", $swap_info);
+	exec("/usr/sbin/swapinfo | /usr/bin/tail -1", $swap_info);
 	$swap_used = "";
 	foreach ($swap_info as $line) {
 		if (preg_match('/(\d+)%$/', $line, $matches)) {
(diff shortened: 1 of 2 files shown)


src/usr/local/www/includes/functions.inc.php
```diff
@@ -177,7 +177,7 @@ function disk_usage($slice = '/') {
 }
 
 function swap_usage() {
-	exec("/usr/sbin/swapinfo", $swap_info);
+	exec("/usr/sbin/swapinfo | /usr/bin/grep Total", $swap_info);
```
jim-p: `swapinfo` does not output a line with `Total` for me, which makes this always empty/blank.

    : /usr/sbin/swapinfo
    Device          1K-blocks     Used    Avail Capacity
    /dev/label/swap0    524284        0   524284     0%

marcelloc:
Forgot to test on systems without more then on swap file.

The code needs more checks

marcelloc:
A tail -1 works on both situations.
Pushed to fix for the fix.

src/usr/local/www/includes/functions.inc.php
```diff
@@ -177,7 +177,7 @@ function disk_usage($slice = '/') {
 }
 
 function swap_usage() {
-	exec("/usr/sbin/swapinfo", $swap_info);
+	exec("/usr/sbin/swapinfo | /usr/bin/tail -1", $swap_info);
```
rbgarga: This will not work for multiple swap devices. Instead use `swapinfo | tain -n  -1` which will just remove header line

src/usr/local/www/widgets/widgets/system_information.widget.php
```diff
@@ -435,7 +435,7 @@
 					<div class="progress-bar progress-bar-striped" role="progressbar" aria-valuenow="<?=$swapusage?>" aria-valuemin="0" aria-valuemax="100" style="width: <?=$swapusage?>%">
 					</div>
 				</div>
-				<span><?=$swapusage?>% of <?= sprintf("%.0f", `/usr/sbin/swapinfo -m | /usr/bin/grep -v Device | /usr/bin/awk '{ print $2;}'`) ?> MiB</span>
+				<span><?=$swapusage?>% of <?= sprintf("%.0f", `/usr/sbin/swapinfo -m | /usr/bin/tail -1 | /usr/bin/awk '{ print $2;}'`) ?> MiB</span>
```
rbgarga: Same here

src/usr/local/www/includes/functions.inc.php
marcelloc: @rbgarga, the fix is to show only one swap info . Or the only one present on system (default) or the total of swap used while using more then one.

I think there is no need to show individual usage od each swap.

rbgarga: Oh, that makes senes. Thanks!

PR #3951, opened 2018-06-22
Provide standard DH groups from RFC 7919.

whislock:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/8582
- [X] Ready for review

In August 2016, IETF issued RFC 7919, _Negotiated Finite Field Diffie-Hellman Ephemeral Parameters for Transport Layer Security (TLS)_, providing audited and standardized parameters for Diffie-Hellman key agreement usage. As these have been audited for fitness, they are considered to be the safest for end users, and it is recommended to use these over those locally generated (e.g. by OpenSSL).

This PR replaces the existing 2048 and 4096-bit DH groups provided with pfSense with those from RFC 7919 (converted into usable PEM form by the provided Python script), as well as adding groups at sizes 3072, 6144, and 8192.

IETF does **not** provide a standard parameter set at the 1024 bit size, as this is considered below the threshold for secure usage. The existing 1024-bit group has been left intact for compatibility purposes.

This will be a transparent change to end users. No additional configuration is required to utilize these parameters (unless selecting a different size group than was previously available). Up to the 4096-bit group size, there are no compatibility issues with any facility within pfSense that makes use of the provided DH parameter files. At the 6144 and 8192 bit sizes, no compatibility issues are known; however, it is extremely uncommon to utilize DH groups at this size, so edge cases may exist. If needed or desired, I can remove them from the PR.


whislock:
Also included the script used to convert the hex-format parameters into PEM files.

whislock:
> Would there be an advantage to keeping the current DH parameters and calling them something like "2048-legacy" and so on, in case a user encounters a problem with these? -- Someone could always manually generate their own if they have a problem. I wouldn't expect much trouble here, though, unless OpenVPN chokes on them in some way.

Not that I can see. I've tested OpenVPN, Apache, and nginx with the RFC 7919 parameters all the way up to the 8192-bit group with no issues. Key negotiation might get bogged down on low-end hardware with groups that large, but I see that as a performance consideration for the administrator to decide.

> Seems like a good time to increase the default DH length for new OpenVPN server instances in src/usr/local/www/vpn_openvpn_server.php and in the OpenVPN wizard to either 2048 or 4096. That could be a separate PR though.

That depends on what the default "security margin" we're targeting is. With modern recommendations for 128-bit security, I'd set the defaults to AES-128, SHA256, and the 3072-bit DH group. 2048-bit DH is roughly approximate to 112-bit security (think 3DES).

jim-p:
Good news about compatibility. Looks like an easy win then.

As for OpenVPN that may need some other consideration. The defaults for Encryption and Auth Digest are not ideal either, but they are based on OpenVPN's own defaults. It may be time to increase them all.

whislock:
> It may be time to increase them all.

Want me to put a PR together for OpenVPN default values?

whislock:
@jim-p I neglected to add the new group sizes to `etc/inc/openvpn.inc`. I originally adjusted the titles displayed to indicate which sizes were RFC-provided, but then realized that nothing was stopping the user from overwriting them, making that label meaningless.

jim-p:
It's OK to keep the descriptions as-is. Like you said, someone could regenerate the parameters and render the description inaccurate. We could add some trickery to detect that and change the description but it doesn't seem like it would be worth the hassle. Otherwise this looks OK to me now. Still waiting on some other internal feedback.

PR #3954, opened 2018-06-23
Update crypto defaults for OpenVPN/IPsec

whislock:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/8594
- [x] Ready for review

Updated default cipher to AES-128, default hash to SHA256, default DH group to 2048 bit (IPsec DH group 14). These are **potential changes**, and are here as a placeholder while the proposal to update them is being discussed. In the mean time, please make sure that I've touched everything necessary to make the UI/admin experience consistent.

### DIFF
--- a/src/usr/local/www/vpn_ipsec_phase1.php
+++ b/src/usr/local/www/vpn_ipsec_phase1.php
@@ -170,8 +170,7 @@ if (isset($p1index) && $a_phase1[$p1index]) {
 // default value for new P1 and failsafe to always have at least 1 encryption item for the Form_ListItem
 if (!is_array($pconfig['encryption']['item']) || count($pconfig['encryption']['item']) == 0) {
 	$item = array();
-	$item['encryption-algorithm'] = array('name' => "aes");
-	$item['encryption-algorithm'] = array('keylen' => 128);
+	$item['encryption-algorithm'] = array('name' => "aes", 'keylen' => 128);
 	$item['hash-algorithm'] = "sha256";
 	$item['dhgroup'] = "14";
 	$pconfig['encryption']['item'][] = $item;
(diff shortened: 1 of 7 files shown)


src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -171,8 +171,9 @@
 if (!is_array($pconfig['encryption']['item']) || count($pconfig['encryption']['item']) == 0) {
 	$item = array();
 	$item['encryption-algorithm'] = array('name' => "aes");
-	$item['hash-algorithm'] = "sha1";
-	$item['dhgroup'] = "2";
+	$item['encryption-algorithm'] = array('keylen' => 128);
```
jim-p: One more little nit upon further review, this should still specify `'name' => "aes"` in addition to the `keylen`. The browser may appear to do the correct thing here, but it shouldn't rely on the default selection behavior of the browser there.

whislock: That was actually my intent, but I somehow missed that I was overwriting the first set of `encryption-algorithm` with the second operation.

whislock:
Commited requested fix.

sdj38:
I noticed one thing that might need to be changed. From the proposal it states:
```
IPsec Phase 2:
Encryption Algorithms: AES, AES-128GCM
Hash Algorithms: SHA256
PFS Key Group: Group 14 (2048 bits)
```

This is what is seen as the default for Phase 2
screen shot 2018-07-05 at 10 34 57 am

If I understand correctly AES and AES-128GCM should be selected. Also `Auto` should probably be `128 bits` to match up. Everything else is looking good so far.

whislock:
Sorry, the original email notification died in email hell. I'll push a fix by the end of the weekend.

whislock:
Changes made. Sorry that took so long!

PR #3959, opened 2018-07-04
Fix #8617

jwsi:
File in /usr/local/share/pear/Auth is actually RADIUS.php not RADIUS.inc.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8617
- [x] Ready for review


rbgarga:
I already pushed a fix. Thanks!

PR #3960, opened 2018-07-04
ipsec, fix Phase1 edit page 'section' class does not have a setHelp() function

PiBa-NL:
ipsec, fix Phase1 edit page 'section' class does not have a setHelp() function

- [x] Redmine Issue: n/a
- [x] Ready for review

fixes the changes from https://github.com/pfsense/pfsense/pull/3939


jim-p:
Tests OK here. Fixes the original error and retains the warning.

PR #3962, opened 2018-07-08
routing, when wan connection is made gatewaymonitor should start before determining default route

PiBa-NL:
routing, when a new pppoe connection is made the gatewaymonitor should be started before decisions about default route can be properly made. also for 'automatic' provide a fallback to the first enabled gateway thats configured, just in case..
As found by w0w:  https://forum.netgate.com/post/776018

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8504
- [x] Ready for review


rbgarga:
@PiBa-NL can you please rebase your fork

PiBa-NL:
@rbgarga , ok rebase done

PR #3965, opened 2018-07-13
changes to mobile ipsec dns to support new features

Hobby-Student:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8644
- [x] Ready for review

discussion started in [redmine 8292](https://redmine.pfsense.org/issues/8292).
[strongswan wiki](https://wiki.strongswan.org/projects/strongswan/wiki/VirtualIp):

> **DNS servers**
> DNS servers and other attributes can be assigned by plugins (e.g. the attr plugin) or **since 5.0.1 directly in ipsec.conf by use of the rightdns option**. In swanctl.conf each pool in the pools section may define a list of attributes to assign to clients.

this change is linked to #3949

Now each client can have an individual assigned DNS.

### DIFF
--- a/src/etc/inc/vpn.inc
+++ b/src/etc/inc/vpn.inc
@@ -522,25 +522,7 @@ EOD;
 
 	if (is_array($a_client) && isset($a_client['enable'])) {
 		$strongswan .= "\t\tattr {\n";
-
-		$cfgservers = array();
-		if (!empty($a_client['dns_server1'])) {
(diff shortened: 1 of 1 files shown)


Hobby-Student:
I did some tests and couldn't see any difference in the "old" behaviour - working as before. Just the possibility to overwrite the rigthdns with extending a connection profile.

Hobby-Student:
little feedback:
have this patch in several pfsenses for testing and everything is working fine.

PR #3974, opened 2018-08-21
Captive Portal : Automatically store the user name linked to a MAC added as pass-through

Augustin-FL:
- [x] Resolve bug mentioned in comments of redmine Issue https://redmine.pfsense.org/issues/3124
- [x] Ready for review

### DIFF
--- a/src/etc/inc/captiveportal.inc
+++ b/src/etc/inc/captiveportal.inc
@@ -1028,7 +1028,7 @@ function captiveportal_prune_old_automac() {
 	global $g, $config, $cpzone, $cpzoneid;
 
 	if (is_array($config['captiveportal'][$cpzone]['passthrumac']) &&
-	    isset($config['captiveportal'][$cpzone]['passthrumacaddusername'])) {
+	    isset($config['captiveportal'][$cpzone]['passthrumacadd'])) {
 		$tmpvoucherdb = array();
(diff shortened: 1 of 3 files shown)


src/etc/inc/captiveportal.inc
```diff
@@ -2456,7 +2455,8 @@ function portal_allow($clientip, $clientmac, $username, $password = null, $attri
 		$my_redirurl = $config['captiveportal'][$cpzone]['redirurl'];
 	}
 
-	if (isset($config['captiveportal'][$cpzone]['logoutwin_enable']) && !$passthrumac) {
+	if (isset($config['captiveportal'][$cpzone]['logoutwin_enable']) && 
+	!isset($config['captiveportal'][$cpzone]['passthrumacadd'])) {
```
rbgarga: Please fix indent here

Augustin-FL:
And about the case where the user login form doesn't require a username.... well```Disable concurrent logins``` makes basically no sense in this situation, so the previous MAC should not be deleted when an user does log in without username.

I added ``` && ($username != 'unauthenticated')``` surrounding the deletion code ([here](https://github.com/pfsense/pfsense/pull/3974/files#diff-764cc5453dc616c068ba7adeb68ba73eR2234)). It should be enough to prevent such case.

PR #3976, opened 2018-09-04
Add RADIUS IP priority to vpn_ipsec_mobile.php. Implements #8160

Amith211:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8160 (feature)
- [x] Ready for review

Hi,

I have to preface this PR by saying that I'm not very experienced with PHP or the inner workings with pfSense, so bear that in mind. 

I have a requirement to be able to have IPSec VPN clients receive their IP addresses via RADIUS for some groups and others from a simple pool. I decided to have a stab at implementing the solution mentioned in Redmine issue 8160.

Some basic manual testing has been done and it seems to be working correctly; enabling and disabling this feature also appears to work as expected. As a user of the Community Edition, I believe I ought to give back to the project by sharing my changes (however poor they may be :laughing:).

In implementing this feature, however, I had to add code to filter.inc (see line 2197) but this does not seem to be the correct place for that.

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2192,10 +2192,6 @@ function filter_nat_rules_generate() {
 		update_filter_reload_status(gettext("Creating outbound NAT rules"));
 		$tonathosts_array = filter_nat_rules_automatic_tonathosts();
 		$tonathosts = implode(" ", $tonathosts_array);
-
-		// Remove %radius from string
-		$tonathosts = str_replace('%radius,', '', $tonathosts);
-
 		$numberofnathosts = count($tonathosts_array);
 
 		$natrules .= "\n# Subnets to NAT \n";
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -2192,6 +2192,10 @@ function filter_nat_rules_generate() {
 		update_filter_reload_status(gettext("Creating outbound NAT rules"));
 		$tonathosts_array = filter_nat_rules_automatic_tonathosts();
 		$tonathosts = implode(" ", $tonathosts_array);
+
+		// Remove %radius from string
+		$tonathosts = str_replace('%radius,', '', $tonathosts);
+
 		$numberofnathosts = count($tonathosts_array);
 
 		$natrules .= "\n# Subnets to NAT \n";
--- a/src/usr/local/www/vpn_ipsec_mobile.php
+++ b/src/usr/local/www/vpn_ipsec_mobile.php
@@ -59,8 +59,15 @@ if (count($a_client)) {
 
 	$pconfig['user_source'] = $a_client['user_source'];
 	$pconfig['group_source'] = $a_client['group_source'];
+   
+	if (strpos($a_client['pool_address'], '%radius') !== false) {
 
-	$pconfig['pool_address'] = $a_client['pool_address'];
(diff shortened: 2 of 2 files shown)


src/etc/inc/filter.inc
```diff
@@ -2192,6 +2192,10 @@ function filter_nat_rules_generate() {
 		update_filter_reload_status(gettext("Creating outbound NAT rules"));
 		$tonathosts_array = filter_nat_rules_automatic_tonathosts();
 		$tonathosts = implode(" ", $tonathosts_array);
+
+		// Remove %radius from string
+		$tonathosts = str_replace('%radius,', '', $tonathosts);
```
Amith211: I don't believe this belongs here, but is required in the current implementation to remove invalid data.

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -252,7 +266,13 @@
 		$client['group_source'] = $pconfig['group_source'];
 
 		if ($pconfig['pool_enable']) {
-			$client['pool_address'] = $pconfig['pool_address'];
+			if($pconfig['radius_ip_priority_enable'])
```
skreii: Brackets are different here.

jim-p: Add a space after `if` ([ref](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style))
Please use ["K&R, BSD KNF variant style" formatting](https://www.netgate.com/docs/pfsense/development/developer-style-guide.html#indent-style).

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -59,8 +59,15 @@
 
 	$pconfig['user_source'] = $a_client['user_source'];
 	$pconfig['group_source'] = $a_client['group_source'];
+   
+	if (strpos($a_client['pool_address'], '%radius') !== false) {
+
```
jim-p: This blank line can be removed

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -76,7 +83,8 @@
 	$pconfig['wins_server2'] = $a_client['wins_server2'];
 	$pconfig['pfs_group'] = $a_client['pfs_group'];
 	$pconfig['login_banner'] = $a_client['login_banner'];
-
+	$pconfig['radius_ip_priority_enable'];
```
jim-p: This does not look like it would do anything. Does it serve a purpose?

Maybe you intended to set this value based on its equivalent in `$a_client`?

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -252,7 +266,13 @@
 		$client['group_source'] = $pconfig['group_source'];
 
 		if ($pconfig['pool_enable']) {
-			$client['pool_address'] = $pconfig['pool_address'];
+			if($pconfig['radius_ip_priority_enable'])
+			{
+				$client['pool_address'] = '%radius,' . $pconfig['pool_address'];
```
jim-p: This is a rather awkward way to do this. Normally you'd check the value of `radius_ip_priority_enable` in the backend code every place that uses `pool_address` and add it there, rather than having it added in the GUI field directly.

sbeaver-netgate:
Closing since no activity in 5 months. Please resubmit if you wish it to be reconsidered.

PR #3980, opened 2018-09-14
Minor fix to Outbound NAT page

luckman212:
Underscores (`_`) are currently replaced with spaces in alias names on the Outbound NAT page. This isn't consistent with the rest of the UI. I updated it to mirror the format used on the other NAT pages.

Also added `gettext()` wrapper around the hover labels for i18n.

image

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8893
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/firewall_nat_out.php
+++ b/src/usr/local/www/firewall_nat_out.php
@@ -351,11 +351,11 @@ print($form);
 
 						if (isset($alias['src'])):
 ?>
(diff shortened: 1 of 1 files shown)


jim-p:
It's done deliberately.

luckman212:
@jim-p Sorry to nitpick but the `_<wbr>` (which is what is used on other Firewall config pages) does still allow the wrapping to take place, while preserving the underscores. I just tested it:

image

Plus there was the other tiny change of the `gettext()` wrapping.

I think this should still be merged.

jim-p:
You are right that one page is inconsistent with the rest. They should all be done the same way. Not going to hold back 2.4.4 for a minor change like this though.

luckman212:
Thank you. 2.4.4-p1 it is then 👍

PR #3981, opened 2018-09-18
NTP status page not honouring NTP ACLs

NanoCaiordo:
- [x] Redmine issue https://redmine.pfsense.org/issues/7609


NanoCaiordo:
I had I thought about adding a break when an IP match, this to stop the loop as soon it found a match improving performances.

It all depends on the philosophy of the first vs last match win. Either won't impose any restriction, but since the ACL page can't reorder any entry, for now, I guess it's better to let the last win.

PR #3983, opened 2018-09-18
hide the "Drag to Change Order" text if roworderdragging is disabled

luckman212:
If `$config['system']['webgui']['roworderdragging']` is true, then dragging to reorder firewall rules is disabled. This simple change hides the header text saying that you can drag, if in fact, you cannot.

- [ ] Redmine Issue: not needed?
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -328,6 +328,12 @@ if (isset($config['interfaces'][$if]['blockbogons'])) {
 	$showblockbogons = true;
 }
 
+if (isset($config['system']['webgui']['roworderdragging'])) {
+	$rules_header_text = 'Rules';
+} else {
+	$rules_header_text = 'Rules (Drag to Change Order)';
+}
+
 /* Load the counter data of each pf rule. */
 $rulescnt = pfSense_get_pf_rules();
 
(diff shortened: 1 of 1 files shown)


NanoCaiordo:
Patch doesn't work, as I immediately suspected "print" inside "gettext" will let the string be printer instead of processed for translation. Moreover when the option is enabled the order of the strings is  reversed and misunderstood.

Spanish, original:
Reglas (arrastrar para cambiar el orden)

Patched Spanish, original:
- (Drag to Change Order)Rules1

luckman212:
I did not create the `roworderdragging` flag so I wasn't the one who named it. I agree it's "backwards" and should be named `disable_row_dragging` or something like that. But, I didn't want to change it since it was already in use elsewhere.

Not quite sure how to fix the translation issue, I would have thought the inner text would have been executed first and then the entire string evaluated by `gettext()` but I guess that's not how it works? Will try another method...

luckman212:
I changed my language to Spanish just to test this. I see that even on the General config page, there are translations missing there. I am not too familiar with adding translations, but if someone can point me in the right direction I can try to help. Is [Zanata](https://www.netgate.com/blog/pfsense-software-translations-with-zanata.html) still the right method?

image

luckman212:
@NanoCaiordo can you try this now? I just pushed an update that should work.

image

luckman212:
@NanoCaiordo I have a new unified commit if you want to try, since I imagine it was a pain with the previous being split over 2 commits
try: fd4dc22e73a43c644e8972923c98abb86a27889c

luckman212:
closing this one and replacing with the squashed PR: https://github.com/pfsense/pfsense/pull/3990

PR #3984, opened 2018-09-18
reduce the log spew for gateway switch events

luckman212:
When a gateway failure/defgw switch event occurs, there seem to be some code paths that get iterated over dozens of times in a very short span of time. This floods system.log making it hard to read. Specifically, these 2 phrases are repeated over and over:
```
...Gateways status could not be determined, considering all as up/active
...Keep current gateway, its already part of the group members
```

This is part of an effort to try to make the logs readable again, while still allowing verbose logging if needed via a new hidden config option from the developer shell:

    $config['system']['gw-debug'] = true;
    write_config("Enable verbose gateway switch logging");
    exec
    exit

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8914
- [x] Ready for review

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -1257,13 +1257,11 @@ function get_gwgroup_members_inner($group, $gateways_status, $gateways_arr, $vip
 	$tiers_count = count($tiers);
 	if ($tiers_count == 0) {
 		/* Oh dear, we have no members! Engage Plan B */
-		if (!platform_booting()) {
-			if (isset($config['system']['gw-debug'])) {
-				$msg = sprintf(gettext('Gateways status could not be determined, considering all as up/active. (Group: %s)'), $group['name']);
-				log_error($msg);
-				notify_via_growl($msg);
-				//notify_via_smtp($msg);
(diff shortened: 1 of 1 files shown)


src/etc/inc/gwlb.inc
```diff
@@ -1256,10 +1258,12 @@ function get_gwgroup_members_inner($group, $gateways_status, $gateways_arr, $vip
 	if ($tiers_count == 0) {
 		/* Oh dear, we have no members! Engage Plan B */
 		if (!platform_booting()) {
-			$msg = sprintf(gettext('Gateways status could not be determined, considering all as up/active. (Group: %s)'), $group['name']);
-			log_error($msg);
-			notify_via_growl($msg);
-			//notify_via_smtp($msg);
+			if (isset($config['system']['gw-debug'])) {
```
rbgarga: Please move this conditionar to prior `if (!platform_booting() && ...)` and reduce unnecessary indent

luckman212: Done!

luckman212:
Thanks, I made those changes.

PR #3985, opened 2018-09-20
fix a few bugs in system.inc and system.php

luckman212:
I've been working on fixing a few things related to static route creation/deletion from DNS server bindings on System > General. There were at least 2 bugs there, which could cause dangling static routes to remain on the system without indication. This has the potential to cause serious issues.

Summary for this PR:

- Fixes bug that was causing static routes to not be deleted after removing DNS servers from your list (see repro steps on the redmine)
- Fixes bug that caused orphan `<dnsNgw>` entries to remain in config.xml even after deleting those gateways
- Reduces log spew for route deletes where route did not exist (uses `-q` flag). This was happening every time a DNS server was deleted that had its gw set to `none` since there was never a route created in the first place
- Fixes truncated IPv6/long gateway names by adjusting field widths using `->setWidth()` on some form elements
- Better worded and more detailed error message when trying to set a gateway for a DNS server that lies within a directly-connected subnet
- Fixes broken logic for `$multiwan` detection (previous code would enable "multiwan" mode even if there was just a single interface configured with a gateway)

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8922
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -139,11 +139,11 @@ foreach ($timezonedesc as $idx => $desc) {
 	    sprintf(ngettext('(%1$s hour %2$s GMT)', '(%1$s hours %2$s GMT)', intval($hr_offset)), $hr_offset, $direction_str);
 }
 
-$multiwan = false;
+$multiwan = 0;
 $interfaces = get_configured_interface_list();
 foreach ($interfaces as $interface) {
 	if (interface_has_gateway($interface)) {
-		$multiwan = true;
+		$multiwan++;
 	}
 }
 
@@ -221,7 +221,7 @@ if ($_POST) {
 		if ($_POST[$dnsgwname] && ($_POST[$dnsgwname] <> "none")) {
 			foreach ($direct_networks_list as $direct_network) {
 				if (ip_in_subnet($_POST[$dnsname], $direct_network)) {
-					$input_errors[] = sprintf(gettext("A gateway can not be assigned to DNS '%s' server which is on a directly connected network."), $_POST[$dnsname]);
+					$input_errors[] = sprintf(gettext("A gateway cannot be specified for %s because that IP is part of a directly connected subnet %s. To use that nameserver, change its Gateway to `none`."), $_POST[$dnsname], $direct_network);
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


luckman212:
Congratulations on the release of 2.4.4 today. I hope this one can make it in for 2.4.4-p1

src/usr/local/www/system.php
```diff
@@ -506,7 +526,7 @@ function is_timezone($elt) {
 		$dnsserver
 	))->setHelp(($is_last_dnsserver) ? $dnsserver_help:null);
 
-	if ($multiwan)	{
+	if ($multiwan > 1)	{
```
rbgarga: Replace TAB by space between `)` and `{`

src/usr/local/www/system.php
```diff
@@ -221,7 +221,7 @@ function is_timezone($elt) {
 		if ($_POST[$dnsgwname] && ($_POST[$dnsgwname] <> "none")) {
 			foreach ($direct_networks_list as $direct_network) {
 				if (ip_in_subnet($_POST[$dnsname], $direct_network)) {
-					$input_errors[] = sprintf(gettext("A gateway can not be assigned to DNS '%s' server which is on a directly connected network."), $_POST[$dnsname]);
+					$input_errors[] = sprintf(gettext("A gateway cannot be specified for %s because that IP is part of a directly connected subnet %s. To use that nameserver, change its Gateway to `none`."), $_POST[$dnsname], $direct_network);
```
jim-p: Change "IP" to "IP address" or "address" ([ref](https://www.netgate.com/docs/pfsense/references/style-guide.html#what-to-avoid))

src/usr/local/www/system.php
```diff
@@ -242,6 +242,22 @@ function is_timezone($elt) {
...
+		  	$dns_servers_arr = $config['system']['dnsserver'];
+	 		foreach ($dns_servers_arr as $arr_index => $this_dnsserver) {
+	   			$i = (int)$arr_index + 1;
+	   			$this_dnsgw = $config['system']['dns'.$i.'gw'];
+				unset($gatewayip);
+				unset($inet6);
+				if ((!empty($this_dnsgw)) && ($this_dnsgw != 'none') && (!empty($this_dnsserver))) {
+					$gatewayip = lookup_gateway_ip_by_name($this_dnsgw);
+					$inet6 = is_ipaddrv6($gatewayip) ? '-inet6 ' : '';
+					mwexec("/sbin/route -q delete -host {$inet6}{$this_dnsserver} {$gatewayip}");
```
jim-p: You do not need to specify the gateway when deleting a route, and doing so could be more error prone than leaving it off in this case.

luckman212: Got it. Fixed.

src/usr/local/www/system.php
luckman212: Done.

src/usr/local/www/system.php
luckman212: Fixed

luckman212:
I also made a tiny correction on the helptext (capitalize "NAT" and remove the `.`)

src/usr/local/www/system.php
```diff
@@ -139,11 +139,11 @@ function is_timezone($elt) {
 	    sprintf(ngettext('(%1$s hour %2$s GMT)', '(%1$s hours %2$s GMT)', intval($hr_offset)), $hr_offset, $direction_str);
 }
 
-$multiwan = false;
+$multiwan = 0;
 $interfaces = get_configured_interface_list();
 foreach ($interfaces as $interface) {
 	if (interface_has_gateway($interface)) {
-		$multiwan = true;
+		$multiwan++;
```
rbgarga: Why not break foreach here instead of changing variable to int?

luckman212: That's how it was before, but you're right it's not efficient. I just pushed a small change to break out of the loop as soon as its >1.

I wonder if the whole "hide the gateways menu if there is only 1" is even worth it. There doesn't seem to be much downside to me of always showing the popup menus, and they will just default to "none" on single-WAN systems.

It could even be useful—I wonder what happens when you have a gateway selected for a DNS server and then that gateway is removed. Does the DNS change back to "none"? Or does the system prevent you from deleting it? If not, then the config might end up with a DNS server bound to a gateway and since the GUI is now hidden, no way to change it back aside from deleting/re-adding (I haven't tested this).

luckman212:
@rbgarga @jim-p Anything else holding this one up? I made the requested changes some time ago...

rbgarga:
@luckman212 sorry for the long delay. Could you please rebase your fork?

sbeaver-netgate:
Please resolve the conflicts as noted above.

luckman212:
@rbgarga I rebased my fork, is that good to go now, or do I have to also rebase the branch?

PR #3988, opened 2018-09-20
gateways, put user back in control of gateway order / precedence

PiBa-NL:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8910
- [x] Ready for review

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -545,10 +545,10 @@ function return_gateways_status_text($byname = false, $brief = false) {
 }
 
 function compare_gateway_order_configured($a, $b) {
-	/* XXX WAN always has precedence */
-	if ($a['friendlyiface'] == "wan") {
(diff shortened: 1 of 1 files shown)


jim-p:
Copying my question here from the Redmine ticket:
> Why would that matter now? What loss of functionality is there? You can select it specifically, or setup a gateway group with it on Tier 1.
>
> Not selecting WAN by default was breaking a number of people in unexpected ways based on whether or not another interface with a gateway came before WAN alphabetically.

I'm not seeing what this PR would solve. It's entirely possible I'm overlooking it though. I just need some clarification on what the problem is an the intended effect of this will be.

PiBa-NL:
Thanks jim-p, reply added in redmine.

sbeaver-netgate:
Apparently no longer required. Please resubmit if still required.

sbeaver-netgate:
No longer required

PR #3990, opened 2018-09-25
Hide "Drag to Change Order" text when roworderdragging is disabled

luckman212:
If `$config['system']['webgui']['roworderdragging']` is true, then dragging to reorder firewall rules is disabled. This simple change hides the header text saying that you can drag, if in fact, you cannot.

(This is a redo of https://github.com/pfsense/pfsense/pull/3983 in a single commit)

- [ ] Redmine Issue: not needed?
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -328,6 +328,12 @@ if (isset($config['interfaces'][$if]['blockbogons'])) {
 	$showblockbogons = true;
 }
 
+if (isset($config['system']['webgui']['roworderdragging'])) {
+	$rules_header_text = 'Rules';
+} else {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/firewall_rules.php
```diff
@@ -328,6 +328,12 @@ function delete_nat_association($id) {
 	$showblockbogons = true;
 }
 
+if (isset($config['system']['webgui']['roworderdragging'])) {
+	$rules_header_text = 'Rules';
+} else {
+	$rules_header_text = 'Rules (Drag to Change Order)';
+}
```
rbgarga: Please add `gettext()` calls on strings and remove from variable

luckman212: Ok, you got it

PR #3991, opened 2018-09-25
attempt to show last IA_PD received from ISP by reading dhcpd.log

luckman212:
When DHCP6 is enabled on an interface and Prefix Delegation is requested, there is nowhere in the GUI to see:

- whether an IA_PD was received from the ISP
- what the prefix was

This is an attempt to get this info by parsing `dhcpd.log` and display it on the DHCPv6 Server page. I don't know if there is any other way but I am of course open to input!

_related forum thread: https://forum.netgate.com/topic/135723/ipv6-ra-prefix-doesn-t-match-interface-prefix-id_

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8946
- [x] Ready for review

_after applying this patch:_
image

### DIFF
--- a/src/usr/local/www/services_dhcpv6.php
+++ b/src/usr/local/www/services_dhcpv6.php
@@ -668,6 +668,12 @@ $group->add($f2);
 
 $section->add($group);
 
+exec("/usr/local/sbin/clog /var/log/dhcpd.log | /usr/bin/awk '/IA_PD prefix:\ /{print $8}' | tail -n1", $ia_pd);
+$section->addInput(new Form_StaticText(
+	'Last PD Received',
+    $ia_pd ? implode($ia_pd):'No IA_PD found. Try a DHCP release/renew on your WAN interface.'
+    ));
+
 $f1 = new Form_Input(
 	'prefixrange_from',
 	null,


rbgarga:
I'm wondering if there is a way to detect it during the process and save int on a file to be used by other processes. That would be a much better approach

jim-p:
I'd rather see another approach used if we do it at all. This method also won't work unless the user runs the client in debug mode, which isn't the default, so for most people this wouldn't show anything.

We have a script that runs but it [doesn't appear that dhcp6c exposes the prefix information](https://www.freebsd.org/cgi/man.cgi?query=dhcp6c&apropos=0&sektion=0&manpath=FreeBSD+11.2-RELEASE+and+Ports&arch=default&format=html) in the environment so there wouldn't be a way to grab it that I see. If it did that, however, lots of other problems would be solved, too (like using track prefixes for other purposes like VPNs and NPt). That would require changes in dhcp6c though.

luckman212:
Yes I realized this morning that this only works if dhcp6c is started in debug mode. I'm looking into directly modifying the dhcp6c code to stuff this into a file.

Would `/tmp/<ifname>_iapd` be a good place for this? Or `/var/etc/...` ?

rbgarga:
If we change dhcp6c to make it to expose this in the environment, the script used could write down the file. It would make dhcp6c change more generic and maybe accepted upstream.

sbeaver-netgate:
net/dhcp6/files/dhcp6c.in perhaps?

rbgarga:
> Started down this road but got stuck immediately. I am looking for the source for `dhcp6c.c` but can't find it anywhere in the ports tree.
> 
> I expected to find it at:
> https://github.com/pfsense/FreeBSD-ports/tree/devel/net/dhcp6/files
> 
> But I only see a few helper scripts (`*.in` files) there?
> 
> The last commit ([pfsense/FreeBSD-ports@4541846](https://github.com/pfsense/FreeBSD-ports/commit/454184635b694e9ee316229dd307991174848e7d)) says the port wasn't being actively maintained and was changed to track [hrs-allbsd/wide-dhcpv6](https://github.com/hrs-allbsd/wide-dhcpv6/) ... Is that still true though?
> 
> ...because https://www.freshports.org/net/dhcp6/ shows recent commits...(as of 9/2/2018)
> 
> ...and, the `PORTREVISION` of [FreeBSD-ports/net/dhcp6/Makefile](https://github.com/pfsense/FreeBSD-ports/blob/devel/net/dhcp6/Makefile) = `20080615.2`
> 
> ...but upstream is now a few commits ahead at `20080615.2_2`
> 
> Meanwhile, the last commit to [hrs-allbsd/wide-dhcpv6](https://github.com/hrs-allbsd/wide-dhcpv6/commits/freebsd) was on3/1/2017 and there are open PR's sitting almost 1 year old...
> 
> If nothing else changes, I assume the work would need to begin with:
> https://github.com/hrs-allbsd/wide-dhcpv6/blob/freebsd/dhcp6c.c
> 
> **But given the above confusion, not sure where to go from here.**

This is the right place - https://github.com/hrs-allbsd/wide-dhcpv6

@hrs-allbsd is a FreeBSD committer who maintains dhcp6 code for FreeBSD ports, and we are downstream in this case

jim-p:
I suggest we close this and wait for a new set of PRs based on the alternate methods we described above.

luckman212:
No problem, I agree.

PR #3957, opened 2018-06-27
Enable setting AllowAgentForwarding value

ssbarnea:
Fixes #8590

Signed-off-by: Sorin Sbarnea <ssbarnea@redhat.com>

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8590
- [ ] Ready for review

PR #3993, opened 2018-10-04
Enable setting AllowAgentForwarding value

ssbarnea:
Fixes #8590

- [x] Redmine Issue: https://redmine.pfsense.org/issues/NNNN
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/system_advanced_admin.php
+++ b/src/usr/local/www/system_advanced_admin.php
@@ -261,7 +261,7 @@ if ($_POST) {
 
 		$sshd_agentforwarding = isset($config['system']['ssh']['sshdagentforwarding']);
 		if ($_POST['sshdagentforwarding']) {
-			$config['system']['ssh']['sshdagentforwarding'] = true;
+			$config['system']['ssh']['sshdagentforwarding'] = 'enabled';
 		} else {
 			unset($config['system']['ssh']['sshdagentforwarding']);
 		}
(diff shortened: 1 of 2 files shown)


ssbarnea:
As requested to be raised against master branch on https://github.com/pfsense/pfsense/pull/3957

src/usr/local/www/system_advanced_admin.php
```diff
@@ -252,15 +259,24 @@
...
+		}
+
 		$sshd_port = $config['system']['ssh']['port'];
 		if ($_POST['sshport']) {
 			$config['system']['ssh']['port'] = $_POST['sshport'];
 		} else if (isset($config['system']['ssh']['port'])) {
 			unset($config['system']['ssh']['port']);
 		}
 
+
```
jim-p: There is an extraneous blank line here

src/usr/local/www/system_advanced_admin.php
```diff
@@ -252,15 +259,24 @@
 			unset($config['system']['ssh']['sshdkeyonly']);
 		}
 
+		$sshd_agentforwarding = isset($config['system']['ssh']['sshdagentforwarding']);
+		if ($_POST['sshdagentforwarding']) {
+			$config['system']['ssh']['sshdagentforwarding'] = true;
```
jim-p: For consistency this should be ` = "enabled";`

src/usr/local/www/system_advanced_admin.php
ssbarnea: Not really because this separates the new block from next one, check the suroundings and you will see that the newline is desirable (github diff can confuse here).

src/usr/local/www/system_advanced_admin.php
ssbarnea: Right!

src/usr/local/www/system_advanced_admin.php
jim-p: There was already one blank line, this adds another one

PR #3994, opened 2018-10-04
8122: enable use of auth script

ssbarnea:

- [x] Redmine Issue: https://redmine.pfsense.org/issues/8122
- [x] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1111,6 +1111,11 @@ function openvpn_reconfigure($mode, $settings) {
 			if (!($settings['auth_user'] && $settings['auth_pass'])) {
 				$userpass .= "\n";
 			}
+
+			// if auth_script is not empty, we use its output instead of user/pass
+			if ($settings['auth_script']) {
+				$userpass = shell_exec($settings['auth_script']);
+			}
 			file_put_contents($up_file, $userpass);
 		}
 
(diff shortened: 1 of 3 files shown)


ssbarnea:
@jim-p Yep, but these comments didn't make much sense because nobody can check if the body of a shell command exists, it does not make sense. 

Let me give you an example of body that chan be written there:
```
echo -n "foo"
seq -s '' 3
```
This is a perfect valid shell script, that would run and produce `foo123` output. What can you test about it? That echo and seq can be called? 

The user is free to put whatever he wants there and there is no way to assure that the commands are valid or if they exist. The same concern applies to cron entries as they could also contain any kind of commands inside and I am sure the pfsense doendn't do any kind of extra checking on them. If it would no it would clearly be wrong.

jim-p:
The biggest problems are:

1. It can't effectively work for OTP because it only updates the password when the config is synced, which does not happen on every reconnect of the VPN. The moment the connection gets bumped, it falls apart.
2. The fact that it gives the user a way to run a shell exec to an arbitrary script is a security concern.
3. The other concerns about the location and content of that script.

If you want to keep using it as a custom patch, feel free, but I don't see it being integrated.

OpenVPN itself needs this function internally so it can be called on each connect, it can't be solved properly with only config changes.

ssbarnea:
I asked on openvpn ticket and I will check with them, maybe a socket approach would work but I need further testing. Thanks.

PR #3998, opened 2018-10-07
Improved domain validation

NanoCaiordo:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9023
- [x] Ready for review


NanoCaiordo:
Hello Jim, none of both are in fact FQDN and so should fail but from what I just discovered this function does not check, despite it's name, for FQDN intended as host names, but if it also needs to validate primary domain and domain DNS name the entire approach would be different, but fqdn doesn't include primary domain.

I've just tested is_domain() from utils.inc and it seems to do what you need, but "_._" and "_test._com" returns true, not sure if intended or not.

I would need some additional clarifications about the each function scope to help further.

NanoCaiordo:
I've uploaded a text file comparing if_fqdn, is_domain and my last changes (not yet pulled), please let me know if something isn't right.

It doesn't check maximum string (should be 253) or group (63) length, it also doesn't check how many groups there are.

[is_fqdn.txt](https://github.com/pfsense/pfsense/files/2464624/is_fqdn.txt)

NanoCaiordo:
Requested changes were applied, cheers.

jim-p:
Looks OK now but it needs testing, and it should wait for 2.4.5, not be included in 2.4.4-pX.

PR #3999, opened 2018-10-09
Added tlsauth keydir options to openvpn client and server

vpiserchia:
This is an updated version of #3702

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -504,7 +504,7 @@ function openvpn_get_keydirlist() {
 		''  => gettext('Use default direction'),
 		'0' => gettext('Direction 0'),
 		'1' => gettext('Direction 1'),
-		'2' => gettext('No keydir direction'),
+		'2' => gettext('Both irections'),
 	);
 	return $keydirs;
 }
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -499,6 +499,16 @@ function openvpn_remap_digest($digest) {
 	return $digest;
 }
 
+function openvpn_get_keydirlist() {
+	$keydirs = array(
+		''  => gettext('Use default direction'),
+		'0' => gettext('Direction 0'),
+		'1' => gettext('Direction 1'),
+		'2' => gettext('No keydir direction'),
+	);
+	return $keydirs;
+}
+
 function openvpn_get_engines() {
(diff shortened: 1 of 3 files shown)


src/etc/inc/openvpn.inc
```diff
@@ -499,6 +499,16 @@ function openvpn_remap_digest($digest) {
 	return $digest;
 }
 
+function openvpn_get_keydirlist() {
+	$keydirs = array(
+		''  => gettext('Use default direction'),
+		'0' => gettext('Direction 0'),
+		'1' => gettext('Direction 1'),
+		'2' => gettext('Both irections'),
```
jim-p: Missing 'd'

jim-p:

For more information, see https://www.netgate.com/docs/pfsense/development/submitting-a-pull-request-via-github.html

Thanks!

src/etc/inc/openvpn.inc
vpiserchia: Agreed

src/etc/inc/openvpn.inc
```diff
@@ -1188,9 +1198,33 @@ function openvpn_reconfigure($mode, $settings) {
...
+						switch($settings['tlsauth_keydir']){
+							case '1':
+								$tlsopt = 1;
+								break;
+							case '2':
+								$tlsopt = '';
+								break;
+							case '':
+							case '0':
+							default:
```
rbgarga: Since you are using `default:` here you don't need to explicit add `case '':` and `case '0':`

src/etc/inc/openvpn.inc
```diff
@@ -1188,9 +1198,33 @@ function openvpn_reconfigure($mode, $settings) {
...
+						switch($settings['tlsauth_keydir']){
+							case '0':
+								$tlsopt = 0;
+								break;
+							case '2':
+								$tlsopt = '';
+								break;
+							case '':
+							case '1':
+							default:
```
rbgarga: Since you are using default: here you don't need to explicit add case '': and case '1':

vpiserchia: in !f93ec3853fc0c01760606994422e9e8fc0d645c9

src/etc/inc/openvpn.inc
vpiserchia: in https://github.com/pfsense/pfsense/pull/3999/commits/f93ec3853fc0c01760606994422e9e8fc0d645c9

PR #4001, opened 2018-10-25
Well known ports: order by number instead of name

Slamdunk:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9065
- [x] Ready for review

To bring the change small, I've only reordered the array.

The downside is that in the dropdown menus it would be a bit confusing to have the name before the port, because the list would appear unordered.

If you prefer I can also rearrange this code:

https://github.com/pfsense/pfsense/blob/4de15854384e28004b0dc571dc8a40fda7eae694/src/usr/local/www/firewall_rules_edit.php#L1382-L1385

to show the port number first.


jim-p:
The problem with this change is that people don't who do **not** know the numbers will want to find them by name, and with this change they are not in a logical order for those folks.

If you know the number, just choose 'other' and type in the number directly, the drop down isn't meant to cover that case.

PR #4003, opened 2018-10-30
optimize parsing of the openvpn server status socket

xenofree:
If your common name or username contains 'END' like 'BACKEND', it is matched by 

if (strstr($line, "END") || strstr($line, "ERROR"))
break;

and it never appear in the status page of the openvpn.

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1691,7 +1691,7 @@ function openvpn_get_server_status($server, $socket) {
 			}
 
 			/* parse end of output line */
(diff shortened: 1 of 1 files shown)


NanoCaiordo:
Bad luck there isn't a redmine issue about this, but why would you use common names or usernames entirely in upper case in first case.

It will also match TERROR while you could still create an account called ERROR.

Etiquette would say to use standard CNs as: My Name, email, domain.name or, why not, bACkEnD.

sbeaver-netgate:
If this issue really needs to be fixed, it needs to be fixed in a comprehensive manner, such as by validating the Name or CN to ensure it does not contain reserved strings.

PR #4004, opened 2018-11-01
Removed js warnings

NanoCaiordo:
- [x] Ready for review

I did hope for better results on mobile devices when upgrading to passive events but I didn't notice any, maybe real slower mobiles will. However, on lists, newSeperator warnings and, on traffic graphs, touchmove passive violations are now gone. 

I'm adding here human readable changes:
```
function sPE() {
	var ret=false;
	try{
		var n=()=>{}
		,opts=Object.defineProperty({},'passive',{
			get:function get(){
				ret={passive:true}
			}
		});
		this.addEventListener("testPESupport",n,opts);
		this.removeEventListener("testPESupport",n,opts);
	}
	finally {
		return ret
	}
};
var sPE=sPE();

function X(n,t,e) {
	function r(){
		var t=this[o];
		t&&(this.removeEventListener(n,t,t.$),delete this[o])
	}
	function i() {
		var i=l(t,co(arguments));
		r.call(this),this.addEventListener(n,this[o]=i,i.$="touchmove"==n?sPE:e),i._=t
	}
```

sPE (support Passive Events) function is new.
The no-op listener idea, instead of the problematic null come from detect-it.
In the existing X function the only change is `"touchmove"==n?sPE:e`

Cheers


NanoCaiordo:
Scrap all the above since the issue goes up to jquery and any tweak I could make only created leaks everywhere, d3 file restored but kept only the newSperatator change.

PR #4013, opened 2018-11-09
The Shabang is crashing the check.

Hugues-Antoine:
I don't think there's any need, the script is executed only by /usr/local/sbin/ovpn_auth_verify


sbeaver-netgate:
Please clarify "Crashing the check"

joshuasign:
can't reproduce.
should be closed

jim-p:
No response from OP and nobody else is having the problem, so it is probably not needed.

PR #4017, opened 2018-11-22
Allow no username for freedns-v6

stephenw10:


stephenw10:
https://redmine.pfsense.org/issues/8924

PR #4018, opened 2018-11-26
Sort diag_states_summary.php by states

teunis90:
- [-] Redmine Issue: No redmine issue
- [x] Ready for review

In my human opinion the diag_states_summary.php page is more helpful when the entries are sorted by states ($seen). In this case you see the top talkers on top of the page.

Please merge if helpful.

### DIFF
--- a/src/usr/local/www/diag_states_summary.php
+++ b/src/usr/local/www/diag_states_summary.php
@@ -27,8 +27,6 @@
 ##|*MATCH=diag_states_summary.php*
 ##|-PRIV
 
-ini_set('memory_limit','1024M');
-
 exec("/sbin/pfctl -s state", $states);
 
 $srcipinfo = array();
@@ -109,10 +107,6 @@ if (count($states) > 0) {
 	}
 }
 
-function sort_by_ip($a, $b) {
-	return ip2ulong($a) < ip2ulong($b) ? -1 : 1;
-}
-
 function sort_by_seen($a, $b) {
 	return $a['seen'] > $b['seen'] ? -1 : 1;
 }
@@ -140,7 +134,6 @@ function print_summary_table($label, $iparr, $sort = TRUE) {
     }
 	if ($sort) {
 		usort($iparr, "sort_by_seen");
-		//uksort($iparr, "sort_by_ip");
 	}
 
 ?>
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_states_summary.php
```diff
@@ -129,8 +129,11 @@ function build_port_info($portarr, $proto) {
 }
 
 function print_summary_table($label, $iparr, $sort = TRUE) {
+    foreach ($iparr as $ip => $ipinfo) {
+        $iparr[$ip]['ip'] = $ip;
```
jim-p: Why do this? The IP address is already there as a key. This will slow the page down and consume a lot more memory with large state tables.

src/usr/local/www/diag_states_summary.php
```diff
@@ -157,7 +160,7 @@ function print_summary_table($label, $iparr, $sort = TRUE) {
 						</tr>
 					</thead>
 					<tbody>
-<?php foreach ($iparr as $ip => $ipinfo):
+<?php foreach ($iparr as $index => $ipinfo):
```
jim-p: See above, this appears to be an unnecessary change

src/usr/local/www/diag_states_summary.php
```diff
@@ -167,9 +170,10 @@ function print_summary_table($label, $iparr, $sort = TRUE) {
 	}
 ?>
 						<tr>
-							<td<?= $rowSpan ?>><?=$ip;?></td>
+							<td<?= $rowSpan ?>><?=$ipinfo['ip'];?></td>
```
jim-p: See above, this appears to be an unnecessary change

src/usr/local/www/diag_states_summary.php
```diff
@@ -129,8 +129,11 @@ function build_port_info($portarr, $proto) {
 }
 
 function print_summary_table($label, $iparr, $sort = TRUE) {
+    foreach ($iparr as $ip => $ipinfo) {
+        $iparr[$ip]['ip'] = $ip;
+    }
 	if ($sort) {
-		uksort($iparr, "sort_by_ip");
+		usort($iparr, "sort_by_seen");
```
jim-p: This should be @uasort()@, not @usort()@.

teunis90:
@jim-p I filed an issue as requested: Redmine: 9152
I was not aware of `uasort()`, that's definitely an improvement of this little code change.

Please merge.

sbeaver-netgate:
While useful, I don't think we should impose a particular order without user input. Could you provide a checkbox or select control to allow the user to choose which order the states are listed in?

teunis90:
I agree, but sorry, that is not something where I would have time for in the near future.

Since the current order doesn't make any sense at all, I do think a static order is 'better' than the current implementation.

rbgarga:
I've opened a redmine ticket to make columns sortable

https://redmine.pfsense.org/issues/9718

PR #4019, opened 2018-12-06
Fix DigitalOcean DynDNS client

GrantSheehan:
Fixes the check on the return value since it's been updated to use
HTTP/2 syntax. Also adds logic to allow using `@` to denote updating the
root domain A record as well.

Updating my settings to log the header object, this is what ends up in logs:
```
/services_dyndns_edit.php: phpDynDNS (@): PAYLOAD: HTTP/2 200 ^M date: Thu, 06 Dec 2018 05:27:10 GMT^M content-type: application/json; charset=utf-8^M content-length: 154^M set-cookie: __cfduid=<snip>; expires=Fri, 06-Dec-19 05:27:10 GMT; path=/; domain=.digitalocean.com; HttpOnly^M cache-control: max-age=0, private, must-revalidate^M etag: "edb266b12ef88418d31c470fd39d5359"^M ratelimit-limit: 5000^M ratelimit-remaining: 4982^M ratelimit-reset: 1544072841^M x-content-type-options: nosniff^M x-frame-options: SAMEORIGIN^M x-gateway: Edge-Gateway^M x-request-id: db9b2919-ce43-4bb6-b7f3-61a19c1abf03^M x-response-from: service^M x-runtime: 0.207846^M x-xss-protection: 1; mode=block^M expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"^M server: cloudflare^M cf-ray: 484c57811ae35011-DEN^M ^M
```

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9171
- [x] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -998,7 +998,9 @@
 						$output->domain_records = array();
 					}
 					foreach($output->domain_records as $dnsRecord) {
-						if ($this->_dnsHost == $dnsRecord->name) {
+						// NS records are named @ in DO's API, so check type as well 
+						// https://redmine.pfsense.org/issues/9171
+						if ($this->_dnsHost == $dnsRecord->name && $dnsRecord->type == 'A') {
 							$recordID = $dnsRecord->id;
 							break;
 						}
@@ -1845,7 +1847,9 @@
 					}
 					break;
 				case 'digitalocean':
-					if (preg_match("/\s200\sOK/i", $header)) {
+					// Creating new records returns an HTTP 201, updating existing records get 200
+					// https://redmine.pfsense.org/issues/9171
+					if (preg_match("/HTTP\/2\s20[0,1]/i", $header)) {
 						$status = $status_intro . $success_str . gettext("IP Address Updated Successfully!");
 						$successful_update = true;
 					} else {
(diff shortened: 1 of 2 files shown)


src/etc/inc/dyndns.class
```diff
@@ -998,7 +998,7 @@
 						$output->domain_records = array();
 					}
 					foreach($output->domain_records as $dnsRecord) {
-						if ($this->_dnsHost == $dnsRecord->name) {
+						if ($this->_dnsHost == $dnsRecord->name && $dnsRecord->type == 'A') {
```
GrantSheehan: When trying to update the `@` record, without checking type here, as written would try to update the NS records instead.

src/etc/inc/dyndns.class
```diff
@@ -1845,7 +1845,7 @@
 					}
 					break;
 				case 'digitalocean':
-					if (preg_match("/\s200\sOK/i", $header)) {
+					if (preg_match("/HTTP\/2\s20[0,1]/i", $header)) {
```
GrantSheehan: Newly created records get a `201`, updating existing records would get a `200`

GrantSheehan:
Sure thing, pushed an update with some comments.

PR #4020, opened 2018-12-07
Regression: openvpn.auth-user.php: calling_station_id was removed

wokis:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9178
- [x] Ready for review


Augustin-FL:
The best would be that the ```calling_station_id``` contains the MAC address linked to the interface used by OpenVPN, and not the IP of WAN...to comply with [rfc3580](https://tools.ietf.org/html/rfc3580#section-3.21). 

But because we don't have an easy way to get the interface used by the tap/tun, let's keep that for now

jim-p:
RFC3580 is for 802.1x, not OpenVPN or RADIUS in general. There is no need to follow it for other purposes.

PR #4021, opened 2018-12-14
Set Hyperscan Ports Module to default to native arch

demospace:
The Hyperscan port configuration does not set the NATIVE option which allows for HS to be built with advanced instruction sets (AVX2/AVX512). Enabling this flag does not effect those without AVX2/AVX512, but does enable it by default within HS for those CPUs that support it.

This change will help with performance in both Suricata and Snort which both implement HS. HS also has internal checks to determine instruction set support after it's built.

- [x] Ready for review

I am quite new to FreeBSD, Ports, and Poudriere and will not be offended if this is rejected or ignored. It would be great if this flag could be changed so I don't have to rebuild HS on every pfSense update.


NanoCaiordo:
No redmine link?

I like this pull, my Scalable will love it.

tools/conf/pfPorts/poudriere_bulk
```diff
@@ -7,6 +7,7 @@ databases/php72-pdo_mysql
 databases/php72-pgsql
 devel/aws-sdk-php
 devel/gdb
+devel/hyperscan
```
rbgarga: This line should be removed. Hyperscan is built as dependency of suricata package

rbgarga:
@bmeeks8 what do you think about it?

bmeeks8:
I don't see a problem for Snort or Suricata with this change.  I'm OK with it.

PR #4025, opened 2018-12-18
webConfigurator/RADIUS: Set NAS Identifier to 'webConfigurator'

kantlivelong:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9209
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1942,7 +1942,7 @@ function session_auth() {
 	}
 
 	/* Validate incoming login request */
-	$attributes = array('nas_identifier' => 'webConfigurator');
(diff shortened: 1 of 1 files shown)


kantlivelong:
Hi @jim-p,

Perhaps I'm missing something but isn't the hostname already provided via the "Client Friendly Name" attribute?

I based this change off of how OpenVPN set the attrib.

jim-p:
Maybe so but we have to keep POLA in mind. People are already using and relying on the current behavior. Changing it too much is worse than leaving it as is. As with the recent captive portal changes, we ended up changing them to include the zone name as a means to keep them more unique, because people were relying on the zone name being included as it was in the past.

kantlivelong:
That's fair. I'll push the change in a few.

PR #4026, opened 2018-12-18
Openvpn radius acl enhancements

kantlivelong:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9206
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.attributes.php
+++ b/src/etc/inc/openvpn.attributes.php
@@ -19,12 +19,7 @@
  * limitations under the License.
  */
 
-global $username;
-
-$devname = getenv("dev");
-if (empty($devname)) {
-	$devname = "openvpn";
-}
+global $username, $dev, $untrusted_port;
 
 function cisco_to_cidr($addr) {
 	if (!is_ipaddr($addr)) {
@@ -51,7 +46,7 @@ function cisco_extract_index($prule) {
 }
 
 function parse_cisco_acl($attribs) {
-	global $devname, $attributes;
+	global $dev, $attributes;
 	if (!is_array($attribs)) {
 		return "";
 	}
@@ -87,11 +82,11 @@ function parse_cisco_acl($attribs) {
 			$index = 0;
 			$isblock = false;
 			if ($rule[$index] == "permit") {
-				$tmprule = "pass {$dir} quick on {$devname} ";
+				$tmprule = "pass {$dir} quick on {$dev} ";
 			} else if ($rule[$index] == "deny") {
 				//continue;
 				$isblock = true;
-				$tmprule = "block {$dir} quick on {$devname} ";
+				$tmprule = "block {$dir} quick on {$dev} ";
 			} else {
 				continue;
 			}
@@ -181,11 +176,8 @@ function parse_cisco_acl($attribs) {
 
 $rules = parse_cisco_acl($attributes);
 if (!empty($rules)) {
-	$pid = posix_getpid();
-	$filename = "{$g['tmp_path']}/ovpn_{$pid}{$username}.rules";
+	$filename = "{$g['tmp_path']}/ovpn_{$dev}_{$username}_{$untrusted_port}.rules";
 	@file_put_contents($filename, $rules);
-	mwexec("/sbin/pfctl -a " . escapeshellarg("openvpn/{$username}") . " -f " . escapeshellarg($filename));
-	@unlink($filename);
 }
 
 ?>
--- a/src/etc/inc/openvpn.attributes.php
+++ b/src/etc/inc/openvpn.attributes.php
@@ -21,7 +21,15 @@
 
 global $username, $dev, $untrusted_port;
 
+if (empty($dev)) {
+    $dev = "openvpn";
+}
+
 function cisco_to_cidr($addr) {
+    if (!is_ipaddr($addr)) {
+        throw new Exception('Invalid IP Addr');
+    }
+
     $mask = decbin(~ip2long($addr));
     $mask = substr($mask, -32);
     $k = 0;
--- a/src/etc/inc/openvpn.attributes.php
+++ b/src/etc/inc/openvpn.attributes.php
@@ -22,162 +22,256 @@
 global $username, $dev, $untrusted_port;
 
 function cisco_to_cidr($addr) {
-	if (!is_ipaddr($addr)) {
-		return 0;
-	}
-	$mask = decbin(~ip2long($addr));
-	$mask = substr($mask, -32);
-	$k = 0;
-	for ($i = 0; $i <= 32; $i++) {
-		$k += intval($mask[$i]);
-	}
-	return $k;
+    $mask = decbin(~ip2long($addr));
+    $mask = substr($mask, -32);
+    $k = 0;
+    for ($i = 0; $i <= 32; $i++) {
+        $k += intval($mask[$i]);
+    }
+    return $k;
 }
 
 function cisco_extract_index($prule) {
+    $index = explode("#", $prule);
+    if (is_numeric($index[1])) {
+        return intval($index[1]);
+    } else {
+        syslog(LOG_WARNING, "Error parsing rule {$prule}: Could not extract index");
+    }
+    return -1;;
+}
 
-	$index = explode("#", $prule);
-	if (is_numeric($index[1])) {
-		return intval($index[1]);
-	} else {
-		syslog(LOG_WARNING, "Error parsing rule {$prule}: Could not extract index");
-	}
-	return -1;;
+function parse_cisco_acl_rule($rule, $devname, $dir) {
+    $rule_orig = $rule;
+    $rule = explode(" ", $rule);
+    $tmprule = "";
+    $index = 0;
+
+    if ($rule[$index] == "permit") {
+        $tmprule = "pass {$dir} quick on {$devname} ";
+    } else if ($rule[$index] == "deny") {
+        $tmprule = "block {$dir} quick on {$devname} ";
+    } else {
+        return;
+    }
+
+    $index++;
+
+    switch ($rule[$index]) {
+        case "ip":
+            $tmprule .= "inet ";
+            break;
+        case "icmp":
+        case "tcp":
+        case "udp":
+            $tmprule .= "proto {$rule[$index]} ";
+            break;
+        default:
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid protocol.");
+            return;
+    }
+    $index++;
+
+    /* Source */
+    if (trim($rule[$index]) == "host") {
+        $index++;
+        $tmprule .= "from {$rule[$index]} ";
+        $index++;
+    } else if (trim($rule[$index]) == "any") {
+        $tmprule .= "from any ";
+        $index++;
+    } else {
+        $network = $rule[$index];
+        $netmask = $rule[++$index];
+
+
+        if(!is_ipaddr($network)) {
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid network.");
+            return;
+        }
+
+        try {
+            $netmask = cisco_to_cidr($netmask);
+        } catch(Exception $e) {
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid netmask.");
+            return;
+        }
+        $tmprule .= "from {network}/{$netmask}";
+
+        $index++;
+    }
+
+    /* Source Operator */
+    if (in_array(trim($rule[$index]), array("lt", "gt", "dq", "neq"))) {
+        switch(trim($rule[$index])) {
+            case "lt":
+                $operator = "<";
+                break;
+            case "gt":
+                $operator = ">";
+                break;
+            case "dq":
+                $operator = "=";
+                break;
+            case "neq":
+                $operator = "!=";
+                break;
+        }
+
+        $port = $rule[++$index];
+        if (is_numeric($port)) {
+            $tmprule .= "port {$operator} {$port} ";
+        } else {
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
+            return;
+        }
+        $index++;
+    } else if (trim($rule[$index]) == "range") {
+        $port = array($rule[++$index], $rule[++$index]);
+        if (is_numeric($port[0]) && is_numeric($port[1])) {
+            $port[0]--;
+            $port[1]++;
+            $tmprule .= "port {$port[0]} >< {$port[1]} ";
+        } else {
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
+            return;
+        }
+        $index++;
+    }
+
+    /* Destination */
+    if (trim($rule[$index]) == "host") {
+        $index++;
+        $tmprule .= "to {$rule[$index]} ";
+        $index++;
+    } else if (trim($rule[$index]) == "any") {
+        $tmprule .= "to any ";
+        $index++;
+    } else {
+        $network = $rule[$index];
+        $netmask = $rule[++$index];
+
+
+        if(!is_ipaddr($network)) {
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid network.");
+            return;
+        }
+
+        try {
+            $netmask = cisco_to_cidr($netmask);
+        } catch(Exception $e) {
+            syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid netmask.");
+            return;
+        }
+        $tmprule .= "to {network}/{$netmask}";
+
+        $index++;
+    }
+
+    /* Destination Operator */
+    if (in_array(trim($rule[$index]), array("lt", "gt", "dq", "neq"))) {
+        switch(trim($rule[$index])) {
+            case "lt":
+                $operator = "<";
+                break;
+            case "gt":
+                $operator = ">";
+                break;
+            case "dq":
+                $operator = "=";
+                break;
+            case "neq":
+                $operator = "!=";
+                break;
+        }
+
+        $port = $rule[++$index];
+        if (is_numeric($port)) {
+            $tmprule .= "port {$operator} {$port} ";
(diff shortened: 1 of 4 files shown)


src/etc/inc/openvpn.attributes.php
```diff
@@ -19,173 +19,259 @@
  * limitations under the License.
  */
 
-global $username;
-
-$devname = getenv("dev");
-if (empty($devname)) {
```
jim-p: Why remove this check? It should still default to the OpenVPN group rules if it can't locate a device.

src/etc/inc/openvpn.attributes.php
```diff
@@ -19,173 +19,259 @@
...
-	$devname = "openvpn";
-}
+global $username, $dev, $untrusted_port;
 
 function cisco_to_cidr($addr) {
-	if (!is_ipaddr($addr)) {
```
jim-p: Why remove this safety check?

src/usr/local/sbin/openvpn.attributes.sh
```diff
@@ -18,16 +18,41 @@
...
+
+    rm "${lockfile}"
 elif [ "$script_type" = "client-disconnect" ]; then
-	command="/sbin/pfctl -a 'openvpn/$common_name' -F rules"
-	eval $command
-	/sbin/pfctl -k $ifconfig_pool_remote_ip
-	/sbin/pfctl -K $ifconfig_pool_remote_ip
+    while [ -f "${lockfile}" ];
+    do
+        sleep 1
```
jim-p: There should be some kind of alternate escape plan for this loop in case another process gets stuck.

src/usr/local/sbin/openvpn.attributes.sh
```diff
@@ -18,16 +18,41 @@
...
+anchorname="openvpn/${dev}_${username}_${trusted_port}"
+
 if [ "$script_type" = "client-connect" ]; then
-	if [ -f /tmp/$common_name ]; then
-		/bin/cat /tmp/$common_name > $1
-		/bin/rm /tmp/$common_name
-	fi
+    while [ -f "${lockfile}" ];
+    do
+        sleep 1
```
jim-p: There should be some kind of alternate escape plan for this loop in case another process gets stuck.

src/etc/inc/openvpn.attributes.php
kantlivelong: Ah yes good point. I don't use the group interface and overlooked this. Will correct.

src/usr/local/sbin/openvpn.attributes.sh
kantlivelong: Do you have any kind of suggestion other than just adding a timeout? I wasn't too crazy about the way this was implemented but couldn't find any other viable/better solution.

jim-p: Not really. Pick a time (like 10s) and then count that many sleeps and if it's still locked, exit/die

kantlivelong: Alright. Hoping to have this cleared up by Friday.

Thanks :+1:

kantlivelong:
@jim-p 

I've made the requested changes. Let me know if there's anything else I can do.

kantlivelong:
Thanks! :+1:

src/usr/local/sbin/ovpn_auth_verify_async
```diff
@@ -53,14 +53,14 @@ nas_port=$5
 # Single quoting $password breaks getting the value from the variable.
 # Base64 and urlEncode usernames and passwords
 password=$(printf "%s" "${password}" | "${openssl}" enc -base64 | \
-    "${sed}" -e 's_=_%3D_g;s_+_%2B_g;s_/_%2F_g')
+	"${sed}" -e 's_=_%3D_g;s_+_%2B_g;s_/_%2F_g')
 username=$(printf "%s" "${username}" | "${openssl}" enc -base64 | \
-    "${sed}" -e 's_=_%3D_g;s_+_%2B_g;s_/_%2F_g')
+	"${sed}" -e 's_=_%3D_g;s_+_%2B_g;s_/_%2F_g')
```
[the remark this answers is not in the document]
kantlivelong: Done

src/usr/local/sbin/ovpn_auth_verify_async
```diff
@@ -53,14 +53,14 @@ nas_port=$5
 # Single quoting $password breaks getting the value from the variable.
 # Base64 and urlEncode usernames and passwords
 password=$(printf "%s" "${password}" | "${openssl}" enc -base64 | \
-    "${sed}" -e 's_=_%3D_g;s_+_%2B_g;s_/_%2F_g')
+	"${sed}" -e 's_=_%3D_g;s_+_%2B_g;s_/_%2F_g')
```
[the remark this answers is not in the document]
kantlivelong: Done

kantlivelong:
@rbgarga I pushed 083390ba61ca7243f284515a5984f4b640ccc661 which adjusts the indentation as requested. :smile:

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid netmask.");
+			return;
+		}
+		$tmprule .= "from {network}/{$netmask}";
+
+		$index++;
+	}
+
+	/* Source Operator */
+	if (in_array(trim($rule[$index]), array("lt", "gt", "dq", "neq"))) {
```
hatclub:
```suggestion
	if (in_array(trim($rule[$index]), array("lt", "gt", "eq", "neq"))) {
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+		switch(trim($rule[$index])) {
+			case "lt":
+				$operator = "<";
+				break;
+			case "gt":
+				$operator = ">";
+				break;
+			case "dq":
```
hatclub:
```suggestion
			case "eq":
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid netmask.");
+			return;
+		}
+		$tmprule .= "to {network}/{$netmask}";
+
+		$index++;
+	}
+
+	/* Destination Operator */
+	if (in_array(trim($rule[$index]), array("lt", "gt", "dq", "neq"))) {
```
hatclub:
```suggestion
	if (in_array(trim($rule[$index]), array("lt", "gt", "eq", "neq"))) {
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+		switch(trim($rule[$index])) {
+			case "lt":
+				$operator = "<";
+				break;
+			case "gt":
+				$operator = ">";
+				break;
+			case "dq":
```
hatclub:
```suggestion
			case "eq":
```

kantlivelong:
@hatclub 

Well that's funny! I must have made a typo and copied it forward. I've have been using "dq" in my rules in prod. :rofl: 

Fixed in 0ac270fbc9fe53247c18301c2348f20d7087948e

hatclub:
Okay, with those changes this now all looks like what we have rolled out into production by replacing the files on 2.4.4-p3 and FWIW it all works 'as intended' for us under normal use (I haven't tried aggressively cycling connections off and on or mass load testing or anything, so I can't speak to that)

I had literally only just discovered pfsense supported RADIUS supplied ACLs and then seen how minimal the implementation was; was gearing myself up to have to write these patches myself when I uncovered this PR, so, thanks for your effort here @kantlivelong !

matsimon:
Thanks for that patch @kantlivelong, I eventually came up with some issues and found this pull request.

I've tested it and have a small suggestion as far as the log output goes: What about adding "source" or "destination" into the various log outputs? i.e. "Invalid network" -> "Invalid destination network"
This could definitely help in debugging wrongly written rules.

And on the other side I've yet failed to get some more specific rules to work since they (seem to) result into wron pf rules:

  * FreeRADIUS: Cisco-AVPair += ```inacl#10=permit tcp any 10.1xx.2xx.2xx 255.255.255.255 eq 22```
   * Resulting pf rule: ```pass in quick on openvpn proto tcp from any to {network}/0port = 22```

Additionnaly I will check with 2 OpenVPN clients being connected if these rules apply only to the client connected or to any client connected since not all should get these rules applied. The pf rule might suggest that all openvpn clients might get these rules applied?

hatclub:
Hi @matsimon,

Do you actually need a rule that is "from any" instead of "from {clientip}"? Are your OpenVPN tunnels for routed blocks rather than just single IP clients? The literal string {clientip} will be expanded to the IP assigned to the openvpn user. (also, is the format of your avpair data correct? AIUI the format is ip:inacl#NUM=)

So instead you might have: 
ip:inacl#1=permit ip host {clientip} host 10.100.200.200 eq 22
ip:inacl#2=deny ip host {clientip} any

This will allow the openvpn client to ssh to 10.100.200.200 but not access anything else (and also not affect any other openvpn users)

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			case "neq":
+				$operator = "!=";
+				break;
+		}
+
+		$port = $rule[++$index];
+		if (is_numeric($port)) {
+			$tmprule .= "port {$operator} {$port} ";
+		} else {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
```
hatclub:
```suggestion
			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid source port: '$port' not a numeric value between 0 and 65535.");
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+		}
+		$index++;
+	} else if (trim($rule[$index]) == "range") {
+		$port = array($rule[++$index], $rule[++$index]);
+		if (is_numeric($port[0]) && is_numeric($port[1])) {
+			$port[0]--;
+			$port[1]++;
+			$tmprule .= "port {$port[0]} >< {$port[1]} ";
+		} else {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
```
hatclub:
```suggestion
			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid source ports: '$port[0]' & '$port[1]' one or both are not a numeric value between 0 and 65535.");
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+	} else if (trim($rule[$index]) == "any") {
+		$tmprule .= "to any ";
+		$index++;
+	} else {
+		$network = $rule[$index];
+		$netmask = $rule[++$index];
+
+
+		if(!is_ipaddr($network)) {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid network.");
```
hatclub:
```suggestion
			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid destination network '$network'.");
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			return;
+		}
+
+		try {
+			$netmask = cisco_to_cidr($netmask);
+		} catch(Exception $e) {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid netmask.");
```
hatclub:
```suggestion
			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid destination netmask '$netmask'.");
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			case "neq":
+				$operator = "!=";
+				break;
+		}
+
+		$port = $rule[++$index];
+		if (is_numeric($port)) {
+			$tmprule .= "port {$operator} {$port} ";
+		} else {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
```
hatclub:
```suggestion
			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid destination port: '$port' not a numeric value between 0 and 65535.");
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+		}
+		$index++;
+	} else if (trim($rule[$index]) == "range") {
+		$port = array($rule[++$index], $rule[++$index]);
+		if (is_numeric($port[0]) && is_numeric($port[1])) {
+			$port[0]--;
+			$port[1]++;
+			$tmprule .= "port {$port[0]} >< {$port[1]} ";
+		} else {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
```
hatclub:
```suggestion
			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid destination ports: '$port[0]' '$port[1]' one or both are not a numeric value between 0 and 65535.");
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			case "eq":
+				$operator = "=";
+				break;
+			case "neq":
+				$operator = "!=";
+				break;
+		}
+
+		$port = $rule[++$index];
+		if (is_numeric($port)) {
```
hatclub:
```suggestion
		if (is_numeric($port) && (0 <= $port && $port <= 65535)) {
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			$tmprule .= "port {$operator} {$port} ";
+		} else {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
+			return;
+		}
+		$index++;
+	} else if (trim($rule[$index]) == "range") {
+		$port = array($rule[++$index], $rule[++$index]);
+		if (is_numeric($port[0]) && is_numeric($port[1])) {
```
hatclub:
```suggestion
		if (is_numeric($port[0]) && is_numeric($port[1]) && (0 <= $port[0] && $port[0] <= 65535) && (0 <= $port[1] && $port[1] <= 65535)) {
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			case "eq":
+				$operator = "=";
+				break;
+			case "neq":
+				$operator = "!=";
+				break;
+		}
+
+		$port = $rule[++$index];
+		if (is_numeric($port)) {
```
hatclub:
```suggestion
		if (is_numeric($port) && (0 <= $port && $port <= 65535)) {
```

src/etc/inc/openvpn.attributes.php
```diff
@@ -50,8 +49,176 @@ function cisco_extract_index($prule) {
...
+			$tmprule .= "port {$operator} {$port} ";
+		} else {
+			syslog(LOG_WARNING, "Error parsing rule {$rule_orig}: Invalid port.");
+			return;
+		}
+		$index++;
+	} else if (trim($rule[$index]) == "range") {
+		$port = array($rule[++$index], $rule[++$index]);
+		if (is_numeric($port[0]) && is_numeric($port[1])) {
```
hatclub:
```suggestion
		if (is_numeric($port[0]) && is_numeric($port[1]) && (0 <= $port[0] && $port[0] <= 65535) && (0 <= $port[1] && $port[1] <= 65535)) {
```

c0olcast:
Hello,

First off, this an awesome feature you guys are working on. I'd really like to add this feature to a running 2.4.4, the current support for Cisco ACL is basically useless. Any pointers on how to merge the changes to my current install would be greatly appreciated.

rbgarga:
@hatclub there is a function called `is_port()` in util.inc that does all the validations you are doing here. Could you please review them and make use you use it?

hatclub:
Ooft. That's what I get for making changes when I haven't looked at the wider codebase. I'll look at it in a sec.

hatclub:
I've revised my suggested changes to replace the previous use of is_numeric (and my additions for range checking) to use is_port from util.inc instead 

(for those playing along at home: util.inc is required by config.inc which is required by openvpn.auth-user.php which is what includes openvpn.attributes.php where the checks occur)

matsimon:
@hatclub Thanks for your suggestion, I've only been able to come back now.

Yes, my intention would be to permit certain access rules depending on the VPN client:
> So instead you might have:
> ip:inacl#1=permit ip host {clientip} host 10.100.200.200 eq 22
> ip:inacl#2=deny ip host {clientip} any

is {clientip} the syntax to define a specific client? If so, that should be documented. :-)
Thanks for the suggestions and hopefully this PR can be merged into an upcoming release.

hatclub:
{clientip} was added as a macro that is substituted with the connected client's tunnel IP as part of this PR, it doesn't exist without the changes in this PR, so there's not really any documentation for it as such beyond reading the source code! https://redmine.pfsense.org/issues/9206 (linked at the top of this PR) does specifically mention it, though.

matsimon:
@hatclub Thanks for pointing that out!

rbgarga:
Superseded by #4192

PR #4027, opened 2018-12-31
Fix hostname not showing with some UIDs

int3l:
Proper parsing for the end of the individual DHCP lease block.
In some cases the uid contains } symbol, which breaks the old parsing and 
excludes the client-hostname from being displayed in the Web UI.

Before:
No hostname shown

After:
Hostname shown

- [x] Redmine Issue: https://redmine.pfsense.org/issues/3500
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -114,10 +114,10 @@ $awk = "/usr/bin/awk";
 /* this pattern sticks comments into a single array item */
 $cleanpattern = "'{ gsub(\"#.*\", \"\");} { gsub(\";\", \"\"); print;}'";
 /* We then split the leases file by } */
-$splitpattern = "'BEGIN { RS=\"}\";} {for (i=1; i<=NF; i++) printf \"%s \", \$i; printf \"}\\n\";}'";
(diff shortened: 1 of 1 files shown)


src/usr/local/www/status_dhcp_leases.php
```diff
@@ -114,10 +114,10 @@ function remove_duplicate($array, $field) {
 /* this pattern sticks comments into a single array item */
 $cleanpattern = "'{ gsub(\"#.*\", \"\");} { gsub(\";\", \"\"); print;}'";
 /* We then split the leases file by } */
-$splitpattern = "'BEGIN { RS=\"}\";} {for (i=1; i<=NF; i++) printf \"%s \", \$i; printf \"}\\n\";}'";
+$splitpattern = "'BEGIN{ RS=\"#\";} {for (i=1; i<=NF; i++) printf \"%s \", \$i; printf \"}\\n\";}'";
```
int3l: Reuse the same character for comments since it is already cleaned, so the awk parsing doesn't break.

joshuasign: please take a look at https://redmine.pfsense.org/issues/3500#note-4
it may be less 'uid' field character dependent

rbgarga:
I've reimplemented the function that parses dhcpd.leases and did it in PHP this time. Thanks!

PR #4029, opened 2019-01-04
pfSense DNS Resolver - Unbound Python module integration

BBcan177:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9251
- [x] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -2213,10 +2213,10 @@ function services_unbound_configure($restart_dhcp = true) {
 				// Add DNS Resolver python integration
 				if (empty($validate)) {
 					if (!is_dir("{$g['unbound_chroot_path']}{$base_folder}{$dir}")) {
-						exec("/bin/mkdir -p {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1");
+						safe_mkdir("{$g['unbound_chroot_path']}{$base_folder}{$dir}");
 					}
-					exec("/sbin/mount_nullfs -o ro /usr/local{$dir} {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1", $output, $retval);
-
+					exec("/sbin/mount_nullfs -o ro " . escapeshellarg("/usr/local{$dir}") . " " . escapeshellarg("{$g['unbound_chroot_path']}{$base_folder}{$dir}") . " 2>&1", $output, $retval);
+					
 					// Disable Unbound python on mount failure
 					if ($retval != 0) {
 						$config['unbound']['python'] = '';
@@ -2229,10 +2229,10 @@ function services_unbound_configure($restart_dhcp = true) {
 
 			// Remove DNS Resolver python integration
 			elseif (!empty($validate)) {
-				exec("/sbin/umount -t nullfs {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1", $output, $retval);
+				exec("/sbin/umount -t nullfs " . escapeshellarg("{$g['unbound_chroot_path']}{$base_folder}{$dir}") . " 2>&1", $output, $retval);
 				if ($retval == 0) {
 					foreach (array( "/usr/local{$dir}", '/usr/local', '/usr') as $folder) {
-						if (is_dir("{$g['unbound_chroot_path']}{$folder}")) {
+						if (!empty($g['unbound_chroot_path']) && $g['unbound_chroot_path'] != '/' && is_dir("{$g['unbound_chroot_path']}{$folder}")) {
 							rmdir("{$g['unbound_chroot_path']}{$folder}");
 						}
 
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -2204,19 +2204,35 @@ function services_unbound_configure($restart_dhcp = true) {
 			}
 		}
 
+		$python_mode = false;
+		if (isset($config['unbound']['python']) && !empty($config['unbound']['python_script'])) {
+			$python_mode = true;
+		}
+
+		/* Include any additional functions as defined by python script include file */
+		if (file_exists("{$g['unbound_chroot_path']}/{$config['unbound']['python_script']}_include.inc")) {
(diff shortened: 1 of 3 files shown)


src/etc/inc/unbound.inc
```diff
@@ -156,14 +167,26 @@ function unbound_generate_config_text($unboundcfg = NULL, $cfgsubdir = "") {
...
+
+	// Setup Python module (pre validator)
+	if (isset($unboundcfg['python']) && !empty($unboundcfg['python_script']) && $unboundcfg['python_order'] == 'pre_validator') {
+		$module_config .= 'python ';
+	}
+
 	// Setup DNSSEC support
 	if (isset($unboundcfg['dnssec'])) {
-		$module_config = "validator iterator";
+		$module_config .= 'validator ';
```
joshuasign: why don't you double quotes this string ? as it was...

any string shoud be double quoted i think...
you should check all over your code and change it
even if i know, it is not wrong syntaxically

BBcan177: https://stackoverflow.com/questions/3446216/what-is-the-difference-between-single-quoted-and-double-quoted-strings-in-php

src/etc/inc/services.inc
```diff
@@ -2204,6 +2204,50 @@ function services_unbound_configure($restart_dhcp = true) {
 			}
 		}
 
+		/* DNS Resolver python integration */
+		$base_folder = '/usr/local';
+		foreach (array('/bin', '/lib') as $dir) {
+			$validate = exec("/sbin/mount | /usr/bin/grep '{$g['unbound_chroot_path']}{$base_folder}{$dir}' 2>&1");
```
jim-p: Change this:
```$validate = exec("/sbin/mount | /usr/bin/grep '{$g['unbound_chroot_path']}{$base_folder}{$dir}' 2>&1");```
To this:
```$validate = exec("/sbin/mount | /usr/bin/grep " . escapeshellarg("{$g['unbound_chroot_path']}{$base_folder}{$dir}") . " 2>&1");```

File: src/etc/inc/unbound.inc (hunk 1 above)
Comment (joshuasign) [reply src/etc/inc/unbound.inc]: i know all of that, 
it is just a question of cosmetic and to respect the existing code convention.

File: src/etc/inc/services.inc [hunk 3]
```diff
@@ -2204,6 +2204,50 @@ function services_unbound_configure($restart_dhcp = true) {
...
+			if (isset($config['unbound']['python']) && !empty($config['unbound']['python_script'])) {
+
+				// Add DNS Resolver python integration
+				if (empty($validate)) {
+					if (!is_dir("{$g['unbound_chroot_path']}{$base_folder}{$dir}")) {
+						exec("/bin/mkdir -p {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1");
```
jim-p: Why not use `safe_mkdir("{$g['unbound_chroot_path']}{$base_folder}{$dir}");`?

src/etc/inc/services.inc
```diff
@@ -2204,6 +2204,50 @@ function services_unbound_configure($restart_dhcp = true) {
...
+					}
+					exec("/sbin/mount_nullfs -o ro /usr/local{$dir} {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1", $output, $retval);
```
jim-p: Change this:
```exec("/sbin/mount_nullfs -o ro /usr/local{$dir} {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1", $output, $retval);```
To this:
```exec("/sbin/mount_nullfs -o ro " . escapeshellarg("/usr/local{$dir}") . " " . escapeshellarg("{$g['unbound_chroot_path']}{$base_folder}{$dir}") . " 2>&1", $output, $retval);```

File: src/etc/inc/services.inc [hunk 5]
```diff
@@ -2204,6 +2204,50 @@ function services_unbound_configure($restart_dhcp = true) {
...
+						$log_msg = "[Unbound-pymod]: Disabling Unbound python due to failed mount";
+						write_config($log_msg);
+						log_error($log_msg);
+					}
+				}
+			}
+
+			// Remove DNS Resolver python integration
+			elseif (!empty($validate)) {
+				exec("/sbin/umount -t nullfs {$g['unbound_chroot_path']}{$base_folder}{$dir} 2>&1", $output, $retval);
```
jim-p: Use `escapeshellarg()` as described previously.

src/etc/inc/services.inc
```diff
@@ -2204,6 +2204,50 @@ function services_unbound_configure($restart_dhcp = true) {
...
+				if ($retval == 0) {
+					foreach (array( "/usr/local{$dir}", '/usr/local', '/usr') as $folder) {
+						if (is_dir("{$g['unbound_chroot_path']}{$folder}")) {
```
jim-p: Please check that `$g['unbound_chroot_path']` has a sane value before taking a destructive action here. Even though it would fail since the directory wouldn't be empty, it's still better to be safe.

src/etc/inc/unbound.inc
BBcan177: Its a style preference.... pfSense older code is mostly double quotes, and most newer code is single quotes. Lets see what the devs say and can change if they wish.

jim-p: Single quotes are safer, unless you have a need to substitute variables in the string, it's better to err on the side of caution and use single quotes. We don't have any stated policy in the developer guide about which to use, so there is no need to follow any kind of tradition here.

src/etc/inc/services.inc
BBcan177: Is that equivalent to the "-p" arg?

jim-p: It has the same net effect as `-p`.

https://github.com/pfsense/pfsense/blob/master/src/etc/inc/util.inc#L2023

src/etc/inc/services.inc
jim-p: At a minimum, preferably also not `/`.

BBcan177: I'd like to hardcode it myself... but that is following the other code in Unbound which uses the $g variable setting.

BBcan177:
Friendly bump ...

PR #4033, opened 2019-01-09
Losing the default gateway after performing a reboot

Raafat:
Hello there! I've noticed that after performing a reboot the default gateway gets lost. I'm using the new feature to set the default gateway and I'm using an entry under the gateway groups to group two of my WAN connections together. So I've create an gateway group containing two WAN connections, let's call them W1, and W2. The gateway group is called WANs. Then using the recently-added feature "default gateway IPv4" I choose the default gateway to be "WANs"(which is the gateway group). So with that being said every I reboot pfsense the default gateway needs to be chosen selectively before it is being added into the IP routing table.


jim-p:
This does not appear to be a valid pull request, and this site is not for support or diagnostic discussion.

For assistance in solving problems, please post on the [Netgate Forum](https://forum.netgate.com) or the [pfSense Subreddit](https://www.reddit.com/r/pfSense/).

See [Reporting Issues with pfSense Software](https://www.netgate.com/docs/pfsense/development/bug-reporting.html) for more information.

PR #4034, opened 2019-01-11
#9004 set the default gateway when system start and a gateway_group is default IPV4 gateway

kkr0kk:
Correct BUG 9004 -> set the default gateway when system start and a gateway_group is default IPV4 gateway

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9004
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -1109,7 +1109,7 @@ function fixup_default_gateway($ipprotocol, $gateways_status, $gateways_arr) {
 			$currentdefaultgwip = getcurrentdefaultgatewayip($ipprotocol);
 			$found_current = false;
 			foreach($gwg_members as $gwgroupitem) {
-				if ($gwgroupitem['gwip'] == $currentdefaultgwip) {
+				if (!empty($currentdefaultgwip) AND $gwgroupitem['gwip'] == $currentdefaultgwip) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/gwlb.inc
```diff
@@ -1109,7 +1109,7 @@ function fixup_default_gateway($ipprotocol, $gateways_status, $gateways_arr) {
 			$currentdefaultgwip = getcurrentdefaultgatewayip($ipprotocol);
 			$found_current = false;
 			foreach($gwg_members as $gwgroupitem) {
-				if ($gwgroupitem['gwip'] == $currentdefaultgwip) {
+				if (!empty($currentdefaultgwip) AND $gwgroupitem['gwip'] == $currentdefaultgwip) {
```
jim-p: Change `AND` to `&&` and place parenthesis around the `==` comparison, so:
```
if (!empty($currentdefaultgwip) && ($gwgroupitem['gwip'] == $currentdefaultgwip)) {
```

lubo:
@jim-p What's the status? Is this going to be included in version 2.4.5?

jim-p:
Needs some testing before it gets merged, if it's OK, then it is likely to be included in the next release.

lubo:
I've just tested this on 2.4.4-RELEASE-p2. Works as expected.

PR #4037, opened 2019-01-16
CP - add trim() to $_POST['auth_user'] & $_POST['auth_user2']

heper:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9274
- [ ] Ready for review


src/usr/local/captiveportal/index.php
```diff
@@ -189,11 +189,11 @@
 			$passwd = $cpcfg['radmac_secret'];
 			$context = 'radmac'; // Radius MAC authentication
 		} elseif (!empty($_POST['auth_user2'])) {
```
joshuasign: this shouldn't be better like that ?
`} elseif (!empty(trim($_POST['auth_user2']))) { `
because : empty(" ") is false

PR #4040, opened 2019-01-21
Add options for resolv.conf

joshuasign:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/790
- [x] Ready for review


joshuasign:
I agree with you, the place to be for this is in "Advanced options", 
As far as users know what they are doing in "Advanced options", there is no problem for me to use free-form.

Even if we can be exhaustive in the options we purpose by this feature, it will make it very hard to manage and maintain. 
Thats why i try to do something very simple (sometimes the less is the best).

joshuasign:
Switched to "Advanced options Network" page. 
Link to man page added.

rbgarga:
Closing after almost one year with no activity.  @joshuasign if you want to work on this change again please re-submit

PR #4041, opened 2019-01-21
#9218 SNMP sysDescr does not display hostname and patch version

gitdevmod:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9218
- [x] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -2317,16 +2317,9 @@ trap := "{$config['snmpd']['trapstring']}"
 EOD;
 		}
 		
-		if (!empty($g['product_version_string'])) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/services.inc
```diff
@@ -2316,10 +2316,17 @@ trap := "{$config['snmpd']['trapstring']}"
 
 EOD;
 		}
-
-		$sysDescr = "{$g['product_name']} " . php_uname("n") .
-			" {$g['product_version']} {$g['platform']} " . php_uname("s") .
-			" " . php_uname("r") . " " . php_uname("m");
+		
+		if (!empty($g['product_version_string'])) {
```
rbgarga: `$g['product_version_string']` is always defined and not empty.  Consider always using it

PR #4043, opened 2019-01-22
add an option to the DHCP server to disable the ping check feature

artooro:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9285
- [x] Ready for review


artooro:
I've renamed the option to disablepingcheck to reflect what it does, still using the isset method to keep consistent with the rest of the dhcp service code.

PR #4044, opened 2019-01-27
Routing, actually show the "(default)" mark on the default route as it is present on the OS

PiBa-NL:
Routing, actually show the "(default)" mark on the default route as it is present on the OS

Most obvious problem was when manually switching from WANGW1 to WANGW2 it showed both as (default) after saving the setting and before applying it. Also after applying it would require another page refresh to show the updated situation. Also add a little information box that shows what (Default) means for the user.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9292
- [x] Ready for review

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -671,11 +671,11 @@ function return_gateways_array($disabled = false, $localhost = false, $inactive 
 			}
 
 			if ($gateway['ipprotocol'] == "inet" &&
-					($config['gateways']['defaultgw4'] == $gateway['name'] || $gateway['gateway'] == $cgw4)) {
+					($gateway['gateway'] == $cgw4)) {
 				$gateway['isdefaultgw'] = true;
 				$found_defaultv4 = 1;
 			} else if ($gateway['ipprotocol'] == "inet6" &&
-					($config['gateways']['defaultgw6'] == $gateway['name'] || $gateway['gateway'] == $cgw6)) {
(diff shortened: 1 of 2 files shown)


jim-p:
I seem to recall this coming up before. Seems like no matter what we do here, there is a potential source of confusion:

1. Now it can show 'default' for both configured and active default gateways, until the user clicks apply.
2. With this, it only shows the 'default' as indicated in the routing table, which can be confusing to a user since if they marked a gateway as default in the configuration, they'd expect it to be marked as default after clicking save. It isn't obvious they need to click apply to see that flag change.
3. Going the other way, if we showed 'default' only next to the configured default, then it would be confusing if that doesn't match the routing table, since the user would have to check to know that it wasn't actually active.

The optimal behavior probably needs more thought. A few random ideas to discuss: Perhaps a warning box stating that the configured and active defaults don't match, or some way to distinguish between both configured and active while both are visible. Maybe use an icon instead of '(default)', then use different icons for different things (active default, configured default, member of a default gateway group...).

PiBa-NL:
With this change it will show the OS default as (default), and the configured default in the "Default" column.   If the current default needs to be shown with a little icon, i can probably do that.. p.s. what do the current 'checkcircle' icons mean.? perhaps i could use those to show the current OS default.?

jim-p:
The check circle indicates if the gateway is enabled/disabled (it will be an a circled x if it's disabled), and it could be used for that with another icon. There are plenty of icons to choose from.

PR #4047, opened 2019-02-01
Fix behaviour of determining dhcp primary/secondary

fredronnv:
When there is more than one CARP address per interface, with different
advskews which is typical in active/active configuration we need to
check only the first CARP if and skew in order to determine primary.

Previously if you had two CARP vips where VIP a had skew <20 on host A
and VIP b had skew <20 on host B, both hosts would be configured as
primary, and dhcp failover would not be configured correctly.

The caveat to this fix is that VIPs need to be configured in the same
order on both hosts.

- [ ] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -602,8 +602,8 @@ EOD;
 							/* this is the interface! */
 							if (is_numeric($vipent['advskew']) && (intval($vipent['advskew']) < 20)) {
 								$skew = 0;
-								break;
 							}
+							break;
 						}
 					}
 				}


fredronnv:
Maybe this behaviour should be further clarified also in the descriptive text "Failover peer IP", to ensure that VIPs are configured correnctly for people who run such a setup?

jim-p:
Active/active is not a supported configuration, all skews on the master should be identical.

fredronnv:
> Active/active is not a supported configuration, all skews on the master should be identical.

Sure, but then the patch could simply be reworded as "All skews should be identical, no need to check more than one.", if that helps. No mention of active/active, and slightly more efficient code :)

Currently serving 20k clients from these boxes, we need to load-balance clients which we accomplish using separate dhcp pools with different default routes (CARP VIPs). We've been applying this patch for two years, so its reasonably battle tested. That's the motivation, just would rather not have to apply the patch every update.

PR #4048, opened 2019-02-06
Host-Uniq

NanoCaiordo:
Host-Uniq untested

- [x] Redmine Issue: https://redmine.pfsense.org/issues/7618
- [ ] Ready for review

Code is "almost" fully tested in master but barely tested on 2.4.4-p2 since I'd like not to risk damaging RJ-45 pins and I'm still waiting for a RJ-11 RJ-45 cable.

Patch accept either plain text HostUniq or the hexadecimal value as 0xhostuniq.

I'm pretty sure help text or error message can be improved, please advise.

### DIFF
--- a/src/usr/local/www/interfaces.php
+++ b/src/usr/local/www/interfaces.php
@@ -130,6 +130,9 @@ if ($wancfg['if'] == $a_ppps[$pppid]['if']) {
 		$pconfig['pppoe_username'] = $a_ppps[$pppid]['username'];
 		$pconfig['pppoe_password'] = base64_decode($a_ppps[$pppid]['password']);
 		$pconfig['provider'] = $a_ppps[$pppid]['provider'];
+		if (isset($a_ppps[$pppid]['pppoe_host-uniq'])) {
+			$pconfig['pppoe_host-uniq'] = $a_ppps[$pppid]['pppoe_host-uniq'];
+		}
 		$pconfig['pppoe_dialondemand'] = isset($a_ppps[$pppid]['ondemand']);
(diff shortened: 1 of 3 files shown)


src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -671,6 +680,13 @@ function build_country_list() {
 					'Check the "Configure NULL" box to configure a blank Service name.');
 
 	$section->add($group);
+
+	$section->addInput(new Form_Input(
+		'pppoe_host-uniq',
+		'Host-Uniq',
+		'text',
+		$pconfig['pppoe_host-uniq']
+	))->setHelp('This field can usually be left empty.');
```
sbeaver-netgate: I'd like to see a description of what this option does rather than just "Can usually be left blank"

src/usr/local/www/interfaces.php
```diff
@@ -2800,6 +2812,13 @@ function build_port_list() {
 	$pconfig['provider']
 ))->setHelp('This field can usually be left empty.');
 
+$section->addInput(new Form_Input(
+	'pppoe_host-uniq',
+	'Host-Uniq',
+	'text',
+	$pconfig['pppoe_host-uniq']
+))->setHelp('This field can usually be left empty.');
```
jim-p: Needs a better description of what the field is and why someone might need to set it.

src/usr/local/www/interfaces_ppps_edit.php
jim-p: Needs a better description of what the field is and why someone might need to set it.

rbgarga:
Closing it due to inactivity since February. If you still want to work on it please re-submit

PR #4050, opened 2019-02-14
Option to hide Kernel PTI from sysinfo widget

Rybo-W:
Add the option to hide the Kernel PTI entry from the system information dashboard widget.

To gain more viewing space I have two sysinfo widgets to split the information on the dashboard, one showing mostly the progress bars (visible) and the other (out of view) showing everything else. Because the Kernel PTI item cannot be hidden I'm forced to have that item visible in BOTH widgets.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9323
- [x] Ready for review


jim-p:
I implemented this a slightly different way for 2.5.0. It's in the repo now. See Redmine for more info.

PR #4051, opened 2019-02-16
Increase number of parallel IKEv1 Phase 2 rekeys

plappermaul:
In certain situations remote firewalls might issue a rekey for
all CHILD_SAs in parallel. At least Watchguard firewall clusters
have been seen flooding rekeys after they failover.

strongSwan usually only allows for 3 concurrent rekeys. Nevertheless
it already has the option max_ikev1_exchanges to increase that
value. This patch adds the setting to /etc/inc/vpn.inc and makes
it persistent for future releases.

More details at https://redmine.pfsense.org/issues/9331

### DIFF
--- a/src/etc/inc/vpn.inc
+++ b/src/etc/inc/vpn.inc
@@ -434,6 +434,7 @@ starter {
 charon {
 # number of worker threads in charon
 	threads = 16
+	max_ikev1_exchanges = 22
 	ikesa_table_size = 32
 	ikesa_table_segments = 4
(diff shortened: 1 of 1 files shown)


jim-p:
Rather than trading one arbitrary value for another, I'd prefer to see this as a GUI option under the IPsec Advanced options tab. When blank/empty the directive can be omitted to retain the current behavior.

plappermaul:
will resend ...

PR #4060, opened 2019-04-02
Fix a Bootstrap conversion error and empty lines in forms

PeterFeicht:
I noticed that there are a couple empty lines in forms on two pages in the Certificate Manager area, caused by using `addInput` instead of `addGlobal` to add hidden form inputs. This PR fixes them, along with all other such cases I could find throughout the GUI.

There's also a small logic error introduced during Bootstrap conversion, which is fixed as well.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9449
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/diag_smart.php
+++ b/src/usr/local/www/diag_smart.php
@@ -231,7 +231,7 @@ switch ($action) {
 
 		$section = new Form_Section('Information');
 		$group = new Form_Group('Select a drive and type:');
-		$group->add(new Form_Input(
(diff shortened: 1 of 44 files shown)


PeterFeicht:
Done, I added the Redmine link to the PR description.

Should I generally assign a new issue to myself if I already made a PR or am about to?

jim-p:
You can leave the assignment blank, usually with a PR that would be set to whoever is testing it or merging it, or back to you if there are questions/problems that need additional development work.

PR #4062, opened 2019-04-04
Add support for encrypted CA and certificate private keys

PeterFeicht:
I've been working on adding support for encrypted private keys, I've added support to the CA and certificate managers, as well as the user manager. I'm working on the OpenVPN wizard now, but wanted to get some feedback on what I have so far. I'm not a PHP programmer at all, so I just tried to stick to the existing code; any feedback in that regard is more than welcome.

I've not added support to the CRL manager yet, because I think that would require a substantial change to its interface. I'd be happy to discuss that, but I wasn't sure if a PR was the right place. Since I don't think I'm going to use CRLs at all, I could also just leave that part out for someone else to implement, if the need is there.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/1257
- [x] Ready for review

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -106,6 +106,22 @@ function & lookup_cert_by_name($name) {
 	}
 }
 
+// Get a map from ID to description for all certs with a private key,
+// optionally including certificates with an encrypted private key.
+function get_cert_list($include_encrypted = false) {
+	global $config;
+
+	$result = array();
+	if (is_array($config['cert'])) {
+		foreach ($config['cert'] as $cert) {
+			if ($cert['prv'] && ($include_encrypted || !is_encrypted_key($cert['prv']))) {
+				$result[$cert['refid']] = $cert['descr'];
+			}
+		}
+	}
+	return $result;
+}
+
 function & lookup_crl($refid) {
 	global $config;
 
@@ -855,6 +871,22 @@ function is_encrypted_key($str_key, $decode = true) {
 		$str_key = base64_decode($str_key);
 	}
 	return openssl_pkey_get_private($str_key) === false;
+}
+
+// Check whether the cert with the specified ID is valid for use on the firewall,
+// i.e. it exists and has a private key that is not encrypted.
+function is_valid_firewall_cert($certref) {
+	if (empty($certref)) {
+		return false;
+	}
+	
+	$cert = lookup_cert($certref);
+	if (!$cert || !$cert['prv']) {
+		return false;
+	} elseif (is_encrypted_key($cert['prv'])) {
+		return false;
+	}
+	return true;
 }
 
 function is_user_cert($certref) {
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -319,7 +319,10 @@ function cert_import(& $cert, $crt_str, $key_str) {
 	return true;
 }
 
-function cert_create(& $cert, $caref, $keylen, $lifetime, $dn, $type = "user", $digest_alg = "sha256") {
+function cert_create(& $cert, $caref, $keylen, $lifetime, $dn, $type = "user", $digest_alg = "sha256", $capass = "", $keypass = null) {
+	if (empty($keypass)) {
+		$keypass = null;
+	}
 
 	$cert['type'] = $type;
 
@@ -333,7 +336,7 @@ function cert_create(& $cert, $caref, $keylen, $lifetime, $dn, $type = "user", $
 		$ca_str_crt = base64_decode($ca['crt']);
 		$ca_str_key = base64_decode($ca['prv']);
 		$ca_res_crt = openssl_x509_read($ca_str_crt);
-		$ca_res_key = openssl_pkey_get_private(array(0 => $ca_str_key, 1 => ""));
+		$ca_res_key = openssl_pkey_get_private($ca_str_key, $capass);
 		if (!$ca_res_key) {
 			return false;
 		}
@@ -389,7 +392,7 @@ function cert_create(& $cert, $caref, $keylen, $lifetime, $dn, $type = "user", $
 	}
 
 	// export our certificate data
-	if (!openssl_pkey_export($res_key, $str_key) ||
+	if (!openssl_pkey_export($res_key, $str_key, $keypass) ||
 	    !openssl_x509_export($res_crt, $str_crt)) {
 		return false;
 	}
@@ -401,7 +404,10 @@ function cert_create(& $cert, $caref, $keylen, $lifetime, $dn, $type = "user", $
 	return true;
 }
 
-function csr_generate(& $cert, $keylen, $dn, $type = "user", $digest_alg = "sha256") {
+function csr_generate(& $cert, $keylen, $dn, $type = "user", $digest_alg = "sha256", $keypass = null) {
+	if (empty($keypass)) {
+		$keypass = null;
+	}
 
 	$cert_type = cert_type_config_section($type);
 
@@ -434,7 +440,7 @@ function csr_generate(& $cert, $keylen, $dn, $type = "user", $digest_alg = "sha2
 	}
 
 	// export our request data
-	if (!openssl_pkey_export($res_key, $str_key) ||
+	if (!openssl_pkey_export($res_key, $str_key, $keypass) ||
 	    !openssl_csr_export($res_csr, $str_csr)) {
 		return false;
 	}
@@ -446,14 +452,14 @@ function csr_generate(& $cert, $keylen, $dn, $type = "user", $digest_alg = "sha2
 	return true;
 }
 
-function csr_sign($csr, & $ca, $duration, $type = "user", $altnames, $digest_alg = "sha256") {
+function csr_sign($csr, & $ca, $duration, $type = "user", $altnames, $digest_alg = "sha256", $capass = "") {
 	global $config;
 	$old_err_level = error_reporting(0);
 
 	// Gather the information required for signed cert
 	$ca_str_crt = base64_decode($ca['crt']);
 	$ca_str_key = base64_decode($ca['prv']);
-	$ca_res_key = openssl_pkey_get_private(array(0 => $ca_str_key, 1 => ""));
+	$ca_res_key = openssl_pkey_get_private($ca_str_key, $capass);
 	if (!$ca_res_key) {
 		return false;
 	}
@@ -476,7 +482,7 @@ function csr_sign($csr, & $ca, $duration, $type = "user", $altnames, $digest_alg
 	);
 
 	// Sign the new cert and export it in x509 format
-	openssl_x509_export(openssl_csr_sign($csr, $ca_str_crt, $ca_str_key, $duration, $args, $ca_serial), $n509);
+	openssl_x509_export(openssl_csr_sign($csr, $ca_str_crt, $ca_res_key, $duration, $args, $ca_serial), $n509);
 	error_reporting($old_err_level);
 
 	return $n509;
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -199,7 +199,10 @@ function ca_import(& $ca, $str, $key = "", $serial = "") {
 	return true;
 }
 
-function ca_create(& $ca, $keylen, $lifetime, $dn, $digest_alg = "sha256") {
+function ca_create(& $ca, $keylen, $lifetime, $dn, $digest_alg = "sha256", $password = null) {
+	if (empty($password)) {
+		$password = null;
+	}
 
 	$args = array(
 		"x509_extensions" => "v3_ca",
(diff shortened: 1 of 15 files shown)


src/etc/inc/certs.inc
```diff
@@ -238,15 +241,19 @@ function ca_create(& $ca, $keylen, $lifetime, $dn, $digest_alg = "sha256") {
...
+
 	// Create Intermediate Certificate Authority
 	$signing_ca =& lookup_ca($caref);
 	if (!$signing_ca) {
 		return false;
 	}
 
 	$signing_ca_res_crt = openssl_x509_read(base64_decode($signing_ca['crt']));
-	$signing_ca_res_key = openssl_pkey_get_private(array(0 => base64_decode($signing_ca['prv']) , 1 => ""));
+	$signing_ca_res_key = openssl_pkey_get_private(array(0 => base64_decode($signing_ca['prv']) , 1 => $capass));
```
PeterFeicht: Why are the arguments passed as an array to `openssl_pkey_get_private`, does that have any significance?

src/usr/local/www/system_camanager.php
```diff
@@ -99,6 +99,7 @@
 	$pconfig['serial'] = $a_ca[$id]['serial'];
 	if (!empty($a_ca[$id]['prv'])) {
 		$pconfig['key'] = base64_decode($a_ca[$id]['prv']);
+		$pconfig['encryptkey'] = (strpos($pconfig['key'], 'ENCRYPTED') !== false);
```
PeterFeicht: I don't know if that's the right way to check for encrypted keys. It was this way before, but there's the very slim chance that the base64 encoded string might contain `ENCRYPTED`. Maybe this should be a function in `certs.inc`?

src/usr/local/www/system_usermanager.php
```diff
@@ -705,7 +721,7 @@ function build_cert_table() {
 		'act',
 		null,
 		'hidden',
-		''
+		$act
```
PeterFeicht: I think this should have been `$act` before, but please have a look if it's right the way it is now.

src/usr/local/www/system_camanager.php
```diff
@@ -616,6 +688,11 @@
 	$pconfig['caref'],
 	$allCas
 ));
+$group->add(new Form_Input(
```
PeterFeicht: Should this be in a separate section? I couldn't find a way to make it look nice when both are toggled, so I added the password field to the group.

PeterFeicht:
Added a few comments to the diff.

src/etc/inc/certs.inc
```diff
@@ -622,6 +647,40 @@ function cert_get_publickey($str_crt, $decode = true, $type = "crt") {
...
+	if ($decode) {
+		$str_key = base64_decode($str_key);
+	}
+
+	$keyfn = tempnam('/tmp', 'CEK');
+	$passfn = tempnam('/tmp', 'CEK');
+	file_put_contents($keyfn, $str_key);
+	file_put_contents($passfn, $password);
+
+	exec("/usr/bin/openssl pkey -in {$keyfn} -des3 -passout file:{$passfn} 2>/dev/null", $out);
```
jim-p: Why 3DES (`-des3`) here, and not something more secure, like `-aes128` or `-aes256`?

src/etc/inc/certs.inc
```diff
@@ -622,6 +647,40 @@ function cert_get_publickey($str_crt, $decode = true, $type = "crt") {
 			break;
 	}
 	unlink($certfn);
+	return ($out !== false ? implode("\n", $out) : false);
```
jim-p: The comparison here should be inside parenthesis, so `return (($out !== false) ? implode("\n", $out) : false);` for example

src/etc/inc/certs.inc
```diff
@@ -601,15 +614,27 @@ function cert_get_issuer($str_crt, $decode = true) {
...
 	switch ($type) {
 		case 'prv':
-			exec("/usr/bin/openssl pkey -in {$certfn} -pubout", $out);
+			if ($password) {
+				$passfn = tempnam('/tmp', 'CGPK');
+				file_put_contents($passfn, $password);
+
+				exec("/usr/bin/openssl pkey -in {$certfn} -passin file:{$passfn} -pubout 2>/dev/null", $out, $status);
+
+				unlink($passfn);
```
jim-p: It's safer to use `unlink_if_exists()` here instead of `unlink()`. Will be the same for other uses throughout this code.

src/etc/inc/certs.inc
```diff
@@ -622,6 +647,40 @@ function cert_get_publickey($str_crt, $decode = true, $type = "crt") {
...
+	return implode("\n", $out);
+}
+
+function cert_decrypt_key($str_key, $password, $decode = true) {
+	if ($decode) {
+		$str_key = base64_decode($str_key);
+	}
+
+	$keyfn = tempnam('/tmp', 'CDK');
+	$passfn = tempnam('/tmp', 'CDK');
```
jim-p: Same as above, use unique prefixes.

src/usr/local/www/system_camanager.php
jim-p: IIRC you can try the openssl function `openssl_pkey_get_private()` on the key with an empty password and if it fails/returns empty then it's safe to assume that it's encrypted. I haven't tested that so I'm not sure exactly what/how it returns, but that should be reliable.

src/usr/local/www/system_camanager.php
jim-p: Have to check it out when testing to be sure but it's probably OK like this

src/usr/local/www/system_certmanager.php
```diff
@@ -195,6 +195,20 @@
 }
 
 if ($_POST['save']) {
+	// If passwords are given, check that they match and verify their length
+	function validatePasswords(& $input_errors, $pw1, $pw2) {
```
jim-p: Why not put this function somewhere common (like `certs.inc`, or `guiconfig.inc`) and use it here and on `system_camanager.php`? It would need a more specific name, though, such as `validate_cert_passwords()`.

src/usr/local/www/system_usermanager.php
jim-p: It should be empty as it was.

src/etc/inc/certs.inc
PeterFeicht: Will do.

src/usr/local/www/system_camanager.php
PeterFeicht: I'll try that, and I'll factor the check out into a separate function in any case.

src/usr/local/www/system_usermanager.php
PeterFeicht: But then there's the problem that the empty POST parameter overrides the non-empty GET parameter, and the form won't show right when you submit but there's an error.

src/usr/local/www/system_camanager.php
PeterFeicht: 👍 It looks fine on desktop, it even flows nicely on small screens.

src/etc/inc/certs.inc
```diff
@@ -601,15 +614,27 @@ function cert_get_issuer($str_crt, $decode = true) {
...
 	if ($decode) {
 		$str_crt = base64_decode($str_crt);
 	}
 	$certfn = tempnam('/tmp', 'CGPK');
 	file_put_contents($certfn, $str_crt);
 	switch ($type) {
 		case 'prv':
-			exec("/usr/bin/openssl pkey -in {$certfn} -pubout", $out);
+			if ($password) {
+				$passfn = tempnam('/tmp', 'CGPK');
```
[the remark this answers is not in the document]
PeterFeicht: Though the documentation says unique, apparently there have been problems. I'll change it.

src/usr/local/www/system_certmanager.php
PeterFeicht: That's not a bad idea considering that it'll be needed for the OpenVPN wizard too.

src/etc/inc/certs.inc
PeterFeicht: Should I go ahead and change all uses of `unlink`?

src/etc/inc/certs.inc
PeterFeicht: ACK

src/etc/inc/certs.inc
PeterFeicht: I just always used `-des3`, I don't have a preference either way.

PeterFeicht:
I've thought about that too, I agree there needs to be a check at least. Since certificates are already filtered in such lists based on whether they even have a private key, I'll just add another check there whether the key is encrypted. I'll add a corresponding check as well.

It would also be nice to have some kind of indication whether a certificate/CA has an encrypted private key in the respective lists. I don't quite know where to put that, a separate column would make the lists a little too crowded I think. For the certificate list I'm thinking a padlock symbol after the purpose string maybe? For CAs the same after the name?

Regarding encrypted export, since encrypted keys are exported as they are anyway, I'd prefer to leave that for a separate PR.

src/etc/inc/certs.inc
jim-p: I'd prefer to see this as at least `-aes128`, otherwise this looks OK to merge and test.

PeterFeicht:
I finally got around to finishing up this PR.

 * Added a new function `get_cert_list` for getting a list of certificates (including or excluding encrypted ones), which is used in most places that display a list of certificates. There are checks against encrypted certificates at those places too.
 * The OpenVPN wizard now supports creation of new certificates using an encrypted CA; encrypted certificates are filtered from all wizard certificate lists (currently only used in the OpenVPN wizard).
 * I also added an indication for encrypted private keys to the CA and certificate lists.

There's only one open question: https://github.com/pfsense/pfsense/pull/4062#discussion_r274039904

jim-p:
I didn't see that reply. I guess we can try populating that, as long as it doesn't negatively affect existing behavior.

PeterFeicht:
Just had a quick look and I think it needs to be set in other places too. I'll get back to you when I've had some time to look it over in detail.

rbgarga:
@PeterFeicht do you still have plans to work on this PR?

rbgarga:
If yes, please rebase your fork. The change is conflicting and cannot be merged

PeterFeicht:
Yeah, sorry for the long silence, I'm still on it. I'll have a look at it in the next couple of days I think.

When I rebase my branch, would it make sense to work the review changes into my commits to clean up the branch, or how do guys usually handle this?

rbgarga:
If you can squash your commits and split them in commits that make sense is better since we don't use the squash function from github's GUI.

PeterFeicht:
So I just rebased onto the master branch and squashed review changes, the diff didn't change at all.
I'll use fixup commits for future review changes (since I know how to use those now, really handy I'll tell you) and squash the branch when it's ready to be merged.

I still have to look at the `$act` thing in detail.

rbgarga:
> So I just rebased onto the master branch and squashed review changes, the diff didn't change at all.
> I'll use fixup commits for future review changes (since I know how to use those now, really handy I'll tell you) and squash the branch when it's ready to be merged.
> 
> I still have to look at the `$act` thing in detail.

great. I'll wait to hear from you about `$act` change and then test it

PeterFeicht:
I've had a look at the `$act` thing now and I'm pretty sure it's fine the way it is now (setting the POST `act` variable to `$act` in [system_usermanager.php:753](https://github.com/pfsense/pfsense/pull/4062/files#diff-adf66c0efb366e75ef2329bafa9d239bR753)).

During my testing I discovered two other things that didn't work quite right and I fixed them too. Please let me know if I should file an issue and move those commits to a separate branch/PR instead.
I also removed a redundant if statement, this could go in a separate PR as well.

Let me know if/when you're ready to merge the branch, then I will squash that fixup commit one last time.

src/usr/local/www/system_usermanager.php
```diff
@@ -742,307 +742,306 @@ function build_cert_table() {
 	exit;
 }
 
+// ($act == "new" || $act == "edit" || $input_errors) == true
```
rbgarga: Please remove this commented out line

PeterFeicht: Done.

rbgarga:
@PeterFeicht do you still plan to work on it?

PeterFeicht:
@rbgarga yes I do. I'm sorry for the long delay, I haven't been well lately.
I should be able to work on it again in the next couple of weeks. This currently targets 2.5, right? Do you have any estimate on when that will be? I'm asking because I tend to get things done better with a deadline looming ;-)

rbgarga:
@PeterFeicht do you still have plans to work on this change?

PeterFeicht:
Yes I do, sorry for the long inactivity.

rbgarga:
Closing due to inactivity since Sep 2020

Dantali0n:
This is very disappointing to see, I was looking forward to this new feature :'(

PR #4063, opened 2019-04-09
Allow import of PKCS12 (pfx) certificates

miken32:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8645
- [x] Ready for review

Allows user to select a PKCS #12 certificate store to import a certificate. Optionally allows for import of bundled intermediate certificates. Tested with a couple of different certificates.

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -168,6 +168,7 @@ function ca_import(& $ca, $str, $key = "", $serial = "") {
 	}
 	$subject = cert_get_subject($str, false);
 	$issuer = cert_get_issuer($str, false);
+	$serialNumber = cert_get_serial($str, false);
 
 	// Find my issuer unless self-signed
 	if ($issuer <> $subject) {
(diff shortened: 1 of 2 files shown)


miken32:
Annoyances: the browser prompts to save password, and the system will allow import of duplicate CAs.

jim-p:
Not much can be done about the browser password, unfortunately. Browsers have decided that web pages can't be trusted to tell them what fields they shouldn't save, so traditional controls to tell them not to auto-fill are ignored these days.

The duplicate CA issue could be solved by checking existing CAs and comparing the subject/serial/etc. Even a basic check using `lookup_ca_by_subject()` and seeing if anything is found should solve most cases.

Though this PR looks OK, it conflicts with another open PR, #4062 -- Depending on which one is accepted first, the other will need fixed to accommodate the changes of the other.

miken32:
I thought about checking the subject, but that's not enough, so serial number would probably need to be checked as well. Ideally these checks would be in a centralized place so that duplicate CAs and certs can't be imported anywhere. I'm not crazy about the way `import_ca()` works (modifying by reference and never returning anything useful) because it makes this more difficult than it should be.

Happy to work around the changes in the other PR if it comes to that.

miken32:
Updated checks; duplicates were not imported during my testing.

Re password field, I'd propose that it doesn't need to be a password field at all. The only thing a password field prevents is shoulder surfing; given this is an administrative backend to manage a system and not a login page, the security implications for almost any user would be negligible. I don't expect anyone to agree with me though.

jim-p:
Closing in favor of #4175

PR #4066, opened 2019-05-15
DHCP Patch for Correct MAC Limitations

reb00tz:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/1605
- [X] Ready for review


src/etc/inc/services.inc
```diff
@@ -841,7 +844,12 @@ EOPP;
 				$dhcpdconf .= "		$deny_action dynamic bootp clients;\n";
 			}
 
-			if (isset($poolconf['denyunknown'])) {
+			// set pool MAC limitations
+			if ($poolconf['denyunknown'] == "class") {
+				$dhcpdconf .= "		allow members of \"s_{$dhcpif}\";\n";
+				$dhcpdconf .= "		$deny_action known-clients;\n";
+				$dhcpdconf .= "		$deny_action unknown-clients;\n";
+			} else if (isset($poolconf['denyunknown'])) {
```
rbgarga: This field has changed from a boolean checkbox to a select form. IT must be checked comparing its value with the string `"disabled"` from now.

Also, old configuration items should be converted to the new field type.

src/usr/local/www/services_dhcp.php
```diff
@@ -581,7 +591,15 @@ function validate_partial_mac_list($maclist) {
 		$dhcpdconf['domain'] = $_POST['domain'];
 		$dhcpdconf['domainsearchlist'] = $_POST['domainsearchlist'];
 		$dhcpdconf['ignorebootp'] = ($_POST['ignorebootp']) ? true : false;
-		$dhcpdconf['denyunknown'] = ($_POST['denyunknown']) ? true : false;
+
+		if ($_POST['denyunknown'] == "enabled") {
+			$dhcpdconf['denyunknown'] = "enabled";
+		} else if ($_POST['denyunknown'] == "class") {
+			$dhcpdconf['denyunknown'] = "class";
+		} else if (isset($_POST['denyunknown'])) {
```
jim-p: Why test that last case? It would be better to just unset it unconditionally if the first two cases fail.

Or consider rewriting it to a more adaptable method with less duplication of text:

```
if (in_array($_POST['denyunknown'], array("enabled", "class"))) {
	$dhcpdconf['denyunknown'] = $_POST['denyunknown'];
} else {
	unset($dhcpdconf['denyunknown']);
}
```

src/usr/local/www/services_dhcp.php
```diff
@@ -868,12 +886,19 @@ function build_pooltable() {
...
-$section->addInput(new Form_Checkbox(
+$section->addInput(new Form_Select(
 	'denyunknown',
 	'Deny unknown clients',
-	'Only the clients defined below will get DHCP leases from this server.',
-	$pconfig['denyunknown']
-));
+	$pconfig['denyunknown'],
+	array(
+		"disabled" => "Allow Any (any MACs, including any not listed anywhere)",
```
jim-p: Not sure this needs the parenthetical part. `Allow all clients` gets the point across clearly and concisely.

src/usr/local/www/services_dhcp.php
```diff
@@ -868,12 +886,19 @@ function build_pooltable() {
...
+		"enabled" => "Allow Known Clients (MACs listed on ANY interface)",
```
jim-p: Consider: `Allow known clients from any interface`

src/usr/local/www/services_dhcp.php
```diff
@@ -868,12 +886,19 @@ function build_pooltable() {
...
+		"class" => "Allow Only Listed Below (only MACs listed below)",
```
jim-p: Consider: `Allow known clients from only this interface`

src/usr/local/www/services_dhcp.php
```diff
@@ -868,12 +886,19 @@ function build_pooltable() {
...
+	)
+))->setHelp('When set to %3$sAllow Any%4$s, any DHCP client will get an IP address within this scope/range on this interface. '.
+	'If set to %3$sAllow Known Clients%4$s, any DHCP client with a MAC address listed on %1$s%3$sany%4$s%2$s scope(s)/interface(s) will get an IP address. ' .
+	'If set to %3$sAllow Only Listed Below%4$s, only MAC addresses listed below (i.e. for this interface) will get an IP address within this scope/range.',
+	'<i>', '</i>', '<b>', '</b>');
```
jim-p: These are fine, but adjust to new suggested option labels.

src/etc/inc/services.inc
```diff
@@ -841,7 +844,14 @@ EOPP;
...
-			if (isset($poolconf['denyunknown'])) {
+			// set pool MAC limitations
+			if ($poolconf['denyunknown'] == "class") {
+				$dhcpdconf .= "		allow members of \"s_{$dhcpif}\";\n";
+ 			//	$dhcpdconf .= "		$deny_action known-clients;\n";        // in combination with above and below options, this will disable all entries WITHOUT FIXED IPs!
+				$dhcpdconf .= "		$deny_action unknown-clients;\n";
+			} else if ($poolconf['denyunknown'] == "disabled") {
+				// add nothing to $dhcpdconf; condition added to prevent next condition applying if ever engine changes such that: isset("disabled") == true
+			} else if (isset($poolconf['denyunknown'])) {	// "catch-all" covering "enabled" value post-PR#4066, and covering non-upgraded boolean option (i.e. literal value "enabled")
 				$dhcpdconf .= "		$deny_action unknown-clients;\n";
```
rbgarga: It doesn't make much sense to check if `$poolconf['denyunknown']` is set at this point. IMO this whole block should be simplified using something like:
```
if (isset($poolconf['denyunknown']) &&
    $poolconf['denyunknown'] != "disabled") {
	if ($poolconf['denyunknown'] == "class") {
		$dhcpdconf .= "		allow members of \"s_{$dhcpif}\";\n";
	}
	$dhcpdconf .= "		$deny_action unknown-clients;\n";
}
```

reb00tz: Apologies for the delay. I disagree:

1. the `isset` is to check for backwards (as-yet-not-upgraded) `denyunknown` values
2. without considering _frequency_, the universe of `!= "disabled"` (negative tests) is far larger than positive tests, and would also be clearer and more concise if/when future values are added

src/etc/inc/services.inc
```diff
@@ -841,7 +844,14 @@ EOPP;
 				$dhcpdconf .= "		$deny_action dynamic bootp clients;\n";
 			}
 
-			if (isset($poolconf['denyunknown'])) {
+			// set pool MAC limitations
+			if ($poolconf['denyunknown'] == "class") {
+				$dhcpdconf .= "		allow members of \"s_{$dhcpif}\";\n";
+ 			//	$dhcpdconf .= "		$deny_action known-clients;\n";        // in combination with above and below options, this will disable all entries WITHOUT FIXED IPs!
```
rbgarga: Please delete commented out line

reb00tz: Done.

jim-p:
Note: The Redmine issue for this is actually https://redmine.pfsense.org/issues/1605 (the newer one was a duplicate)

src/etc/inc/services.inc
```diff
@@ -770,6 +770,9 @@ EOPP;
 			}
 		}
 
+		// instantiate class before pool definition
+		$dhcpdconf .= "class \"s_{$dhcpif}\" {\n	match hardware;\n}\n";
```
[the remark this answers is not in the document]
reb00tz: Followed other code lines using the same style. (Search for regex `$dhcpdconf .= \"\t`). `\t` would actually make it harder to read here (i.e. `{\n\tmatch hardware;\n}\n\";`).

src/etc/inc/services.inc
```diff
@@ -841,7 +844,13 @@ EOPP;
 				$dhcpdconf .= "		$deny_action dynamic bootp clients;\n";
 			}
 
-			if (isset($poolconf['denyunknown'])) {
+			// set pool MAC limitations
+			if ($poolconf['denyunknown'] == "class") {
```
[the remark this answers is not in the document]
reb00tz: Done!

src/etc/inc/services.inc
```diff
@@ -1165,6 +1174,12 @@ EOD;
 				}
 
 				$dhcpdconf .= "}\n";
+
+				// subclass for DHCP limiting
+				// unknown if the "1:" prefix is still needed, so adding it just in case
+				$dhcpdconf .= "subclass \"s_{$dhcpif}\" 1:{$sm['mac']};\n";
+				$dhcpdconf .= "subclass \"s_{$dhcpif}\" {$sm['mac']};\n";
```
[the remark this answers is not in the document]
reb00tz: Done!

rbgarga:
Superseded by #4066

reb00tz:
@rbgarga: I think this should be "Superseded by [#4184](https://github.com/pfsense/pfsense/pull/4184)".

PR #4068, opened 2019-05-21
Issue 9539

jwsi:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9539
- [ ] Ready for review

These changes allow for the ability for the admin account to be synced between HA pairs.

### The following functionality is introduced:

1. All admin account changes are synced to a partner.
2. When the admin password is changed on the primary node, the secondary is updated accordingly and the primary HA sync password is also updated to prevent future auth errors occuring during sync.

### Important Notes:
- When an admin account password is changed, the config.xml file will be updated accordingly on the secondary. In xmlrpc.php there is a function called `auth` which authenticates the entity making changes to the device. Obviously this cannot be called if the config.xml file has been updated with the new credentials, but the PHP session is still using the old credentials. Therefore, I have made `filter_configure` a private function to remove this possibility.

James.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -447,7 +447,9 @@ function local_sync_accounts($u2add, $u2del, $g2add, $g2del) {
 	}
 
 	foreach($u2del as $user) {
-		if ($user['uid'] < 2000 || $user['uid'] > 65000) {
+		if ($user['uid'] > 65000) {
+			continue;
+		} else if ($user['uid'] < 2000 && !in_array($user, $u2add)) {
 			continue;
 		}
 
@@ -774,9 +776,14 @@ function local_user_del($user) {
 }
 
 function local_user_set_password(&$user, $password) {
+	global $config;
+
 	unset($user['password']);
 	unset($user['md5-hash']);
 	$user['bcrypt-hash'] = password_hash($password, PASSWORD_BCRYPT);
+	if ($user['name'] == $config['hasync']['username']) {
+		$config['hasync']['new_password'] = $password;
+	}
 }
 
 function local_user_get_groups($user, $all = false) {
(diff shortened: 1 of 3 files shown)


jim-p:
I'm not convinced this is the best course of action. Some people like to keep separate admin account passwords on each node, and this breaks that behavior. I can see syncing some account properties such as SSH keys but I don't believe we should clobber the credentials.

Changing the behavior in this way by default will catch a number of people off guard in a negative way (POLA violation).

Another option is to not change anything, but document on the XMLRPC sync settings and elsewhere that admin will not sync.

jwsi:
In all fairness I'd agree with the above. I just assumed from the issue tracker that this functionality was requested. Personally, I think that if there is a requirement for keeping separate admin account passwords on each node, then in my opinion it would make sense to segregate the SSH keys and other ancillary account properties too. Otherwise, a hybrid solution feels as though it's neither segregated nor synced.

In addition to the above, it's not really that much effort to log into a secondary node and add an SSH key versus the potential ramification of unitentially allowing SSH access to a pfSense instance without realising.

Open to suggestions, but maybe it's worth just documenting that the admin account will not sync.

jim-p:
Yeah it's a bit confusing as-is but maybe that could be solved with documentation. I had initially thought it seemed like a bug but the more I thought about it, the more the current behavior made sense in that context. What we could do, however, is have a separate checkbox in the sync options to sync the admin account. Disabled by default, then someone could opt into the behavior.

jwsi:
Yeah, the more I delve into this, the more it seems that the current implementation is rather well defined.

I think documentation is a good addition regardless of the approach chosen. However, I like the idea of having a separate checkbox in the sync options specifically for the admin account. Seems like the best of both worlds if a specific option is available to users. This then clearly communicates the behaviour of the config sync wrt. admin accounts and ensures that a user consents to the admin account being synced to a secondary node if they wish to do so.

rbgarga:
No recent updates since May 2019.  Closing.

PR #4070, opened 2019-06-07
Fix handing of DNSimple API response

paul:
It seems DNSimple started using HTTP/2, which broke the regex the dnsimple updater was using to check for success. I changed it to use the CURLINFO_HTTP_CODE instead.

I noticed several other providers are using the regex status match instead of HTTP_CODE, but I didn't touch those. I haven't written any PHP in 20 years, and I don't use those providers to test them. This change works for me for DNSimple now.

The HTTP/2 response the API was sending back (missing the "OK" the regex was looking for):

```
HTTP/2 200 \r\n
server: nginx
...
```

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9580
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -2005,10 +2005,12 @@
 					   Responds with HTTP 4xx on error.
 					   Returns JSON data as body */
 ;
-					if (preg_match("/\s200\sOK/i", $header)) {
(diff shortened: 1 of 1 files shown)


paul:
@jim-p Ok I created a Redmine ticket, and linked it here in the review app and in the commit message.

jim-p:
Looks good, thanks!

PR #4072, opened 2019-06-19
Issue 9595

jwsi:
# OpenVPN Gateway Group Fix

Implementation now checks if OpenVPN client/server running on gateway group should resync when IP changes occur or if cables are unplugged/replugged.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9595
- [x] Ready for review

### VirtualBox
1. Use VirtualBox with 3 NAT adapters (WAN1, WAN2, and a monitoring link).
2. Set up a gateway group with WAN1 (T1) and WAN2 (T2).
3. Define ovpn client on gwgroup interface.
4. Alternate adapters within virtualbox between "NAT" and "Not Attached", one will notice that OpenVPN never resyncs to WAN1 when interface comes back online.

### Physical Device
1. Set up gateway group with 2 interfaces on differing tiers.
2. Unplug WAN1 cable, openvpn resyncs to WAN2.
3. Replug WAN1 cable, openvpn hugs WAN2 still.

## Fix Information
Fix brings the codebase more in line with the currrent implementation defined in [rc.openvpn](https://github.com/pfsense/pfsense/blob/48fee584e6c2cf7cb1d0821dd45288da2fd8e126/src/etc/rc.openvpn) which accounts for situations where OpenVPN instances run on gateway groups.

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1584,6 +1584,8 @@ function openvpn_resync_all($interface = "") {
 				if (trim(file_get_contents($fpath), " \t\n") != get_failover_interface($settings['interface'])) {
 					openvpn_resync('server', $settings);
 				}
+			} else {
+				openvpn_resync('server', $settings);
 			}
 		}
 	}
@@ -1595,6 +1597,8 @@ function openvpn_resync_all($interface = "") {
 				if (trim(file_get_contents($fpath), " \t\n") != get_failover_interface($settings['interface'])) {
 					openvpn_resync('client', $settings);
 				}
+			} else {
+				openvpn_resync('client', $settings);
 			}
 		}
 	}
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1581,11 +1581,11 @@ function openvpn_resync_all($interface = "") {
 			$mode_id = "server" . $settings['vpnid'];
 			$fpath = "{$g['varetc_path']}/openvpn/{$mode_id}.interface";
 			$gw_group_change = FALSE;
-			$ip_change = ($interface === $settings['interface']);
-			if (file_exists($fpath)) {
+			$ip_change = ($interface === "") || ($interface === $settings['interface']);
+			if (!$ip_change && file_exists($fpath)) {
 				$gw_group_change = (trim(file_get_contents($fpath), " \t\n") != get_failover_interface($settings['interface']));
 			}
-			if ($interface === "" || $ip_change || $gw_group_change) {
+			if ($ip_change || $gw_group_change) {
 				openvpn_resync('server', $settings);
 			}
 		}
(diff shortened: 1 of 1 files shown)


jwsi:
Is it possible to get a review on this soon?

src/etc/inc/openvpn.inc
```diff
@@ -1575,21 +1575,33 @@ function openvpn_resync_all($interface = "") {
...
 
+	// Check if OpenVPN clients and servers require a resync.
 	if (is_array($config['openvpn']['openvpn-server'])) {
 		foreach ($config['openvpn']['openvpn-server'] as & $settings) {
-			if ($interface <> "" && $interface != $settings['interface']) {
-				continue;
+			$mode_id = "server" . $settings['vpnid'];
+			$fpath = "{$g['varetc_path']}/openvpn/{$mode_id}.interface";
+			$gw_group_change = FALSE;
+			$ip_change = ($interface === $settings['interface']);
```
rbgarga: This check is reversed, `$ip_change` should be true when `$interface` differs from `$settings['interface']`.

src/etc/inc/openvpn.inc
```diff
@@ -1575,21 +1575,33 @@ function openvpn_resync_all($interface = "") {
...
+			if (file_exists($fpath)) {
```
rbgarga: if `$ip_change` is true the check for `$gw_group_change` could be skipped

src/etc/inc/openvpn.inc
```diff
@@ -1575,21 +1575,33 @@ function openvpn_resync_all($interface = "") {
...
 	}
-
 	if (is_array($config['openvpn']['openvpn-client'])) {
 		foreach ($config['openvpn']['openvpn-client'] as & $settings) {
-			if ($interface <> "" && $interface != $settings['interface']) {
-				continue;
+			$mode_id = "client" . $settings['vpnid'];
+			$fpath = "{$g['varetc_path']}/openvpn/{$mode_id}.interface";
+			$gw_group_change = FALSE;
+			$ip_change = ($interface === $settings['interface']);
```
rbgarga: Same comment for previous definition of `$ip_change`

src/etc/inc/openvpn.inc
```diff
@@ -1575,21 +1575,33 @@ function openvpn_resync_all($interface = "") {
...
+			if (file_exists($fpath)) {
```
rbgarga: Same comment for previous, it could be skipped if `$ip_change` is true

src/etc/inc/openvpn.inc
jwsi: @rbgarga I think you may be mistaken. The code prior to this change will NOT resync when `$interface != $settings['interface']` via use of a continue statement. The change above is syntactic sugar because when `$interface === $settings['interface']` a re-sync occurs via marking the `$ip_change` variable as `TRUE`.

More fundamentally, it's pretty clear that when an interface IP changes that an OpenVPN instance is bound to, a resync must occur...

src/etc/inc/openvpn.inc
jwsi: Noted, I will improve the efficiency here.

src/etc/inc/openvpn.inc
jwsi: See comment above.

src/etc/inc/openvpn.inc
jwsi: See comment above

src/etc/inc/openvpn.inc
rbgarga: actually I believe we are both mistaken here. I was re-reading the code and when $interface is not empty it means an interface was passed to the function and only tunnels configured to use that interface will be resync'd. I believe that if statement with a continue for the case where interface is not empty and differs from tunnel should be kept as it is today

rbgarga: @jwsi You are right and I need more coffee :)

src/etc/inc/openvpn.inc
jwsi: This has been improved in changeset 15f8062b42b3b2849d5dd7fdde9170d4785e84e4

PR #4073, opened 2019-06-21
Display interfaces in asc order

pauloklaus:
- [ ] No redmine Issue =[
- [ ] Ready for review


jim-p:
We've been down this road a few times in the past. The order of the interfaces is significant as they are. They cannot be alphabetized in these lists.

When it comes to things like HA it is critical that the order match exactly, and this will only compound confusion.

pauloklaus:
Hi @jim-p,

We have a large infrastructure, with many, many VLANs. The order that was created represents nothing to us. But ok, thank you for the explanation.

PR #4077, opened 2019-08-06
Adding OMAPI functionality to the DHCP Server settings. 

Erik10206:
Supports setting a port and defining or generating a key. Service is enabled when settings are defined. Validation logic added on saving to verify the port is valid and not in use as well as that a key is defined or being generated. If validation fails, the user is presented with a descriptive error message.

Feature#7304

Also updated the write_settings call on the DHCP server page. No arguments were being passed so changes came up as unknown on the config history screen. Config history will now reflect the UI section and interface where change was made.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/7304
- [x] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -566,11 +566,11 @@ EOD;
 	if (isset($dhcpifconf['omapi_port']) && is_numeric($dhcpifconf['omapi_port'])) {
 		$dhcpdconf .= <<<EOD
 
-omapi-port {$dhcpifconf['omapi_port']};
 key omapi_key {
-  algorithm hmac-md5;
+  algorithm {$dhcpifconf['omapi_key_algorithm']};
   secret "{$dhcpifconf['omapi_key']}";
 };
+omapi-port {$dhcpifconf['omapi_port']};
 omapi-key omapi_key;
 
 EOD;
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -562,6 +562,20 @@ EOD;
 		$dhcpdconf .= "always-broadcast on\n";
 	}
 
+	// OMAPI Settings
+	if (isset($dhcpifconf['omapi_port']) && is_numeric($dhcpifconf['omapi_port'])) {
+		$dhcpdconf .= <<<EOD
+
+omapi-port {$dhcpifconf['omapi_port']};
+key omapi_key {
+  algorithm HMAC-MD5;
+  secret "{$dhcpifconf['omapi_key']}";
+};
+omapi-key omapi_key;
+\n
+EOD;
+	}
+
 	$dhcpdifs = array();
 	$enable_add_routers = false;
 	$gateways_arr = return_gateways_array();
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -568,7 +568,7 @@ EOD;
 
 omapi-port {$dhcpifconf['omapi_port']};
 key omapi_key {
-  algorithm HMAC-MD5;
+  algorithm hmac-md5;
   secret "{$dhcpifconf['omapi_key']}";
 };
 omapi-key omapi_key;
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -572,7 +572,7 @@ key omapi_key {
   secret "{$dhcpifconf['omapi_key']}";
 };
 omapi-key omapi_key;
-\n
+
 EOD;
 	}
 
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1136,12 +1136,9 @@ function is_port($port) {
 
 /* returns true if $port is in use */
 function is_port_in_use($port, $proto = "tcp", $ip_version = 4) {
-	exec("/usr/bin/netstat -an{$ip_version}p {$proto} | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
+	exec("/usr/bin/netstat -an " . escapeshellarg('-' . $ip_version) . " -p " . escapeshellarg($proto) . " | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
 	
-	if (in_array($port, $port_info)) {
-		return true;
-	}
-	return false;
+	return in_array($port, $port_info);
 }
 
 /* returns true if $portrange is a valid TCP/UDP portrange ("<port>:<port>") */
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1134,6 +1134,16 @@ function is_port($port) {
 	return false;
 }
 
+/* returns true if $port is in use */
+function is_port_in_use($port, $proto = "tcp", $ip_version = 4) {
+	exec("/usr/bin/netstat -an{$ip_version}p {$proto} | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
+	
+	if (in_array($port, $port_info)) {
+		return true;
+	}
+	return false;
+}
+
 /* returns true if $portrange is a valid TCP/UDP portrange ("<port>:<port>") */
 function is_portrange($portrange) {
 	$ports = explode(":", $portrange);
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1136,7 +1136,16 @@ function is_port($port) {
 
 /* returns true if $port is in use */
 function is_port_in_use($port, $proto = "tcp", $ip_version = 4) {
-	exec("/usr/bin/netstat -an " . escapeshellarg('-' . $ip_version) . " -p " . escapeshellarg($proto) . " | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
+	$port_info = array();
+	exec("/usr/bin/netstat --libxo json -an " . escapeshellarg('-' . $ip_version) . " -p " . escapeshellarg($proto), $rawdata, $rc);
+	if ($rc == 0) {
+		$netstatarr = json_decode(implode(" ", $rawdata), JSON_OBJECT_AS_ARRAY);
+		$netstatarr = $netstatarr['statistics']['socket'];
+		
+		foreach($netstatarr as $index => $portstats){
+			array_push($port_info, $portstats['local']['port']);
+		}
+	}
 	
 	return in_array($port, $port_info);
 }
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -267,19 +267,23 @@ if (isset($_POST['save'])) {
 				$input_errors[] = gettext("OMAPI key must be at least 256 bits.");
 			}
 		} elseif($_POST['omapi_gen_key'] == "yes") {
-			//Build the command to generate and capture the key
-			$cmd = "/usr/local/sbin/dnssec-keygen -r /dev/urandom -a HMAC-MD5 -b 512 -n HOST omapi_key; "; //Generate the key
-			$cmd .= "/bin/cat Komapi_key.+*.private | /usr/bin/grep ^Key | /usr/bin/cut -d ' ' -f2-; "; //Capture the key
-			$cmd .= "/bin/rm -f Komapi_key*"; //Cleanup key files
+		    // Set the output directory
+            $output_dir = "/tmp/dnssec-keygen";
 
-			//Execute the command
+			// Build the command to generate and capture the key
+            $cmd = "/bin/mkdir {$output_dir}; "; // Create output directory
+            $cmd .= "/usr/local/sbin/dnssec-keygen -r /dev/urandom -K {$output_dir} -a HMAC-MD5 -b 512 -n HOST omapi_key; "; // Generate the key
+			$cmd .= "/bin/cat {$output_dir}/Komapi_key.+*.private | /usr/bin/grep ^Key | /usr/bin/cut -d ' ' -f2-; "; // Capture the key
+			$cmd .= "/bin/rm -Rf {$output_dir}"; // Cleanup key files
+
+			// Execute the command
 			exec($cmd, $cmd_output);
 
-			//Set the key
+			// Set the key
 			$_POST['omapi_key'] = $cmd_output[1];
 			$pconfig['omapi_key'] = $cmd_output[1];
 
-			//Uncheck the generate box
+			// Uncheck the generate box
 			unset($_POST['omapi_gen_key']);
 			unset($pconfig['omapi_gen_key']);
 		} else {
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -264,32 +264,26 @@ if (isset($_POST['save'])) {
 
 		// Generate a key if checked
 		if ($_POST['omapi_gen_key'] == "yes") {
-			// Check to see if the utility is installed
-			$cmd = "/usr/bin/which ddns-confgen";
+			// Set some vars
+			$key_bit_len = 256;
+			$key_bytes_len = $key_bit_len / 8; // 8 bits = 1 Byte
 
-			// Execute the command
-			$cmd_output = exec_command($cmd);
+			// Generate random bytes based on key length
+			$ran_bytes = openssl_random_pseudo_bytes($key_bytes_len);
 
-			if (strstr($cmd_output, "ddns-confgen")) {
-				// Generate and capture the key
-				$cmd = "/usr/local/sbin/ddns-confgen -a HMAC-MD5 -k omapi_key -q | /usr/bin/grep secret | /usr/bin/cut -d'\"' -f2;";
+			// Encode the bytes to get the key string
+			$key_str = base64_encode($ran_bytes);
 
-				// Execute the command
-				$cmd_output = exec_command($cmd);
-
-				// Set the key
-				$_POST['omapi_key'] = $cmd_output;
-				$pconfig['omapi_key'] = $cmd_output;
-			} else {
-				$input_errors[] = gettext("Key generation utility not found. Please go to System -> Package Manager and install the bind package.");
-			}
+			// Set the key
+			$_POST['omapi_key'] = $key_str;
+			$pconfig['omapi_key'] = $key_str;
 
 			// Uncheck the generate box
 			unset($_POST['omapi_gen_key']);
 			unset($pconfig['omapi_gen_key']);
 		} elseif (!empty($_POST['omapi_key'])) { // Check the key if it's not being generated
-			if (strlen($_POST['omapi_key']) >= 25) {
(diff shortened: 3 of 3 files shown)


src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +241,52 @@ function validate_partial_mac_list($maclist) {
...
+		}
+
+		// Check the key entry
+		if($_POST['omapi_key'] && $_POST['omapi_key'] != "") {
+			if (strlen($_POST['omapi_key']) < 45) {
+				$input_errors[] = gettext("OMAPI key must be at least 256 bits.");
+			}
+		} elseif($_POST['omapi_gen_key'] == "yes") {
+			//Build the command to generate and capture the key
+			$cmd = "/usr/local/sbin/dnssec-keygen -r /dev/urandom -a HMAC-MD5 -b 512 -n HOST omapi_key; "; //Generate the key
```
jim-p: These should be using an absolute path for the key files. Even if the files are cleaned up after, there is no way to be certain about which directory these are landing in, which could lead to unintended issues.

Erik10206: No problem. Added a new commit to address this

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +241,56 @@ function validate_partial_mac_list($maclist) {
...
 
+	/*
+	 * Check the OMAPI settings
+	 * - Make sure that if the port is defined, that it is valid and isn't in use
+	 * - Make sure the key length is at least 256bit (45 characters)
+	 * - Generate a new key if selected
+	 */
+	if($_POST['omapi_port'] && $_POST['omapi_port'] != "") {
+		// Check the port entry
+		if(in_array($_POST['omapi_port'], range(1, 65535))) {
```
jim-p: We have an `is_port($port)` function in `util.inc` that does a better check to tell if a submitted value is a port number, you should use it instead.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +241,56 @@ function validate_partial_mac_list($maclist) {
...
+			// Get the list of open or listening ports locally
+			exec("/usr/bin/netstat -an4p tcp | awk '{print $4}' | egrep -o '*([0-9]{1,5})$' | sort | uniq", $port_info);
```
jim-p: The `awk`, `egrep`, and `sort` commands should use full paths to binaries. You can remove `uniq` and use `sort -u` instead of just `sort`.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +241,56 @@ function validate_partial_mac_list($maclist) {
...
+		if($_POST['omapi_key'] && $_POST['omapi_key'] != "") {
+			if (strlen($_POST['omapi_key']) < 45) {
+				$input_errors[] = gettext("OMAPI key must be at least 256 bits.");
+			}
+		} elseif($_POST['omapi_gen_key'] == "yes") {
+			// Set the output directory
+			$output_dir = "/tmp/dnssec-keygen";
+
+			// Build the command to generate and capture the key
+			$cmd = "/bin/mkdir {$output_dir}; "; // Create output directory
```
jim-p: You should use `safe_mkdir($path)` here, or at least `mkdir -p`.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +241,56 @@ function validate_partial_mac_list($maclist) {
...
+		}
+
+		// Check the key entry
+		if($_POST['omapi_key'] && $_POST['omapi_key'] != "") {
+			if (strlen($_POST['omapi_key']) < 45) {
+				$input_errors[] = gettext("OMAPI key must be at least 256 bits.");
+			}
+		} elseif($_POST['omapi_gen_key'] == "yes") {
+			// Set the output directory
+			$output_dir = "/tmp/dnssec-keygen";
```
jim-p: If you are operating in `/tmp` you should really be making a random name and not a predictable path.

Erik10206:
Good looking out Jim. Commit added.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +242,63 @@ function validate_partial_mac_list($maclist) {
...
+			$rand = substr(md5(microtime()), rand(0, 26), 25);
+
+			// Set the output directory
+			$output_dir = "/tmp/{$rand}";
+
+			// Create output directory
+			mkdir($output_dir);
+
+			// Build the command to generate and capture the key
+			$cmd = "/usr/local/sbin/dnssec-keygen -r /dev/urandom -K {$output_dir} -a HMAC-MD5 -b 512 -n HOST omapi_key > /dev/null; "; // Generate the key
```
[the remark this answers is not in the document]
Erik10206: I am currently on 2.4.3. I wasn't aware that I could download it yet so I'll pull a copy of 2.5 dev to a vm and test. I will have some time tonight to work on that.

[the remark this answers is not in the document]
Erik10206: So, the issue is that the dnssec-keygen version that is on the 2.5 images has discontinued support for HMAC key generation. I am working to see if there is another utility that I can use to generate the key. The recommended once is tsig-keygen but that isn't installed.  Is installing that an option?

rbgarga:
@Erik10206 make a base functionality to depend of an additional package being installed is not a good way to go IMO.  Based on what I saw on ISC-dhcpd 4.4.x manpage it supports following algorithms:
```
       The key name, algorithm, and secret must match that being used by the
       DNS server. The DHCP server currently supports the following
       algorithms:

               HMAC-MD5
               HMAC-SHA1
               HMAC-SHA224
               HMAC-SHA256
               HMAC-SHA384
               HMAC-SHA512
```
Can't you generate a key using any of those protocols with dnssec-keygen?

Erik10206:
> @Erik10206 make a base functionality to depend of an additional package being installed is not a good way to go IMO. Based on what I saw on ISC-dhcpd 4.4.x manpage it supports the following algorithms:
> 
> ```
>        The key name, algorithm, and secret must match that being used by the
>        DNS server. The DHCP server currently supports the following
>        algorithms:
> 
>                HMAC-MD5
>                HMAC-SHA1
>                HMAC-SHA224
>                HMAC-SHA256
>                HMAC-SHA384
>                HMAC-SHA512
> ```
> 
> Can't you generate a key using any of those protocols with dnssec-keygen?

I agree that it isn't ideal but I was looking into the options and the version of dnssec-keygen that is included on 2.5 (9.14.3) has depreciated all hmac algorithms. That is when I came across ddns-confgen and while not included, can be added by simply installing the bind package with no further config needed. The package dependency only affects the auto key generation. OMAPI can be configured and setup without the need for the package when the key string is provided by the user. 

This can also be done in code as well. I looked into it a little bit but I can dig into it some more and see if I can eliminate this dependency.

Erik10206:
@rbgarga New commit has been pushed. I dug into the source for dnssec-keygen and just replicated the process in php.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +242,54 @@ function validate_partial_mac_list($maclist) {
...
+			unset($_POST['omapi_gen_key']);
+			unset($pconfig['omapi_gen_key']);
+		} elseif (!empty($_POST['omapi_key'])) { // Check the key if it's not being generated
+            if (strlen($_POST['omapi_key']) < 44) {
+				$input_errors[] = gettext("Please specify a valid OMAPI key. Key must use HMAC-MD5 and be a minimum length of 256 bits.");
+			}
+		} else {
+			$input_errors[] = gettext("A key is required when OMAPI is enabled (port specified).");
+		}
+	}
```
rbgarga: It looks much better this way. Please just fix indentation in the above block

jim-p:
You may want to look at `phpseclib` for making keys, if it has a function like this. We added phpseclib 2.0.x to pfSense 2.5.0 because OpenSSL deprecated some functions for small keys that we needed for captive portal vouchers.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +242,54 @@ function validate_partial_mac_list($maclist) {
...
+			// Set the key
+			$_POST['omapi_key'] = $key_str;
+			$pconfig['omapi_key'] = $key_str;
+
+			// Uncheck the generate box
+			unset($_POST['omapi_gen_key']);
+			unset($pconfig['omapi_gen_key']);
+		} elseif (!empty($_POST['omapi_key'])) { // Check the key if it's not being generated
+			if (strlen($_POST['omapi_key']) < 44) {
+				$input_errors[] = gettext("Please specify a valid OMAPI key. Key must use HMAC-MD5 and be a minimum length of 256 bits.");
```
jim-p: If the server supports various HMAC types up to HMAC-SHA512, why limit the user input to HMAC-MD5? Only because that's what was hardcoded above? Consider adding a GUI option letting the user select the key algorithm as well, which would be more accommodating.

src/usr/local/www/services_dhcp.php
```diff
@@ -1010,6 +1077,36 @@ function build_pooltable() {
...
+	'text',
+	$pconfig['omapi_key']
+))->setAttribute('placeholder', 'OMAPI Key')->setHelp('Set the key used to secure the connection to the OMAPI endpoint. The minimum HMAC-MD5 key length is 256 bits.');
+
+$group->add(new Form_Checkbox(
+	'omapi_gen_key',
+	'',
+	'Generate New Key',
+	$pconfig['omapi_gen_key']
+))->setHelp('Key Bits: 256<br />Algorithm: HMAC-MD5');
```
jim-p: Might not be as viable here, but take a look at the captive portal voucher page and see how this is handled there. There is a button to generate the key, rather than creating it in the background when saving. Either way is fine, though.

src/etc/inc/services.inc
```diff
@@ -562,6 +562,20 @@ EOD;
 		$dhcpdconf .= "always-broadcast on\n";
 	}
 
+	// OMAPI Settings
+	if (isset($dhcpifconf['omapi_port']) && is_numeric($dhcpifconf['omapi_port'])) {
+		$dhcpdconf .= <<<EOD
+
+omapi-port {$dhcpifconf['omapi_port']};
+key omapi_key {
+  algorithm hmac-md5;
```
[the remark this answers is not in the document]
Erik10206: This was just the default that is specified in all of the documentation. No reason not to allow others.

src/usr/local/www/services_dhcp.php
Erik10206: Yea, this was due to it being hardcoded in services. I have added another input that will be used to dynamically set the alg based on user selection

src/usr/local/www/services_dhcp.php
Erik10206: I took a look at this and I think I am going to stick with the current method.

Erik10206:
@jim-p I took a look at that library. I would prefer to stick to this key string gen workflow unless you require the changes.

src/usr/local/www/services_dhcp.php
```diff
@@ -235,6 +243,76 @@ function validate_partial_mac_list($maclist) {
...
+	 * Check the OMAPI settings
+	 * - Make sure that if the port is defined, that it is valid and isn't in use
+	 * - Make sure the key is defined and the length is appropriate for the selected algorithm
+	 * - Generate a new key if selected
+	 */
+	if (!empty($_POST['omapi_port'])) {
+		// Check the port entry
+		if (is_port($_POST['omapi_port']) && $_POST['omapi_port'] > 1024) {
+			// Get the list of open or listening ports locally
+			exec("/usr/bin/netstat -an4p tcp | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
```
jim-p: The more I think about it, this would be generally useful in its own function so other areas could use it as well. I'd like to see this check moved to `/etc/inc/util.inc` in a function called `is_port_in_use($port)` that returns true/false based on whether the passed port is already in use. Then you could use that function in the if statement here and then add to input errors based on the result.

src/usr/local/www/services_dhcp.php
```diff
@@ -639,7 +717,23 @@ function validate_partial_mac_list($maclist) {
...
+
+			unset($pconfig['omapi_port']);
+			unset($pconfig['omapi_key']);
+			unset($pconfig['omapi_key_algorithm']);
+		} else {
+			$dhcpdconf['omapi_port'] = $_POST['omapi_port'];
+			$dhcpdconf['omapi_key'] = $_POST['omapi_key'];
+			$dhcpdconf['omapi_key_algorithm'] = $_POST['omapi_key_algorithm'];
+		}
+
```
jim-p: There is an extra blank line here

src/usr/local/www/services_dhcp.php
Erik10206: Will do!

Erik10206:
Got those 2 things knocked out. Let me know if there is anything around the function call that you wanted to be changed. I saw that there were other functions that had separate v4 vs v6 definitions but it's the exact same command so that is why I chose it as an arg rather than making v6 and v4 thier own functions. Let me know if you would prefer otherwise.

src/etc/inc/util.inc
```diff
@@ -1134,6 +1134,16 @@ function is_port($port) {
 	return false;
 }
 
+/* returns true if $port is in use */
+function is_port_in_use($port, $proto = "tcp", $ip_version = 4) {
+	exec("/usr/bin/netstat -an{$ip_version}p {$proto} | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
+	
+	if (in_array($port, $port_info)) {
```
jim-p: You can simplify this a bit:  `return in_array($port, $port_info);`

src/usr/local/www/services_dhcp.php
```diff
@@ -252,11 +252,8 @@ function validate_partial_mac_list($maclist) {
 	if (!empty($_POST['omapi_port'])) {
 		// Check the port entry
 		if (is_port($_POST['omapi_port']) && $_POST['omapi_port'] > 1024) {
-			// Get the list of open or listening ports locally
-			exec("/usr/bin/netstat -an4p tcp | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
-
-			// Check to see if the port is in the list
-			if (in_array($_POST['omapi_port'], $port_info) && $_POST['omapi_port'] != $dhcpdconf['omapi_port']){
+		    // Check to see if the port is in the list
+			if (is_port_in_use($_POST['omapi_port']) && $_POST['omapi_port'] != $dhcpdconf['omapi_port']){
```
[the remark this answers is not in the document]
Erik10206: @jim-p This change changes the behavior of the 2 statements. An error when trying to save with the previously specified port. I changed this setup to a switch statement that will avoid the nested  ifs.

src/etc/inc/util.inc
```diff
@@ -1134,6 +1134,13 @@ function is_port($port) {
 	return false;
 }
 
+/* returns true if $port is in use */
+function is_port_in_use($port, $proto = "tcp", $ip_version = 4) {
+	exec("/usr/bin/netstat -an " . escapeshellarg('-' . $ip_version) . " -p " . escapeshellarg($proto) . " | /usr/bin/awk '{print $4}' | /usr/bin/egrep -o '*([0-9]{1,5})$' | /usr/bin/sort -u", $port_info);
```
rbgarga: `netstat` binary supports libxo(3).  It means it can output in json format which is much better to parse instead of running all these other commands. See the following example and remember to remove `,pretty` from the call since it's not needed here.
```
[2.4.4-RELEASE][root@pfgarga.home]/root: netstat --libxo json,pretty -an -4 -p tcp
{
  "statistics": {
    "socket": [
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 44,
        "local": {
          "address": "172.21.4.1",
          "port": "16222"
        },
        "remote": {
          "address": "172.21.4.121",
          "port": "50010"
        },
        "tcp-state": "ESTABLISHED"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "127.0.0.1",
          "port": "953"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "*",
          "port": "53"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "127.0.0.1",
          "port": "19001"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "*",
          "port": "80"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "*",
          "port": "16443"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "127.0.0.1",
          "port": "19000"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      },
      {
        "protocol": "tcp4",
        "receive-bytes-waiting": 0,
        "send-bytes-waiting": 0,
        "local": {
          "address": "*",
          "port": "16222"
        },
        "remote": {
          "address": "*",
          "port": "*"
        },
        "tcp-state": "LISTEN"
      }
    ]
  }
}
```

jim-p: Only downside in this case is that it would require a bit more memory/processing in PHP to handle and parse through the JSON data, but I don't think `netstat` output would ever be large enough for that to be a problem here.

If you want a recent example of how to parse that kind of output, see https://github.com/pfsense/pfsense/blob/master/src/etc/inc/system.inc#L2450 or https://github.com/pfsense/pfsense/blob/master/src/etc/inc/syslog.inc#L722

Erik10206:
Hey @jim-p, I looked into squash and merge and it looks like it requires write permission to do so via github. Is there another way you would recommend?

rbgarga:
@Erik10206 When "key algorithm" changes, shouldn't OMAPI key change as well? If a sha256 key was first generated then algorithm changes to MD5, can it keep using same key?

Erik10206:
@rbgarga yes, the minimum length is deciding factor. The extra characters don't add much but it doesn't hurt anything. However, if you go from MD5 -> sha256, you would be forced to gen a new key if the current one isn't long enough. One thing that I got hung up on initially is that there is really no difference in how the secret (key) is created. The alg is just the method on how the secret is used so it boils down to min char length of each alg's key requirements.

rbgarga:
I'll merge this one manually preserving you as Author just to avoid adding all these 14 commits. Thank you!

rbgarga:
Merged. Thanks!

PR #4081, opened 2019-08-20
restartallwan - pfSsh.php script to restart all wan interfaces

viktor-g-netgate:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9688
- [ ] Ready for review

Description:
This script looks for all WAN interfaces by checking which of them has gateways
and then it simply disable / enable them
useful for testing configuration

### DIFF
--- a/src/etc/phpshellsessions/restartallwan
+++ b/src/etc/phpshellsessions/restartallwan
@@ -1,7 +1,7 @@
 /*
  * restartallwans
- * 
- * This script looks for all WAN interfaces by checking whitch has gateways
+ *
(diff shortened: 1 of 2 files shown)


rbgarga:
I don't like the idea of disabling all interfaces on config.xml and re-enable them afterwards.  If the execution is interrupted in the middle it would end up leaving config broken.  Any thoughts @jim-p?

jim-p:
Agreed. If it's only restarting them, I don't see why they need to be disabled in the middle. It should only need to take actions similar to 'apply' on interfaces.php, i.e. bring it down then configure it and any dependent items.

PR #4084, opened 2019-08-27
Add more color choices for login screen

Godwottery:
I use different colors to distinguish visually between firewalls I manage. I wanted some more variation. 
I checked that chosen colors have a high contrast to the white that is used for the text-entry, so that the text is clearly visible.  

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9706
I read https://docs.netgate.com/pfsense/en/latest/development/submitting-a-pull-request-via-github.html, and I would consider this to be a "very minor typo/wording fixes" so I have not created an issue.
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -662,9 +662,11 @@ $section->addInput(new Form_Select(
 	'logincss',
 	'Login page color',
 	$pconfig['logincss'],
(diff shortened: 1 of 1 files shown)


Godwottery:
Referes to https://redmine.pfsense.org/issues/9706

PR #4085, opened 2019-09-11
Redmine #9747: Fixes malformed JSON

s-fiebig:
Fixes malformed JSON by escaping backslash in case of a domain prefix.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9747
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/widgets/widgets/ipsec.widget.php
+++ b/src/usr/local/www/widgets/widgets/ipsec.widget.php
@@ -125,39 +125,29 @@ if ($_REQUEST && $_REQUEST['ajax']) {
 	}
 
 	// Generate JSON formatted data for the widget to update from
-	$jsondata = "{";
+	$data->overview = "<tr>";
+	$data->overview .= "<td>" . $activecounter . "</td>";
+	$data->overview .= "<td>" . $inactivecounter . "</td>";
+	$data->overview .= "<td>" . (is_array($mobile['pool']) ? htmlspecialchars($mobile['pool'][0]['usage']) : '0') . "</td>";
+	$data->overview .= "</tr>";
 
-	$jsondata .= "\"overview\":\"";
-	$jsondata .= "<tr>";
-	$jsondata .= "<td>" . $activecounter . "</td>";
-	$jsondata .= "<td>" . $inactivecounter . "</td>";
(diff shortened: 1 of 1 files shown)


s-fiebig:
This would be definitely the better solution. Will have a look at that, as soon as I get the time.

jim-p:
Looks like you need to initialize `$data` before its first use or it returns an error and does not render the widget properly.

s-fiebig:
It's a warning and it did actually render the widget at my pfsense, but of course I will change it.

src/usr/local/www/widgets/widgets/ipsec.widget.php
```diff
@@ -125,58 +125,48 @@
...
-			$jsondata .= "<td>" . $ipsec['remote-subnet'] . "<br />(" . htmlspecialchars($ipsec['dest']) . ")</td>";
-			$jsondata .= "<td>" . htmlspecialchars($ipsec['descr']) . "</td>";
-
+			$data->tunnel .= "<tr>";
+			$data->tunnel .= "<td>" . htmlspecialchars($ipsec['src']) . "</td>";
+			$data->tunnel .= "<td>" . $ipsec['remote-subnet'] . "<br />(" . htmlspecialchars($ipsec['dest']) . ")</td>";
+			$data->tunnel .= "<td>" . htmlspecialchars($ipsec['descr']) . "</td>";
 			if ($ipsec['status'] == "true") {
-				$jsondata .= '<td><i class=\"fa fa-arrow-up text-success\"></i></td>';
+				$data->tunnel .= '<td><i class=\"fa fa-arrow-up text-success\"></i></td>';
```
jim-p: These double quotes no longer need escaped, they are getting escaped again which breaks the status icon.

src/usr/local/www/widgets/widgets/ipsec.widget.php
```diff
@@ -125,58 +125,48 @@
...
 			} else {
-				$jsondata .= '<td><i class=\"fa fa-arrow-down text-danger\"></i></td>';
+				$data->tunnel .= '<td><i class=\"fa fa-arrow-down text-danger\"></i></td>';
```
jim-p: These double quotes no longer need escaped, they are getting escaped again which breaks the status icon.

s-fiebig:
Thanks for pointing this out! This was totally under my radar, because my current setup has no tunnels.

PR #4087, opened 2019-09-16
diag_packet_capture.php: allow to input multiple tcp/udp ports

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9766
- [x] Ready for review

Ability to use space-separated list of TCP/UDP ports in diag_packet_capture.php

it uses count() for loop instead of foreach & end() to allow work correct if entered same port number, e.g. "22 80 80 3389"

### DIFF
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -518,7 +518,6 @@ if ($do_tcpdump) :
 		$matches[] = hostport_array_fixer($hosts,'host');
 	}
 
-
 	if ($count != "0") {
 		$searchcount = "-c " . $count;
 	} else {
@@ -533,8 +532,6 @@ if ($do_tcpdump) :
 		print_info_box(gettext('Packet capture is running.'), 'info');
 
 		$cmd = "/usr/sbin/tcpdump -i {$selectedif} {$disablepromiscuous} {$searchcount} -s {$snaplen} -w {$fp}{$fn} " . escapeshellarg($matchstr);
-		// Debug
-		// echo $cmd;
 		mwexec_bg ($cmd);
 	} else {
 ?>
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -30,16 +30,16 @@
 
 $allowautocomplete = true;
 
-function fixup_host_logic($value) {
+function fixup_logic($value) {
 	return str_replace(array(" ", ",", "+", "|", "!"), array("", "and ", "and ", "or ", "not "), $value);
 }
 
-function strip_host_logic($value) {
+function strip_logic($value) {
 	return str_replace(array(" ", ",", "+", "|", "!"), array("", "", "", "", ""), $value);
 }
 
-function get_host_boolean($value, $host) {
-	$value = str_replace(array("!", $host), array("", ""), $value);
+function get_boolean($value, $entry) {
+	$value = str_replace(array("!", $entry), array("", ""), $value);
 	$andor = "";
 	switch (trim($value)) {
 		case "|":
@@ -67,9 +67,9 @@ function strip_not($value) {
 }
 
 function fixup_host($value, $position) {
-	$host = strip_host_logic($value);
+	$host = strip_logic($value);
 	$not = has_not($value) ? "not " : "";
-	$andor = ($position > 0) ? get_host_boolean($value, $host) : "";
+	$andor = ($position > 0) ? get_boolean($value, $host) : "";
 	if (is_ipaddr($host)) {
 		return "{$andor}host {$not}" . $host;
 	} elseif (is_subnet($host)) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_packet_capture.php
```diff
@@ -200,8 +200,11 @@ function fixup_host($value, $position) {
 	}
 
 	if ($port != "") {
-		if (!is_port(strip_not($port))) {
-			$input_errors[] = gettext("Invalid value specified for port.");
+		$ports = explode(" ",$port);
+		foreach($ports as &$port) {
```
jim-p: Needs a space between `foreach` and `(`.

src/usr/local/www/diag_packet_capture.php
```diff
@@ -479,7 +482,14 @@ function fixup_host($value, $position) {
 	}
 
 	if ($port != "") {
-		$matches[] = "port ".fixup_not($port);
+		for($i = 0; $i < count($ports); $i++){
```
jim-p: Needs a space between `for` and `(`.

src/usr/local/www/diag_packet_capture.php
```diff
@@ -479,7 +482,14 @@ function fixup_host($value, $position) {
...
+                    	 if($i == count($ports)-1) {
```
jim-p: Needs a space between `if` and `(`, between the `)` and `-`, and between `-` and `1`).

src/usr/local/www/diag_packet_capture.php
```diff
@@ -479,7 +482,14 @@ function fixup_host($value, $position) {
...
+                	                $ports_matches .= "port ".($ports[$i]);
```
jim-p: This lost the `fixup_not()` used in the previous code, which will make negating ports impossible. Perhaps that's why there are parenthesis where they appear to be unnecessary?

Also needs spaces around the `.`.

src/usr/local/www/diag_packet_capture.php
```diff
@@ -200,8 +200,11 @@ function fixup_host($value, $position) {
 	}
 
 	if ($port != "") {
-		if (!is_port(strip_not($port))) {
-			$input_errors[] = gettext("Invalid value specified for port.");
+		$ports = explode(" ",$port);
```
jim-p: Using space is inconsistent with the other fields on the page. The **Host Address** field supports multiple values and has different characters to denote logic (`,` is `and`, `|` is `or`). This feature should match that. Someone might want to match both source and destination port which would be `and` here.

src/usr/local/www/diag_packet_capture.php
```diff
@@ -479,7 +482,14 @@ function fixup_host($value, $position) {
 	}
 
 	if ($port != "") {
-		$matches[] = "port ".fixup_not($port);
+		for($i = 0; $i < count($ports); $i++){
+                    	 if($i == count($ports)-1) {
+                	                $ports_matches .= "port ".($ports[$i]);
+               	         } else {
+                	                $ports_matches .= "port ".($ports[$i]). " or ";
```
jim-p: Same as above, appears to be missing `fixup_not`, and needs spaces around the `.`(s).

vktg: now same boolean logic as for hosts - 
and, or, not

debug output sample:
/usr/sbin/tcpdump -i vtnet1 -p -c 100 -s 0 -w /root/packetcapture.cap 'ip and tcp and ( port 22 and port 80 and port not 100 or port not 50) and ( host 1.1.1.1 and host 8.8.8.8)'

vktg:
New code, same logic as for hosts:
, | ! 
and or not

src/usr/local/www/diag_packet_capture.php
```diff
@@ -101,6 +101,13 @@ function fixup_host($value, $position) {
 	}
 }
 
+function fixup_port($value, $position) {
```
jim-p: There is a lot of repeated code here that could easily be avoided. Make the `fixup_host()` function generic with an extra parameter to set the type (`host` or `port`) and adjust tests/output as needed to avoid all the copy/pasted code with small changes.

vktg: all repeated code consolidated to generic functions:
fixup_hostport
precheck_hostport
hostport_array_fixer

src/usr/local/www/diag_packet_capture.php
```diff
@@ -479,26 +511,14 @@ function fixup_host($value, $position) {
...
-			}
-		}
-
-		if (!empty($hostmatch)) {
-			$matches[] = "({$hostmatch})";
-		}
+		$matches[] = hostport_array_fixer($hosts,'host');
 	}
 
+
```
rbgarga: Remove extra blank line

src/usr/local/www/diag_packet_capture.php
```diff
@@ -514,7 +534,7 @@ function fixup_host($value, $position) {
 
 		$cmd = "/usr/sbin/tcpdump -i {$selectedif} {$disablepromiscuous} {$searchcount} -s {$snaplen} -w {$fp}{$fn} " . escapeshellarg($matchstr);
 		// Debug
-		//echo $cmd;
+		// echo $cmd;
```
rbgarga: While there, could you please remove this whole commented out block?

src/usr/local/www/diag_packet_capture.php

src/usr/local/www/diag_packet_capture.php
vktg: removed

PR #4089, opened 2019-09-19
add ability to reroot and reboot with fsck to WebGUI

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9771
- [x] Ready for review

Description:
Adds extra buttons for Reroot and Reboot with fsck to Diagnostics / Reboot
code mostly taken from /etc/rc.initial.reboot to diag_reboot.php

### DIFF
--- a/src/usr/local/www/diag_reboot.php
+++ b/src/usr/local/www/diag_reboot.php
@@ -50,25 +50,23 @@ if (($_SERVER['REQUEST_METHOD'] == 'POST') && (empty($_POST['override']) ||
 	if (DEBUG) {
 		print_info_box(gettext("Not actually rebooting (DEBUG is set true)."), 'success');
 	} else {
+		print('<div><pre>');
 		switch ($_POST['rebootmode']) {
 			case 'FSCKReboot':
 				mwexec('/sbin/nextboot -e "pfsense.fsck.force=5"');
-				print('<div><pre>');
 				system_reboot();
-				print('</pre></div>');
 				break;
 			case 'Reroot':
 				if (!is_module_loaded("zfs.ko")) {
-					print('<div><pre>');
 					system_reboot_sync(true);
-					print('</pre></div>');
 				}
 				break;
 			case 'Reboot':
-				print('<div><pre>');
 				system_reboot();
-				print('</pre></div>');
+				break;
+			default:
 		}
+		print('</pre></div>');
 	}
 ?>
 
@@ -128,10 +126,6 @@ if (!is_module_loaded("zfs.ko")) {
 $help .= '.';
 
 $section = new Form_Section('Select reboot method');
-
-foreach (auth_get_authserver_list() as $key => $auth_server) {
-        $serverlist[$key] = $auth_server['name'];
-}
 
 $section->addInput(new Form_Select(
         'rebootmode',
--- a/src/usr/local/www/diag_reboot.php
+++ b/src/usr/local/www/diag_reboot.php
@@ -49,6 +49,17 @@ if (($_SERVER['REQUEST_METHOD'] == 'POST') && (empty($_POST['override']) ||
     ($_POST['override'] != "yes"))):
 	if (DEBUG) {
 		print_info_box(gettext("Not actually rebooting (DEBUG is set true)."), 'success');
+	} elseif ($_POST['Submit'] == 'FSCK Reboot') {
+		mwexec('/sbin/nextboot -e "pfsense.fsck.force=5"');
+		print('<div><pre>');
+		system_reboot();
+		print('</pre></div>');
+	} elseif ($_POST['Submit'] == 'Reroot') {
+		if (!is_module_loaded("zfs.ko")) {
+			print('<div><pre>');
+			system_reboot_sync(true);
+			print('</pre></div>');
+		}
 	} else {
 		print('<div><pre>');
 		system_reboot();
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_reboot.php
```diff
@@ -49,6 +49,17 @@
     ($_POST['override'] != "yes"))):
 	if (DEBUG) {
 		print_info_box(gettext("Not actually rebooting (DEBUG is set true)."), 'success');
+	} elseif ($_POST['Submit'] == 'FSCK Reboot') {
```
jim-p: Just use one keyword here, rather than a word with a space.

src/usr/local/www/diag_reboot.php
```diff
@@ -110,12 +121,33 @@ function startCountdown() {
...
+					<i class="fa fa-refresh"></i>
+					<?=gettext("FSCK Reboot")?>
+				</button>
+<?php
+if (!is_module_loaded("zfs.ko")) {
+				echo '<button type="submit" class="btn btn-danger pull-center" name="Submit" value="Reroot" title="Reroot (Stop processes, remount disks, re-run startup sequence)">
+					<i class="fa fa-refresh"></i>
+					Reroot
+				</button>';
+//        echo "    " . gettext("R/r: Reroot (Stop processes, remount disks, re-run startup sequence)") . "\n";
```
jim-p: No need to leave this commented out line here

src/usr/local/www/diag_reboot.php
```diff
@@ -110,12 +121,33 @@ function startCountdown() {
...
+?>
+</p>
 			<form action="diag_reboot.php" method="post">
 				<button type="submit" class="btn btn-danger pull-center" name="Submit" value="<?=gettext("Reboot")?>" title="<?=gettext("Reboot the system")?>">
 					<i class="fa fa-refresh"></i>
 					<?=gettext("Reboot")?>
 				</button>
+				<button type="submit" class="btn btn-danger pull-center" name="Submit" value="<?=gettext("FSCK Reboot")?>" title="<?=gettext("Reboot and run a filesystem check")?>">
+					<i class="fa fa-refresh"></i>
+					<?=gettext("FSCK Reboot")?>
```
jim-p: Rather than multiple buttons, consider changing to a drop-down selection list where the user can choose the type (e.g. Normal reboot, Reboot with Filesystem Check, Reroot).

I would avoid using the term `fsck` here, but "Filesystem Check". The short term is at best likely to confuse some users.

src/usr/local/www/diag_reboot.php
```diff
@@ -110,12 +121,33 @@ function startCountdown() {
 	</div>
 	<div class="panel-body">
 		<div class="content">
-			<p><?=gettext('Click "Reboot" to reboot the system immediately, or "Cancel" to go to the system dashboard without rebooting. (There will be a brief delay before the dashboard appears.)')?></p>
+			<p>
+<?php 
+if (!is_module_loaded("zfs.ko")) {
+	print('Click "Reboot" to reboot the system immediately, "FSCK Reboot" to reboot and run filesystem check, "Reroot" to stop processes, remount disks and re-run startup sequence or "Cancel" to go to the system dashboard without rebooting. (There will be a brief delay before the dashboard appears.)');
+	} else {
+	print('Click "Reboot" to reboot the system immediately, "FSCK Reboot" to reboot and run filesystem check or "Cancel" to go to the system dashboard without rebooting. (There will be a brief delay before the dashboard appears.)');
```
jim-p: There is a lot of shared/common/repeated text here that could be simplified. For example, if zfs is loaded, just add the reroot part at the end. It's also a bit crammed in there all run together.

src/usr/local/www/diag_reboot.php
```diff
@@ -49,6 +49,17 @@
     ($_POST['override'] != "yes"))):
 	if (DEBUG) {
 		print_info_box(gettext("Not actually rebooting (DEBUG is set true)."), 'success');
+	} elseif ($_POST['Submit'] == 'FSCK Reboot') {
+		mwexec('/sbin/nextboot -e "pfsense.fsck.force=5"');
+		print('<div><pre>');
```
jim-p: There is quite a lot of repeated code here in each section that could be simplified.

A switch statement should be used instead of a long chain of if/elseif/else

src/usr/local/www/diag_reboot.php
vktg: fixed

src/usr/local/www/diag_reboot.php
vktg: fixed

vktg:
all issues fixed

src/usr/local/www/diag_reboot.php
```diff
@@ -50,11 +50,26 @@
 	if (DEBUG) {
 		print_info_box(gettext("Not actually rebooting (DEBUG is set true)."), 'success');
 	} else {
-		print('<div><pre>');
-		system_reboot();
-		print('</pre></div>');
+		switch ($_POST['rebootmode']) {
+			case 'FSCKReboot':
+				mwexec('/sbin/nextboot -e "pfsense.fsck.force=5"');
+				print('<div><pre>');
```
jim-p: This is still repeated too much. It's used in every case. Put the opening `print` before the `switch` and the closing `print` after the `switch`.

src/usr/local/www/diag_reboot.php
```diff
@@ -102,28 +117,40 @@ function startCountdown() {
...
+$modeslist = ['Reboot' => 'Normal reboot', 'FSCKReboot' => 'Reboot with Filesystem Check'];
+if (!is_module_loaded("zfs.ko")) {
+	$help .= ' or "Reroot" to stop processes, remount disks and re-run startup sequence';
+	$modeslist += ['Reroot' => 'Reroot'];
+	} 
+$help .= '.';
+
+$section = new Form_Section('Select reboot method');
+
+foreach (auth_get_authserver_list() as $key => $auth_server) {
```
jim-p: Why is this here? It doesn't appear to be used for anything.

vktg: removed

src/usr/local/www/diag_reboot.php
```diff
@@ -50,11 +50,26 @@
...
+					print('<div><pre>');
+					system_reboot_sync(true);
+					print('</pre></div>');
+				}
+				break;
+			case 'Reboot':
+				print('<div><pre>');
+				system_reboot();
+				print('</pre></div>');
+		}
```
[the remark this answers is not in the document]
vktg: added

src/usr/local/www/diag_reboot.php
vktg: fixed

vktg:
on ARM platform I found this checks in /etc/rc.initial.reboot:
```
$platform = system_identify_specific_platform();
$no_options = array('SG-1100', 'ROGUE-1', 'uFW');
if (!in_array($platform['name'], $no_options)) {
        echo "    " . gettext("S: Reboot into Single User Mode (requires console access!)") . "\n";
        echo "    " . gettext("F: Reboot and run a filesystem check") . "\n";
}
```

Do I need to add these checks here?

jim-p:
Yes, though that kind of ARM-specific change usually only goes on our factory repository.

PR #4094, opened 2019-09-26
prefer-ip6 option added to advanced resolver options

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9797
- [x] Ready for review

from man unbound.conf(5):

```
prefer-ip6: <yes or no>
              If enabled, prefer IPv6 transport for sending DNS queries to internet nameservers. Default is no.
```
this patch will add option to Advanced Resolver Options page


jim-p:
See discussion on the Redmine issue. Seems simple enough but I don't know that we want to keep adding GUI knobs for every single Unbound directive. I'm not seeing this option being used by a significant number of users, and the very rare edge cases that need it could easily add the option in the advanced options box.

vktg:
Not so necessary 
Can be entered in the 'custom options' field

PR #4096, opened 2019-10-02
CVE-2019-16701

flipchan:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-16701
is this cve really a vulnerability ? seems like someone has cve'ed a bad feature

Can someone verify that this CVE is actually a vulnerability?


jim-p:
The line is correct. If the userid is set AND the user is not 0 (admin) AND that user does not have privileges to access the page, then they are denied. That CVE is invalid, it's not a bug, and it's an intentional (not bad) feature. If you grant someone access to run commands and code, don't be surprised when they can, in fact, run commands and code.

flipchan:
thanks for confirming this

PR #4097, opened 2019-10-03
Filter.inc undefined variables - dead openvpn alias code

phil-davis:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9806
- [x] Ready for review

PHPstorm was telling me:
1) `$newaddress` should be set to empty before being used
2) A reference to `$openvpn` array is undefined. It should be `$openvpncfg`

But this code seems to be about "OpenVPN user aliases", and they seem to no longer be an alias type that can be defined (or never were)

Did this exist in the past?
Is there any way to cause this code to be executed on current systems?
Or is it dead code?

See answer below - it is dead code.

So I have removed the whole block of code related to alias type "openvpn".

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -816,28 +816,6 @@ function filter_generate_aliases() {
 
 				$aliases .= "{$aliased['name']} = \"<{$aliased['name']}>\"\n";
 				break;
-			case "openvpn":
-				$openvpncfg = array();
-				if ($config['openvpn']['user']) {
-					/* XXX: Check if we have a correct ip? */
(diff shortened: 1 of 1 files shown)


rbgarga:
@phil-davis after some discussion and log reading we understood it's dead code.  Last change to this feature was in 2011 and said it was unfinished.

If you want to remove it, go ahead. Thanks!!

phil-davis:
I removed the whole block of code, and squashed. Top ost has been updated a bit to reflect what was done. I looked elsewhere in the code base and could not find any other code that seemed to be about `$config['openvpn']['user']` ...

PR #4098, opened 2019-10-04
auto-remove empty zombie fw aliases

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9790
- [x] Ready for review

auto-remove empty zombie fw aliases

If you create Alias table under Firewall / Aliases / IP with FQDNs,
PF table with such name stays in system after you delete alias.
it just became empty after "deleting".

if works as garbage collector, checking which aliases presents in config and removes other

original issue:

If you create Alias table under Firewall / Aliases / IP with FQDNs,
PF table with such name stays in system after you delete alias.
it just became empty after "deleting".

you can see that empty table on Diagnostics / Tables or with pfctl -sT
it's not deleted if you do "Filter reload".

it seems there is needed "pfctl -t <alias_name> -T kill" command

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -492,9 +492,7 @@ function filter_configure_sync($delete_states_if_needed = true) {
 	// get aliases names from pfctl -sT
 	exec("/sbin/pfctl -sT", $pftables);
 	// compare config and pfctl aliases
-	$diff = array_diff($pftables, $aliasesnames);
-	// don't delete special aliases
-	$diff = array_diff($diff, $reserved_table_names);
+	$diff = array_diff($pftables, $aliasesnames, $reserved_table_names);
 	// delete unused aliases
 	foreach ($diff as $alias) {
 		$_grbg = exec("/sbin/pfctl -t " . escapeshellarg($alias) . " -T kill 2>/dev/null");
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -256,6 +256,7 @@ function filter_configure_sync($delete_states_if_needed = true) {
 	}
 	update_filter_reload_status(gettext("Creating aliases"));
 	$aliases = filter_generate_aliases();
+
 	$gateways = filter_generate_gateways();
 	if (platform_booting() == true) {
 		echo ".";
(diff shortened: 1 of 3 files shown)


rbgarga:
IMO it would make more sense to delete pf table together when alias is deleted

src/etc/inc/filter.inc
```diff
@@ -483,6 +483,23 @@ function filter_configure_sync($delete_states_if_needed = true) {
...
+	$aliasesnames = array();
+	foreach ($config['aliases']['alias'] as $alias) {
+		array_push($aliasesnames, $alias['name']);
+	}
+	// get aliases names from pfctl -sT
+	$pftables = explode("\n",shell_exec("/sbin/pfctl -sT"));
+	// compare config and pfctl aliases
+	$diff = array_diff($pftables, $aliasesnames);
+	// don't delete special aliases
+	$diff_exceptions = array('bogons','bogonsv6','snort2c','sshguard','negate_networks','tonatsubnets','virusprot','vpn_networks');
```
jim-p: We already have this list as a global `$reserved_table_names` in `/usr/local/www/guiconfig.inc` -- the list could be relocated to `/etc/inc/globals.inc` so it can be used here as well. No need to duplicate it.

src/etc/inc/filter.inc
```diff
@@ -483,6 +483,23 @@ function filter_configure_sync($delete_states_if_needed = true) {
 		$_grbg = exec("/sbin/pfctl -t bogonsv6 -T kill 2>/dev/null");
 	}
 
+	// get aliases names from config
+	$aliasesnames = array();
+	foreach ($config['aliases']['alias'] as $alias) {
+		array_push($aliasesnames, $alias['name']);
+	}
+	// get aliases names from pfctl -sT
+	$pftables = explode("\n",shell_exec("/sbin/pfctl -sT"));
```
jim-p: Use `exec("/sbin/pfctl -sT", $pftables);` instead of this.  It yields an array in one step without the need to manually run `explode()`.

src/etc/inc/filter.inc
```diff
@@ -483,6 +483,23 @@ function filter_configure_sync($delete_states_if_needed = true) {
...
+	// compare config and pfctl aliases
+	$diff = array_diff($pftables, $aliasesnames);
+	// don't delete special aliases
+	$diff_exceptions = array('bogons','bogonsv6','snort2c','sshguard','negate_networks','tonatsubnets','virusprot','vpn_networks');
+	$diff = array_diff($diff, $diff_exceptions);
+	// delete unused aliases
+	foreach ($diff as $alias) {
+		$_grbg = exec("/sbin/pfctl -t " . $alias . " -T kill 2>/dev/null");
```
jim-p: Needs `escapeshellarg()` around `$alias`.

src/etc/inc/filter.inc
vktg: fixed

src/etc/inc/filter.inc
vktg: fixed

src/etc/inc/filter.inc
vktg: fixed

src/etc/inc/filter.inc
```diff
@@ -483,6 +483,23 @@ function filter_configure_sync($delete_states_if_needed = true) {
...
+	$aliasesnames = array();
+	foreach ($config['aliases']['alias'] as $alias) {
+		array_push($aliasesnames, $alias['name']);
+	}
+	// get aliases names from pfctl -sT
+	exec("/sbin/pfctl -sT", $pftables);
+	// compare config and pfctl aliases
+	$diff = array_diff($pftables, $aliasesnames);
+	// don't delete special aliases
+	$diff = array_diff($diff, $reserved_table_names);
```
rbgarga: Prototype of array_diff `array_diff(array $array1 , array $array2 [, array $... ] )` allows to get many arrays, it will compare `$array1` with other arrays and will identify items present on `$array1` which are not present on any other array.  It seems to be the case here, so you could reduce it to a single call to `array_diff($pftables, $aliasesnames, $reserved_table_names);`

vktg: done

PR #4101, opened 2019-10-17
Pcapstart

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9831
- [ ] Ready for review

Prints packet capture start time in extra field

### DIFF
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -524,7 +524,7 @@ if ($do_tcpdump) :
 		print_info_box(gettext('Packet capture is running'), 'info');
 
 		$cmd = "/usr/sbin/tcpdump -i {$selectedif} {$disablepromiscuous} {$searchcount} -s {$snaplen} -w {$fp}{$fn} " . escapeshellarg($matchstr);
-		mwexec ("touch /root/packetcapture.start");
+		touch("/root/packetcapture.start");
 		// Debug
 		//echo $cmd;
 		mwexec_bg ($cmd);
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -443,6 +443,9 @@ if (($action == gettext("Stop") or $action == "") and $processisrunning != true)
 		null,
 		'fa-stop-circle'
 	))->addClass('btn-warning');
+	if ($action == gettext("Start")) {
+                touch("/root/packetcapture.start");
+        }
 	if (file_exists($fp.$fns)) {
 		$section->addInput(new Form_StaticText(
 			'Last capture start',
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_packet_capture.php
```diff
@@ -510,9 +521,10 @@ function fixup_host($value, $position) {
 	if ($action == gettext("Start")) {
 		$matchstr = implode($matches, " and ");
 
-		print_info_box(gettext('Packet capture is running.'), 'info');
+		print_info_box(gettext('Packet capture is running'), 'info');
 
 		$cmd = "/usr/sbin/tcpdump -i {$selectedif} {$disablepromiscuous} {$searchcount} -s {$snaplen} -w {$fp}{$fn} " . escapeshellarg($matchstr);
+		mwexec ("touch /root/packetcapture.start");
```
jim-p: Use `touch()` here rather than a shell exec.

vktg: touch() fixed

src/usr/local/www/diag_packet_capture.php
```diff
@@ -442,6 +443,10 @@ function fixup_host($value, $position) {
 		null,
 		'fa-stop-circle'
 	))->addClass('btn-warning');
+	$section->addInput(new Form_StaticText(
```
jim-p: This should be wrapped in a `file_exists()` test. When trying this out, I received an error:

```
Warning: filemtime(): stat failed for /root/packetcapture.start in /usr/local/www/diag_packet_capture.php on line 480 
```

This will never display after clicking `Start` because the file is touched after this block of code is run. It does show if the page is refreshed after.

src/usr/local/www/diag_packet_capture.php
```diff
@@ -459,8 +464,14 @@ function fixup_host($value, $position) {
 		'fa-download'
 	))->addClass('btn-primary');
 
+	if (file_exists($fp.$fns) {
```
jim-p: Missing a closing paren (`)`) here

src/usr/local/www/diag_packet_capture.php

src/usr/local/www/diag_packet_capture.php
```diff
@@ -442,6 +443,12 @@ function fixup_host($value, $position) {
 		null,
 		'fa-stop-circle'
 	))->addClass('btn-warning');
+	if (file_exists($fp.$fns)) {
```
jim-p: There is still a problem here where immediately after clicking **Start**, this will display the wrong timestamp. It will either be missing if the file did not exist, or it will be the time of the previous capture, not the capture just started.

The `touch()` performed down lower in the diff needs to happen before this bit of code, not after. But the entire block of code below should not be moved, so perhaps a simple test before this point to check for the start action and run the `touch()` might work.

tested with/without timestamp file, start/stop
now works correctly

PR #4102, opened 2019-10-19
Allow import cert without private key. Issue #9834

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9834
- [x] Ready for review

Adds ability to import certificate without private key
This is necessary when VPN clients use the PKCS#11 token for authentication, and they cannot export the private key.

Imported certificates allow to:
- Add them to CRL
- Shows up these users in OpenVPN Client Export Utility

### DIFF
--- a/src/usr/local/www/system_certmanager.php
+++ b/src/usr/local/www/system_certmanager.php
@@ -246,16 +246,15 @@ if ($_POST['save'] == gettext("Save")) {
 		case 'edit':
 		case 'import':
 			$reqdfields = explode(" ",
-				"descr cert key");
+				"descr cert");
 			$reqdfieldsn = array(
 				gettext("Descriptive name"),
-				gettext("Certificate data"),
-				gettext("Key data"));
+				gettext("Certificate data"));
 			if ($_POST['cert'] && (!strstr($_POST['cert'], "BEGIN CERTIFICATE") || !strstr($_POST['cert'], "END CERTIFICATE"))) {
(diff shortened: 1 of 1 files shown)


rbgarga:
@vktg do you still plan to work on that?

vktg:
For now cert_build_list() hide certs without a private key by default ( see https://github.com/pfsense/pfsense/pull/4132 )
Can this PR be approved now?

vktg:
done
such note:
"In some cases, you can skip this field, for example, if your private key is on the PKCS#11 token and you only need to import and sign its certificate."

src/usr/local/www/system_certmanager.php
```diff
@@ -794,9 +793,9 @@ function list_csrs() {
 
 	$section->addInput(new Form_Textarea(
 		'key',
-		'*Private key data',
+		'Private key data',
 		$pconfig['key']
-	))->setHelp('Paste a private key in X.509 PEM format here.');
+	))->setHelp('Paste a private key in X.509 PEM format here. In some cases, you can skip this field, for example, if your private key is on the PKCS#11 token and you only need to import and sign its certificate.');
```
jim-p: Reword to avoid addressing the user directly and over-explaining. For example:

> This field may remain empty in certain cases, such as when the private key is stored on a PKCS#11 token.

vktg: done

PR #4103, opened 2019-10-19
ECDSA private key parser for sign request

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9745
- [ ] Ready for review

adds ability to parse EC private key on Add/Sign a New Certificate page
Like:

-----BEGIN EC PRIVATE KEY-----
xxxxxx
-----END EC PRIVATE KEY-----

### DIFF
--- a/src/usr/local/www/system_certmanager.php
+++ b/src/usr/local/www/system_certmanager.php
@@ -216,11 +216,13 @@ if ($_POST['save']) {
 				$input_errors[] = gettext("This signing request does not appear to be valid.");
 			}
 
-			if ( (($_POST['csrtosign'] === "new") && (strlen($_POST['keypaste']) > 0)) && (!strstr($_POST['keypaste'], "BEGIN PRIVATE KEY") || !strstr($_POST['keypaste'], "END PRIVATE KEY"))) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_certmanager.php
```diff
@@ -216,11 +216,13 @@
 				$input_errors[] = gettext("This signing request does not appear to be valid.");
 			}
 
-			if ( (($_POST['csrtosign'] === "new") && (strlen($_POST['keypaste']) > 0)) && (!strstr($_POST['keypaste'], "BEGIN PRIVATE KEY") || !strstr($_POST['keypaste'], "END PRIVATE KEY"))) {
+			if ( (($_POST['csrtosign'] === "new") && (strlen($_POST['keypaste']) > 0)) && 
+					((!strstr($_POST['keypaste'], "BEGIN PRIVATE KEY") && !strstr($_POST['keypaste'], "BEGIN EC PRIVATE KEY")) ||
```
jim-p: These indentations should only be four spaces under the `if` line.

vktg: fixed

PR #4105, opened 2019-10-23
arm checks for diag_reboot.php gui

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9771
- [ ] Ready for review

additions for https://github.com/pfsense/pfsense/pull/4089
ARM checks in GUI

on ARM platform I found this checks in /etc/rc.initial.reboot:
```
$platform = system_identify_specific_platform();
$no_options = array('SG-1100', 'ROGUE-1', 'uFW');
if (!in_array($platform['name'], $no_options)) {
        echo "    " . gettext("S: Reboot into Single User Mode (requires console access!)") . "\n";
        echo "    " . gettext("F: Reboot and run a filesystem check") . "\n";
}
```


jim-p:
Need to resolve conflicts. The main part of this was already merged, so all this PR needs to contain is the new change.

vktg:
done

vktg:
why we need it?
I found that "Reboot with filesystem check" works fine on SG-1000..

src/usr/local/www/diag_reboot.php
```diff
@@ -43,6 +43,8 @@
 $guiretry = 20;		// Seconds to try again if $guitimeout was not long enough
 
 $pgtitle = array(gettext("Diagnostics"), gettext("Reboot"));
+$platform = system_identify_specific_platform();
+$no_options = array('SG-1100', 'ROGUE-1', 'uFW');
```
jim-p: This isn't very future-proof, though it is a similar method to that used by `/etc/rc.initial.reboot`. For example it's already out of date since it's missing upcoming models listed in the factory repository. It is probably better to check if the value of `get_single_sysctl('hw.machine')` or `php_uname('m')` is `arm`, and then in the future if some models work we can whitelist the ones known to work.

vktg: done

src/usr/local/www/diag_reboot.php
```diff
@@ -53,8 +54,10 @@
 		print('<div><pre>');
 		switch ($_POST['rebootmode']) {
 			case 'FSCKReboot':
-				mwexec('/sbin/nextboot -e "pfsense.fsck.force=5"');
-				system_reboot();
+				if (get_single_sysctl('hw.machine') != 'arm') {
```
rbgarga: `php_uname('m')` is a better options here since it's less expensive

src/usr/local/www/diag_reboot.php
```diff
@@ -117,8 +120,12 @@ function startCountdown() {
 
 $form = new Form(false);
 
-$help = 'Click "Normal reboot" to reboot the system immediately, "Reboot with Filessystem Check" to reboot and run filesystem check';
-$modeslist = ['Reboot' => 'Normal reboot', 'FSCKReboot' => 'Reboot with Filesystem Check'];
+$help = 'Select "Normal reboot" to reboot the system immediately';
+$modeslist = ['Reboot' => 'Normal reboot'];
+if (get_single_sysctl('hw.machine') != 'arm') {
```
rbgarga: php_uname('m') is a better options here since it's less expensive

vktg: ok, done

PR #4107, opened 2019-10-30
Add wait-interval to diag_ping.php

Godwottery:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9862
- [x] Ready for review

The diagnostic page for ping does not have support for waiting for a different amount of time between pings. I added this function as I wanted to use it. I tested on my 2.4.x version, and it works for me.

### DIFF
--- a/src/usr/local/www/diag_ping.php
+++ b/src/usr/local/www/diag_ping.php
@@ -39,9 +39,12 @@ require_once("guiconfig.inc");
 
 define('MAX_COUNT', 10);
 define('DEFAULT_COUNT', 3);
+define('MAX_WAIT', 10);
+define('DEFAULT_COUNT', 1);
 $do_ping = false;
 $host = '';
 $count = DEFAULT_COUNT;
+$wait = DEFAULT_WAIT;
 
 if ($_POST || $_REQUEST['host']) {
 	unset($input_errors);
@@ -55,7 +58,11 @@ if ($_POST || $_REQUEST['host']) {
 	if (($_REQUEST['count'] < 1) || ($_REQUEST['count'] > MAX_COUNT)) {
 		$input_errors[] = sprintf(gettext("Count must be between 1 and %s"), MAX_COUNT);
 	}
-
+	
+	if (($_REQUEST['wait'] < 1) || ($_REQUEST['wait'] > MAX_WAIT)) {
+		$input_errors[] = sprintf(gettext("Wait must be between 1 and %s"), MAX_WAIT);
+	}
+	
 	$host = trim($_REQUEST['host']);
 	$ipproto = $_REQUEST['ipproto'];
 	if (($ipproto == "ipv4") && is_ipaddrv6($host)) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_ping.php
```diff
@@ -157,6 +168,13 @@
 	array_combine(range(1, MAX_COUNT), range(1, MAX_COUNT))
 ))->setHelp('Select the maximum number of pings.');
 
+$section->addInput(new Form_Select(
+	'wait',
+	'Number of seconds to wait between pings',
+	$wait,
+	array_combine(range(1, MAX_WAIT), range(1, MAX_WAIT))
+))->setHelp('Select the number of seconds to wait between pings.');
+
```
sbeaver-netgate: There seems no point in the label being the same as the help text. Perhaps simplify the label to "Seconds between pings" of just eliminate the help text ?

Godwottery: Changed. I left the help text so that they are consistent. In my opinion consistency trumps removing redundancy in such cases.

src/usr/local/www/diag_ping.php
```diff
@@ -55,7 +58,11 @@
 	if (($_REQUEST['count'] < 1) || ($_REQUEST['count'] > MAX_COUNT)) {
 		$input_errors[] = sprintf(gettext("Count must be between 1 and %s"), MAX_COUNT);
 	}
-
+	
+	if (($_REQUEST['wait'] < 1) || ($_REQUEST['wait'] > MAX_WAIT)) {
```
jim-p: You should also test to make sure it's an integer, e.g. `is_numericint($_REQUEST['wait'])`

src/usr/local/www/diag_ping.php
```diff
@@ -55,7 +58,11 @@
...
+		$input_errors[] = sprintf(gettext("Wait must be between 1 and %s"), MAX_WAIT);
+	}
+
```
jim-p: Remove this line containing only whitespace

src/usr/local/www/diag_ping.php
```diff
@@ -76,6 +83,10 @@
 		if (preg_match('/[^0-9]/', $count)) {
 			$count = DEFAULT_COUNT;
 		}
+		$wait = $_REQUEST['wait'];
+		if (preg_match('/[^0-9]/', $wait)) {
```
jim-p: If your input validation above is better, this `preg_match` is unnecessary. Even if it does match the existing test above it, it's better to improve the code than to match the existing format. 

This whole statement could be one line:
```
$wait = (empty($_REQUEST['wait'])) ? DEFAULT_WAIT : $_REQUEST['wait'];
```

src/usr/local/www/diag_ping.php
```diff
@@ -55,7 +58,11 @@
 	if (($_REQUEST['count'] < 1) || ($_REQUEST['count'] > MAX_COUNT)) {
 		$input_errors[] = sprintf(gettext("Count must be between 1 and %s"), MAX_COUNT);
 	}
-
+
```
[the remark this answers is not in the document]
Godwottery: Done

PR #4110, opened 2019-11-03
Remove Zabbix 2.2 Packages

dbaio:
End of life was August, 2019.
Ports will expire after November, 30 on FreeBSD.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9880
- [x] Ready for review


rbgarga:
Already removed without using this patch. Sorry.

PR #4111, opened 2019-11-11
Reverting changes to RDNSS and DNSSL settings in RADVD

elbinteh:
Reverting this change as I think this requires more thought as per comments in #9302.

Revert "RADVD: In "managed" or "stateless_dhcp" mode, don't use default values for DNS servers etc (these should come from DHCPv6)"

This reverts commit dcc887a355aae49c7df0c29752c04e12922aca83.

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9302
- [ ] Ready for review


elbinteh:
No longer required as this has been reverted in master

PR #4112, opened 2019-11-14
add poly1305-chacha20 to nginx cipher list

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9896
- [x] Ready for review

add poly1305-chacha20 to the nginx cipher list of WebGUI 
as part of NGE

https://tools.ietf.org/html/rfc7905


vktg:
poly1305-chacha20 at the end of the list.

PR #4113, opened 2019-11-14
show ECDSA certs only with prime256v1 and secp384r1 curves

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9897
- [ ] Ready for review

 Only prime256v1 and secp384r1 will work from our list with TLS v1.3.


jim-p:
I'm already working on a more general solution here to cover both this and IPsec

PR #4114, opened 2019-11-16
packet capture ospf/ospfv3 support

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9905
- [x] Ready for review

Adds the ability to select OSPF in the protocol field
It can capture OSPF, OSPFv3 or both, depending of Address Family
Very useful for FRR troubleshooting

### DIFF
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -514,16 +514,15 @@ if ($do_tcpdump) :
 	}
 
 	if (in_array($proto, $protos)) {
-		if ($proto == 'carp') {
-			$matches[] = fixup_not(str_replace('carp', 'proto 112', $proto));
-		} else {
-			if ($fam == 'ip') {
-				$matches[] = fixup_not(str_replace('ospf', 'ip[9] == 89', $proto));
-			} else if ($fam == 'ip6') {
-				$matches[] = fixup_not(str_replace('ospf', 'proto 0x59', $proto));
-			} else {
+		switch ($proto) {
+			case 'carp':
+				$matches[] = fixup_not(str_replace('carp', 'proto 112', $proto));
+				break;
+			case 'ospf':
 				$matches[] = fixup_not(str_replace('ospf', 'proto ospf', $proto));
-			}
+				break;
+			default:
+				break;
 		}
 	}
 
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -521,12 +521,14 @@ if ($do_tcpdump) :
 		switch ($proto) {
 			case 'ospf':
 				$proto = str_replace('ospf', 'proto ospf', $proto);
+				break;
 			case 'carp':
 				$proto = str_replace('carp', 'proto 112', $proto);
+				break;
 			default:
-				$matches[] = fixup_not($proto);
 				break;
 		}
+		$matches[] = fixup_not($proto);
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_packet_capture.php
```diff
@@ -512,7 +514,17 @@ function hostport_array_fixer($items,$type) {
 	}
 
 	if (in_array($proto, $protos)) {
-		$matches[] = fixup_not(str_replace('carp', 'proto 112', $proto));
+		if ($proto == 'carp') {
+			$matches[] = fixup_not(str_replace('carp', 'proto 112', $proto));
+		} else {
+			if ($fam == 'ip') {
+				$matches[] = fixup_not(str_replace('ospf', 'ip[9] == 89', $proto));
```
jim-p: Why is this necessary? Running with `proto ospf` catches OSPF and OSPFv3 (output shortened):
```
: tcpdump -i vmx0 ip6 and proto ospf
08:28:09.061963 IP6 fe80::20d:b9ff:fe33:f71 > ff02::5: OSPFv3, Hello, length 52
08:28:09.062241 IP6 fe80::20c:29ff:fef8:3bdc > ff02::5: OSPFv3, Hello, length 52
08:28:09.062920 IP6 fe80::20c:29ff:fe45:256 > ff02::5: OSPFv3, Hello, length 52

: tcpdump -i ovpns4 ip and proto ospf
08:28:17.229238 IP 10.4.103.2 > ospf-all.mcast.net: OSPFv2, Hello, length 48
```

src/usr/local/www/diag_packet_capture.php
```diff
@@ -512,7 +514,17 @@ function hostport_array_fixer($items,$type) {
 	}
 
 	if (in_array($proto, $protos)) {
-		$matches[] = fixup_not(str_replace('carp', 'proto 112', $proto));
+		if ($proto == 'carp') {
```
jim-p: I'm not fond of this structure. It looks like it's going to get even uglier if anything else is added. If it is necessary, it should be turned into a switch based on the contents of `$proto` with the default action being to not do a `str_replace()`. Or perhaps the switch only alters the contents of `$proto` and then after that block, gets `fuxup_not` done and added to `$matches` just once at the end, rather than repeating it so much.

src/usr/local/www/diag_packet_capture.php
vktg: right, it's unnecessary
fixed

src/usr/local/www/diag_packet_capture.php

extend fixup_not() for !ospf, 
because 'tcpdump -i vmx0 ip and ! ospf' is incorrect, 
should be  'tcpdump -i vmx0 ip and ! proto ospf'

src/usr/local/www/diag_packet_capture.php
```diff
@@ -512,7 +518,15 @@ function hostport_array_fixer($items,$type) {
 	}
 
 	if (in_array($proto, $protos)) {
-		$matches[] = fixup_not(str_replace('carp', 'proto 112', $proto));
+		switch ($proto) {
```
jim-p: I think I'd rather see this have a `break` in each case and then add to `$matches` outside the `switch`. This way works, but is a bit awkward to follow and unnecessarily runs the extra lower steps. Even if they are not matching, it's still doing more than it needs to.

vktg: done

PR #4115, opened 2019-11-16
do not show incompatible ECDSA certs for DNS Resolver

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9907
- [ ] Ready for review

Do not show incompatible ECDSA certs for DNS Resolver
It is difficult to find EC curves supported by each DNS implementation, so
It’s better to use the “safe list” of curves anyway

same as https://redmine.pfsense.org/issues/9897

p.s. 1.1.1.1 and 9.9.9.9 uses secp256r1

### DIFF
--- a/src/usr/local/www/services_unbound.php
+++ b/src/usr/local/www/services_unbound.php
@@ -343,7 +343,7 @@ if ($certs_available) {
 		'sslcertref',
 		'SSL/TLS Certificate',
 		$pconfig['sslcertref'],
-		cert_build_list('cert', 'HTTPS')
+		cert_build_list('cert', 'IPsec')
 	))->setHelp('The server certificate to use for SSL/TLS service. The CA chain will be determined automatically.');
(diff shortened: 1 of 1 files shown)


vktg:
did some tests between two pfSense -
if you try to use non IPsec-safe certs, you got error in resolver logs:
`error: ssl handshake failed crypto error:1417A0C1:SSL routines:tls_post_process_client_hello:no shared cipher `

in case of IPsec-safe certs:
`error: ssl handshake failed crypto error:14094418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca`
- which is ok, because pfSense do not copy CA certs to system ca certs storage

jim-p:
> which is ok, because pfSense do not copy CA certs to system ca certs storage (another issue?)

It can if you configure it to, now: https://redmine.pfsense.org/issues/4068

PR #4116, opened 2019-11-16
Add option to disallow unauthenticated LDAP binds

Augustin-FL:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9909
- [x] Ready for review

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1371,6 +1371,11 @@ function ldap_backed($username, $passwd, $authcfg, &$attributes = array()) {
 		return false;
 	}
 
+	if (!isset($authcfg['ldap_allow_unauthenticated']) && $passwd == '') {
+		$attributes['error_message'] = gettext("Invalid credentials.");
+		return false;
+	}
+
 	if (!function_exists("ldap_connect")) {
 		log_error(gettext("ERROR! unable to find ldap_connect() function."));
 		$attributes['error_message'] = gettext("Internal error during authentication.");
@@ -1959,7 +1964,7 @@ function session_auth() {
 
 	/* Validate incoming login request */
 	$attributes = array('nas_identifier' => 'webConfigurator-' . gethostname());
-	if (isset($_POST['login']) && !empty($_POST['usernamefld']) && !empty($_POST['passwordfld'])) {
+	if (isset($_POST['login']) && !empty($_POST['usernamefld'])) {
 		$authcfg = auth_get_authserver($config['system']['webgui']['authmode']);
 		$remoteauth = authenticate_user($_POST['usernamefld'], $_POST['passwordfld'], $authcfg, $attributes);
 		if ($remoteauth || authenticate_user($_POST['usernamefld'], $_POST['passwordfld'])) {
--- a/src/etc/inc/ipsec.auth-user.php
+++ b/src/etc/inc/ipsec.auth-user.php
@@ -51,7 +51,7 @@ if (isset($_GET['username'])) {
 	$authmodes = explode(",", getenv("authcfg"));
 }
 
-if (!$username || !$password) {
+if (!$username) {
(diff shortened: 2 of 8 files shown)


src/etc/inc/auth.inc
```diff
@@ -1371,6 +1371,11 @@ function ldap_backed($username, $passwd, $authcfg, &$attributes = array()) {
 		return false;
 	}
 
+	if (!isset($authcfg['ldap_allow_unauthenticated']) && $passwd == '') {
```
jim-p: Does this also affect someone submitting a password containing only whitespace? It seems like `empty($passwd)` might be a better choice here.

src/etc/inc/auth.inc
```diff
@@ -1959,7 +1964,7 @@ function session_auth() {
 
 	/* Validate incoming login request */
 	$attributes = array('nas_identifier' => 'webConfigurator-' . gethostname());
-	if (isset($_POST['login']) && !empty($_POST['usernamefld']) && !empty($_POST['passwordfld'])) {
+	if (isset($_POST['login']) && !empty($_POST['usernamefld'])) {
```
jim-p: Why change this here? It would alter existing behavior for non-LDAP logins unrelated to this change.

src/etc/inc/ipsec.auth-user.php
```diff
@@ -51,7 +51,7 @@
 	$authmodes = explode(",", getenv("authcfg"));
 }
 
-if (!$username || !$password) {
```
jim-p: Why change this here? It would alter existing behavior for non-LDAP logins unrelated to this change.

src/etc/inc/openvpn.auth-user.php
```diff
@@ -53,7 +53,7 @@
 	$common_name = getenv("common_name");
 }
 
-if (!$username || !$password) {
```
jim-p: Why change this here? It would alter existing behavior for non-LDAP logins unrelated to this change.

src/usr/local/www/diag_authentication.php
```diff
@@ -40,7 +40,7 @@
 		$input_errors[] =  sprintf(gettext('%s is not a valid authentication server'), $_POST['authmode']);
 	}
 
-	if (empty($_POST['username']) || empty($_POST['password'])) {
```
jim-p: Why change this here? It would alter existing behavior for non-LDAP logins unrelated to this change.

src/etc/inc/auth.inc
Augustin-FL: unauthenticated LDAP binds only refers to binds with an empty string....So this line is fine. (I tested using a Windows AD to make sure).

src/etc/inc/auth.inc
Augustin-FL: Hi,
The idea is remove password checks from here and let *_backed() functions handle password controls. 

I understand that this is a change in a critical part of the code however....Please let me know if you would like me to revert.

src/etc/inc/ipsec.auth-user.php
Augustin-FL: Hi,
The idea is remove password checks from here and let *_backed() functions handle password controls. 

Please let me know if you would like me to revert.

src/etc/inc/openvpn.auth-user.php
Augustin-FL: Hi,
The idea is remove password checks from here and let *_backed() functions handle password controls. 

Please let me know if you would like me to revert.

src/usr/local/www/diag_authentication.php
Augustin-FL: Hi,

Yes, but this is okay. this page is only used for debugging/diags. I believe it is better for this page to give an auth result which matches the return value of *_backed().

PR #4117, opened 2019-11-23
Fix widget refresh

NobleKangaroo:
## Description:

This pull request resolves issues with the widget refresh system:
- The refresh system uses flawed logic, see below for an in-depth explanation.
- Widgets, despite using ajax callbacks, are not truly async - one widget can lock another if it takes too long to run.
- Session timeouts result in the dashboard continuing to attempt ajax callbacks, despite access being blocked due to the aforementioned session timeout.

**History:** This pull request is a simplified rework of #3879 (from my old Github account), but does not not default to a 1-second refresh interval for the Firewall Logs and Gateways widgets; in fact, it doesn't directly modify those widgets at all.

### /usr/local/www/index.php
- Widget locking:
    - Removed global mutex, preventing one long-running widget from hanging up others and ensuring true asynchronous operation.
    - Added per-widget lock, allowing a widget to execute only if it's not already executing.

- Widget refreshing:
  - Renamed function `executewidget()` to `executewidgets()` for the sake of grammar.
  - Changed execution to a `setInterval()` instead of a self-calling `setTimeout()`, as it is cleaner this way; allows for `clearInterval()` upon session timeout, if desired.
  - Fixed refresh logic, as it is logically flawed in its current implementation:
    - The function `executewidget()` is called every second.
    - `freq` is defined as the desired number of seconds between each execution of the `ajaxidx`th widget.
    - If `ajaxcntr % freq` is equal to 0, the `ajaxidx`th widget will run.
    - At the end of `executewidget()`, `ajaxidx` is increased.
      - If `ajaxidx` is greater than the number of widgets, it is reset to 0 and `ajaxcntr` is also increased. If `ajaxcntr` is greater than 4096, is it reset to 0.  
  - The end result is that each call of `executewidget()` takes `n` seconds to run, where `n` is the number of widgets on the dashboard, and each widget will only execute every `t` times `executewidget()` is run, resulting in an *actual* update time of `n * t`.
  - Example, for further illustration:
    - Suppose that there are **8 widgets** present on the pfSense dashboard, and the Firewall Logs widget is configured to have an "Update interval" of **10 seconds**.
    - A single call of `executewidget()` takes **8 seconds** (1 second per widget * 8 widgets total), and Firewall Logs only refreshes every **10 times** `updatewidget()` is called.
    - Thus Firewall Logs only refreshes every 8 * 10 seconds, or **80 seconds**. This is a bug, and incorrect behavior.
    - This has been fixed through the use of a single `for` loop, each time `executewidget()` is called, that "passes over" all widgets and decides on a per-widget basis if they should be executed (based on `ajaxcntr`).
  - In short, this change allows all widgets to operate independently of each other.

- Session timeouts:
  - Don't try to execute ajax callbacks if session timed out
  - Removed `ajaxtimeout` variable as it was not referenced

## Checklist:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/9917
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -567,8 +567,6 @@ function set_widget_checkbox_events(checkbox_panel_ref, all_none_button_id) {
 // ---------------------Centralized widget refresh system -------------------------------------------
 // These need to live outside of the events.push() function to enable the widgets to see them
 var ajaxspecs = new Array();	// Array to hold widget refresh specifications (objects )
-var ajaxidx = 0;
-var ajaxmutex = false;
 var ajaxcntr = 0;
 
 // Add a widget refresh object to the array list
@@ -634,10 +632,9 @@ events.push(function() {
 	});
 
 	// --------------------- Centralized widget refresh system ------------------------------
-	ajaxtimeout = false;
-
 	function make_ajax_call(wd) {
-		ajaxmutex = true;
+		// Add a lock on the widget to prevent overlapping callbacks
+		wd.locked = true;
 
 		$.ajax({
 			type: 'POST',
@@ -648,52 +645,48 @@ events.push(function() {
 			success: function(data){
 				if (data.length > 0 ) {
 					// If the session has timed out, display a pop-up
-					if (data.indexOf("SESSION_TIMEOUT") === -1) {
-						wd.callback(data);
-					} else {
-						if (ajaxtimeout === false) {
-							ajaxtimeout = true;
-							alert("<?=$timeoutmessage?>");
-						}
+					if (data.indexOf("SESSION_TIMEOUT") > -1) {
+						alert("<?=$timeoutmessage?>");
+						return;
 					}
-				}
 
-				ajaxmutex = false;
+					// Else call the callback
+					wd.callback(data);
+				}
 			},
 
 			error: function(e){
 //				alert("Error: " + e);
-				ajaxmutex = false;
+			},
+
+			complete: function(){
+				// Regardless of success or error, remove the widget's lock
+				wd.locked = false;
 			}
 		});
 	}
 
 	// Loop through each AJAX widget refresh object, make the AJAX call and pass the
 	// results back to the widget's callback function
-	function executewidget() {
-		if (ajaxspecs.length > 0) {
-			var freq = ajaxspecs[ajaxidx].freq;	// widget can specify it should be called freq times around the loop
-
-			if (!ajaxmutex) {
-				if (((ajaxcntr % freq) === 0) && (typeof ajaxspecs[ajaxidx].callback === "function" )) {
-				    make_ajax_call(ajaxspecs[ajaxidx]);
-				}
-
-			    if (++ajaxidx >= ajaxspecs.length) {
-					ajaxidx = 0;
-
-					if (++ajaxcntr >= 4096) {
-						ajaxcntr = 0;
-					}
-			    }
+	function executewidgets() {
+		for (var ajaxidx = 0; ajaxidx < ajaxspecs.length; ajaxidx++) {
+			// If the widget is locked, continue
+			if (ajaxspecs[ajaxidx].locked) {
+				continue;
 			}
 
-		    setTimeout(function() { executewidget(); }, 1000);
-	  	}
+			var freq = ajaxspecs[ajaxidx].freq;	// widget can specify it should be called every freq seconds
+			if (((ajaxcntr % freq) === 0) && (typeof ajaxspecs[ajaxidx].callback === "function" )) {
+				make_ajax_call(ajaxspecs[ajaxidx]);
+			}
+		}
+		ajaxcntr++;
 	}
 
 	// Kick it off
-	executewidget();
+	if (ajaxspecs.length > 0) {
+		setInterval(executewidgets, 1000);
+	}
 
 	//----------------------------------------------------------------------------------------------------
 });


NobleKangaroo:
Hi Steve,

I don't follow - the proposed PR only affects widget refresh speed, nothing
from the frontend (such as what happens after the widget refreshes). If
there's an issue with how the NTP widget works, perhaps that's simply
another issue? I tried replicating this by adding the NTP widget to my
dashboard however I don't see it "flash annoyingly". Any chance you could
perhaps record it happening?

Again, this PR was only to fix the issue where widgets don't refresh as
often as they should - in short, the widget refresh system, in its current
state, is flawed such that instead of each widget refreshing when its
defined time (say, 5 seconds) elapses, it actually takes that long,
multiplied by the number of widgets that use that system. For example: In
the case that you have 7 widgets and you want the Firewall Logs widget to
refresh every 5 seconds, each time the widget refresh code runs, it only
refreshes one of the widgets (the index is determined inside the nested
for... loop), and doesn't give any of the other widgets a chance to
execute. So the end result is that the Firewall Logs widgets, although
being configured to execute every 5 seconds, in actuality only refreshes
every 35 seconds. If you turn on developer tools in Chrome or Firefox or
whatever browser you use, you'll see what I mean.

NobleKangaroo:
> This PR seems OK, however, as currently implemented the code changes somewhat worsen the user experience. Now that the timing of each widget is handled individually screen updates can occur much faster. Try the NTP widget for example. The server time value now updates much more frequently causing the value to flash annoyingly.
> 
> Most of the widget refresh intervals seem to be about five times too fast. Please adjust these so that they are more inline with the current code.

I'm sorry but I don't follow - this PR only affects widget refresh speed, nothing from the frontend (such as what actually happens after the widget refreshes). If there's an issue with how the NTP widget works, perhaps in the behavior of the Javascript that is called upon refresh, that's simply another issue that is now visible since widgets are refreshing as often as they're supposed to be. I tried replicating this by adding the NTP widget to my dashboard however I don't see it "flash annoyingly".

I'm more than willing to take the time to fix anything directly broken by this PR, but in the end this PR was only to fix the issue where widgets don't refresh as often as they should; if you turn on developer tools in Chrome or Firefox or whatever browser you use, you'll see what I mean. It doesn't change _how_ the NTP widget (or any other widget, for that matter) actually works, only _when_.

All of that said - I am not personally experiencing what you mentioned in my own configuration. Any chance you could record it happening? I'm thinking that this may require paying some more attention to how the other code (such as the NTP widget, for example) interacts with the user; there shouldn't be any erratic behavior caused just by this PR. If there is flashing, that could possibly be an artifact of bad Javascript.

sbeaver-netgate:
Call me overly sensitive, but please don't refer to the existing, functional code as "flawed". Your PR provides some potential improvements to the logic and I would be happy to accept those as long as the user experience is not compromised.

When we evaluate PRs, we test them against a huge test suite. Bare metal of all sizes and configurations, different virtualization environments including AWS, Azure, VMWare etc. and dozens of browsers and OS combinations. We are often able to provide feedback to the submitter demonstrating that some issues exist in certain environments. Of course it is not reasonable to expect the submitter to have the same resources. I should have specified the particular situation in which I saw the "flashing" - sorry - but I'm glad you found it.

Before there was a centralized refresh system, we had a widget free-for-all. Each widget did its own thing and there were some hard to reproduce bugs when they would step on each other and do bad things. Providing a central system enabled us to get some sort of predictability in place. The widgets refreshed one at a time and on an approximate schedule. Not much more than "Some widgets refresh more often that others", but at least they weren't stepping on each other or all happening at once. Any changes to the code need to at least maintain the this capability. We don't want all the widgets refreshing at the same time, and we need to allow some to refresh more often that others. 

When I apply this PR I notice that on the first couple of test systems the refresh rate is faster than before by a factor of about 5. (I had NTP, RSS and IPSec widgets enabled with a single System Information instance) Also enabling the developer tools in the browser and looking at the network tab, I can see that this PR causes all of the widgets to refresh at exactly the same time. That is not desirable.

So to be acceptable, this PR really needs two parts:
1. Randomize the AJAX calls by widget
2. Adjust the refresh rates of all widgets so that it approximates the previous refresh rate
3. Accompany it with a second PR to address the NTP widget empty <div> thing

NobleKangaroo:
> Call me overly sensitive, but please don't refer to the existing, functional code as "flawed". Your PR provides some potential improvements to the logic and I would be happy to accept those as long as the user experience is not compromised.
>
> When we evaluate PRs, we test them against a huge test suite. Bare metal of all sizes and configurations, different virtualization environments including AWS, Azure, VMWare etc. and dozens of browsers and OS combinations. We are often able to provide feedback to the submitter demonstrating that some issues exist in certain environments. Of course it is not reasonable to expect the submitter to have the same resources. I should have specified the particular situation in which I saw the "flashing" - sorry - but I'm glad you found it.
> 
> Before there was a centralized refresh system, we had a widget free-for-all. Each widget did its own thing and there were some hard to reproduce bugs when they would step on each other and do bad things. Providing a central system enabled us to get some sort of predictability in place. The widgets refreshed one at a time and on an approximate schedule. Not much more than "Some widgets refresh more often that others", but at least they weren't stepping on each other or all happening at once. Any changes to the code need to at least maintain the this capability. We don't want all the widgets refreshing at the same time, and we need to allow some to refresh more often that others.

I apologize for my use of the term 'flawed', that was an overstatement and I should've chosen a better word or simply stated the facts without reverting to what could be taken as hostile remarks. I didn't realize that it was intentional that the refresh system doesn't refresh in a predictable way, and assumed (incorrectly) that it was an oversight or a result of poor coding. It's now abundantly clear that the reasons you mentioned were more important to the user experience. I understand that there are folks out there that install pfSense on less-powerful machines, and that the lack of resources on some hardware presents challenges when devising solutions.

> When I apply this PR I notice that on the first couple of test systems the refresh rate is faster than before by a factor of about 5. (I had NTP, RSS and IPSec widgets enabled with a single System Information instance) Also enabling the developer tools in the browser and looking at the network tab, I can see that this PR causes all of the widgets to refresh at exactly the same time. That is not desirable.

Understood. Unfortunately I don't know of a great way to implement this, given the following constraints exist:
- Issues arise from two or more widgets refresh synchronously
- The user may in fact want faster refreshes, have pfSense running on hardware that can handle any and all system load resulting from faster refreshes, and not experience any issue with two or more widgets refreshing at a particular time interval

That said, I'll keep the history of everything in mind when working on a solution, and try to avoid that. :)

> So to be acceptable, this PR really needs two parts:
> 
> 1. Randomize the AJAX calls by widget
> 2. Adjust the refresh rates of all widgets so that it approximates the previous refresh rate

I believe one solution to that would be to add a per-widget refresh offset, with the index of the widget determining its offset, which should prevent other widgets from running at the same time. For example:

Widget # | Freq (ms) | Offset (ms)
-- | -- | --
1 | 5000 | 0
2 | 3000 | 100
3 | 10000 | 200
...

What are your thoughts on something like that? The way I envision it, it would be pretty easy to implement, and that would allow both:
- The widgets to run at a specific interval (e.g. widget # 2 to run every 3000 ms)
- The widgets to not run at the same time (e.g. widget #2 would start 100ms AFTER widget #1, but only if they both need to run, etc)

If you're on board with that, I can work on that and submit updated code to this PR. Give me a heads up.

> 3. Accompany it with a second PR to address the NTP widget empty  thing

Agreed. Shouldn't require a lot of effort to resolve this on its own, but would be best suited in its own separate, tracked, issue and PR.

Thanks for the reply!

sbeaver-netgate:
Your approach seems quite reasonable. Another potentially more simple, though perhaps less likely to win an award for algorithmic elegance would be to apply some small random number of mS to the period. For example if the refresh rate is specified as once second, we would actually set the timer to 1009 mS (or 9993, or 1021 or ?) rather than 1000. No one will notice, but the AJAX calls will drift apart as time goes by.

Steve
--

NobleKangaroo:
> Another potentially more simple, though perhaps less likely to win an award for algorithmic elegance
Gave me a good laugh this morning. Thank you for that.

> would be to apply some small random number of mS to the period. For example if the refresh rate is specified as once second, we would actually set the timer to 1009 mS (or 9993, or 1021 or ?) rather than 1000. No one will notice, but the AJAX calls will drift apart as time goes by.
So long as the drift in each period doesn't compound, I'm on board with this. I wouldn't want a widget to start at refreshing 1000ms, then after some time end up refreshing every 10000ms given enough time. Your approach of allowing the widget to potentially refresh earlier than expected would be acceptable, however I would think that given the current code (`refreshWidgets` being called every 1000ms then deciding what to do at that point) it would need to be rewritten to allow for the "ahead of time" execution of widgets. Which is why I proposed a simple "let them run at their normal times, but delay them slightly" approach. We could shorten the `setInterval` to 500ms for example, but would have to delay further (and call many individual `setTimeout` within) if we were to make exceptions for each widget.

Side thought - what is your opinion on shortening `refreshWidgets()` to something like 50ms and allowing it to keep track of when to run the widgets? That would cause the function to run 20 times per second, however in doing so we then have more granularity in execution.

rbgarga:
Closing due to lack of activity for over 6 months

PR #4118, opened 2019-11-23
check user certificates for correct ECDSA curves

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9918
- [x] Ready for review

Show only correct (IPsec = OpenVPN) ECDSA when adding existing certificates to users, 
'Choose an Existing Certificate' on System \ Certificate Manager \ Certificates \ Edit page


jim-p:
I don't think we should limit this here. When creating/assigning the certs, it's really up to the admin. We don't know what they are using them for necessarily.

PR #4121, opened 2019-11-27
Redmine #9931: Change default ntp server to 2.pfsense.pool.ntp.org

imcdona:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9931
- [x] Ready for review


rbgarga:
@imcdona What is the reason for this change?

jim-p:
I didn't see the PR and had already made the change after testing it out locally, it will show up soon.

PR #4122, opened 2019-11-28
renew cert/ca with ECDSA

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9842
- [x] Ready for review

Allow to "reuse key" with any elliptic curve
and use prime256v1 for Strict Security

### DIFF
--- a/src/etc/inc/certs.inc
+++ b/src/etc/inc/certs.inc
@@ -1655,7 +1655,7 @@ function cert_renew(& $cert, $reusekey = true, $strictsecurity = false) {
 		    ($private_key_bits < $cert_strict_values['min_private_key_bits'])) {
 			$private_key_bits = $cert_strict_values['min_private_key_bits'];
 			$reusekey = false;
-		} else if (!in_array($key_details['ec']['curve_name'], $cert_curve_compatible['IPsec'])) {
+		} else if (!in_array($key_details['ec']['curve_name'], $curve_compatible_list)) {
 			$ec_curve = $cert_strict_values['ec_curve'];
 			$reusekey = false;
 		}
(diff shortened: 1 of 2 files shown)


src/etc/inc/certs.inc
```diff
@@ -1652,13 +1644,29 @@ function cert_renew(& $cert, $reusekey = true, $strictsecurity = false) {
...
-	    (!empty($key_details['ec']['curve_name']))) {
+	/* If we are enforcing strict security, then ensure the private key size
+	 * is at least 2048 bits or NIST P-256 elliptic curve*/
+	$private_key_bits = $key_details['bits'];
+	if ($strictsecurity) {
+		if (($key_details['type'] == OPENSSL_KEYTYPE_RSA) &&
+		    ($private_key_bits < $cert_strict_values['min_private_key_bits'])) {
+			$private_key_bits = $cert_strict_values['min_private_key_bits'];
+			$reusekey = false;
+		} else if (!in_array($key_details['ec']['curve_name'], $cert_curve_compatible['IPsec'])) {
```
jim-p: Rather than arbitrarily choosing `IPsec` as the list here, we can check it against all known compatible values using something like `array_unique(call_user_func_array('array_merge', $cert_curve_compatible))` to get the full list of compatible curves for any purpose.

At the moment the lists may be similar but that may not always be the case in the future.

src/usr/local/www/system_certmanager_renew.php
```diff
@@ -249,6 +249,14 @@
 					<td><?=$cert_strict_values['min_private_key_bits']?></td>
 					<td><?=($key_details['bits'] < $cert_strict_values['min_private_key_bits']) ? gettext('Yes') : gettext('No') ?></td>
 				</tr>
+<?php endif; ?>
+<?php if ($key_details['type'] == OPENSSL_KEYTYPE_EC): ?>
+				<tr>
+					<td><?=gettext("Elliptic Curve Name")?></td>
+					<td><?=htmlspecialchars($key_details['ec']['curve_name'])?></td>
+					<td><?=$cert_strict_values['ec_curve']?></td>
+					<td><?=(!in_array($key_details['ec']['curve_name'], $cert_curve_compatible['IPsec'])) ? gettext('Yes') : gettext('No') ?></td>
```
jim-p: See above about not using `IPsec` specifically for this purpose.

src/etc/inc/certs.inc
vktg: done

PR #4123, opened 2019-12-01
support variable value length in queue stats parser

lucasheld:
Hello,
currently the queue stats parser in the file "/etc/inc/shaper.inc" assumes that the bytes value does not exceed the length of 10. If this limit is exceed the page "/status_queues.php" displays wrong values. After a filter reload the bytes value is reset and the values are correct for a few minutes until the bytes value is too large again.

Decisive line from "/sbin/pfctl -s queue -v" output
Displays correct:
`'  [ pkts:    6653183  bytes: 9993145948  dropped pkts:      0 bytes:      0 ]'`

Displays incorrect:
`'  [ pkts:    6661877  bytes: 10006205865  dropped pkts:      0 bytes:      0 ]'`

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9938
- [x] Ready for review

### DIFF
--- a/src/etc/inc/shaper.inc
+++ b/src/etc/inc/shaper.inc
@@ -284,10 +284,26 @@ function get_queue_stats() {
 			$interfacestats[$currentqueueif][$currentqueuename] = &$newqueue;
 		} else {
 			if ($items[3] == "pkts:") {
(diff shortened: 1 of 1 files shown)


lucasheld:
Sure, I thought for speed reasons string operations should be used. I have adapted it to the suggested regular expression.

PR #4127, opened 2019-12-06
mark certificates with lifetime > 825 days

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9825
- [ ] Ready for review

mark certificates with lifetime > 825 days with asterisk *


jim-p:
This seems unnecessary.

PR #4129, opened 2019-12-13
Add options to disable pushing IPv6 DNS servers, fixes localized queries

luckman212:
**TL;DR— adds options that fix split-horizon DNS for dnsmasq + IPv6 setups.**

Out of the box, a pfSense install with working IPv6 will not be able to use dnsmasq to serve unique IPs (split-horizon DNS) via the [localise-queries](https://www.freebsd.org/cgi/man.cgi?query=dnsmasq) option, since that feature supports IPv4 only:

> **--localise-queries**
>     Return answers to DNS queries [..] which depend on the interface over which the query was received [..] If a name has more than one address associated with it [..] return only the address(es) on that subnet [..] **Currently this facility is limited to IPv4.**

Without the patch, LAN clients will typically submit DNS queries to pfSense's IPv6 address, which results in ALL records being returned instead of just the localised ones. This is unexpected and undesired. This adds the following options to the **DHCPv6 Server & RA** pages, allowing the options that push DNS servers via dhcpd + radvd to be disabled independently:

image
image

This fixes the issue.

Putting something like this in the Custom options section of DNS Forwarder will activate this feature:
```
localise-queries
host-record=xyzzy.foo.lan,xyzzy,192.168.1.234
host-record=xyzzy.foo.lan,xyzzy,10.20.30.254
```
or you can use an external file:
```
localise-queries
addn-hosts=/usr/local/etc/dnsmasq-addn-hosts
```
Somewhat related discussion:
https://redmine.pfsense.org/issues/9302

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9302
- [x] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -249,7 +249,7 @@ function services_radvd_configure($blacklist = array()) {
 		$radvdconf .= "\t};\n";
 
 		/* add DNS servers */
-		if (($dhcpv6ifconf['enable-radvd-dns']) != 'disabled') {
+		if (($dhcpv6ifconf['radvd-dns']) != 'disabled') {
 			$dnslist = array();
 			if (isset($dhcpv6ifconf['rasamednsasdhcp6']) && is_array($dhcpv6ifconf['dnsserver']) && !empty($dhcpv6ifconf['dnsserver'])) {
 				foreach ($dhcpv6ifconf['dnsserver'] as $server) {
@@ -371,7 +371,7 @@ function services_radvd_configure($blacklist = array()) {
 		$radvdconf .= "\t};\n";
 
 		/* add DNS servers */
-		if (($dhcpv6ifconf['enable-radvd-dns']) != 'disabled') {
+		if (($dhcpv6ifconf['radvd-dns']) != 'disabled') {
 			$dnslist = array();
 			if (isset($config['dnsmasq']['enable']) || isset($config['unbound']['enable'])) {
 				$dnslist[] = $ifcfgipv6;
@@ -1438,7 +1438,7 @@ EOD;
 			$dnscfgv6 .= "	do-forward-updates false;\n";
 		}
 
-		if (($dhcpv6ifconf['enable-dhcp6c-dns']) != 'disabled') {
+		if (($dhcpv6ifconf['dhcp6c-dns']) != 'disabled') {
 			if (is_array($dhcpv6ifconf['dnsserver']) && ($dhcpv6ifconf['dnsserver'][0])) {
 				$dnscfgv6 .= "	option dhcp6.name-servers " . join(",", $dhcpv6ifconf['dnsserver']) . ";";
 			} else if (((isset($config['dnsmasq']['enable'])) || isset($config['unbound']['enable'])) && (is_ipaddrv6($ifcfgipv6))) {
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -249,50 +249,52 @@ function services_radvd_configure($blacklist = array()) {
 		$radvdconf .= "\t};\n";
 
 		/* add DNS servers */
-		$dnslist = array();
-		if (isset($dhcpv6ifconf['rasamednsasdhcp6']) && is_array($dhcpv6ifconf['dnsserver']) && !empty($dhcpv6ifconf['dnsserver'])) {
-			foreach ($dhcpv6ifconf['dnsserver'] as $server) {
-				if (is_ipaddrv6($server)) {
-					$dnslist[] = $server;
+		if (!isset($dhcpv6ifconf['disable-radvd-dns'])) {
+			$dnslist = array();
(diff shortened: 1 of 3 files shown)


luckman212:
Got it @jim-p — I understand the reasoning. I'll make the changes & resubmit.

luckman212:
@jim-p I just pushed a little update that hopefully satisfies the new requirement.

src/etc/inc/services.inc
```diff
@@ -249,51 +249,53 @@ function services_radvd_configure($blacklist = array()) {
 		$radvdconf .= "\t};\n";
 
 		/* add DNS servers */
-		$dnslist = array();
-		if (isset($dhcpv6ifconf['rasamednsasdhcp6']) && is_array($dhcpv6ifconf['dnsserver']) && !empty($dhcpv6ifconf['dnsserver'])) {
-			foreach ($dhcpv6ifconf['dnsserver'] as $server) {
-				if (is_ipaddrv6($server)) {
-					$dnslist[] = $server;
+		if (($dhcpv6ifconf['enable-radvd-dns']) != 'disabled') {
```
jim-p: Calling this `enable-radvd-dns` when it will contain a value of `enabled` or `disabled` seems a bit redundant, why not just `radvd-dns`?

src/etc/inc/services.inc
```diff
@@ -369,24 +371,26 @@ function services_radvd_configure($blacklist = array()) {
...
 
 		/* add DNS servers */
-		$dnslist = array();
-		if (isset($config['dnsmasq']['enable']) || isset($config['unbound']['enable'])) {
-			$dnslist[] = $ifcfgipv6;
-		} elseif (is_array($config['system']['dnsserver']) && !empty($config['system']['dnsserver'])) {
-			foreach ($config['system']['dnsserver'] as $server) {
-				if (is_ipaddrv6($server)) {
-					$dnslist[] = $server;
+		if (($dhcpv6ifconf['enable-radvd-dns']) != 'disabled') {
```
jim-p: Same as above for `enable-dhcp6c-dns`, use `dhcp6c-dns` or something similar to avoid redundancy.

luckman212: @jim-p Names fixed

src/etc/inc/services.inc
```diff
@@ -249,51 +249,53 @@ function services_radvd_configure($blacklist = array()) {
 		$radvdconf .= "\t};\n";
 
 		/* add DNS servers */
-		$dnslist = array();
-		if (isset($dhcpv6ifconf['rasamednsasdhcp6']) && is_array($dhcpv6ifconf['dnsserver']) && !empty($dhcpv6ifconf['dnsserver'])) {
-			foreach ($dhcpv6ifconf['dnsserver'] as $server) {
-				if (is_ipaddrv6($server)) {
-					$dnslist[] = $server;
+		if (($dhcpv6ifconf['radvd-dns']) != 'disabled') {
```
jim-p: One last little nit I didn't catch before, the variable doesn't need to be surrounded by ``()`` on its own like that.

src/etc/inc/services.inc
```diff
@@ -369,24 +371,26 @@ function services_radvd_configure($blacklist = array()) {
...
 
 		/* add DNS servers */
-		$dnslist = array();
-		if (isset($config['dnsmasq']['enable']) || isset($config['unbound']['enable'])) {
-			$dnslist[] = $ifcfgipv6;
-		} elseif (is_array($config['system']['dnsserver']) && !empty($config['system']['dnsserver'])) {
-			foreach ($config['system']['dnsserver'] as $server) {
-				if (is_ipaddrv6($server)) {
-					$dnslist[] = $server;
+		if (($dhcpv6ifconf['radvd-dns']) != 'disabled') {
```
jim-p: Remove the extra ``()`` around the variable

src/etc/inc/services.inc
```diff
@@ -1434,20 +1438,24 @@ EOD;
...
-		if (is_array($dhcpv6ifconf['dnsserver']) && ($dhcpv6ifconf['dnsserver'][0])) {
-			$dnscfgv6 .= "	option dhcp6.name-servers " . join(",", $dhcpv6ifconf['dnsserver']) . ";";
-		} else if (((isset($config['dnsmasq']['enable'])) || isset($config['unbound']['enable'])) && (is_ipaddrv6($ifcfgipv6))) {
-			$dnscfgv6 .= "	option dhcp6.name-servers {$ifcfgipv6};";
-		} else if (is_array($syscfg['dnsserver']) && ($syscfg['dnsserver'][0])) {
-			$dns_arrv6 = array();
-			foreach ($syscfg['dnsserver'] as $dnsserver) {
-				if (is_ipaddrv6($dnsserver)) {
-					$dns_arrv6[] = $dnsserver;
+		if (($dhcpv6ifconf['dhcp6c-dns']) != 'disabled') {
```
jim-p: Remove the extra ``()`` around the variable

luckman212:
@jim-p  Thansk. Zapped 'em.

PR #4132, opened 2019-12-17
cert_build_list(): by default don't show certs without prv key

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9972
- [ ] Ready for review

by default don't show certs without prv key
allow to show it with using extra $noprv arg (can be used for CA certs)

it will also allow  https://github.com/pfsense/pfsense/pull/4102


vktg:
> There are very few, if any, places that will require a private key as a CA, and those that do, don't use this function.

only Squid

done

src/etc/inc/certs.inc
```diff
@@ -2517,7 +2517,9 @@ function cert_build_list($type = 'cert', $consumer = '', $selectsource = false,
 	/* Create a hashed array with the certificate refid as the key and
 	 * descriptive name as the value. Exclude incompatible certificates. */
 	foreach ($config[$type] as $cert) {
-		if (cert_check_pkey_compatibility($cert['prv'], $consumer)) {
+		if (empty($cert['prv']) && $type == 'cert') {
```
jim-p: `$type == 'cert'` should be surrounded by parenthesis.

vktg: done

PR #4133, opened 2019-12-17
Captive Portal and SAML2 Integration

maurobraggio:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9970
- [ ] Ready for review


jim-p:
Typically third-party libraries like that should not be added directly to the base system if it can be avoided. We prefer to have them setup as separate FreeBSD ports so they can be updated independently.

Augustin-FL:
in addition to Jim P comments....It has been decided to integrate the captive portal with the user manager of pfSense.

please integrate SAML mechanism as "authentication server" instead of creating new fields in the captive portal config.

using an external library is fine (...we are using an external library for RADIUS messages, for instance) but have to be integrated as external packages, typically using FreeBSD Ports.

rbgarga:
@maurobraggio The comments from @jim-p and @Augustin-FL are a good sumary about this change.  We would be glad to review and integrate the feature but it would need to be completely reworked using:

- External library being installed from FreeBSD-ports, as a dependency of `security/pfSense` metaport
- Integrate SAML to "authentication server" mechanism, which will automatically make it available to captive portal

I'm closing this one since the proper implementation will be completely different.

maurobraggio:
Hi,
sorry for delay, I'm integrating SAML to "authentication server" mechanism, but I got stuck with freebsd port.
Could you give me an example of one freebsd port that include php source.
These php files must be accessible from the network via http(s), in particular inside /usr/local/captiveportal. Could I make it with freebsd port?
Thanks

Vdogue:
Hello people :),

This feature interests me and i would like to know if you (@maurobraggio ) have found any solution for now? :)

I'm letting my actual issue here just to show how this feature could be great. I hope its not a problem...
I'm currently changing my company's firewall with a netgate pfsense which will handle VPN connexions as well. My problem is that i want to handle anthentifications with our Azure AD (SAML protocol). Pfsense VPN availables are : 
 - OpenVPN which uses LDAP. It seems that azure AD DS needs to be implemented (watch link). 
 https://docs.microsoft.com/en-gb/azure/active-directory-domain-services/overview
 - L2TP which needs RADIUS. I don't know this last one but i believe that it implies to have a RADIUS auth server installed (watch link), another authentification layer, it's not really a good solution.
 https://docs.microsoft.com/en-gb/azure/active-directory/fundamentals/auth-radius
 - IPSec , i don't have much documentation about this one.
Perhaps i'm wrong, but none of these solutions seems to be comfortables. Do one of you guys have experienced that kind of issue? 
Thanks ! :)

resposit:
Hi @maurobraggio
did you make any progress with this ?

PR #4137, opened 2019-12-20
system_crlmanager.php: fixed incorrect input validation

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/3258
- [ ] Ready for review

invalid input validation, multiple serials not allowed

### DIFF
--- a/src/usr/local/www/system_crlmanager.php
+++ b/src/usr/local/www/system_crlmanager.php
@@ -127,9 +127,6 @@ switch ($act) {
 			$input_errors[] = gettext("Cannot revoke certificates for an imported/external CRL.");
 		}
 		foreach (explode(' ', $pconfig['revokeserial']) as $serial) {
-			if (!is_numeric($serial)) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_crlmanager.php
```diff
@@ -148,6 +137,17 @@
 				$input_errors[] = gettext("Invalid serial in list (Must be ASN.1 integer compatible decimal or hex string).");
 			}
 		}
+		if (empty($pconfig['certref']) && empty($revoke_list)) {
```
jim-p: This should be after the block below it, shouldn't it? Otherwise `$revoke_list` will always be empty if they selected a certificate and didn't enter a serial number.

vktg: now it check if(!empty($pconfig['revokeserial'])) and check each value for ASN.1 compatibility

working ok now

PR #4138, opened 2019-12-21
force noatime mount option

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9483
- [ ] Ready for review

On a clean CE install using the default options the / filesystem (UFS) is not mounted noatime.
ZFS is ok

this PR will force noatime option for all filesystems from fstab

### DIFF
--- a/src/etc/pfSense-rc
+++ b/src/etc/pfSense-rc
@@ -106,12 +106,12 @@ if [ ${FSCK_ACTION_NEEDED} = 1 ]; then
 	/sbin/fsck -y -t ufs
 fi
 
-/sbin/mount -a 2>/dev/null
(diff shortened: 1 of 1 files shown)


rbgarga:
We should respect /etc/fstab instead of adding more custom code to approach it.  What I have in mind is to change bsdinstall code to add noatime by default when installing a new system and make a change on one of our core package's post-install script to add it to /etc/fstab when upgrading.

What do you think @jim-p ?

jim-p:
That sounds like a better approach to me.

rbgarga:
Closing this since changes will be done in FreeBSD-src and FreeBSD-ports repositories

PR #4139, opened 2019-12-23
fix pfsense book link

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9996
- [ ] Ready for review

Help -> pfSense Book points to https://www.pfsense.org/j.php?jumpto=book,
which redirects to https://www.netgate.com/blog/pfsense-book-now-available-for-purchase.html page

it should redirects to https://docs.netgate.com/pfsense/en/latest/book/

this PR just point it directly to https://docs.netgate.com/pfsense/en/latest/book/


jim-p:
This needs fixed on the server (In `pfsense.org/j.php`), not here.

PR #4140, opened 2019-12-26
fix vti remoteid hiding in WebGUI

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9720
- [ ] Ready for review

fix vti remoteid hiding in WebGUI when you switch to Transport mode and then to VTI ( see https://youtu.be/avQWWjNl53o )
It also fix localid type to Address for VTI mode

### DIFF
--- a/src/usr/local/www/vpn_ipsec_phase2.php
+++ b/src/usr/local/www/vpn_ipsec_phase2.php
@@ -785,7 +785,9 @@ events.push(function() {
 		} else if (value == 'vti') {
 			hideClass('opt_localid', false);
 			hideClass('opt_natid', true);
-			$('#localid_type').val('network');
+			hideClass('opt_remoteid', false);
+			$('#localid_type').val('address');
+			disableInput('localid_type', true);
 			typesel_change_local(30);
 			$('#remoteid_type').val('address');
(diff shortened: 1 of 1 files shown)


fenichelar:
I think this broke the creation of a new Phase 2 in VTI mode. When you create a new Phase 2 and select Routed (VTI) as the mode, the Local Network Type is automatically set to Address and is disabled.

vktg:
> I think this broke the creation of a new Phase 2 in VTI mode. When you create a new Phase 2 and select Routed (VTI) as the mode, the Local Network Type is automatically set to Address and is disabled.

This is the correct behavior - the network type for VTI interfaces must be 'Address'

fenichelar:
@vktg I don't believe this behavior is correct. From https://docs.netgate.com/pfsense/en/latest/vpn/ipsec/ipsec-routed.html:

> Enter 10.6.106.1/30 for the Local Network Address

I have always used a /30 for Local Network Address. In 2.4.4, Local Network Address could be a subnet.

jim-p:
The code assumes /30 for IPv4. You do not need to pick the mask manually.

fenichelar:
@jim-p Ahh, that makes sense.

PR #4141, opened 2019-12-28
set correct MTU for static and default IPv6 routes. Issue #6868

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/6868
- [ ] Ready for review

If I adjust the MTU of an assigned interface, only link route for IPv6 on the interface has its MTU adjusted, not others.

This PR set correct MTU for static and default IPv6 routes of interface

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3995,6 +3995,7 @@ function interface_configure($interface = "wan", $reloadall = false, $linkupeven
 	} else {
 		if ($wantedmtu != get_interface_mtu($mtuif)) {
 			pfSense_interface_mtu($mtuif, $wantedmtu);
+			set_ipv6routes_mtu($mtuif, $wantedmtu);
 		}
 	}
 	/* XXX: What about gre/gif/.. ? */
@@ -6700,6 +6701,7 @@ function set_interface_mtu($interface, $mtu) {
 		}
 	} else {
 		pfSense_interface_mtu($interface, $mtu);
+		set_ipv6routes_mtu($interface, $mtu);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -6700,6 +6701,37 @@ function set_interface_mtu($interface, $mtu) {
...
+			}
+			if (($gate['interface'] == $interface) && $gate['isdefaultgw']) {
+				$tgt = "default";
+				$gateway = $gate['gateway'];
+				$ipv6mturoutes[$tgt] = $gateway;
+			}
+		}
+		if (!empty($ipv6mturoutes)) {
+			foreach ($ipv6mturoutes as $tgt => $gateway) {
+				mwexec("/sbin/route change -6 -mtu {$mtu} " . escapeshellarg($tgt) . " " . escapeshellarg($gateway));
```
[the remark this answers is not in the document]
vktg: done

src/etc/inc/interfaces.inc
```diff
@@ -6700,6 +6701,37 @@ function set_interface_mtu($interface, $mtu) {
 		}
 	} else {
 		pfSense_interface_mtu($interface, $mtu);
+		set_ipv6routes_mtu($interface, $mtu);
+	}
+}
+
+function set_ipv6routes_mtu($interface, $mtu) {
```
jim-p: This function probably belongs in `/etc/inc/util.inc` instead of here, since it's not technically an interface function. Move it after `function route_add_or_change()` near line 2483

vktg: done

PR #4145, opened 2020-01-02
openvpn server ocsp support

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/7767
- [ ] Ready for review

Based on https://github.com/OpenVPN/openvpn/blob/master/contrib/OCSP_check/OCSP_check.sh
but for latest openssl 1.1.1a output
shell script uses $config env to find correct vpnid from config

Do we need 'nonce' option here?
Do we need to check all certs in chain?
Bypass users on 'soft fail'? (server is not available)

Needs more testing!

### DIFF
--- a/src/etc/inc/openvpn.tls-verify.php
+++ b/src/etc/inc/openvpn.tls-verify.php
@@ -72,6 +72,36 @@ if (isset($allowed_depth) && ($cert_depth > $allowed_depth)) {
 	}
 }
 
+preg_match('/\/var\/etc\/openvpn\/server(\d+)\/config\.ovpn/', $_GET['config'], $current_vpnid);
+foreach ($config['openvpn']['openvpn-server'] as $ovpns) {
+	if (($ovpns['vpnid'] == $current_vpnid['1']) && ($ovpns['ocspcheck'] == 'yes')) {
+		$capath = "/var/etc/openvpn/server{$ovpns['vpnid']}/ca/";
+		$ca = lookup_ca($ovpns['caref']);
+		$cert_contents = base64_decode($ca['crt']);
+        	$cert_details = openssl_x509_parse($cert_contents);
+	        $issuer = $capath . $cert_details['hash'] . ".0";
+		$serial = $_GET['serial'];
+		$status = exec("/usr/bin/openssl ocsp -issuer " . escapeshellarg($issuer) 
+			. " -no_nonce"
+			. " -CApath " . escapeshellarg($capath)
+			. " -url " . escapeshellarg($ovpns['ocspurl'])
+			. " -serial " . escapeshellarg($serial));
+		if (preg_match('/(error|fail)/', $status)) {
+			echo "FAILED";
+			closelog();
+			return;
+		} else if (preg_match('/Cert Status: good/', $status) {
+			if (preg_match('/^Response verify OK/', $status) {
+				break;
+			}
+		} else {
+			echo "FAILED";
+			closelog();
+			return;
+		}
+	}
+}
+
 // Debug
 //syslog(LOG_WARNING, "Found certificate {$argv[2]} with depth {$cert_depth}");
 
--- a/src/usr/local/sbin/ovpn_auth_verify
+++ b/src/usr/local/sbin/ovpn_auth_verify
@@ -22,7 +22,11 @@
 
 
 if [ "$1" = "tls" ]; then
-	RESULT=$(/usr/local/sbin/fcgicli -f /etc/inc/openvpn.tls-verify.php -d "servercn=$2&depth=$3&certdepth=$4&certsubject=$5")
+	for ((check_depth=$3; check_depth>=0; check_depth--));
+	do
+		eval serial="\$tls_serial_${check_depth}"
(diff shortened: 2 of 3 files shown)


src/etc/inc/openvpn.tls-verify.php
```diff
@@ -72,6 +74,37 @@
...
+preg_match('/\/var\/etc\/openvpn\/server(\d+)\/config\.ovpn/', $_GET['config'], $current_vpnid);
+foreach ($config['openvpn']['openvpn-server'] as $ovpns) {
+	if (($ovpns['vpnid'] == $current_vpnid['1']) && ($ovpns['ocspcheck'] == 'yes')) {
+		$capath = "/var/etc/openvpn/server{$ovpns['vpnid']}/ca/";
+		$ca = lookup_ca($ovpns['caref']);
+		$cert_contents = base64_decode($ca['crt']);
+        	$cert_details = openssl_x509_parse($cert_contents);
+	        $issuer = $capath . $cert_details['hash'] . ".0";
+		$serial = $_GET['serial'];
+		$status = exec("/usr/bin/openssl ocsp -issuer {$issuer}
```
jim-p: Each of these variables needs `escapeshellarg()`

src/usr/local/www/firewall_aliases_edit.php
```diff
@@ -121,6 +121,12 @@ function alias_same_type($name, $type) {
 	$origname = $pconfig['name'];
 }
 
+if ($_REQUEST['exportaliases']) {
```
jim-p: Changes to this file seem to have snuck in from another PR. Remove these and squash the commits (or make a fresh PR)

src/usr/local/www/vpn_openvpn_server.php
```diff
@@ -929,6 +937,20 @@
 		));
 	}
 
+	$section->addInput(new Form_Checkbox(
+		'ocspcheck',
+		'OCSP Check',
+		'Check clients certificates with OCSP',
```
jim-p: s/clients/client/

src/usr/local/www/vpn_openvpn_server.php
```diff
@@ -602,6 +606,10 @@
 			}
 			$server['caref'] = $pconfig['caref'];
 			$server['crlref'] = $pconfig['crlref'];
+			if ($pconfig['ocspcheck']) {
+				$server['ocspcheck'] = "yes";
+			}
+			$server['ocspurl'] = $pconfig['ocspurl'];
```
jim-p: The URL must be validated before it is saved, especially since it is passed to a shell command. There is a function for this already in `util.inc`, called `is_URL($url)`, which should work.

src/usr/local/sbin/ovpn_auth_verify
```diff
@@ -22,7 +22,11 @@
 
 
 if [ "$1" = "tls" ]; then
-	RESULT=$(/usr/local/sbin/fcgicli -f /etc/inc/openvpn.tls-verify.php -d "servercn=$2&depth=$3&certdepth=$4&certsubject=$5")
+	# Depth in the certificate chain where the cert to verify is.
+	# "0" is the usual value here, where the client certificate is
+	check_depth=0
```
jim-p: Should `check_depth` always be `0` here? Or should it be what the user selected for depth in the OpenVPN server?

src/etc/inc/openvpn.tls-verify.php
vktg: done

src/usr/local/www/firewall_aliases_edit.php
vktg: done

src/usr/local/www/vpn_openvpn_server.php
vktg: done

src/usr/local/www/vpn_openvpn_server.php
vktg: but it's already checked with Form/Input placeholder type 'url'
do we need another check?

jim-p: Yes, that check is only client-side in JavaScript. It must also be checked when submitted, since malicious clients can submit things other ways, or someone may have JS disabled.

src/usr/local/sbin/ovpn_auth_verify
vktg: done

src/usr/local/www/vpn_openvpn_server.php
vktg: ok, done

rbgarga:
@vktg please squash all commits into a single one with a meaning commit log message

vktg:
done
squashed

src/etc/inc/openvpn.tls-verify.php
```diff
@@ -34,6 +34,8 @@
 require_once("config.inc");
 require_once("interfaces.inc");
 
+global $config;
```
rbgarga: global declaration is not needed outside a function

vktg: Hm, right
fixed

rbgarga:
I'll merge it as a 2.5.0 only feature. Thanks!

merkan414:
Hi everyone 
we cannot check ocsp result for client authentication but using crl checking is worked. What can we check ocsp checking

PR #4146, opened 2020-01-03
fix double proposal on vpn_ipsec_phase1.php

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10151
- [ ] Ready for review

Phase 1 Proposal (Encryption Algorithm) is doubled on latest pfSense 2.5


jim-p:
Change was made manually in another commit.

PR #4147, opened 2020-01-03
check private key export pass length

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/1192
- [ ] Ready for review

The password must be between 3 and 1023 characters long, otherwise you will get an openssl error:
```
$ openssl rsa -in shortpass.key -out p.key
Enter pass phrase for shortpass.key:
140194656773248:error:28078065:UI routines:UI_set_result_ex:result too small:../crypto/ui/ui_lib.c:905:You must type in 4 to 1023 characters
```

### DIFF
--- a/src/usr/local/www/system_certmanager.php
+++ b/src/usr/local/www/system_certmanager.php
@@ -163,21 +163,37 @@ switch ($act) {
 		/* Exporting a private key */
 		$keyout = base64_decode($thiscert['prv']);
 		if (isset($_POST['exportpass']) && !empty($_POST['exportpass'])) {
-			$res_key = openssl_pkey_get_private($keyout);
-			if ($res_key) {
-				openssl_pkey_export($res_key, $keyout, $_POST['exportpass']);
-			} else {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_certmanager.php
```diff
@@ -163,21 +163,37 @@
...
 		if (isset($_POST['exportpass']) && !empty($_POST['exportpass'])) {
-			$res_key = openssl_pkey_get_private($keyout);
-			if ($res_key) {
-				openssl_pkey_export($res_key, $keyout, $_POST['exportpass']);
-			} else {
-				$savemsg = gettext("Unable to export password-protected private key.");
-				$class = 'danger';
-			}
-		}
+                        if ((strlen($_POST['exportpass']) < 4) or (strlen($_POST['exportpass']) > 1023)) {
```
jim-p: The indentation of all this needs fixed, it's using spaces and should be tabs

vktg: hm, console copy/paste behavior
fixed

PR #4148, opened 2020-01-03
strip ipsec hash algo from swanctl.conf (phase2 only)

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9726
- [ ] Ready for review

strip ipsec hash algo from swanctl.conf Phase 2 proposals


rbgarga:
@vktg just tagging you case you missed @jim-p comment

vktg:
@rbgarga thanks, PR is in progress

vktg:
done

> It would also be useful to hide or unset+disable the hash controls when only GCM algorithms are selected.

It’s better not to uncheck the checkboxes - this will help users quickly return to the previous configuration.

PR #4151, opened 2020-01-04
enable urltable alias to return > 1 IPs

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8531
- [ ] Ready for review

Patch that enables URL Table aliases to return a mix of IPv4/IPv6 entries as well as accomodate multiple entries per FQDN
Original patch by Luke Hamburg:
https://github.com/luckman212/pfsense/commit/063fa3a4a0832da86f9076040417903aa5d4ca7c

it's a bit modified by creating separate function fqdn_to_ip()

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -1867,23 +1867,43 @@ function get_uptime_sec() {
 	return $uptime;
 }
 
-function fqdn_to_ip($hostname, $ipv6 = false) {
-	$domrecords = array();
-	$domips = array();
-	$qtype = ($ipv6 ? 'AAAA' : 'A');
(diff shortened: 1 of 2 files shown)


vktg:
while checking this code I found that add_hostname_to_watch() works only with IPv4 addresses,
and it's used only in system.inc: https://github.com/pfsense/pfsense/blob/ed6f2195b282008e7d84a087759dbc6b4fc23eed/src/etc/inc/system.inc#L935

in foreach loop through static routes, but System / Routing / Static Routes don't allow you to use FQDN as destination, only IPv4/IPv6 addresses!

vktg:
Done
switched to resolve_host_addresses () func

PR #4153, opened 2020-01-04
Improve building of non pfSense builds

rschell:
* Remove redirection on jail delete so user receives delete prompt
* Fix editing ports that don't contain a pkg-descr file by testing for file's existence first 
* Fix renaming of pfSense.c file in php module
* Renaming of php_pfSense.h references in pfSense.c and dummynet.c
* Allow for renaming of pfSense files to include both files and directories

A pull request for sysutils/pfSense-repo/makefile is also suggested in FreeBSD-ports.

See 2.5 Development Snapshot "Minor issues with non-pfSense rename logic"  discussion and redmine #10161

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10161
- [ ] Ready for review

### DIFF
--- a/tools/builder_common.sh
+++ b/tools/builder_common.sh
@@ -1541,6 +1541,7 @@ poudriere_rename_ports() {
 		local _pdir=$(dirname ${d})
 		local _pname=$(echo $(basename ${d}) | sed "s,pfSense,${PRODUCT_NAME},")
 		local _plist=""
+		local _pdescr=""
 
 		if [ -e ${_pdir}/${_pname} ]; then
 			rm -rf ${_pdir}/${_pname}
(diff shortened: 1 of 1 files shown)


tools/builder_common.sh
```diff
@@ -1552,31 +1553,36 @@ poudriere_rename_ports() {
...
 				  -e "/ZEND_GET_MODULE/ s,pfSense,${PRODUCT_NAME}," \
 				  -e "/PHP_PFSENSE_WORLD_EXTNAME/ s,pfSense,${PRODUCT_NAME}," \
 				${_pdir}/${_pname}/files/pfSense.c \
+				${_pdir}/${_pname}/files/dummynet.c \
 				${_pdir}/${_pname}/files/php_pfSense.h
 		fi
 
 		if [ -d ${_pdir}/${_pname}/files ]; then
-			for fd in $(find ${_pdir}/${_pname}/files -type d -name '*pfSense*'); do
+			for fd in $(find ${_pdir}/${_pname}/files d -name '*pfSense*'); do
```
rbgarga: It looks like you removed `-type` parameter by accident here

rschell: I was trying to remove "-type d" so that both files and directories get renamed.  I managed not to delete the "d".

PR #4156, opened 2020-01-08
XMLRPC: fix last shaper/limiter removing. Issue #9468-9469

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9468
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9469
- [ ] Ready for review

With an HA cluster synchronizing settings via XMLRPC, when the last limiter is deleted that change is not reflected on the secondary. The last limiter remains in the configuration on the secondary.

With an HA cluster synchronizing settings via XMLRPC, when the last ATLQ traffic shaper queue is deleted that change is not reflected on the secondary. The last queue remains in the configuration on the secondary. If the user clicks the button to remove the traffic shaper settings, then all settings remain on the secondary.

We've hit issues like this in the past in other areas, where the sync only happens when the section is not empty, so that last change gets skipped.


src/etc/rc.filter_synchronize
```diff
@@ -266,10 +266,16 @@ if (is_array($config['hasync'])) {
 		}
 		$sections[] = 'wol';
 	}
-	if ($hasync['synchronizetrafficshaper'] != "" and is_array($config['shaper'])) {
+	if ($hasync['synchronizetrafficshaper'] != "") {
+		if (!is_array($config['shaper'])) {
```
jim-p: You can safely do `init_config_arr(array('shaper'))` and skip the manual check. So, you could just do this:
```
if ($hasync['synchronizetrafficshaper'] != "") {
	init_config_arr(array('shaper'));
	$sections[] = 'shaper';
}
```

src/etc/rc.filter_synchronize
```diff
@@ -266,10 +266,16 @@ if (is_array($config['hasync'])) {
...
+			$config['shaper'] = array();
+		}
 		$sections[] = 'shaper';
 	}
-	if ($hasync['synchronizetrafficshaperlimiter'] != "" and is_array($config['dnshaper'])) {
+	if ($hasync['synchronizetrafficshaperlimiter'] != "") {
+		if (!is_array($config['dnshaper'])) {
```
jim-p: See above note, same can be done here (but for `dnshaper`)

src/etc/rc.filter_synchronize
vktg: ok, done

src/etc/rc.filter_synchronize
vktg: done

PR #4157, opened 2020-01-09
encrypt exported key with AES-256. Issue #1192

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/1192
- [ ] Ready for review

By default openssl export uses DES-EDE3-CBC for encryption

This is not necessary, but if we want to avoid using any outdated encryption standards (to comply Suite B, CNSA), we should avoid the old algorithms throughout the all system.

### DIFF
--- a/src/usr/local/www/system_certmanager.php
+++ b/src/usr/local/www/system_certmanager.php
@@ -170,7 +170,8 @@ switch ($act) {
 			} else {
 				$res_key = openssl_pkey_get_private($keyout);
 				if ($res_key) {
-					openssl_pkey_export($res_key, $keyout, $_POST['exportpass']);
+					$args = array('encrypt_key_cipher' => OPENSSL_CIPHER_AES_256_CBC);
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_certmanager.php
```diff
@@ -170,7 +170,9 @@
 			} else {
 				$res_key = openssl_pkey_get_private($keyout);
 				if ($res_key) {
-					openssl_pkey_export($res_key, $keyout, $_POST['exportpass']);
+					$args = array();
```
jim-p: You could do this in one line:
`$args = array('encrypt_key_cipher' => OPENSSL_CIPHER_AES_256_CBC);`
Actually if `$args` does not need to be passed by reference to `openssl_pkey_export` you don't need to define it at all, just use that array straight in the function call. But that can be more difficult to read, so it's not necessary.

vktg: fixed

PR #4159, opened 2020-01-10
don't kill states when failover gateway is down

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8555
- [ ] Ready for review

This PR changes filter_delete_states_for_down_gateways() by adding failover group checking -
in this case if, for example Tier 1 is up but Tier 2 is down, it doesn't kill all states

It kill states only if all tiers in failover group are equal (loadbalance configuration)
or if failed Tier == Tier with lower value (high priority)

This avoid state killing in case when primary (Tier 1) link works fine, but secondary (Tier 2) link is flapping

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -200,6 +200,7 @@ function filter_delete_states_for_down_gateways() {
 
 	$any_gateway_down = false;
 	$a_gateways = return_gateways_status();
+	$a_gateway_groups = &$config['gateways']['gateway_group'];
 	if (is_array($GatewaysList)) {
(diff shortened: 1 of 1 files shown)


vktg:
What if we put checkbox for this feature on the System \ Routing \ Gateway Groups \ Edit page?

jim-p:
I think it would confuse users and give them false hope that it might behave in other ways.

The right fix, as discussed on the Redmine issue, would be for pfctl to know how to kill by gateway since it would have access to that information in the state table directly. There isn't a scalable way to fix this in scripting logic.

PR #4160, opened 2020-01-11
fix incorrect encrypted backup restore error handling. Issue #10179

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10179
- [ ] Ready for review

If you trying to restore encrypted backup, but select invalid file, you got blank page after pressing "restore" ( caused by 'return 1;' in code )
If you enter incorrect decryption password, you got error message: 
'A full configuration restore was selected but a pfsense tag could not be located.'
which is incorrect, it should be:
'File decryption failed. Incorrect password or file is invalid.'

### DIFF
--- a/src/usr/local/www/diag_backup.php
+++ b/src/usr/local/www/diag_backup.php
@@ -233,18 +233,17 @@ if ($_POST) {
 					/* read the file contents */
 					$data = file_get_contents($_FILES['conffile']['tmp_name']);
 					if (!$data) {
-						log_error(sprintf(gettext("Warning, could not read file %s"), $_FILES['conffile']['tmp_name']));
-						return 1;
-					}
-
-					if ($_POST['decrypt']) {
+						$input_errors[] = gettext("Warning, could not read file {$_FILES['conffile']['tmp_name']}");
+					} elseif ($_POST['decrypt']) {
 						if (!tagfile_deformat($data, $data, "config.xml")) {
 							$input_errors[] = gettext("The uploaded file does not appear to contain an encrypted pfsense configuration.");
-							return 1;
+						} else {
+							$data = decrypt_data($data, $_POST['decrypt_password']);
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_backup.php
```diff
@@ -233,18 +233,17 @@ function check_and_returnif_section_exists($section) {
 					/* read the file contents */
 					$data = file_get_contents($_FILES['conffile']['tmp_name']);
 					if (!$data) {
-						log_error(sprintf(gettext("Warning, could not read file %s"), $_FILES['conffile']['tmp_name']));
-						return 1;
-					}
-
-					if ($_POST['decrypt']) {
+						$input_errors[] = gettext("Warning, could not read file {$_FILES['conffile']['tmp_name']}");
+					} else if ($_POST['decrypt']) {
```
jim-p: This should be `elseif` ([ref](https://docs.netgate.com/pfsense/en/latest/development/developer-style-guide.html#general-rules))

src/usr/local/www/diag_backup.php
```diff
@@ -277,7 +276,7 @@ function check_and_returnif_section_exists($section) {
 								$savemsg = gettext("The configuration area has been restored. The firewall may need to be rebooted.");
 							}
 						}
-					} else {
+					} else if (!$input_errors) {
```
jim-p: This should be `elseif` ([ref](https://docs.netgate.com/pfsense/en/latest/development/developer-style-guide.html#general-rules))

src/usr/local/www/diag_backup.php
vktg: fixed

src/usr/local/www/diag_backup.php
vktg: fixed

PR #4161, opened 2020-01-11
OpenVPN: allow to fully disable compression

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10177
- [ ] Ready for review

From openvpn(8):
If  the  algorithm  parameter  is  empty, compression will be turned off, but the packet framing for compression will still be enabled, **allowing a different setting to be pushed later**.

This PR allow to completely remove 'compress' option from config file


vktg:
default is no compression, so, 
simply rename "Omit Preference (Use OpenVPN Default)" to  "Fully Disable Compression, prevent from pushing settings later."

jim-p:
This is not necessary.

PR #4164, opened 2020-01-16
inverse matching of tags. Issue #10186

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10186
- [ ] Ready for review

allow to inverse tagged rules
i.e.
`block out quick on RED_NET ! tagged RED_NET_OK`

### DIFF
--- a/src/usr/local/www/firewall_rules_edit.php
+++ b/src/usr/local/www/firewall_rules_edit.php
@@ -101,6 +101,7 @@ function is_aoadv_used($rule_config) {
 	    (isset($rule_config['disablereplyto'])) ||
 	    ($rule_config['tag'] != "") ||
 	    ($rule_config['tagged'] != "") ||
+	    (isset($rule_config['nottagged'])) ||
 	    ($rule_config['max'] != "") ||
 	    ($rule_config['max-src-nodes'] != "") ||
 	    ($rule_config['max-src-conn'] != "") ||
@@ -273,6 +274,9 @@ if (isset($id) && $a_filter[$id]) {
 	}
 	if (isset($a_filter[$id]['disablereplyto'])) {
 		$pconfig['disablereplyto'] = true;
+	}
+	if (isset($a_filter[$id]['nottagged'])) {
+		$pconfig['nottagged'] = true;
 	}
 
 	/* advanced */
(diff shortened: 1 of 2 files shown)


src/usr/local/www/firewall_rules_edit.php
```diff
@@ -521,6 +525,10 @@ function is_aoadv_used($rule_config) {
 		$input_errors[] = gettext("Invert match cannot be selected with 'any'.");
 	}
 
+	if ((isset($_POST['nottagged']) && empty($_POST['tagged']))) {
```
jim-p: There is an extra set of parenthesis on the outside this test

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1570,12 +1583,25 @@ function build_flag_table() {
...
-	'Tagged',
-	'text',
-	$pconfig['tagged']
-))->setHelp('A packet can be matched on a mark placed before on another rule.');
+$group = new Form_Group('Tagged');
+
+$group->add(new Form_Checkbox(
+	'nottagged',
+	'nottagged',
+	'Invert',
```
jim-p: This should be `Invert match`

src/usr/local/www/firewall_rules_edit.php
```diff
@@ -1570,12 +1583,25 @@ function build_flag_table() {
...
+))->setWidth(1);
+
+$group->add(new Form_Input(
+'tagged',
+'Tagged',
+'text',
+$pconfig['tagged']
+))->setWidth(4);
+
+$group->setHelp('A packet can be matched on a mark placed before on another rule.');
```
jim-p: Change this to:
> Match a mark placed on a packet by a different rule with the Tag option. Check Invert match to match packets which do not contain this tag.

src/usr/local/www/firewall_rules_edit.php
vktg: fixed

src/usr/local/www/firewall_rules_edit.php
vktg: done

src/usr/local/www/firewall_rules_edit.php
vktg: done,
but now GUI is not so pretty

jim-p: I see that now, it forces a word wrap. I guess it's OK to move it back to just 'Invert' and adjust the text below to match.

vktg: done

PR #4165, opened 2020-01-16
IPsec IPv6 dynamic FQDN Remote Gateways, resolve_retry() IPv6 support. Issue #9405

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9405
- [ ] Ready for review

IPsec IPv6 dynamic FQDN Remote Gateways support by adding IPv6 FQDN resolution ability to resolve_retry()

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -1932,22 +1932,55 @@ function mac_format($clientmac) {
 	}
 }
 
-function resolve_retry($hostname, $retries = 5) {
+function resolve_retry($hostname, $protocol = 'inet', $retries = 5, $numrecords = 1) {
 
-	if (is_ipaddr($hostname)) {
-		return $hostname;
-	}
-
+	$recresult = array();
+	$returnres = array();
 	for ($i = 0; $i < $retries; $i++) {
-		// FIXME: gethostbyname does not work for AAAA hostnames, boo, hiss
-		$ip = gethostbyname($hostname);
-
-		if ($ip && $ip != $hostname) {
-			/* success */
-			return $ip;
+		switch ($protocol) {
+			case 'any':
+				$checkproto = 'is_ipaddr';
+				$dnsproto = DNS_ANY;
+				$dnstype = array('A', 'AAAA');
+				break;
+			case 'inet6':
+				$checkproto = 'is_ipaddrv6';
+				$dnsproto = DNS_AAAA;
+				$dnstype = array('AAAA');
+				break;
+			case 'inet': 
+			default:
+				$checkproto = 'is_ipaddrv4';
+				$dnsproto = DNS_A;
+				$dnstype = array('A');
+				break;
+		}
+		if ($checkproto($hostname)) {
+			return $hostname;
+		}
+		$dnsresult = dns_get_record($hostname, $dnsproto);
+		foreach ($dnsresult as $dnsrec => $ip) {
+			if (is_array($ip)) {
+				if (in_array($ip['type'], $dnstype)) {
+				    if ($checkproto($ip['ip'])) { 
+					    $recresult[] = $ip['ip'];
+				    }
+				    if ($checkproto($ip['ipv6'])) { 
+					    $recresult[] = $ip['ipv6'];
+				    }
+				}
+			}
 		}
 
 		sleep(1);
+	}
+
+	if (!empty($recresult)) {
+		if ($numrecords == 1) {
+			return $recresult[0];
+		} else {
+			return array_slice($recresult, 0, $numrecords);
+		}
 	}
 
 	return false;
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -351,7 +351,7 @@ function ipsec_get_phase1_dst(& $ph1ent) {
 	$rg = $ph1ent['remote-gateway'];
 	if (!is_ipaddr($rg)) {
 		if (!platform_booting()) {
-			return resolve_retry($rg);
+			return resolve_retry($rg, $ph1ent['protocol']);
 		}
 	}
 	if (!is_ipaddr($rg)) {
@@ -1209,7 +1209,7 @@ function ipsec_setup_gwifs() {
 				$filterdns_list[] = "{$rg}";
 				add_hostname_to_watch($rg);
 				if (!platform_booting()) {
-					$rg = resolve_retry($rg);
+					$rg = resolve_retry($rg, $ph1ent['protocol']);
 				}
 				if (!is_ipaddr($rg)) {
 					continue;


src/etc/inc/util.inc
```diff
@@ -1932,19 +1932,30 @@ function mac_format($clientmac) {
...
 
-		if ($ip && $ip != $hostname) {
-			/* success */
-			return $ip;
+		// IPv4 or any, return IPv4 address if resolved
+		if ($protocol != 'inet6') {
+			if (is_ipaddrv4($hostname)) {
+                                return $hostname;
+                        }
+                        $ip = dns_get_record($hostname, DNS_A)[0]['ip'];
```
jim-p: This should probably be more than one step here, since if it fails the return false will be `false` and then this attempt would fail with a PHP error since the response would not be an array.

src/etc/inc/util.inc
```diff
@@ -1932,19 +1932,30 @@ function mac_format($clientmac) {
 	}
 }
 
-function resolve_retry($hostname, $retries = 5) {
-
-	if (is_ipaddr($hostname)) {
-		return $hostname;
-	}
+function resolve_retry($hostname, $protocol = 'any', $retries = 5) {
```
jim-p: The protocol defaults to `any` here but there is no special handling for that. `dns_get_record()` could be passed `DNS_A + DNS_AAAA` and it should return whatever it finds in that case. The code below should probably allow that case, and then it wouldn't need to care whether the returned value was IPv4 or IPv6 (it can use `is_ipaddr()` and not the v4/v6 variants).

Safer assumptions for `$protocol` would be to use the explicitly value if set (`any`, `inet`, `inet6`) and perhaps assume `inet` if it's passed an empty value.

cachethrasher:
...anything as long as for a host with both A and AAAA, when the hostname is configured as IPV6 tunnel endpoint, the resolver will return the IPV6 address... :)

jim-p:
> Thanks for working on this guys. ...Anything as long as for a host with both A and AAAA, when the hostname is configured as an IPv6 IPSec tunnel endpoint, the resolver will return the IPV6 address... :)

Yes, that is still accounted for since the call to this from IPsec passes the configured protocol of the tunnel.

src/etc/inc/util.inc
vktg: fixed

src/etc/inc/util.inc
vktg: I found that I got "Warning: dns_get_record(): DNS Query failed " error in any case, 
if I add 
php```
if (dns_get_record($hostname, $dnsp)) {
 ....
} else {
continue
} 
```
I got "Warning: dns_get_record(): DNS Query failed " error anyway

File: src/etc/inc/util.inc [hunk 3]
```diff
@@ -1932,19 +1932,34 @@ function mac_format($clientmac) {
...
+				break;
+			case 'inet6':
+				$checkproto = 'is_ipaddrv6';
+				$dnsproto = array('ipv6' => DNS_AAAA);
+				break;
+			case 'any':
+				$checkproto = 'is_ipaddr';
+				$dnsproto = array('ip' => DNS_A, 'ipv6' => DNS_AAAA);
+				break;
+			default:
```
jim-p: Pair the default case with `inet` to preserve the old behavior if something passes it a null/empty value, just in case. I know nothing else in the base system uses it, but I haven't checked packages.

src/etc/inc/util.inc
```diff
@@ -1932,19 +1932,34 @@ function mac_format($clientmac) {
...
+				break;
+		}
+		if (call_user_func($checkproto, $hostname)) {
```
jim-p: You can do `$checkproto($hostname)` instead of using `call_user_func`. Not that the difference is significant here, but using variable functions is slightly faster, and more intuitive to look at.

src/etc/inc/util.inc
```diff
@@ -1932,19 +1932,34 @@ function mac_format($clientmac) {
...
+			return $hostname;
+		}
+		foreach ($dnsproto as $ipp => $dnsp) {
+			$ip = dns_get_record($hostname, $dnsp)[0][$ipp];
+			if ($ip && call_user_func($checkproto, $ip)) {
```
jim-p: See above note about variable functions.

src/etc/inc/util.inc
```diff
@@ -1932,19 +1932,34 @@ function mac_format($clientmac) {
...
+				$checkproto = 'is_ipaddr';
+				$dnsproto = array('ip' => DNS_A, 'ipv6' => DNS_AAAA);
+				break;
+			default:
+				break;
+		}
+		if (call_user_func($checkproto, $hostname)) {
+			return $hostname;
+		}
+		foreach ($dnsproto as $ipp => $dnsp) {
```
jim-p: The query should be doing `DNS_A + DNS_AAAA` so it only performs a single query, not multiple which will be slower. You can iterate through the resulting array.

src/etc/inc/util.inc
```diff
@@ -1932,19 +1932,34 @@ function mac_format($clientmac) {
...
+			$ip = dns_get_record($hostname, $dnsp)[0][$ipp];
```
jim-p: I know you said you still received an error, but this is still the wrong way to do it. It must get the result and check it before treating it as an array. You could suppress any remaining errors by using `$result = @dns_get_record( [...]` but since that function may not return an array, calling it this way is incorrect.

src/etc/inc/util.inc
vktg: done

src/etc/inc/util.inc
vktg: done

src/etc/inc/util.inc
vktg: done

src/etc/inc/util.inc
vktg: DNS_A + DNS_AAAA is return false if one of protocols is false
see https://stackoverflow.com/questions/26920802/should-i-call-dns-get-record-multiple-times-or-once-with-dns-any

it return false for IPv6 only sites for example

src/etc/inc/util.inc
vktg: done

src/etc/inc/util.inc
jim-p: Then use `DNS_ANY` for the `any` case to fetch everything at once, and perhaps add another parameter to limit how many records are returned (defaulting to `1` to preserve existing behavior)

vktg: You can't control how many records are returned, 
but I set the $anyrecords parameter to control how many returned records need to be checked
it 5 by default, because A/AAAA records is not always first

also comment on https://www.php.net/manual/en/function.dns-get-record.php#120049:

>  Please note that Firewalls and anti malware software detects  (and depending on company policies even blocks) DNS_ANY requests.
> In that case the usage of this function fails.
> This is because DNS_ANY requests can be  exploited for creating "amplification (D)DOS attackes": You send 1 DNS_ANY request and get a huge amount of information back, thus even small requests can result into hugh network load.
> 

That's true,- not all domains are resolved by DNS_ANY

jim-p: You can't control how many records get returned by `dns_get_record()`, but you can control what gets returned by `resolve_retry()`, not how many get checked. I'm saying that the 'any' case should return everything it gets, not just the first entry, up to the number of records requested (defaulting to any/all without a limit)

vktg: ok, changed
it returns single value if $numrecords == 1 (current behavior)  or array if $numrecords > 1

src/etc/inc/util.inc
```diff
@@ -1932,24 +1932,63 @@ function mac_format($clientmac) {
...
+		if ($numrecords == 1) {
+			return $recresult[0];
+		} else {
+			foreach (range(0, $numrecords-1) as $i) {
+				$rec = $recresult[$i];
+				if ($rec) {
+					$returnres[] .= $rec;
+				}
+			}
+			return $returnres;
```
jim-p: This whole block can be done in one line with a built-in function:

`return array_slice($recresult, 0, $numrecords);`

It's safe for `$numrecords` to be larger than the size of `$recresult` when using this function, it will return as many as it can.

src/etc/inc/util.inc
```diff
@@ -1932,24 +1932,63 @@ function mac_format($clientmac) {
...
+		}
+		if ($checkproto($hostname)) {
+			return $hostname;
+		}
+		$dnsresult = dns_get_record($hostname, $dnsproto);
+		foreach ($dnsresult as $dnsrec => $ip) {
+			if (is_array($ip)) {
+				if (in_array($ip['type'], $dnstype)) {
+				    if ($checkproto($ip['ip'])) { 
+					    $recresult[] .= $ip['ip'];
```
jim-p: Why is this using `.=`? There is nothing to add to, it should just be `=`.

src/etc/inc/util.inc
```diff
@@ -1932,24 +1932,63 @@ function mac_format($clientmac) {
...
+				    }
+				    if ($checkproto($ip['ipv6'])) { 
+					    $recresult[] .= $ip['ipv6'];
```
jim-p: Why is this using `.=`? There is nothing to add to, it should just be `=`.

vktg: fixed

src/etc/inc/util.inc
vktg: fixed

src/etc/inc/util.inc
vktg: ok, done

PR #4166, opened 2020-01-17
allow to disable IPsec P1 when P2 is disabled VTI. Issue #10190

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10190
- [ ] Ready for review

The original ipsec_vti() doesn't check that P2 VTI is disabled or not
This PR fix it

### DIFF
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -920,7 +920,7 @@ function ipsec_get_loglevels() {
 	return $levels;
 }
 
-function ipsec_vti($ph1ent, $returnaddresses = false) {
+function ipsec_vti($ph1ent, $returnaddresses = false, $skipdisabled = true) {
 	global $config;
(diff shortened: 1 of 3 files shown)


src/etc/inc/ipsec.inc
```diff
@@ -920,7 +920,7 @@ function ipsec_get_loglevels() {
 	return $levels;
 }
 
-function ipsec_vti($ph1ent, $returnaddresses = false) {
+function ipsec_vti($ph1ent, $returnaddresses = false, $disabled = false) {
```
jim-p: This seems like it's misnamed, shouldn't it be `$skipdisabled` or something similar? Just calling it `$disabled` and then passing it a static value during validation seems awkward/confusing.

vktg: ok, changed

PR #4167, opened 2020-01-21
DHCP6 client discard REQUEST messages. Issue #9634

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9634
- [ ] Ready for review

pfsense sends DHCPv6 Request messages to ff02::1:2 on its WAN interface at an interval of about 7 seconds. As this is a link-local multicast destination, it also receives those messages. This starts a chain reaction: Upon recieving the message, dhcp6c_wan_script.sh calls rc.newwanipv6 which triggers filter reload etc., pushing my APU's load from 9 to >70 %, leading to severe UDP packet loss, causing DNS request to fail, and rendering VoIP calls sustainably unbearable.

RFC 8415 (DHCPv6) clearly states, "Clients MUST discard any received Request messages" (section 16.4)


luckman212:
@vktg Wow, nice find! This one seems pretty high importance. Any reason this patch can't be applied to 2.4.5?

PR #4169, opened 2020-01-21
Use IPv4 connectivity as parent interface for SLAAC. Issue #9324

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9324
- [ ] Ready for review

Use IPv4 (PPPoE) connectivity as parent interface for SLAAC
Same behavior as staticv6  route

### DIFF
--- a/src/usr/local/www/interfaces.php
+++ b/src/usr/local/www/interfaces.php
@@ -283,6 +283,7 @@ switch ($wancfg['ipaddr']) {
 switch ($wancfg['ipaddrv6']) {
 	case "slaac":
 		$pconfig['type6'] = "slaac";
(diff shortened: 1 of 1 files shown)


vktg:
successfully tested with cisco ios 15.5(3) pppoe server
PPPoE part:
```
interface Virtual-Template1
 mtu 1492
 ip address 10.122.122.254 255.255.255.0
 peer default ip address pool CLIENT
 peer default ipv6 pool pool_pppoe
 ipv6 address autoconfig
 ipv6 unnumbered Loopback0
 ipv6 enable
 ipv6 mtu 1492
 ipv6 nd other-config-flag
 no ipv6 nd ra suppress
 ppp authentication chap pap callin
```

PR #4170, opened 2020-01-22
Avoid very slow GUI loads when ews.netgate.com can't be resolved #8987

emmtbot:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/8987
- [x] Ready for review

### DIFF
--- a/src/etc/inc/copyget.inc
+++ b/src/etc/inc/copyget.inc
@@ -73,6 +73,7 @@ function updatecopyright() {
 // Retrieve the copyright data from Netgate.com if
 // the support data file does not exist, or
 // if it is more than a day old and the URL seems resolvable
-if ( !file_exists($copyrightfile) || ( time()-filemtime($copyrightfile) > $refreshinterval && is_url_hostname_resolvable($FQDN) ) ) {
+if (!file_exists($copyrightfile) ||
+    ((time()-filemtime($copyrightfile) > $refreshinterval) && is_url_hostname_resolvable($FQDN))) {
 	updateCopyright();
 }
--- a/src/etc/inc/copyget.inc
+++ b/src/etc/inc/copyget.inc
@@ -70,8 +70,9 @@ function updatecopyright() {
 	}
 }
 
-// Retrieve the copyright data from Netgate.com if the support data file does not exist,
-// or if it is more than a day old
-if (!file_exists($copyrightfile) || ( time()-filemtime($copyrightfile) > $refreshinterval)) {
+// Retrieve the copyright data from Netgate.com if
+// the support data file does not exist, or
+// if it is more than a day old and the URL seems resolvable
+if ( !file_exists($copyrightfile) || ( time()-filemtime($copyrightfile) > $refreshinterval && is_url_hostname_resolvable($FQDN) ) ) {
 	updateCopyright();
 }
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -3565,6 +3565,9 @@ function send_user_download($type = 'data', $content, $userfilename = "", $conte
 // Test whether the hostname in a URL can be resolved with a very short timeout
 function is_url_hostname_resolvable($url) {
 	$urlhostname = parse_url($url, PHP_URL_HOST);
+	if (empty($urlhostname)) {
+		return false;
+	}
 	putenv("RES_OPTIONS=timeout:3 attempts:1");
 	$resolvable = ($urlhostname !== gethostbyname($urlhostname));
 	putenv("RES_OPTIONS");
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -3562,4 +3562,13 @@ function send_user_download($type = 'data', $content, $userfilename = "", $conte
 	exit;
 }
 
+// Test whether the hostname in a URL can be resolved with a very short timeout
+function is_url_hostname_resolvable($url) {
+	$urlhostname = explode("/", $url)[2];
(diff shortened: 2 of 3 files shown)


src/etc/inc/copyget.inc
```diff
@@ -70,8 +70,9 @@ function updatecopyright() {
 	}
 }
 
-// Retrieve the copyright data from Netgate.com if the support data file does not exist,
-// or if it is more than a day old
-if (!file_exists($copyrightfile) || ( time()-filemtime($copyrightfile) > $refreshinterval)) {
+// Retrieve the copyright data from Netgate.com if
+// the support data file does not exist, or
+// if it is more than a day old and the URL seems resolvable
+if ( !file_exists($copyrightfile) || ( time()-filemtime($copyrightfile) > $refreshinterval && is_url_hostname_resolvable($FQDN) ) ) {
```
jim-p: You need one more set of parenthesis, around `time()-filemtime($copyrightfile) > $refreshinterval`.

Also, this should be broken up into multiple lines. Additional lines should be indented four spaces in below the `if`.

src/usr/local/www/widgets/widgets/netgate_services_and_support.widget.php
```diff
@@ -40,9 +40,10 @@
 
 if ($_REQUEST['ajax']) {
 
-	// Retrieve the support data from Netgate.com if the support data file does not exist,
-	// or if it is more than a day old
-	if (!file_exists($supportfile) || ( time()-filemtime($supportfile) > $refreshinterval)) {
+	// Retrieve the support data from Netgate.com if
+	// the support data file does not exist, or
+	// if it is more than a day old and the URL seems resolvable
+	if ( !file_exists($supportfile) || ( time()-filemtime($supportfile) > $refreshinterval && is_url_hostname_resolvable($FQDN) ) ) {
```
jim-p: You need one more set of parenthesis, around `time()-filemtime($copyrightfile) > $refreshinterval`.

Also, this should be broken up into multiple lines. Additional lines should be indented four spaces in below the `if`.

src/etc/inc/pfsense-utils.inc
```diff
@@ -3562,4 +3562,13 @@ function send_user_download($type = 'data', $content, $userfilename = "", $conte
 	exit;
 }
 
+// Test whether the hostname in a URL can be resolved with a very short timeout
+function is_url_hostname_resolvable($url) {
+	$urlhostname = explode("/", $url)[2];
+	putenv("RES_OPTIONS=timeout:3 attempts:1");
+	$resolvable = $urlhostname !== gethostbyname($urlhostname);
```
jim-p: `$urlhostname !== gethostbyname($urlhostname)` should be inside a set of parenthesis.

src/etc/inc/pfsense-utils.inc
```diff
@@ -3562,4 +3562,13 @@ function send_user_download($type = 'data', $content, $userfilename = "", $conte
 	exit;
 }
 
+// Test whether the hostname in a URL can be resolved with a very short timeout
+function is_url_hostname_resolvable($url) {
+	$urlhostname = explode("/", $url)[2];
```
jim-p: Assuming that will return an array with 3+ parts is dangerous. You should be using [parse_url()](https://www.php.net/manual/en/function.parse-url.php) and checking for the appropriate field in the results.

src/etc/inc/pfsense-utils.inc
emmtbot: Sure, done

src/usr/local/www/widgets/widgets/netgate_services_and_support.widget.php
emmtbot: Sure, done, and I tried to find an example to match style.

src/etc/inc/copyget.inc
emmtbot: Sure, done, and I tried to find an example to match style.

src/etc/inc/pfsense-utils.inc
emmtbot: Sure, as you can likely tell I'm not really a PHP guy.  Updated to use parse_url and tested that it still works as expected.

src/etc/inc/pfsense-utils.inc
```diff
@@ -3562,4 +3562,13 @@ function send_user_download($type = 'data', $content, $userfilename = "", $conte
 	exit;
 }
 
+// Test whether the hostname in a URL can be resolved with a very short timeout
+function is_url_hostname_resolvable($url) {
+	$urlhostname = parse_url($url, PHP_URL_HOST);
```
jim-p: I would like to see one more test after this to ensure the result from `parse_url()` is sane. It can return `false` if there is a problem. Something as easy as this would suffice, on the line under this one:

    	if (empty($urlhostname)) {
    		return false;
    	}

That way it won't try to run a potentially invalid value through `gethostbyname()`.

emmtbot: I just added those lines verbatim

Fohdeesha:
thank you so much for this! this has been driving me nuts for YEARS

ikifar2012:
I seem to be experiencing this issue again, when my internet goes down I can't seem to load my dashboard  on 2.5.1-RELEASE

vertex-github:
2.6 is showing this behavior again - this makes pfsense useless to diagnose when we have WAN outages or disconnects.  The moment we have an internet issue, the first thing we look at is pfsense - but that takes so long to respond that its practically unusable.

patrickpesegodinski:
Hello.

Is this procedure valid for version 2.6.0?

Line in green that I should insert in my firewall?

PR #4173, opened 2020-01-27
enable gateway duplicates on ipsec

f-bor:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10214
- [x] Ready for review

This PR enables ipsec phases 1 to share the same endpoint threw a new option.
When the option is enabled, the static route to the endpoint is not added and the automatic rules for the ipsec tunnel are not generated.

### DIFF
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -1675,7 +1675,7 @@ function ipsec_setup_bypasslan() {
  * RESULT
  *   Static routes in the OS routing table for IPsec peers
  ******/
-function ipsec_setup_routes($interface, $family, $sourcehost) {
+function ipsec_setup_routes($interface, $family, $sourcehost, $duplicates) {
 	if (substr($interface, 0, 4) == "_vip") {
 		$vpninterface = get_configured_vip_interface($interface);
 		if (substr($vpninterface, 0, 4) == "_vip") {
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -1219,7 +1219,7 @@ function ipsec_setup_gwifs() {
 					continue;
 				}
 			}
-			if (array_search($rg, $rgmap)) {
+			if (!isset($ph1ent['gw_duplicates']) && array_search($rg, $rgmap)) {
 				log_error(sprintf(gettext("The remote gateway %s already exists on another phase 1 entry"), $rg));
 				continue;
 			}
@@ -1671,6 +1671,7 @@ function ipsec_setup_bypasslan() {
  *   $interface : The interface upon which routes will be configured.
  *   $family    : The address family ('inet' or 'inet6')
  *   $sourcehost: The local source address used for initiating connections.
+ *   $duplicates: If the ipsec tunnel allows duplicates remote peers
  * RESULT
  *   Static routes in the OS routing table for IPsec peers
  ******/
@@ -1700,7 +1701,7 @@ function ipsec_setup_routes($interface, $family, $sourcehost) {
 			/* if the remote gateway is in the local subnet, then don't add a route */
 			if (is_ipaddrv4($sourcehost) &&
 			    !ip_in_subnet($sourcehost, "{$subnet_ip}/{$subnet_bits}")) {
-				if (is_ipaddrv4($gatewayip)) {
+				if (is_ipaddrv4($gatewayip) && !$duplicates) {
 					// log_error("IPSEC interface is not WAN but {$ifacesuse}, adding static route for VPN endpoint {$rgip} via {$gatewayip}");
 					route_add_or_change("-host {$sourcehost} {$gatewayip}");
 				}
@@ -1715,7 +1716,7 @@ function ipsec_setup_routes($interface, $family, $sourcehost) {
 			/* if the remote gateway is in the local subnet, then don't add a route */
 			if (is_ipaddrv6($sourcehost) &&
 			    !ip_in_subnet($sourcehost, "{$subnet_ip}/{$subnet_bits}")) {
-				if (is_ipaddrv6($gatewayip)) {
+				if (is_ipaddrv6($gatewayip) && !$duplicates) {
 					// log_error("IPSEC interface is not WAN but {$ifacesuse}, adding static route for VPN endpoint {$rgip} via {$gatewayip}");
 					route_add_or_change("-inet6 -host {$sourcehost} {$gatewayip}");
 				}
@@ -2018,7 +2019,7 @@ function ipsec_setup_tunnels() {
 		} else {
 			$remote_spec = $ph1ent['remote-gateway'];
 			$sourcehost = (is_ipaddr($remote_spec)) ? $remote_spec : $rgmap[$remote_spec];
-			$ifacesuse = ipsec_setup_routes($ph1ent['interface'], $ph1ent['protocol'], $sourcehost);
+			$ifacesuse = ipsec_setup_routes($ph1ent['interface'], $ph1ent['protocol'], $sourcehost, isset($ph1ent['gw_duplicates']));
 		}
 
 		/* Setup IKE proposals */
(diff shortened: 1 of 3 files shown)


src/etc/inc/ipsec.inc
```diff
@@ -1671,6 +1671,7 @@ function ipsec_setup_bypasslan() {
  *   $interface : The interface upon which routes will be configured.
  *   $family    : The address family ('inet' or 'inet6')
  *   $sourcehost: The local source address used for initiating connections.
+ *   $duplicates: If the ipsec tunnel allows duplicates remote peers
```
rbgarga: This variable was not added to function prototype

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -115,6 +115,12 @@
 	$pconfig['descr'] = $a_phase1[$p1index]['descr'];
 	$pconfig['nat_traversal'] = $a_phase1[$p1index]['nat_traversal'];
 	$pconfig['mobike'] = $a_phase1[$p1index]['mobike'];
+	$pconfig['gw_duplicates'] = $a_phase1[$p1index]['gw_duplicates'];
+
+	if (isset($a_phase1[$p1index]['gw_duplicates'])) {
+		$pconfig['gw_duplicates'] = true;
+	}
```
rbgarga: I didn't get the idea here, first you attribute the value of `$a_phase1[$p1index]['gw_duplicates']` to `$pconfig['gw_duplicates']` and then, if former is set you attribute true to`$pconfig['gw_duplicates']`.  What are you trying to achieve?

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -500,6 +506,14 @@
 		$ph1ent['descr'] = $pconfig['descr'];
 		$ph1ent['nat_traversal'] = $pconfig['nat_traversal'];
 		$ph1ent['mobike'] = $pconfig['mobike'];
+
+		if ( isset($pconfig['gw_duplicates'])) {
+			$ph1ent['gw_duplicates'] = true;
+		}
+		else {
```
rbgarga: Move else to the same line of `}`

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -933,6 +947,13 @@ function build_eal_list() {
 	array('on' => gettext('Enable'), 'off' => gettext('Disable'))
 ))->setHelp('Set this option to control the use of MOBIKE');
 
+$section->addInput(new Form_Checkbox(
+	'gw_duplicates',
+	'Gateway duplicates',
+	'Enable this to allow multiple phase 1 configurations with the same endpoint. When enabled, you are responsible for doing proper source routing configuration to the endpoints. When disabled, pfSense will create a static route and routing rules to the remote gateway.',
```
rbgarga: Please break this huge line to make it easy to read

f-bor:
@rbgarga Hi, I fixed the requested changes. Sorry about that.

rbgarga:
> @rbgarga Hi, I fixed the requested changes. Sorry about that.

Don't apologize.  You are sending a contribution and it's very welcome.  Thanks!

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -933,6 +945,15 @@ function build_eal_list() {
 	array('on' => gettext('Enable'), 'off' => gettext('Disable'))
 ))->setHelp('Set this option to control the use of MOBIKE');
 
+$section->addInput(new Form_Checkbox(
+	'gw_duplicates',
+	'Gateway duplicates',
+	'Enable this to allow multiple phase 1 configurations with the same endpoint. ' .
+		'When enabled, you are responsible for doing proper source routing configuration to the endpoints. '.
```
jim-p: Please reword this to avoid addressing the user directly ("You"). Also this is not "source routing".

For example, replace it with "When enabled, pfSense does not manage routing to the remote gateway and traffic will follow the default route without regard for the chosen interface. Static routes can override this behavior."

f-bor: Thanks. "traffic will follow the default route without regard for the chosen interface" may be confusing.

It took me some time to figure out how to properly setup my pfSenses for the routing part and the solution was to set the upstream gateway on the chosen interfaces and doing nothing else (which is probably obvious for an experimented pfSense user). I assumed this is because the left endpoint of the phase 1 is set with the address of the chosen interface, so the pass out route-to rule is applied and the traffic is correctly routed without further configuration.

Mentionning static routes may be confusing too (at least for me since I don't see how to set them to have something working with the same endpoint).

Maybe 'When enabled, it is advised to set an upstream gateway on the chosen interfaces for the duplicates phase 1." ?

jim-p: That wording has its own problems, since that may not be enough. The most reliable way to make sure it always works is with static routes. Setting an upstream gateway is always required for WAN type interfaces, so if you didn't do that, that was an unrelated misconfiguration. That particular behavior isn't specific to IPsec or routing to peers. It may have affected this, but it wouldn't necessarily solve the root problem, which is ensuring that all traffic of all types going to the remote peer use the appropriate interface. You have to keep in mind that different tunnels in different environments may behave in other ways (With/without NAT-T, ESP, different types of interfaces, etc). We have to cover as many scenarios as possible, and what worked in your environment may not work for others.

We could maybe replace "traffic will follow the default route without regard for the chosen interface" with "traffic will follow the system routing table", but I'd still prefer to see the text appear as I originally wrote.

f-bor: > The most reliable way to make sure it always works is with static routes. 
That's a way I don't understand (and I'd like to). Since we can have only one static route for a given destination, do you mean using multiple routing tables ?

> but I'd still prefer to see the text appear as I originally wrote.
Ok.

jim-p: > That's a way I don't understand (and I'd like to). Since we can have only one static route for a given destination, do you mean using multiple routing tables ?

You can only have one route to one destination, which is why the original restriction was put in place. But it can't happen automatically with this option enabled, so the user has to add the equivalent route manually. In other words, if you allow multiple connections to the same peer, they have to all use the same interface, or it won't route correctly. There is a chance that it may work with multiple interfaces when set as "responder only" since `reply-to` will help in that case, but when initiating, you can't necessarily ensure the traffic will take the correct path with interface binding alone.

It may work for some people in some scenarios, but it won't be consistent. One of many reasons I hesitate to allow changes like this through. It may appear to work, but it may not work consistently or for many different scenarios.

But since we can't put that much warning text in the GUI, at least a routing warning will have to do.

f-bor: Thanks for your answers. I agree for the routing warning, it's the difficulty here. I have followed your wording.

For what it worth, in my setup it's working well with two simple ethernet ipv4 wan interfaces per pfSense, each connected to a wan router (same thing on the remote endpoint), for a total of 4 tunnels between the 2 endpoints using all interface combinations. I have tested etablishing the tunnels from both sides on the 4 tunnels and the routing was correct, using a different combination each time (I checked with tcpdump). Something similar could probably be achieved without upstream gateways but using rules with gateways.

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -933,6 +945,15 @@ function build_eal_list() {
 	array('on' => gettext('Enable'), 'off' => gettext('Disable'))
 ))->setHelp('Set this option to control the use of MOBIKE');
 
+$section->addInput(new Form_Checkbox(
+	'gw_duplicates',
+	'Gateway duplicates',
+	'Enable this to allow multiple phase 1 configurations with the same endpoint. ' .
+		'When enabled, pfSense does not manage routing to the remote gateway and traffic will follow the default route '.
+		'without regard for the chosen interface. Static routes can override this behavior.'.
```
jim-p: This line should end in a comma outside the quotes, not a period.

vktg:
tested on 2.5.0.a.20200305.2255:
VPN with two ISP WAN links, and two PH1 entries with the same remoteip,
In this case, remoteip vpn can initiate an IPsec connection with any of these WAN interfaces.
Can be useful for some scenarios.

PR #4177, opened 2020-01-31
Set correct default MTU for GRE,GIF and GRE/IPsec. Issue #10222

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10222
- [ ] Ready for review

Default GRE MTU value is 1476
This is impossible in most cases when GRE/IPsec is used
It can be useful to set default MTU value for such cases to 1400

When you first create GRE/GIF interfaces, pfSense sets the correct MTU for it - 1476/1280
But with any change on the interface edit page, it set MTU to 1500
This PR also fixes this

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -29,6 +29,7 @@
 require_once("globals.inc");
 require_once("util.inc");
 require_once("gwlb.inc");
+require_once("ipsec.inc");
 
 function interfaces_bring_up($interface) {
 	if (!$interface) {
@@ -267,21 +268,6 @@ function interface_is_qinq($if = NULL) {
 		}
 	}
 
-	return (NULL);
-}
-
-function interface_is_lagg($if = NULL) {
-	global $config;
-
-	if (!isset($config['laggs']['lagg']) || !is_array($config['laggs']['lagg'])) {
-		return (NULL);
-	}
-
-	foreach ($config['laggs']['lagg'] as $lagg) {
-		if ($lagg['laggif'] == $if) {
-			return ($lagg);
-		}
-	}
 	return (NULL);
 }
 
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -1156,6 +1142,60 @@ function interface_gre_configure(&$gre, $grekey = "") {
...
+		}
+	}
+	return (NULL);
+}
+
+function is_greipsec($if) {
+	global $config;
+
+	if (ipsec_enabled() && is_array($config['gres']) && is_array($config['ipsec']['phase2'])) {
+		foreach ($config['gres']['gre'] as $gre) {
```
jim-p: `$config['gres']` may be an array (for the previous test) but that doesn't guarantee `$config['gres']['gre']` is. It either needs tested or setup with `init_config_arr()` before this foreach.

src/etc/inc/interfaces.inc
```diff
@@ -3950,6 +3990,17 @@ function interface_configure($interface = "wan", $reloadall = false, $linkupeven
 	if ($wantedmtu == 0 && interface_is_vlan($mtuif) != NULL && interface_isppp_type($interface)) {
 		$wantedmtu = interface_mtu_wanted_for_pppoe($mtuhwif);
 	}
+	if ($wantedmtu == 0 && interface_is_type($mtuif, 'gre')) {
```
jim-p: The first part here needs its own set of parenthesis. I know it matches the block above, but the block above is also wrong and should be fixed. Each test should be isolated in parenthesis.

src/etc/inc/interfaces.inc
```diff
@@ -3950,6 +3990,17 @@ function interface_configure($interface = "wan", $reloadall = false, $linkupeven
...
+		/* set MTU to 1400 for GRE over IPsec */
+		if (is_greipsec($mtuif)) {
+			$wantedmtu = 1400;
+		} else {
+			$wantedmtu = 1476;
+		}
+	}
+	if ($wantedmtu == 0 && interface_is_type($mtuif, 'gif')) {
```
jim-p: See previous note about parenthesis.

vktg: fixed

src/etc/inc/interfaces.inc
vktg: fixed

src/etc/inc/interfaces.inc
vktg: is_array($config['gres']['gre']) added

PR #4178, opened 2020-01-31
Create key and zone section for static DHCP mappings

ableischwitz:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10224
- [X] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1106,7 +1106,9 @@ EOD;
 		if (is_array($dhcpifconf['staticmap'])) {
 
 			$i = 0;
+			$sm_newzone[] = array();
 			foreach ($dhcpifconf['staticmap'] as $sm) {
+				$need_sm_ddns_updates = false;
 				$dhcpdconf .= "host s_{$dhcpif}_{$i} {\n";
 
 				if ($sm['mac']) {
@@ -1154,6 +1156,7 @@ EOD;
 				if (isset($sm['ddnsupdate'])) {
 					if (($sm['ddnsdomain'] <> "") && ($sm['ddnsdomain'] != $dhcpifconf['ddnsdomain'])) {
 						$smdnscfg .= "		ddns-domainname \"{$sm['ddnsdomain']}\";\n";
+				 		$need_sm_ddns_updates = true;	
 					}
 					$smdnscfg .= "		ddns-update-style interim;\n";
 				}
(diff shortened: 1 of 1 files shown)


src/etc/inc/services.inc
```diff
@@ -1153,9 +1155,10 @@ EOD;
 
 				if (isset($sm['ddnsupdate'])) {
 					if (($sm['ddnsdomain'] <> "") && ($sm['ddnsdomain'] != $dhcpifconf['ddnsdomain'])) {
-						$smdnscfg .= "		ddns-domainname \"{$sm['ddnsdomain']}\";\n";
+						$smdnscfg .= "        ddns-domainname \"{$sm['ddnsdomain']}\";\n";
```
jim-p: Why change from tabs to spaces? The rest of the file uses tabs.

src/etc/inc/services.inc
```diff
@@ -1153,9 +1155,10 @@ EOD;
...
+				 		$need_sm_ddns_updates = true;	
 					}
-					$smdnscfg .= "		ddns-update-style interim;\n";
+					$smdnscfg .= "        ddns-update-style interim;\n";
```
jim-p: Why change from tabs to spaces? The rest of the file uses tabs.

src/etc/inc/services.inc
```diff
@@ -1190,6 +1193,18 @@ EOD;
 				}
 
 				$dhcpdconf .= "}\n";
+				// add zone DDNS key/server to $ddns_zone[] if required
+				if ($need_sm_ddns_updates) {
+					$sm_newzone['dns-servers'] = array($sm['ddnsdomainprimary']);
+					$sm_newzone['domain-name'] = $sm['domain'];
```
jim-p: If multiple static maps use the same domain, won't this make multiple entries for the same domain? Is that allowed and functional in ISC DHCPD?

src/etc/inc/services.inc
ableischwitz: I changed it because the resulting output was indented differently than the rest. With that change all configuration options are in line.

src/etc/inc/services.inc
ableischwitz: I changed it because the resulting output was indented differently than the rest. With that change all configuration options are in line.

src/etc/inc/services.inc
jim-p: This looks like it would result in a mix of spaces and tabs, which is not desirable. If something else is using spaces, it should be changed to tabs and not the other way around.

src/etc/inc/services.inc
ableischwitz: I was thinking about this as well, but even though a new zone-entry for `$ddns_zones[]` is added multiple times, it will get filtered by `function dhcpdzones($ddns_zones)`:
```
function dhcpdzones($ddns_zones) {
...
                        // We don't need to add zones multiple times.
                        if ($zone['domain-name'] && !in_array($zone['domain-name'], $added_zones)) {
...
```

jim-p: If it will be skipped, then we should not bother to add it in the first place, rather than stuffing the array full of potentially redundant information. Also if it's duplicated across multiple interfaces/entries then it probably needs to be made clear to the user that only the first defined set of options per domain will be honored.

ableischwitz: In case the information entered on a per-node basis is not added to that array, it will be skipped entirely. We could add a check if an entry is already pushed to that array, but it would be duplicated code. If you come up with a better solution, I'm fine with it. The current implementation offers options, which are simply ignored as of now.

rbgarga:
@ableischwitz do you still plan to work on this change?  If so, please rebase your fork and fix conflicts

rbgarga:
Closing due to lack of activity for 6 months

PR #4182, opened 2020-02-01
Make disabled firewall rules red instead of grey in dark theme

lindhe:
With this change, in the dark theme, firewall rules are marked with red text instead of light grey. The light grey text had very little contrast to the slightly darker grey background, so it was hard to see. The red text is much more noticeable.

I spent hours yesterday debugging an issue with my network. Turned out that I had disabled some firewall rules, and forgot to turn them back on again. I checked the rules multiple times just to make sure they made sense, but the very slight difference in contrast made me not register that some rules were disabled. I've got perfectly good eye-sight with no color blindness or anything, it was just that the contrast difference was not clear enough for me to register.

I don't think it looks too bad either:
**Old style:**
Old style
**New style:**
New style

- [x] Ready for review

### DIFF
--- a/src/usr/local/www/css/pfSense-dark.css
+++ b/src/usr/local/www/css/pfSense-dark.css
@@ -284,12 +284,12 @@ caption {
 
 .table-striped>tbody>tr.disabled:nth-of-type(odd) {
     background-color: rgba(48, 48, 48, .5);
-    color: rgba(255,255,255,.5);
+    color: rgba(255,55,55,1);
 }
 
 .table-striped>tbody>tr.disabled:nth-of-type(even) {
     background-color: rgba(66, 66, 66, .5);
-    color: rgba(255,255,255,.5);
+    color: rgba(255,55,55,1);
 }
 
 .table-hover>tbody>tr.disabled:hover, .table-striped>tbody>tr.disabled:nth-of-type(odd):hover, .table-striped>tbody>tr.disabled:nth-of-type(even):hover {


jim-p:
Light red on grey background has even worse contrast. At least to my eyes, that is much more difficult to read.

Maybe try changing the background color instead?

lindhe:
Thanks for the input. I haven't had time to try stuff, and quite honestly have no eye for design at all. I have no idea what colors to try. Any suggestions?

rbgarga:
@jdillard can you suggest what color is best to be used in this case?

jdillard:
Would a left border work as a signifier of the status? Here are two options along with code sample (I personally prefer the first, but both are valid):

Screen Shot 2020-09-14 at 3 24 17 PM

For this one, just add the following CSS rule (don't include the font color change rule that is in place now):

```
.table-striped>tbody>tr.disabled>td:first-child {
    border-left: 3px solid #F44336;
    padding-left: 7px;
}
```

Screen Shot 2020-09-14 at 3 23 52 PM

For this one, add the following rule in addition to, and below, the previous CSS rule:

```
.table-striped>tbody>tr:not(.disabled)>td:first-child {
    border-left: 3px solid #4CAF50;
} 
```

lindhe:
That's a huge improvement! Let's go for that, then. Since I'm not very active in this project, it's quite a threshold for me to bring it up again and remember how to change and test things. So either I incorporate this change (which I probably have time for this year), or you can do it and close this PR.

First commit wins. ;)

jdillard:
I added the first rule here: https://github.com/pfsense/pfsense/commit/8774f2c7d9e0d3b77f7a8f71129e60a5b59ac781

It should show up in the next build. Closing this due to the above commit that supersedes this PR.

PR #4184, opened 2020-02-07
DHCP group known clients by interface. Issue #1605

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/1605
- [ ] Ready for review

 When set to 'Allow all clients', any DHCP client will get an IP address within this scope/range on this interface. 
 If set to 'Allow known clients from any interface', any DHCP client with a MAC address listed on 'any' scope(s)/interface(s) will get an IP address. 
 If set to 'Allow known clients from only this interface', only MAC addresses listed below (i.e. for this interface) will get an IP address within this scope/range

This is squashed & tested PR https://github.com/pfsense/pfsense/pull/4066
Tested on pfSense 2.5.0.a.20200206.2005

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -789,6 +789,9 @@ EOPP;
 			}
 		}
 
+		// instantiate class before pool definition
+		$dhcpdconf .= "class \"s_{$dhcpif}\" {\n	match pick-first-value (option dhcp-client-identifier, hardware);\n}\n";
+
 		$dhcpdconf .= "subnet {$subnet} netmask {$subnetmask} {\n";
 
 		// Setup pool options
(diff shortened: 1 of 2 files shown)


src/etc/inc/services.inc
```diff
@@ -1190,6 +1201,10 @@ EOD;
 				}
 
 				$dhcpdconf .= "}\n";
+
+				// subclass for DHCP limiting
+				$dhcpdconf .= "subclass \"s_{$dhcpif}\" 1:{$sm['mac']};\n"; // assuming ALL addresses are ethernet hardware type ("1:" prefix)
```
jim-p: What if the MAC is empty? Entries can be specified based on MAC or Client ID. Either one could be empty.

src/etc/inc/services.inc
```diff
@@ -789,6 +789,9 @@ EOPP;
 			}
 		}
 
+		// instantiate class before pool definition
+		$dhcpdconf .= "class \"s_{$dhcpif}\" {\n	match hardware;\n}\n";
```
jim-p: What if the MAC is empty? Entries can be specified based on MAC or Client ID. Either one could be empty.

The `dhcpd.conf` man page has an example that might work:

    match pick-first-value	(option	dhcp-client-identifier,	hardware);

src/usr/local/www/services_dhcp.php
```diff
@@ -163,7 +163,17 @@
 	list($pconfig['wins1'], $pconfig['wins2']) = $dhcpdconf['winsserver'];
 	list($pconfig['dns1'], $pconfig['dns2'], $pconfig['dns3'], $pconfig['dns4']) = $dhcpdconf['dnsserver'];
 	$pconfig['ignorebootp'] = isset($dhcpdconf['ignorebootp']);
-	$pconfig['denyunknown'] = isset($dhcpdconf['denyunknown']);
+
+	if (isset($dhcpdconf['denyunknown'])) {
+		if ($dhcpdconf['denyunknown'] == "") {
```
jim-p: `empty()` would be better here.

Or even collapse this whole inner if/else to a single line:

    $pconfig['denyunknown'] = empty($dhcpdconf['denyunknown']) ? "enabled" : $dhcpdconf['denyunknown'];

vktg: done

src/etc/inc/services.inc
vktg: done

src/etc/inc/services.inc
vktg: MAC or Client ID, or both are added to config now

PR #4190, opened 2020-02-12
Remove extra 00 padding of VTI interface names. Issue #9592

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9592
- [x] Ready for review

FreeBSD will not accept an interface number greater than 32767, therefore this padding limits the number of VTI interfaces to 32.
This change will increase this to 3276.

This is squashed & resolved PR https://github.com/pfsense/pfsense/pull/4071


src/etc/inc/interfaces.inc
```diff
@@ -1356,6 +1356,57 @@ function interface_gif_configure(&$gif, $gifkey = "") {
...
+	}
+	
+	if ((($ikeid < 33) && ($idx < 10)) || (($ikeid < 10) && ($idx < 100))) {
+		$oldformat = "{$ikeid}00{$idx}";
+		$config['ipsec']['vtimaps'][]=array(
+			"reqid" => $ikeid,
+			"index" => $idx,
+			"ifnum" => $oldformat
+		);
+		write_config(gettext("New ipsec vtimap added"));
```
rbgarga: I still don't like the idea of this function to change config.  It would be better if the conversion from old format to new format happens on upgrade_config.inc with a version config bump

vktg: how can I fix this?
there is no upgrade_config.inc in the master branch
I only see this on installed pfSense

rbgarga: Yes, there is an upgrade_config.inc on master branch, at:

https://github.com/pfsense/pfsense/blob/master/src/etc/inc/upgrade_config.inc

vktg: ok, done
now upgrade_201_to_202() converts all IPsec VTI (including disabled) to new <vtimap> format

rbgarga: You need to bump `latest_config` at globals.inc to 20.2 otherwise that function will never be called.
https://github.com/pfsense/pfsense/blob/master/src/etc/inc/globals.inc#L73

Also, since upgrade_config.inc will take care of converting it, this block can be removed from `get_ipsecifnum()`

vktg: this block is used to create new <vtimap> entries
but there is no any mechanism to remove unused <vtimap>
In Progress...

vktg: All done

Now it correctly updates config.xml when upgrading,
Deletes unused vtimaps after deleting P1/P2 in WebGUI,
Deletes/creates vtimaps on switching from IKEv1/IKEv2 with split connections to IKEv2 and vise-versa

jim-p: Why is it still doing `config_write()` calls in the function (here and once more below)? If it's happening on upgrade, it should be unnecessary to do a write in the function. It should only be done when creating/deleting a VTI, shouldn't it?

src/etc/inc/interfaces.inc
```diff
@@ -1356,6 +1356,80 @@ function interface_gif_configure(&$gif, $gifkey = "") {
...
+/*
+ * $ipsecifnum = get_ipsecifnum($ikeid, $idx, $create=true);
+ * locates or creates and returns an ipsecifnum in the config.
+ */
+function get_ipsecifnum($ikeid, $idx, $create = true) {
+	global $config;
+
+	if (is_array($config['ipsec']['vtimap'])) {
+		$matches = array_filter($config['ipsec']['vtimap'], function($val) use($ikeid, $idx){
+			return ($val['reqid'] == $ikeid and $val['index'] == $idx);
```
jim-p: Each test should be in its own set of parenthesis, plus `and` should probably be `&&`. I don't see anything here that would require using the lower precedence of `and` vs `&&`.

vktg: In progress
too many config writes, need to move it to vpn_ipsec_phase1.php and vpn_ipsec_phase2.php

src/etc/inc/ipsec.inc
```diff
@@ -2316,6 +2324,17 @@ function ipsec_setup_tunnels() {
 				$children[$cname]['local_ts'] = implode(",", $local_ts_spec);
 			}
 			if ($has_vti) {
+				// cleanup all vtimaps after IKEv1 or Split Connections
+				if (is_array($config['ipsec']['vtimap'])) {
+					$a_vtimap = &$config['ipsec']['vtimap'];
+					foreach ($a_vtimap as $id => $vtimap) {
+						if ($vtimap['reqid'] == $ph1ent['ikeid']) {
+							unset($a_vtimap[$id]);
```
jim-p: If this is making changes, should this not also set `$vtimap_save` to `true` so those changes get stored at the end?

vktg: Now it's completely rewritten

No any extra write_config()
Correct creating/deleting vtimaps on IKEv1/IKEv2 switching
Correct creating/deleting vtimaps on Phase 2 mode changing

rbgarga:

vktg:
> @vktg please rebase your fork and fix conflicts

Rebased

rbgarga:

vktg:
Fixed!

PR #4078, opened 2019-08-13
Feature #9661

kiokoman:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9661
- [x] Ready for review

PR #4191, opened 2020-02-12
Increase DHCP/DHCPv6 NTP server options to three. Issue #9661

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9661
- [ ] Ready for review

This is squashed and tested PR https://github.com/pfsense/pfsense/pull/4078

It is considered a general best practice to use at least three NTP servers to help identify "falsetickers" (bad clock sources). Unfortunately the pfSense DHCP/DHCPv6 web UI (v2.4.4-RELEASE-p3 at the time of writing) doesn't allow you to specify more than two NTP servers in the "Other Options".


vktg:
works fine on 2.5.0.a.20200211.1811,
I can see three NTP servers in the packet capture

jim-p:
The other PR is identical (except the commits), is still open, is approved, has no unresolved conflicts, etc. This one has less commits but that alone doesn't seem like enough reason to warrant a copy of the PR.

PR #4192, opened 2020-02-12
OpenVPN RADIUS ACL enhancements. Issue #9206

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9206
- [ ] Ready for review

* fixed a race condition where disconnecting and reconnecting a session when using AVPair ACLs would result in no rules being inserted.
* Updated the ACL parser to support IP and ICMP protocol options
* Updated the ACL parser to support port operators.
* Added a template variable "{clientip}" that is replaced with the connecting clients VPN IP.

This is squashed, resolved (@hatclub notes) and tested PR https://github.com/pfsense/pfsense/pull/4026

tested on pfSense 2.5.0.a.20200211.1811 with freeradius3 0.15.7_10
OpenVPN 2.4.7 client (Debian 10)

### DIFF
--- a/src/usr/local/sbin/openvpn.attributes.sh
+++ b/src/usr/local/sbin/openvpn.attributes.sh
@@ -20,16 +20,53 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+lockfile="/tmp/ovpn_${dev}_${username}_${trusted_port}.lock"
+rulesfile="/tmp/ovpn_${dev}_${username}_${trusted_port}.rules"
+anchorname="openvpn/${dev}_${username}_${trusted_port}"
+
 if [ "$script_type" = "client-connect" ]; then
+	i=1
+	while [ -f "${lockfile}" ]; do
+		if [ $i -ge 30 ]; then
+			/bin/echo "Timeout while waiting for lockfile"
+			exit 1
+		fi
+
+		/bin/sleep 1
+		i=$(( i + 1 ))
+	done
+	/usr/bin/touch "${lockfile}"
+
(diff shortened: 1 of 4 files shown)


src/usr/local/sbin/openvpn.attributes.sh
```diff
@@ -20,16 +20,53 @@
...
+			/bin/echo "Timeout while waiting for lockfile"
+			exit 1
+		fi
+
+		/bin/sleep 1
+		i=$(( i + 1 ))
+	done
+	/usr/bin/touch "${lockfile}"
+
+	/bin/cat "${rulesfile}" | sed "s/{clientip}/${ifconfig_pool_remote_ip}/g" > "${rulesfile}.tmp" && mv "${rulesfile}.tmp" "${rulesfile}"
```
jim-p: Full paths are missing from `sed` and `mv`

hatclub:
```suggestion
	/bin/cat "${rulesfile}" | /usr/bin/sed "s/{clientip}/${ifconfig_pool_remote_ip}/g" > "${rulesfile}.tmp" && /bin/mv "${rulesfile}.tmp" "${rulesfile}"
```

hatclub: Whoops, I created a new change instead of putting it here - sorry. :\

hatclub:
```suggestion
	/bin/cat "${rulesfile}" | /usr/bin/sed "s/{clientip}/${ifconfig_pool_remote_ip}/g" > "${rulesfile}.tmp" && /bin/mv "${rulesfile}.tmp" "${rulesfile}"
```

vktg: thanks!
I added it to the squashed commit

PR #4194, opened 2020-02-15
Accomodate both RADIUS and pool IP addresses in IPsec. Issue #8160

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8160
- [ ] Ready for review

fixed and tested copy of https://github.com/pfsense/pfsense/pull/3976: by @Amith211

> I have a requirement to be able to have IPSec VPN clients receive their IP addresses via RADIUS for some groups and others from a simple pool. I decided to have a stab at implementing the solution mentioned in Redmine issue 8160.

What done:
* Adopted to swanctl.conf https://redmine.pfsense.org/issues/9603,
From https://wiki.strongswan.org/projects/strongswan/wiki/VirtualIp#RADIUS-backend:

> the order in which they are queried for virtual IPs depends on the plugin load order (in-memory pools are provided by the stroke and vici plugins, respectively). The order in rightsourceip or pools is irrelevant unless multiple in-memory pools from the same backend are defined.

this is why we need increase the priority of eap-radius plugin by `$ssconf['charon']['plugins']['eap-radius']['load'] = "2";`
see https://wiki.strongswan.org/projects/strongswan/wiki/PluginLoad:

> Besides simply enabling/disabling plugins the load setting accepts a numeric priority
> value, which the plugin loader uses to decide in which order plugins are loaded. Plugins with the same priority are loaded
> according to the default load order, unknown plugins with the same priority are loaded first and in alphabetical order.
> The default priority is 1, and can also be negative to simplify moving a plugin to the end of the list.

* removed line 
`$ssconf['starter']['config_file'] = "{$g['varetc_path']}/ipsec/ipsec.conf";`
no such file

* Fixed creating correct config file when using EAP-RADIUS without pool ( ipsec_setup_pools() used the ipsec.conf-style %radius)

p.s. as I found you can't see mobile clients that get their IP via RADIUS on the Status / IPsec/ Leases page (only under child sa), and their IPs are not added to the tonatsubnets table

### DIFF
--- a/src/usr/local/www/vpn_ipsec_mobile.php
+++ b/src/usr/local/www/vpn_ipsec_mobile.php
@@ -69,18 +69,23 @@ if (count($a_client)) {
 	$pconfig['wins_server2'] = $a_client['wins_server2'];
 	$pconfig['pfs_group'] = $a_client['pfs_group'];
 	$pconfig['login_banner'] = $a_client['login_banner'];
-
+	$pconfig['radius_ip_priority_enable'] = $a_client['radius_ip_priority_enable'];
+	
 	if (isset($pconfig['enable'])) {
 		$pconfig['enable'] = true;
 	}
 
-	if ($pconfig['pool_address']&&$pconfig['pool_netbits']) {
+	if ($pconfig['pool_address'] && $pconfig['pool_netbits']) {
 		$pconfig['pool_enable'] = true;
 	} else {
 		$pconfig['pool_netbits'] = 24;
 	}
 
-	if ($pconfig['pool_address_v6']&&$pconfig['pool_netbits_v6']) {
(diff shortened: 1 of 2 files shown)


src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -231,6 +236,12 @@
 		}
 	}
 
+	if ($pconfig['radius_ip_priority_enable']) {
+		if (!(isset($mobileph1) && $mobileph1['authentication_method'] == 'eap-radius')) {
```
jim-p: Second test on this line needs parenthesis

vktg: fixed

PR #4195, opened 2020-02-17
Add support for Cloudflare DDNS with API token

netpok:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9639
- [x] Ready for review

This pull request adds support for [Cloudflare API tokens](https://blog.cloudflare.com/api-tokens-general-availability/) in the Dynamic DNS section. API tokens are great because it's scope can be limited to a specific resource unlike in the case of the global api key.

Api tokens can be used by entering the zone id and api token as username/password combination instead of the email and api token.

The zone id is required to reduce the required permissions only to the DNS edit permission, otherwise it would require read permission on the zone and on the account. This information is readily available on the dashboard.

A related question: Why does one to confirm the password, this is more like a login form than a password update form?

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -212,7 +212,7 @@
 				log_error(gettext("Dynamic DNS: updatedns() starting"));
 			}
 
-			$dyndnslck = lock("DDNS".$dnsID, LOCK_EX);
+			$dyndnslck = lock("DDNS" . $dnsID, LOCK_EX);
 
 			if (!$dnsService) $this->_error(2);
 			switch ($dnsService) {
@@ -454,7 +454,7 @@
 				case 'glesys':
 					$needsIP = TRUE;
 					$server = 'https://api.glesys.com/domain/updaterecord/format/json';
-					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser.':'.$this->_dnsPass);
+					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser . ':' . $this->_dnsPass);
 					$post_data['recordid'] = $this->_FQDN;
 					$post_data['data'] = $this->_dnsIP;
 					curl_setopt($ch, CURLOPT_URL, $server);
@@ -467,7 +467,7 @@
 					if (isset($this->_dnsWildcard) && $this->_dnsWildcard != "OFF") {
 						$this->_dnsWildcard = "ON";
 					}
-					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser.':'.$this->_dnsPass);
+					curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser . ':' . $this->_dnsPass);
 					$server = "https://members.dyndns.org/nic/update";
 					$port = "";
 					if ($this->_dnsServer) {
@@ -476,7 +476,7 @@
 					if ($this->_dnsPort) {
 						$port = ":" . $this->_dnsPort;
 					}
-					curl_setopt($ch, CURLOPT_URL, $server .$port . '?system=dyndns&hostname=' . $this->_dnsHost . '&myip=' . $this->_dnsIP . '&wildcard='.$this->_dnsWildcard . '&mx=' . $this->_dnsMX . '&backmx=NO');
+					curl_setopt($ch, CURLOPT_URL, $server . $port . '?system=dyndns&hostname=' . $this->_dnsHost . '&myip=' . $this->_dnsIP . '&wildcard=' . $this->_dnsWildcard . '&mx=' . $this->_dnsMX . '&backmx=NO');
 					break;
 				case 'dhs':
 					// DHS is disabled in the GUI because the following doesn't work.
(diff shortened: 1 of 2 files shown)


src/etc/inc/dyndns.class
```diff
@@ -798,17 +798,28 @@
...
-						'X-Auth-Key: '.$this->_dnsPass.'',
-						'Content-Type: application/json'
-					));
 
-					// Get zone ID
-					$getZoneId = "https://{$dnsServer}/client/v4/zones/?name={$this->_dnsDomain}";
-					curl_setopt($ch, CURLOPT_URL, $getZoneId);
-					$output = json_decode(curl_exec($ch));
-					$zone = $output->result[0]->id;
+					if(strpos($this->_dnsUser, '@') !== false){
```
rbgarga: Add missing space between if and (

src/etc/inc/dyndns.class
```diff
@@ -798,17 +798,28 @@
...
+						curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+							'X-Auth-Email: '.$this->_dnsUser.'',
+							'X-Auth-Key: '.$this->_dnsPass.'',
```
rbgarga: You are concatenating an empty string, please remove it

src/etc/inc/dyndns.class
```diff
@@ -798,17 +798,28 @@
...
-						'X-Auth-Key: '.$this->_dnsPass.'',
-						'Content-Type: application/json'
-					));
 
-					// Get zone ID
-					$getZoneId = "https://{$dnsServer}/client/v4/zones/?name={$this->_dnsDomain}";
-					curl_setopt($ch, CURLOPT_URL, $getZoneId);
-					$output = json_decode(curl_exec($ch));
-					$zone = $output->result[0]->id;
+					if (strpos($this->_dnsUser, '@') !== false){
```
jim-p: Needs a space between `)` and `{` at the end of the line.

src/etc/inc/dyndns.class
```diff
@@ -798,17 +798,28 @@
...
+						curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+							'X-Auth-Email: '.$this->_dnsUser,
+							'X-Auth-Key: '.$this->_dnsPass,
```
jim-p: Should have a space on either side of the `.` operator on both lines here.

src/etc/inc/dyndns.class
```diff
@@ -798,17 +798,28 @@
...
+						));
+
+						// Get zone ID
+						$getZoneId = "https://{$dnsServer}/client/v4/zones/?name={$this->_dnsDomain}";
+						curl_setopt($ch, CURLOPT_URL, $getZoneId);
+						$output = json_decode(curl_exec($ch));
+						$zone = $output->result[0]->id;
+					} else {
+						curl_setopt($ch, CURLOPT_HTTPHEADER, array(
+							'Authorization: Bearer '.$this->_dnsPass,
```
jim-p: Put spaces around `.` here as well

src/etc/inc/dyndns.class
netpok: I did not put there because it wasn't there originally, I fixed all the concatenations in this file based on your instruction.

cpick:
In May Cloudflare [improved their API](https://community.cloudflare.com/t/bug-zone-detail-by-name-requires-zone-list-permission/128042/14) so that a token scoped to a particular zone now has permission to `GET https://api.cloudflare.com/client/v4/zones/?name=<specific_zone_name_here>`.

As a result, a further update could be made so that the zone id doesn't have to be provided as the username (perhaps the password would be assumed to be a token if the username was blank?), but instead be queried like it is in the old API key scenario.

netpok:
So you could replace the zone id with the name of the zone?

cpick:
Basically I think we could change [these lines](https://github.com/pfsense/pfsense/blob/master/src/etc/inc/dyndns.class#L870-L875) from:
```perl
if (strpos($this->_dnsUser, '@') !== false) {
	curl_setopt($ch, CURLOPT_HTTPHEADER, array(
		'X-Auth-Email: ' . $this->_dnsUser,
		'X-Auth-Key: ' . $this->_dnsPass,
		'Content-Type: application/json'
	));
```

To something like:
```perl
if ((!$this->_dnsUser) || (strpos($this->_dnsUser, '@') !== false)) {
	if (strpos($this->_dnsUser, '@') !== false) {
		curl_setopt($ch, CURLOPT_HTTPHEADER, array(
			'X-Auth-Email: ' . $this->_dnsUser,
			'X-Auth-Key: ' . $this->_dnsPass,
			'Content-Type: application/json'
		));
	} else {
		curl_setopt($ch, CURLOPT_HTTPHEADER, array(
			'Authorization: Bearer ' . $this->_dnsPass,
			'Content-Type: application/json'
		));
	}
	// Keep going to query Zone ID ...
```

And [this argument validation](https://github.com/pfsense/pfsense/blob/master/src/etc/inc/dyndns.class#L239) earlier on would need to be loosened in the Cloudflare cases:
```perl
if (!$dnsUser) $this->_error(3);
```

Then if the user leaves the user field blank, the password will be assumed to be a token.  The token can still be scoped to the specific Zone and the Zone ID will be queried automatically like it is when  a global API key is used.

(I'm hesitating to make a pull request with the changes myself because I've never written any perl, I'm sort of copying and pasting here.)

netpok:
I checked* and I think you're right and don't worry about your lack of perl knowledge, since this is written in php.

* not tested

cpick:
I've given it a crack in https://github.com/pfsense/pfsense/pull/4482

PR #4199, opened 2020-02-21
Complete IPv6 router address in radvd.conf prefix. Issue #9710

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9710
- [x] Ready for review

When setting router mode to "unmanaged" and not specifying any prefix, pfSense does not send the advertising interface's address in the RA's prefix option even though the "R" flag is being set. Instead it only announces the prefix, with all host bits set to zero. This is in violation of RFC 6275, which states on page 65 regarding the "R" flag: "When set, indicates that the Prefix field contains a complete IP address assigned to the sending router."

see https://tools.ietf.org/html/rfc6275#page-65:

This fix set full IPv6 router address in prefix field, and removes 'AdvRouterAddr on' for subnets, because (radvd.conf(5)):

> AdvRouterAddr on|off
> When set, indicates that the address of interface is sent instead of network prefix, as is required by Mobile IPv6. When set, minimum limits specified by Mobile IPv6 are used for MinRtrAdvInterval and MaxRtrAdvInterval.

> Default: off

tested with Cisco IOS 15.5(3) client:

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -158,7 +158,12 @@ function services_radvd_configure($blacklist = array()) {
 				$radvdconf .= "\tAdvOtherConfigFlag on;\n";
 				break;
 		}
-		$radvdconf .= "\tprefix {$subnetv6}/{$ifcfgsnv6} {\n";
+		if ($dhcpv6ifconf['ramode'] == "unmanaged") {
+			$prefixnet = $ifcfgipv6;
+		} else {
+			$prefixnet = $subnetv6;
+		}
+		$radvdconf .= "\tprefix {$prefixnet}/{$ifcfgsnv6} {\n";
 		if ($racarpif == true) {
 			$radvdconf .= "\t\tDeprecatePrefix off;\n";
 		} else {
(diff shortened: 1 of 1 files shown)


jim-p:
If the misbehavior only affects "unmanaged" mode, why is it being changed unconditionally for all modes?

IsaacFL:
The reason it would affect all modes, is that Router Announcements are required for all modes, it is only the flag settings that change change for the different modes.

In this case, the AdvRouterAddr flag should be set to 0 for all of the pfsense supported modes. 

As far as I can tell this flag is to be used for the Mobility Support RFCs 6275 and 5555.

RFC 8504 IPv6 Node Requirements Best Current Practice 220 Par 12, basically says "more usage and deployment experience is needed with mobility before any specific approach can be recommended for broad implementation in all hosts and routers"

IsaacFL:
That being said, the Prefix itself, (once you remove the AdvRouterAddr entry) should be all zeros after the prefix as it currently is.
RFC 4861 Par 4.6.2 "The bits in the prefix after the prefix length are reserved and MUST be initialized to zero by the sender and ignored by the receiver."

IsaacFL:
I am not sure what the change in src/etc/inc/services.inc line 161, does, but if it is sending the interface address instead of the prefix only, I don't think that is correct per RFC 4861.

src/etc/inc/services.inc
```diff
@@ -158,7 +158,7 @@ function services_radvd_configure($blacklist = array()) {
 				$radvdconf .= "\tAdvOtherConfigFlag on;\n";
 				break;
 		}
-		$radvdconf .= "\tprefix {$subnetv6}/{$ifcfgsnv6} {\n";
+		$radvdconf .= "\tprefix {$ifcfgipv6}/{$ifcfgsnv6} {\n";
```
jim-p: Why use the interface address here? As mentioned in the main comments, that doesn't seem to be correct

vktg: Fixed
Now it uses full ip address only for unmanaged mode

PR #4200, opened 2020-02-21
DNS64 support. Issue #10274

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10274
- [ ] Ready for review

Add DNS64 configuration to DNS Resolver WebGUI

from https://github.com/monero-project/unbound/blob/master/doc/README.DNS64:

> To enable DNS64 functionality in Unbound, two directives in unbound.conf must
> be edited:
> 
> 1. The "module-config" directive must start with "dns64". For example:
> 
>     module-config: "dns64 validator iterator" 
> 
> 2. The "dns64-prefix" directive indicates your DNS64 prefix. For example:
> 
>     dns64-prefix: 64:FF9B::/96
> 
> The prefix must be a /96 or shorter.

if there is no 'dns64-prefix: xxx' option, default 64:FF9B::/96 prefix is used

During this coding this PR, I also found issue with addMask() -
if you are trying to use max netmask shorter that 128 for IPv6 or shorter that 32 for IPv4, js code resets netmask size to 128/32, 
this not allow to use /96 max size for dns64-prefix, or, you can see this issue on 1:1 NAT page - try to enter any IPv4 address to the destination network field, and you can that it allow you to select /32 range (but only /31 is allowed in code)
IpAddress.class.php fixed

### DIFF
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -175,6 +175,17 @@ function unbound_generate_config_text($unboundcfg = NULL, $cfgsubdir = "") {
 		$module_config .= 'python ';
 	}
 
+	// Setup DNS64 support
+	if (isset($unboundcfg['dns64'])) {
+		$module_config .= 'dns64 ';
+		$dns64_conf = 'dns64-prefix: ';
+		if (is_subnetv6($unboundcfg['dns64prefix'] . '/' . $unboundcfg['dns64netbits'])) {
+			$dns64_conf .= $unboundcfg['dns64prefix'] . '/' . $unboundcfg['dns64netbits'];
+		} else {
+			$dns64_conf .= '64:ff9b::/96';
+		}
+	}
+
 	// Setup DNSSEC support
 	if (isset($unboundcfg['dnssec'])) {
 		$module_config .= 'validator ';
@@ -487,6 +498,7 @@ serve-expired: {$dns_record_cache}
 # DNS Rebinding
 {$private_addr}
 {$private_domains}
+{$dns64_conf}
 
 # Access lists
 include: {$g['unbound_chroot_path']}{$cfgsubdir}/access_lists.conf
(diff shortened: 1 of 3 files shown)


src/usr/local/www/services_unbound_advanced.php
```diff
@@ -143,6 +149,13 @@
 		if (isset($_POST['dnssecstripped']) && !isset($config['unbound']['dnssec'])) {
 			$input_errors[] = gettext("Harden DNSSEC Data option can only be enabled if DNSSEC support is enabled.");
 		}
+		if (isset($_POST['dns64']) && !empty($_POST['dns64prefix']) && !is_ipaddrv6($_POST['dns64prefix'])) {
+			$input_errors[] = gettext("DNS64 Prefix should be valid IPv6 prefix.");
```
jim-p: Change "should" to "must"

src/usr/local/www/services_unbound_advanced.php
```diff
@@ -143,6 +149,13 @@
...
+		}
+		if (isset($_POST['dns64']) && isset($_POST['dns64netbits']) && 
+		    (($_POST['dns64netbits'] > 96) || ($_POST['dns64netbits'] < 1))) {
+			$input_errors[] = gettext("DNS64 subnet should be not more than 96 and not less that 1.");
```
jim-p: Change "should" to "must"

src/etc/inc/unbound.inc
```diff
@@ -175,6 +175,14 @@ function unbound_generate_config_text($unboundcfg = NULL, $cfgsubdir = "") {
 		$module_config .= 'python ';
 	}
 
+	// Setup DNS64 support
+	if (isset($unboundcfg['dns64'])) {
+		$module_config .= 'dns64 ';
+		if (!empty($unboundcfg['dns64prefix']) && !empty($unboundcfg['dns64netbits'])) {
+			$dns64_conf = 'dns64-prefix: ' . $unboundcfg['dns64prefix'] . '/' . $unboundcfg['dns64netbits'];
```
jim-p: The GUI portion mentions using a default value for the prefix if the user does not specify a prefix value. The linked doc from unbound doesn't say that it assumes one if the directive is missing, so it looks like this needs to have a check if the supplied values are empty/invalid here which then adds the directive with default values instead.

Something like this:

    $dns64_conf = 'dns64-prefix: ';
    if (is_subnetv6($unboundcfg['dns64prefix'] . '/' . $unboundcfg['dns64netbits'])) {
    	$dns64_conf .= gen_subnetv6($unboundcfg['dns64prefix'], $unboundcfg['dns64netbits']);
    } else {
    	$dns64_conf .= "64:ff9b::/96";
    }

vktg: all fixed

PR #4201, opened 2020-02-22
NTP WebGUI checkbox fix. Issue #10276

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10276
- [ ] Ready for review

When you add a new server(s) on the Services / NTP / Settings page, it does not save the values of the checkboxes flags for the first time, you need to set them again and click "save".

This is caused by incorrect processing of POST values.

### DIFF
--- a/src/usr/local/www/services_ntpd.php
+++ b/src/usr/local/www/services_ntpd.php
@@ -36,6 +36,7 @@ require_once("shaper.inc");
 
 global $ntp_poll_min_default, $ntp_poll_max_default;
 $ntp_poll_values = system_ntp_poll_values();
+$auto_pool_suffix = "pool.ntp.org";
 
 if (!is_array($config['ntpd'])) {
 	$config['ntpd'] = array();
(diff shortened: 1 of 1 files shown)


vktg:
I added an additional input check to prevent the simultaneous use of "No Select" and "Is a Pool" to resolve the redmine issue.

src/usr/local/www/services_ntpd.php
```diff
@@ -70,6 +70,12 @@
 		$input_errors[] = gettext("The supplied value for Maximum Poll Interval is invalid.");
 	}
 
+	for ($i = 0; $i < NUMTIMESERVERS; $i++) {
+		if (isset($pconfig["servselect{$i}"]) && isset($pconfig["servispool{$i}"])) {
```
jim-p: If you look at https://github.com/pfsense/pfsense/blob/master/src/etc/inc/system.inc#L1776 and https://github.com/pfsense/pfsense/blob/master/src/usr/local/www/services_ntpd.php#L282 you can see that the system assumes an address is a pool if it ends with the value in `$auto_pool_suffix` (currently `pool.ntp.org`), so this validity check should also perform the same test.

You will need to move the definition of `$auto_pool_suffix` up higher (or perhaps make it global in globals.inc and change all references to it so it reads from there).

vktg: ok, fixed

PR #4202, opened 2020-02-22
Register OpenVPN client IPv6 address in DNS Resolver. Issue #8624

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8624
- [ ] Ready for review

Currently it only handles the IPv4 of the client and ignores the IPv6 of a client if it exists.
It would be nice if it would resolve both IPv4 and IPv6 when present.

This PR also changes unbound client conf filename syntax to `openvpn.client.${CN}.${SUFFIX}.conf`;
Like: openvpn.client.user1.ipv4.conf and openvpn.client.user1.ipv6.conf


vktg:
I think it would be nice to have an extra directory for openvpn client records,
like /var/unbound/openvpn

PR #4203, opened 2020-02-26
EDNS buffer size configuration. Issue #10293

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10293
- [ ] Ready for review

https://dnsflagday.net/2020/:

> Message Size Considerations
> 
> The optimum DNS message size to avoid IP fragmentation while minimizaing the use of TCP will depend on the Maximum Transmission Unit (MTU) of the physical network links connecting two network endpoints. Unfortunately, there is not yet a standard mechanism for DNS server implementors to access this information. Until such a standard exists, we recommend that the EDNS buffer size should, by default, be set to a value small enough to avoid fragmentation on the majority of network links in use today.
> 
> An EDNS buffer size of 1232 bytes will avoid fragmentation on nearly all current networks. This is based on an MTU of 1280, which is required by the IPv6 specification, minus 48 bytes for the IPv6 and UDP headers.
> 
> Note that this recomendation is for a default value, to be used when better information is not available. Operators may still configure larger values if their networks support larger data frames and they are certain there is no risk of IP fragmentation. DNS server vendors may use higher (or lower) packet sizes if better information about the MTU is available from the kernel.

### DIFF
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -30,6 +30,8 @@ require_once("config.inc");
 require_once("functions.inc");
 require_once("filter.inc");
 require_once("shaper.inc");
+require_once("interfaces.inc");
+require_once("util.inc");
 
 function create_unbound_chroot_path($cfgsubdir = "") {
 	global $config, $g;
@@ -317,7 +319,11 @@ EOF;
 	$dns_record_cache = isset($unboundcfg['dnsrecordcache']) ? "yes" : "no";
 	$outgoing_num_tcp = isset($unboundcfg['outgoing_num_tcp']) ? $unboundcfg['outgoing_num_tcp'] : "10";
 	$incoming_num_tcp = isset($unboundcfg['incoming_num_tcp']) ? $unboundcfg['incoming_num_tcp'] : "10";
-	$edns_buffer_size = (!empty($unboundcfg['edns_buffer_size'])) ? $unboundcfg['edns_buffer_size'] : "4096";
+	if (empty($unboundcfg['edns_buffer_size']) || ($unboundcfg['edns_buffer_size'] == 'auto')) {
+		$edns_buffer_size = unbound_auto_ednsbufsize();
+	} else {
+		$edns_buffer_size = $unboundcfg['edns_buffer_size'];
+	}
 	$num_queries_per_thread = (!empty($unboundcfg['num_queries_per_thread'])) ? $unboundcfg['num_queries_per_thread'] : "4096";
 	$jostle_timeout = (!empty($unboundcfg['jostle_timeout'])) ? $unboundcfg['jostle_timeout'] : "200";
 	$cache_max_ttl = (!empty($unboundcfg['cache_max_ttl'])) ? $unboundcfg['cache_max_ttl'] : "86400";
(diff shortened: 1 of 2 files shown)


vktg:
now it can calculate EDNS buffer size limit by taking the smallest MTU - max IPv4 header (60 bytes IP + 8 bytes UDP = 68).
does not allow too small sizes:
minimum for IPv4-only setup = 512
minimum for +IPv6 setup = 1232

extra sizes on services_unbound_advanced.php page:
1232 (IPv6 minimum)
1432 (1500 - 68)
1480 - seems incorrect, although recommended by unbound and CoreDNS, left for backward compatibility. 

see interesting discussion:
https://github.com/dns-violations/dnsflagday/issues/125
and https://tools.ietf.org/html/draft-fujiwara-dnsop-avoid-fragmentation-01

src/usr/local/www/services_unbound_advanced.php
```diff
@@ -335,10 +335,12 @@
 	'edns_buffer_size',
 	'EDNS Buffer Size',
 	$pconfig['edns_buffer_size'],
-	array_combine(array("512", "1480", "4096"), array("512", "1480", "4096"))
-))->setHelp('Number of bytes size to advertise as the EDNS reassembly buffer size. This is the value that is used in UDP datagrams sent to peers. ' .
-			'RFC recommendation is 4096 (which is the default). If fragmentation reassemble problems occur, usually seen as timeouts, then a value of 1480 should help. ' .
-			'The 512 value bypasses most MTU path problems, but it can generate an excessive amount of TCP fallback.');
+	array_combine(array("auto", "512", "1232", "1432", "1480", "4096"), array("Auto", "512", "1232", "1432", "1480", "4096"))
```
jim-p: The array of choices would be better as a global set at the start of the page, with the choices as keys and some explanation about why users might choose each one in the value.

Something like this:

```
$unbound_edns_sizes = array(
	"auto" => "Automatic value based on active interface MTUs",
	"512"  => "512: IPv4 Minimum",
	"1220" => "1220: NSD IPv6 EDNS Minimum",
	"1232" => "1232: IPv6 Minimum",
	"1432" => "1432: 1500 Byte MTU",
	"1480" => "1480: NSD IPv4 EDNS Minimum",
	"4096" => "4096: Unbound Default"
);
```

And then you can use that array when doing the tests (e.g. `array_key_exists( ... )` in the validity check, and using the array directly here in the field.

Then you can simplify the description text below as well since you don't have to explain them all.

vktg: ok, done
I see no reason to simplify the description text - it does not explain all available sizes, but provides useful info for fragmentation and MTU issues.

src/etc/inc/unbound.inc
```diff
@@ -947,4 +953,38 @@ function unbound_local_zone_types() {
 	);
 }
 
+// Autoconfig EDNS buffer size
+function unbound_auto_ednsbufsize() {
+	global $g, $config;
```
rbgarga: `$g` variable is not being used in this function

vktg: fixed

PR #4206, opened 2020-02-26
DHCPv6 range is not mandatory for Stateless DHCP. Issue #9596

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9596
- [ ] Ready for review

"The firewall will send out RA packets and addresses can be assigned to clients by SLAAC while providing additional information such as DNS and NTP from DHCPv6."

I think, in this case, DHCPv6 range should not be a mandatory field in DHCPv6 Server tab.
The DHCPv6 should only pass additional info to clients, not the IP address.

### DIFF
--- a/src/usr/local/www/services_dhcpv6.php
+++ b/src/usr/local/www/services_dhcpv6.php
@@ -227,10 +227,10 @@ if (isset($_POST['apply'])) {
 
 	// Note: if DHCPv6 Server is not enabled, then it is OK to adjust other parameters without specifying range from-to.
 	if ($_POST['enable']) {
-		$reqdfields = explode(" ", "range_from range_to");
-		$reqdfieldsn = array(gettext("Range begin"), gettext("Range end"));
-
-		do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_dhcpv6.php
```diff
@@ -227,10 +227,10 @@ function dhcpv6_apply_changes($dhcpdv6_enable_changed) {
 
 	// Note: if DHCPv6 Server is not enabled, then it is OK to adjust other parameters without specifying range from-to.
 	if ($_POST['enable']) {
-		$reqdfields = explode(" ", "range_from range_to");
-		$reqdfieldsn = array(gettext("Range begin"), gettext("Range end"));
-
-		do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
+		if (empty($_POST['range_from']) && empty($_POST['range_to']) &&
```
jim-p: The logic here seems wrong. It should be checking the mode and if it's not stateless then it should error if _either_ the range from or to is empty (not just both)

vktg: right, fixed

PR #4210, opened 2020-02-28
IPsec IPv4/IPv6 pools parameters. Issue #10296

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10296
- [ ] Ready for review

After last commit https://redmine.pfsense.org/projects/pfsense/repository/revisions/407a5c28093d46cb39cc1bba75740523a1ee97e6
The mobile-pool-v4 and mobile-pool-v6 pools are created as expected.
But seems that "addrs" is required for each pool, so the mobile-pool is not created (or referenced in connections).
Maybe the commons attrs should be included in the v4 pool or both for now?

This PR adds common attrs to each pool + description fix for vpn_ipsec_mobile.php, because RADIUS IP address priority applies to both ipv4 and ipv6

### DIFF
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -1466,7 +1466,17 @@ function ipsec_setup_pools() {
 		return;
 	}
 	$scconf['pools']['mobile-pool'] = array();
+	$pool_addrs = array();
 	$pool_common =& $scconf['pools']['mobile-pool'];
+	if (!empty($a_client['pool_address'])) {
+		$pool_addrs[] = "{$a_client['pool_address']}/{$a_client['pool_netbits']}";
+	}
+	if (!empty($a_client['pool_address_v6'])) {
+		$pool_addrs[] = "{$a_client['pool_address_v6']}/{$a_client['pool_netbits_v6']}";
+	}
+	if (!empty($pool_addrs)) {
+		$pool_common['addrs'] = implode(',', $pool_addrs);
+	}
 
 	$rightdnsservers = array();
 	if (!empty($a_client['dns_server1'])) {
--- a/src/usr/local/www/vpn_ipsec_mobile.php
+++ b/src/usr/local/www/vpn_ipsec_mobile.php
@@ -532,13 +532,6 @@ $group->add(new Form_Select(
 $section->add($group);
 
 $section->addInput(new Form_Checkbox(
-	'radius_ip_priority_enable',
-	'RADIUS IP address priority',
-	'IPv4 address pool is used if IP is not supplied by RADIUS server',
-	$pconfig['radius_ip_priority_enable']
-));
-
-$section->addInput(new Form_Checkbox(
 	'pool_enable_v6',
 	'Virtual IPv6 Address Pool',
 	'Provide a virtual IPv6 address to clients',
@@ -571,9 +564,16 @@ $group->add(new Form_Select(
 	'',
 	$pconfig['pool_netbits_v6'],
 	$netBitsv6
-))->setWidth(3);
+))->setWidth(2);
 
 $section->add($group);
+
+$section->addInput(new Form_Checkbox(
+	'radius_ip_priority_enable',
+	'RADIUS IP address priority',
+	'IPv4/IPv6 address pool is used if address is not supplied by RADIUS server',
+	$pconfig['radius_ip_priority_enable']
+));
 
 $section->addInput(new Form_Checkbox(
 	'net_list_enable',


vktg:
ok, I simply revert back that v4+v6 mobile pool code

vktg:
What change?

jim-p:
Strange, it didn't post my review comment. I'll do it again.

src/etc/inc/ipsec.inc
```diff
@@ -1466,6 +1466,16 @@ function ipsec_setup_pools() {
...
+	if (!empty($a_client['pool_address'])) {
+		$pool_addrs[] = "{$a_client['pool_address']}/{$a_client['pool_netbits']}";
+	}
+	if (!empty($a_client['pool_address_v6'])) {
+		$pool_addrs[] = "{$a_client['pool_address_v6']}/{$a_client['pool_netbits_v6']}";
+	}
+	if (count($pool_addrs)) {
+		$scconf['pools']['mobile-pool']['addrs'] = implode(',', $pool_addrs);
+	}
 	$pool_common =& $scconf['pools']['mobile-pool'];
```
jim-p: If you move `$pool_common =& $scconf['pools']['mobile-pool'];` above these changes, then when setting the pool `addrs` value you can do it this way:

    $pool_common['addrs'] = implode(',', $pool_addrs)

Which makes it more consistent with the way the rest of the settings are handled.

vktg: ok, done

src/etc/inc/ipsec.inc
```diff
@@ -1466,7 +1466,15 @@ function ipsec_setup_pools() {
...
 	$scconf['pools']['mobile-pool'] = array();
+	$pool_addrs = array();
 	$pool_common =& $scconf['pools']['mobile-pool'];
+	if (!empty($a_client['pool_address'])) {
+		$pool_addrs[] = "{$a_client['pool_address']}/{$a_client['pool_netbits']}";
+	}
+	if (!empty($a_client['pool_address_v6'])) {
+		$pool_addrs[] = "{$a_client['pool_address_v6']}/{$a_client['pool_netbits_v6']}";
+	}
+	$scconf['pools']['mobile-pool']['addrs'] = implode(',', $pool_addrs);
```
jim-p: This still needs a check to ensure it isn't added with an empty value. And the variable reference didn't get changed.

So it should be:

    if (!empty($pool_addrs)) {
    	$pool_common['addrs'] = implode(',', $pool_addrs);
    }

vktg: oh, ok, fixed

emes:
Tested (EAP-TLS and EAP-RADIUS) and this works. But, the DNS info from mobile-pool is not passed.

vktg:
If I revert back https://github.com/pfsense/pfsense/commit/fb0827f350e417cf29e0f3bb8d2f8ecefb2bb928 DNS is passed.
strongswan inherit bug?

jim-p:
I'm also curious to know if it's just DNS that isn't passed down, or if other fields like the default domain get omitted.

We could copy all of that into the other pools (v4, v6, per-user pools) but it seems broken to have to repeat it all when the references/templating are supposed to allow for that.

emes:
If mobile-pool doesn't have addresses then it's not being referenced at all as I understand it.  If mobile-pool does not have other utility (I have no clue) then seems that it would be better to apply the dns info to each of the v4/v6 pools.  That doesn't seem broken to me, since it would would work regardless of whether assignments were v4, v6 or both.

jim-p:
In the swanctl/strongswan.conf style syntax you are supposed to be able to setup templates with base settings and then make reference to them so others inherit the "defaults" from there and add their own options which differ. The mobile pool wouldn't be used directly in this case, it's a container for default values. Others reference it to obtain the common settings.

We _can_ copy the defaults to every pool but that's inefficient and redundant. It should be working this way, and if it's not, there is likely a bug in strongSwan.

emes:
Aha.  Sorry I missed that.  

Since the mobile-pool is defined as a real pool within the pools section, not just as a top-level default, it's reference should be "pools.mobile-pool", no?

For example:

```
1552c1552
<               $scconf['pools']['mobile-pool-v4 : mobile-pool'] = array();
---
>               $scconf['pools']['mobile-pool-v4 : pools.mobile-pool'] = array();
```

jim-p:
That's part of it, but it looks like per-user settings are also broken in other ways. I'm working on it now, it will supplant any changes made here, so I'm closing this out.

jim-p:
For those following along here, this should be OK on the next snapshot. See also: https://redmine.pfsense.org/issues/10314 and https://redmine.pfsense.org/issues/10296

PR #4211, opened 2020-02-29
Replace double quote character in DHCP client ID. Issue #10295

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10295
- [ ] Ready for review

The following is allowed by the webgui in a static mapping: Client Identifier: 32" Sony Trinitron
That creates a configuration that cannot be loaded:
```
host s_lan_0 {
        hardware ethernet 01:02:03:04:05:06;
        option dhcp-client-identifier "32" Sony Trinitron";
    option host-name "32-Sony";
    option domain-name-servers 172.25.228.1;
}
```

This fix replaces the double quote character with '&quot';
There is no issue with other special characters (',./!@#$%^&*()~)

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1118,6 +1118,7 @@ EOD;
 
 			$i = 0;
 			foreach ($dhcpifconf['staticmap'] as $sm) {
+				$cid = '';
 				$dhcpdconf .= "host s_{$dhcpif}_{$i} {\n";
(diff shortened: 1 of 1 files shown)


jim-p:
Does escaping the double quote as `\"` work? That seems like a better choice if it does.

vktg:
It works with `\"` too
Replaced

PR #4213, opened 2020-02-29
NTPd Autoset GPS device baud rate. Issue #7284

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/7284
- [ ] Ready for review

It would be nice to have the option to attempt to auto configure the baud rate for a GPS device.
Useful if you don't know the baud rate of your device.
Or for some devices that lose their configurations during power outages.
Or when changing the baud rate.

This is resolved copy of the original PR https://github.com/pfsense/pfsense/pull/3561 by @jskyboo

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
 	}
 }
 
+function check_gps_speed($device) {
+	usleep(1000);
+	// Set timeout to 5s
+	$timeout=microtime(true)+5;
+	if ($fp = fopen($device, 'r')) {
+		stream_set_blocking($fp, 0);
+		stream_set_timeout($fp, 5);
+		$contents = "";
+		$cnt = 0;
+		$buffersize = 256;
+		do {
+			$c = fread($fp, $buffersize - $cnt);
+
+			// Wait for data to arive
+			if (($c === false) || (strlen($c) == 0)) {
+				usleep(500);
+				continue;
+			}
+
+			$contents.=$c;
+			$cnt = $cnt + strlen($c);
+		} while (($cnt < $buffersize) && (microtime(true) < $timeout));
+		fclose($fp);
+
+		$nmeasentences = ['RMC', 'GGA', 'GLL', 'ZDA', 'ZDG', 'PGRMF'];
+		foreach ($nmeasentences as $sentence) {
+			if (strpos($contents, $sentence) > 0) {
+				return true;
+			}
+		}
+		if (strpos($contents, '0') > 0) {
+			$filters = ['`', '?', '/', '~'];
+			foreach ($filters as $filter) {
+				if (strpos($contents, $filter) !== false) {
+					return false;
+				}
+			}
+			return true;
+		}
+	}
+	return false;
+}
+
 /* Generate list of possible NTP poll values
  * https://redmine.pfsense.org/issues/9439 */
 global $ntp_poll_min_value, $ntp_poll_max_value;
@@ -1548,6 +1591,8 @@ function system_ntp_fixup_poll_value($type, $configvalue, $default) {
 
 function system_ntp_setup_gps($serialport) {
 	global $config, $g;
+
+	init_config_arr(array('ntpd', 'gps', 'speed'));
 	$gps_device = '/dev/gps0';
 	$serialport = '/dev/'.$serialport;
 
@@ -1560,28 +1605,42 @@ function system_ntp_setup_gps($serialport) {
 	@symlink($serialport, $gps_device);
 
 	$gpsbaud = '4800';
-	if (is_array($config['ntpd']) && is_array($config['ntpd']['gps']) && !empty($config['ntpd']['gps']['speed'])) {
-		switch ($config['ntpd']['gps']['speed']) {
-			case '16':
-				$gpsbaud = '9600';
-				break;
-			case '32':
-				$gpsbaud = '19200';
-				break;
-			case '48':
-				$gpsbaud = '38400';
-				break;
-			case '64':
-				$gpsbaud = '57600';
-				break;
-			case '80':
-				$gpsbaud = '115200';
-				break;
-		}
+	$speeds = array(
+		0 => '4800', 
+		16 => '9600', 
+		32 => '19200', 
+		48 => '38400', 
+		64 => '57600', 
+		80 => '115200'
+	);
+	if (!empty($config['ntpd']['gps']['speed']) && array_key_exists($config['ntpd']['gps']['speed'], $speeds)) {
+		$gpsbaud = $speeds[$config['ntpd']['gps']['speed']];
 	}
 
-	/* Configure the serial port for raw IO and set the speed */
-	mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
+	system_ntp_setup_rawspeed($serialport, $gpsbaud);
+
+	$autospeed = ($config['ntpd']['gps']['speed'] == 'autoalways' || $config['ntpd']['gps']['speed'] == 'autoset');
+	if ($autospeed || ($config['ntpd']['gps']['autobaudinit'] && !check_gps_speed($gps_device))) {
+		$found = false;
+		foreach ($speeds as $baud) {
+			system_ntp_setup_rawspeed($serialport, $baud);
+			if ($found = check_gps_speed($gps_device)) {
+				if ($autospeed) {
+					$saveconfig = ($config['ntpd']['gps']['speed'] == 'autoset');
+					$config['ntpd']['gps']['speed'] = array_search($baud, $speeds);
+					$gpsbaud = $baud;
+					if ($saveconfig) {
+						write_config(sprintf(gettext('Autoset GPS baud rate to %s'), $baud));
+					}
+				}
+				break;
+			}
+		}
+		if ($found === false) {
+			log_error(gettext("Could not find correct GPS baud rate."));
+			return false;
+		}
+	}
 
 	/* Send the following to the GPS port to initialize the GPS */
 	if (is_array($config['ntpd']) && is_array($config['ntpd']['gps']) && !empty($config['ntpd']['gps']['type'])) {
@@ -1594,13 +1653,27 @@ function system_ntp_setup_gps($serialport) {
 	@file_put_contents('/tmp/gps.init', $gps_init);
 	mwexec("cat /tmp/gps.init > {$serialport}");
 
+	if ($found && $config['ntpd']['gps']['autobaudinit']) {
+		system_ntp_setup_rawspeed($serialport, $gpsbaud);
+	}
+
+	/* Remove old /etc/remote entry if it exists */
+	if (mwexec("/usr/bin/grep -c '^gps0' /etc/remote")) {
+		mwexec("/usr/bin/sed -i '' -n '/gps0/!p' /etc/remote");
+	}
+
 	/* Add /etc/remote entry in case we need to read from the GPS with tip */
-	if (intval(`grep -c '^gps0' /etc/remote`) == 0) {
+	if (intval(`/usr/bin/grep -c '^gps0' /etc/remote`) == 0) {
 		@file_put_contents("/etc/remote", "gps0:dv={$serialport}:br#{$gpsbaud}:pa=none:\n", FILE_APPEND);
 	}
 
-
 	return true;
+}
+
+// Configure the serial port for raw IO and set the speed
+function system_ntp_setup_rawspeed($serialport, $baud) {
+	mwexec("/bin/stty -f " .  escapeshellarg($serialport) . " raw speed " . escapeshellarg($baud));
+	mwexec("/bin/stty -f " .  escapeshellarg($serialport) . ".init raw speed " . escapeshellarg($baud));
 }
 
 function system_ntp_setup_pps($serialport) {
@@ -1637,6 +1710,12 @@ function system_ntp_configure() {
 	if (!is_array($config['ntpd'])) {
 		$config['ntpd'] = array();
 	}
+
+	/* if ntpd is running, kill it */
+	while (isvalidpid("{$g['varrun_path']}/ntpd.pid")) {
+		killbypid("{$g['varrun_path']}/ntpd.pid");
+	}
+	@unlink("{$g['varrun_path']}/ntpd.pid");
 
 	$ntpcfg = "# \n";
 	$ntpcfg .= "# pfSense ntp configuration file \n";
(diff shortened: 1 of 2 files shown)


src/etc/inc/system.inc
```diff
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
 	}
 }
 
+function check_gps_speed($device) {
+	usleep(1000);
+	// Set timeout to 5s
+	$timeout=microtime(true)+5;
+	if($fp = fopen($device, 'r')) {
```
jim-p: Needs a space between `if` and `(`

src/etc/inc/system.inc
```diff
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
...
+		stream_set_blocking($fp, 0);
+		stream_set_timeout($fp, 5);
+		$contents = "";
+		$cnt = 0;
+		$buffersize = 256;
+		do {
+			$c = fread($fp, $buffersize - $cnt);
+
+			// Wait for data to arive
+			if($c === false || strlen($c) == 0) {
```
jim-p: Needs a space between `if` and `(`.
Each test in the `if` needs its own set of parenthesis.

src/etc/inc/system.inc
```diff
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
...
+				usleep(500);
+				continue;
+			}
+
+			$contents.=$c;
+			$cnt = $cnt + strlen($c);
+		} while($cnt < $buffersize && microtime(true)<$timeout);
```
jim-p: Needs a space between `while` and `(`.
Needs spaces around `<` in the second test.
Each test needs its own set of parenthesis.

src/etc/inc/system.inc
```diff
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
...
+		fclose($fp);
+
+		$nmeasentences = ['RMC', 'GGA', 'GLL', 'ZDA', 'ZDG', 'PGRMF'];
+		foreach ($nmeasentences as $sentence) {
+			if(strpos($contents, $sentence) > 0) {
```
jim-p: Needs a space between `if` and `(`

src/etc/inc/system.inc
```diff
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
...
+				return true;
+			}
+		}
+		if (strpos($contents, '0') > 0) {
+			$filters = ['`', '?', '/', '~'];
+			foreach ($filters as $filter) {
+				if(strpos($contents, $filter) !== false) {
```
jim-p: Needs a space between `if` and `(`

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
-				$gpsbaud = '38400';
-				break;
-			case '64':
-				$gpsbaud = '57600';
-				break;
-			case '80':
-				$gpsbaud = '115200';
-				break;
-		}
+	$speeds=[0 => '4800', 16 => '9600', 32 => '19200', 48 => '38400', 64 => '57600', 80 => '115200'];
```
jim-p: That's not the proper way to define a new inline array. It should at least be:

    $speeds = array(0 => '4800', 16 => '9600', 32 => '19200', 48 => '38400', 64 => '57600', 80 => '115200');

It would look better with each new value on its own line with proper indentation, though.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+	$revspeeds = array_flip($speeds);
+	if($config['ntpd']['gps']['speed'] && array_key_exists($config['ntpd']['gps']['speed'], $speeds)) {
```
jim-p: Needs a space between `if` and `(`.
The test on `$config['ntpd']['gps']['speed']` should probably be an `is_array()` test, or even better, initialize it with `init_config_arr()` before this and check that it isn't empty.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+		$gpsbaud = $speeds[$config['ntpd']['gps']['speed']];
 	}
 
 	/* Configure the serial port for raw IO and set the speed */
+	mwexec("stty -f {$serialport} raw speed {$gpsbaud}");
```
jim-p: In this line and the line below it, I'd like to see `$serialport` and `$gpsbaud` run through `escapeshellarg()` for good measure. Also, use the full path to `stty`.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
 	mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
 
+	$autospeed = ($config['ntpd']['gps']['speed'] == 'autoalways' || $config['ntpd']['gps']['speed'] == 'autoset');
+	if($autospeed || ($config['ntpd']['gps']['autobaudinit'] && !check_gps_speed($gps_device))) {
```
jim-p: Needs a space between `if` and `(`.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+		$found = false;
+		foreach ($speeds as $baud) {
+			mwexec("stty -f {$serialport} raw speed {$baud}");
```
jim-p: See previous comments about `stty` and the variables in the command.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+			mwexec("stty -f {$serialport}.init raw speed {$baud}");
+			if($found = check_gps_speed($gps_device)) {
```
jim-p: Needs a space between `if` and `(`.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+				if($autospeed) {
```
jim-p: Needs a space between `if` and `(`.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+					$saveconfig = ($config['ntpd']['gps']['speed'] == 'autoset');
+					$config['ntpd']['gps']['speed'] = $revspeeds[$baud];
+					$gpsbaud = $baud;
+					if($saveconfig) {
```
jim-p: Needs a space between `if` and `(`.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+						write_config(sprintf(gettext('Autoset GPS baud rate to %s'), $baud));
+					}
+				}
+				break;
+			}
+		}
+		if($found === false) {
```
jim-p: Needs a space between `if` and `(`.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
 	if (is_array($config['ntpd']) && is_array($config['ntpd']['gps']) && !empty($config['ntpd']['gps']['type'])) {
 		$gps_init = base64_decode($config['ntpd']['gps']['initcmd']);
 	} else {
 		$gps_init = base64_decode('JFBVQlgsNDAsR1NWLDAsMCwwLDAqNTkNCiRQVUJYLDQwLEdMTCwwLDAsMCwwKjVDDQokUFVCWCw0MCxaREEsMCwwLDAsMCo0NA0KJFBVQlgsNDAsVlRHLDAsMCwwLDAqNUUNCiRQVUJYLDQwLEdTViwwLDAsMCwwKjU5DQokUFVCWCw0MCxHU0EsMCwwLDAsMCo0RQ0KJFBVQlgsNDAsR0dBLDAsMCwwLDANCiRQVUJYLDQwLFRYVCwwLDAsMCwwDQokUFVCWCw0MCxSTUMsMCwwLDAsMCo0Ng0KJFBVQlgsNDEsMSwwMDA3LDAwMDMsNDgwMCwwDQokUFVCWCw0MCxaREEsMSwxLDEsMQ==');
 	}
 
-	/* XXX: Why not file_put_contents to the device */
-	@file_put_contents('/tmp/gps.init', $gps_init);
-	mwexec("cat /tmp/gps.init > {$serialport}");
+	@file_put_contents($gps_device, $gps_init);
```
jim-p: Does that actually work as expected?

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+
+	if ($found && $config['ntpd']['gps']['autobaudinit']) {
+		mwexec("stty -f {$serialport} raw speed {$gpsbaud}");
```
jim-p: See previous comments about `stty` and variables in the command.

src/etc/inc/system.inc
```diff
@@ -1560,46 +1603,64 @@ function system_ntp_setup_gps($serialport) {
...
+		mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
+	}
+
+	/* Remove old /etc/remote entry if it exists */
+	if (mwexec("/usr/bin/grep -c '^gps0' /etc/remote")) {
+		mwexec("/usr/bin/sed -i '' -n '/gps0/!p' /etc/remote");
+	}
 
 	/* Add /etc/remote entry in case we need to read from the GPS with tip */
 	if (intval(`grep -c '^gps0' /etc/remote`) == 0) {
```
jim-p: Fix this grep to use the full path while you're at it

src/usr/local/www/services_ntpd_gps.php
```diff
@@ -198,9 +198,18 @@ function autocorrect_initcmd($initcmd) {
 
 	if (!empty($_POST['gpsspeed'])) {
 		$config['ntpd']['gps']['speed'] = $_POST['gpsspeed'];
+		if($_POST['gpsspeed'] == 'autoalways') {
```
jim-p: Needs a space between `if` and `(`.

src/etc/inc/system.inc
vktg: Unable to check now (no GPS device available)
I can leave this strings as-is, or test it later

src/usr/local/www/services_ntpd_gps.php
vktg: all fixed

src/etc/inc/system.inc
```diff
@@ -1503,6 +1503,49 @@ function system_timezone_configure() {
...
+		stream_set_blocking($fp, 0);
+		stream_set_timeout($fp, 5);
+		$contents = "";
+		$cnt = 0;
+		$buffersize = 256;
+		do {
+			$c = fread($fp, $buffersize - $cnt);
+
+			// Wait for data to arive
+			if ($c === false || strlen($c) == 0) {
```
jim-p: Each test requires its own set of parenthesis

src/etc/inc/system.inc
```diff
@@ -1560,28 +1605,46 @@ function system_ntp_setup_gps($serialport) {
...
-		}
+	$speeds = array(
+		0 => '4800', 
+		16 => '9600', 
+		32 => '19200', 
+		48 => '38400', 
+		64 => '57600', 
+		80 => '115200'
+	);
+	$revspeeds = array_flip($speeds);
```
jim-p: Why do an `array_flip()` here? You only use this variable once below. It would be better to skip this and use `array_search()` later.

src/etc/inc/system.inc
```diff
@@ -1560,28 +1605,46 @@ function system_ntp_setup_gps($serialport) {
...
+	if (!empty($config['ntpd']['gps']['speed']) && array_key_exists($config['ntpd']['gps']['speed'], $speeds)) {
+		$gpsbaud = $speeds[$config['ntpd']['gps']['speed']];
 	}
 
 	/* Configure the serial port for raw IO and set the speed */
-	mwexec("stty -f {$serialport}.init raw speed {$gpsbaud}");
+	mwexec("/bin/stty -f " .  escapeshellarg($serialport) . " raw speed " . escapeshellarg($gpsbaud));
+	mwexec("/bin/stty -f " .  escapeshellarg($serialport) . ".init raw speed " . escapeshellarg($gpsbaud));
```
jim-p: Since these two lines are practically the same here and when used again, it would probably be beneficial to make a function that takes the port and baud as arguments, and then call that from both places. That way if the init process changes in the future, only one place needs to be updated.

src/etc/inc/system.inc
```diff
@@ -1560,28 +1605,46 @@ function system_ntp_setup_gps($serialport) {
...
+
+	$autospeed = ($config['ntpd']['gps']['speed'] == 'autoalways' || $config['ntpd']['gps']['speed'] == 'autoset');
+	if ($autospeed || ($config['ntpd']['gps']['autobaudinit'] && !check_gps_speed($gps_device))) {
+		$found = false;
+		foreach ($speeds as $baud) {
+			mwexec("/bin/stty -f " .  escapeshellarg($serialport) . " raw speed " . escapeshellarg($baud));
+			mwexec("/bin/stty -f " .  escapeshellarg($serialport) . ".init raw speed " . escapeshellarg($baud));
```
jim-p: See previous note about making this a function

src/etc/inc/system.inc
```diff
@@ -1590,16 +1653,23 @@ function system_ntp_setup_gps($serialport) {
...
 	}
 
-	/* XXX: Why not file_put_contents to the device */
-	@file_put_contents('/tmp/gps.init', $gps_init);
-	mwexec("cat /tmp/gps.init > {$serialport}");
+	@file_put_contents($gps_device, $gps_init);
+
+	if ($found && $config['ntpd']['gps']['autobaudinit']) {
+		mwexec("/bin/stty -f " .  escapeshellarg($serialport) . " raw speed " . escapeshellarg($gpsbaud));
+		mwexec("/bin/stty -f " .  escapeshellarg($serialport) . ".init raw speed " . escapeshellarg($gpsbaud));
```
jim-p: See previous note about making this a function

src/etc/inc/system.inc
vktg: I'll test @file_put_contents ($ gps_device, $ gps_init); when I get the gps device
Now I leave it as-is

src/etc/inc/system.inc
vktg: all fixed

PR #4214, opened 2020-02-29
Special characters in Schedules descr and rangedescr fields. Issue #10305

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10305
- [ ] Ready for review

> When using special characters in Schedules and timeranges description fields
> firewall_schedule.php page shows escaping symbols, i.e.
> `G&uuml;nter` instead of Günter

This PR removes `htmlentities($_POST['descr'], ENT_QUOTES, 'UTF-8');` (legacy code?)
from firewall_schedule_edit.php
and adds `rangedescr` to $cdata_fields array
It also sort $cdata_fields array alphabetically

### DIFF
--- a/src/etc/inc/xmlparse.inc
+++ b/src/etc/inc/xmlparse.inc
@@ -218,11 +218,11 @@ function parse_xml_config_raw($cffile, $rootobj, $isstring = "false") {
 /* Return true if a field should be cdata encoded */
 function is_cdata_entity($ent) {
 	$cdata_fields = array(
-		'auth_prompt', 'certca', 'certname', 'city', 'common_name',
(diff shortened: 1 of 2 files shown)


vktg:
> The `htmlentities` change was specifically to fix a JS issue. See https://redmine.pfsense.org/issues/4964 -- Test the described scenario in a couple different browsers. If it works as expected with this change, then it may be OK.

Successfully tested with IE 11, Chromium 79, Firefox 68.5, Opera 67

> It that change really related to the problem in the linked Redmine issue? It might need to be its own issue and not tacked on here.

This is correct issue - after removing htmlentities() from firewall_schedule_edit.php,  firewall_schedule.php starts displaying special characters

PR #4216, opened 2020-03-05
Do not restart L2TP server after adding/modifying users. Issue #4866

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/4866
- [ ] Ready for review

Full daemon restart drops L2TP clients sessions and may cause issues with other packages (see the redmine link)
This is useless since mpd reads mpd.secret on the fly.

This PR moves creation of the mpd.secret file to a separate function that does not restart mpd daemon.

### DIFF
--- a/src/etc/inc/vpn.inc
+++ b/src/etc/inc/vpn.inc
@@ -407,29 +407,7 @@ EOD;
 			fclose($fd);
 			unset($mpdconf);
 
-			/* write mpd.secret */
-			$fd = fopen("{$g['varetc_path']}/l2tp-vpn/mpd.secret", "w");
-			if (!$fd) {
-				printf(gettext("Error: cannot open mpd.secret in vpn_l2tp_configure().") . "\n");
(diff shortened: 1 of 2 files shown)


src/etc/inc/vpn.inc
```diff
@@ -448,4 +426,33 @@ EOD;
 	return 0;
 }
 
+function vpn_l2tp_updatesecret() {
+	global $config, $g;
+	$l2tpcfg = $config['l2tp'];
+
+	$fd = fopen("{$g['varetc_path']}/l2tp-vpn/mpd.secret", "w");
+	if (!$fd) {
+		printf(gettext("Error: cannot open mpd.secret in vpn_l2tp_configure().") . "\n");
```
jim-p: This description isn't accurate anymore (the function name changed).

vktg: ok, fixed

PR #4218, opened 2020-03-06
Notifications page input validation. Issue #8522

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8522
- [ ] Ready for review

Bug:
When I clicked the "Test SMTP Settings" button, I got a green message "SMTP testing e-mail successfully sent" but I never received the email. I then realized that I forgot to type anything in the "E-Mail server" textbox.

Expected behavior:
The GUI should have a red message saying "Could not send a test email because you haven't specified the SMTP Server and port."

This PR adds extra input validation

### DIFF
--- a/src/usr/local/www/system_advanced_notifications.php
+++ b/src/usr/local/www/system_advanced_notifications.php
@@ -32,6 +32,7 @@ require_once("guiconfig.inc");
 require_once("notices.inc");
 require_once("pfsense-utils.inc");
 
+global $smtp_authentication_mechanisms;
 $pconfig = array();
 init_config_arr(array('notifications', 'certexpire'));
 init_config_arr(array('notifications', 'smtp'));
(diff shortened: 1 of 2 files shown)


src/usr/local/www/system_advanced_notifications.php
```diff
@@ -123,7 +143,12 @@
 			}
 		}
 
-		$config['notifications']['smtp']['authentication_mechanism'] = $_POST['smtpauthmech'];
+		if (!in_array($_POST['smtpauthmech'], $smtp_authentication_mechanisms)) {
```
jim-p: Seems like `$smtp_authentication_mechanisms` is missing a global declaration near the top of this file (and also missing in `/etc/inc/notices.inc` before it's defined).

Also this test should be using `array_key_exists()` rather than `in_array()`. It may work at the moment but only by coincidence since currently the key and value pairs in that array are identical.

vktg: done, fixed

rbgarga: global declaration outside a function makes no sense.  All variables are available on that scope

jim-p: It's defined in another file, not here, and isn't declared global before it's defined as global in either place. We've seen situations before where that has caused problems and the variable ends up empty in certain rare consequences.

PR #4221, opened 2020-03-07
Synchronize admin accounts. Issue #9622

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9622
- [ ] Ready for review

This is copy of the original PR https://github.com/pfsense/pfsense/pull/4068 by @jwsi 
with checkbox on the System \ High Availability Sync to turn on/off this feature

from original PR:

These changes allow for the ability for the admin account to be synced between HA pairs.

**The following functionality is introduced**:
1. All admin account changes are synced to a partner.
2. When the admin password is changed on the primary node, the secondary is updated accordingly and the primary HA sync password is also updated to prevent future auth errors occuring during sync.

**Important Notes**:
* When an admin account password is changed, the config.xml file will be updated accordingly on the secondary. In xmlrpc.php there is a function called auth which authenticates the entity making changes to the device. Obviously this cannot be called if the config.xml file has been updated with the new credentials, but the PHP session is still using the old credentials. Therefore, I have made filter_configure a private function to remove this possibility.
James.

### DIFF
--- a/src/usr/local/www/system_hasync.php
+++ b/src/usr/local/www/system_hasync.php
@@ -33,6 +33,7 @@ $a_hasync = &$config['hasync'];
 
 $checkbox_names = array(
 	'pfsyncenabled',
+	'adminsync',
 	'synchronizeusers',
 	'synchronizeauthservers',
 	'synchronizecerts',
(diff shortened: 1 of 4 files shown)


src/usr/local/www/system_hasync.php
```diff
@@ -173,6 +174,15 @@ interfaces_sync_setup();
 ))->setHelp('Enter the webConfigurator password of the system entered above for synchronizing the configuration.%1$s' .
 			'Do not use the Synchronize Config to IP and password option on backup cluster members!', '<br />');
 
+$section->addInput(new Form_Checkbox(
+	'adminsync',
+	'Synchronize admin',
+	'synchronize admin accounts and autoupdate sync password.',
+	($pconfig['adminsync'] === 'on'),
+	'on'
+))->setHelp('By default, admin accounts does not sync, and you can have different passwords on both nodes.%1$s' .
```
jim-p: Reword to fix grammar and avoid addressing the user directly ("you"). For example:

    'By default, the admin account does not synchronize, and each node may have a different admin password.%1$s' .
    'This option automatically updates XMLRPC Remote System Password when the password is changed on the Remote System Username account.'

vktg: ok, fixed

PR #4223, opened 2020-03-08
NTP status/widget long IPv6 address. Issue #10307

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10307
- [ ] Ready for review

if I have a) an IPv6 NTP server, and b) the NTP status widget on the dashboard, the widget doesn't display the stratum correctly:
```
NTP Status 
Server Time    10:08:58 CST
Sync Source    2620:19:4000:100::86 (stratum )
```

Also NTP status page shows truncated IPv6 address 

from ntpq(1):
>      -w, --wide
>              Display the full 'remote' value.
> 
>              Display the full value of the 'remote' value.  
>              If this requires more than 15 characters, display the full value, emit
>              a newline, and continue the data display properly indented on the next line.

This PR fixes both issues

### DIFF
--- a/src/usr/local/www/status_ntpd.php
+++ b/src/usr/local/www/status_ntpd.php
@@ -52,7 +52,7 @@ if ($allow_query) {
 		$inet_version = " -4";
 	}
 
-	exec("/usr/local/sbin/ntpq -pn $inet_version | /usr/bin/tail +3", $ntpq_output);
+	exec('/usr/local/sbin/ntpq -pnw ' . $inet_version . ' | /usr/bin/tail +3 | /usr/bin/sed -E \'N;s/\n[[:blank:]]+/ /;P;D\'', $ntpq_output);
 
 	$ntpq_servers = array();
 	foreach ($ntpq_output as $line) {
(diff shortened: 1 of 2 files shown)


jim-p:
Applying the patch as-is didn't work, the lines with a long hostname (like pool placeholders) had an empty ref column and other values were shifted over. See my review comment for a much simpler alternative fix.

vktg:
Testing `/usr/local/sbin/ntpq -pnw ' . $inet_version . ' | /usr/bin/tail +3 | /usr/bin/sed \'N;s/\n\s*//;P;D\'` I found that it returns only 'oversized', but not 'normal' peer string

jim-p:
Hmm, I saw both. But I did find another system where it failed in a different way. Try this instead:

    exec('/usr/local/sbin/ntpq -pnw ' . $inet_version . ' | /usr/bin/tail +3 | /usr/bin/sed -E \'N;s/\n[[:blank:]]+/ /;P;D\'', $ntpq_output);

rschell:
This solution is still not quite right.  I'm losing the last peer which happens to be the active one.

for example:
image

results in:
image

the -pnw full results are something like:
image

PR #4224, opened 2020-03-09
 Feature #10323

kiokoman:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10323
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1652,11 +1652,11 @@ function system_ntp_configure() {
 		$ntpcfg .= '12';
 	}
 	/* Add Maximum candidate NTP peers */
+	$ntpcfg .= ' maxclock ';
 	if (!empty($config['ntpd']['ntpmaxpeers'])) {
-		$ntpcfg .= ' maxclock ';
 		$ntpcfg .= $config['ntpd']['ntpmaxpeers'];
 	} else {
-		$ntpcfg .= ' maxclock 5';
+		$ntpcfg .= '5';
 	}
 	$ntpcfg .= "\n";
 
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1653,6 +1653,16 @@ function system_ntp_configure() {
 	}
 	$ntpcfg .= "\n";
 
+	/* Add min-max ntp peers */
+        $ntpcfg .= "# Min-Max peers\n";
+        $ntpcfg .= 'tos minclock 3 maxclock ';
+        if (!empty($config['ntpd']['ntpmaxpeers'])) {
+                $ntpcfg .= $config['ntpd']['ntpmaxpeers'];
+        } else {
+                $ntpcfg .= '4';
+        }
+        $ntpcfg .= "\n";
+	
 	/* Add PPS configuration */
 	if (is_array($config['ntpd']['pps']) && !empty($config['ntpd']['pps']['port']) &&
 	    file_exists('/dev/'.$config['ntpd']['pps']['port']) &&
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1644,22 +1644,19 @@ function system_ntp_configure() {
 	$ntpcfg .= "tinker panic 0 \n";
 
 	/* Add Orphan mode */
-	$ntpcfg .= "# Orphan mode stratum\n";
+	$ntpcfg .= "# Orphan mode stratum and Maximum candidate NTP peers\n";
 	$ntpcfg .= 'tos orphan ';
 	if (!empty($config['ntpd']['orphan'])) {
 		$ntpcfg .= $config['ntpd']['orphan'];
 	} else {
 		$ntpcfg .= '12';
 	}
-	$ntpcfg .= "\n";
-
-	/* Add min-max ntp peers */
-	$ntpcfg .= "# Min-Max peers\n";
-	$ntpcfg .= 'tos minclock 3 maxclock ';
+	/* Add Maximum candidate NTP peers */
 	if (!empty($config['ntpd']['ntpmaxpeers'])) {
+		$ntpcfg .= ' maxclock ';
 		$ntpcfg .= $config['ntpd']['ntpmaxpeers'];
 	} else {
-		$ntpcfg .= '4';
+		$ntpcfg .= ' maxclock 5';
 	}
 	$ntpcfg .= "\n";
 
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1654,15 +1654,15 @@ function system_ntp_configure() {
 	$ntpcfg .= "\n";
 
 	/* Add min-max ntp peers */
-        $ntpcfg .= "# Min-Max peers\n";
-        $ntpcfg .= 'tos minclock 3 maxclock ';
-        if (!empty($config['ntpd']['ntpmaxpeers'])) {
-                $ntpcfg .= $config['ntpd']['ntpmaxpeers'];
-        } else {
-                $ntpcfg .= '4';
-        }
-        $ntpcfg .= "\n";
-	
+	$ntpcfg .= "# Min-Max peers\n";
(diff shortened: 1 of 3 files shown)


src/etc/inc/system.inc
```diff
@@ -1653,6 +1653,16 @@ function system_ntp_configure() {
...
 
+	/* Add min-max ntp peers */
+        $ntpcfg .= "# Min-Max peers\n";
+        $ntpcfg .= 'tos minclock 3 maxclock ';
+        if (!empty($config['ntpd']['ntpmaxpeers'])) {
+                $ntpcfg .= $config['ntpd']['ntpmaxpeers'];
+        } else {
+                $ntpcfg .= '4';
+        }
+        $ntpcfg .= "\n";
```
rbgarga: Please indent using TABs instead of spaces

kiokoman:
done

src/etc/inc/system.inc
```diff
@@ -1653,6 +1653,16 @@ function system_ntp_configure() {
 	}
 	$ntpcfg .= "\n";
 
+	/* Add min-max ntp peers */
+	$ntpcfg .= "# Min-Max peers\n";
+	$ntpcfg .= 'tos minclock 3 maxclock ';
```
jim-p: There is already a line for `tos` which sets up orphan mode. Does this work with `tos` commands on two different lines, or does one override the other?

The best thing to do would be to combine them in one line, IMO.

Also, the default value for `minclock` in `ntpd` is already `3` so that can be omitted.

src/usr/local/www/services_ntpd.php
```diff
@@ -59,6 +59,10 @@
 	unset($input_errors);
 	$pconfig = $_POST;
 
+	if (!is_numericint($_POST['ntpmaxpeers']) || ($_POST['ntpmaxpeers'] < 4) || ($_POST['ntpmaxpeers'] > 10)) {
+		$input_errors[] = gettext("Max candidate NTP peers must be a number between 4 and 10");
```
jim-p: Where did you come up with these minimum and maximum numbers? I don't see any mention of limits on the values in the NTP docs or source. The default is 10, but someone could potentially want something higher if it's a completely internal setup.

src/etc/inc/system.inc
kiokoman: ok i will work on it

src/usr/local/www/services_ntpd.php
kiokoman: based on this statement With three servers, you have the minimum number of time sources needed to allow ntpd to dectect if one time source is a "falseticker". However ntpd will then be in the position of choosing from the two remaining sources.This configuration provides no redundancy.

**if someone put only 1 pool and set minclock to less than 4 this can lead to trouble.**

With three servers, you have the minimum number of time sources needed to allow ntpd to dectect if one time source is a "falseticker". However ntpd will then be in the position of choosing from the two remaining sources.This configuration provides no redundancy.

With at least four upstream servers, one (or more) can be a "falseticker", or just unreachable, and ntpd will have a sufficient number of sources to choose from.

Specify the maximum number of servers retained by the server discovery schemes. The default is 10, which should typically be changed. This should be an odd number typically two or three more than minclock, plus the number of pool entries. The pool entries must be added as maxclock, but not minclock, also counts the pool entries themselves. 

**For example, tos maxclock 4 with 1 pool lines would keep 3 associations.**

that said i think that maxclock should start from 4 and default set to 5

ok, 10 was my arbitrary upper limit but i think that a limit should be present, there wasn't a knob before but if people start to play with it and set it to 9999 or a random higher value multiplied by all devices out there, it can become a real problem for the ntp pool. maybe we can set to 20 and see if people ask for more? or you tell me

jim-p: I read the same info but I was mostly curious if you had seen a limit mentioned elsewhere that I missed. `4` seems like a sane minimum (Since `minclock` will always be `3`, and it should be higher). `5` is probably the best default since it should be `minclock+2`. The upper bound is less clear. I guess make it `25` for now just for sanity's sake. If someone needs it higher than that, they can patch the source or open a bug report to have it changed.

Since the values are used in multiple places you should probably define the min/max near the top of the file as a variable or constant, and then reference the variables in the later instances. That way if we decide to change them in the future, only one line must be changed.

src/etc/inc/system.inc
```diff
@@ -1644,13 +1644,20 @@ function system_ntp_configure() {
...
+	$ntpcfg .= "# Orphan mode stratum and Maximum candidate NTP peers\n";
 	$ntpcfg .= 'tos orphan ';
 	if (!empty($config['ntpd']['orphan'])) {
 		$ntpcfg .= $config['ntpd']['orphan'];
 	} else {
 		$ntpcfg .= '12';
 	}
+	/* Add Maximum candidate NTP peers */
+	if (!empty($config['ntpd']['ntpmaxpeers'])) {
+		$ntpcfg .= ' maxclock ';
```
jim-p: Rather than repeating `' maxclock '` in both branches of this test since it's added either way, add it to the string before the `if`, and then add either the user-supplied value or `5`. Similar to what is done just above for `orphan`.

src/usr/local/www/services_ntpd.php
```diff
@@ -59,6 +60,10 @@
 	unset($input_errors);
 	$pconfig = $_POST;
 
+	if (!is_numericint($_POST['ntpmaxpeers']) || ($_POST['ntpmaxpeers'] < 4) || ($_POST['ntpmaxpeers'] > $max_candidate_peers)) {
```
jim-p: `4` should be in its own similar variable, like `$min_candidate_peers` -- here and in the option text. Same reason as max was, so it could be changed later if need be.

Also please break up this long line -- After the `||`, start a new line and intent it the same level as the `if` + four spaces. (https://docs.netgate.com/pfsense/en/latest/development/developer-style-guide.html#indent-style)

src/usr/local/www/services_ntpd.php
```diff
@@ -60,8 +61,10 @@
 	unset($input_errors);
 	$pconfig = $_POST;
 
-	if (!is_numericint($_POST['ntpmaxpeers']) || ($_POST['ntpmaxpeers'] < 4) || ($_POST['ntpmaxpeers'] > $max_candidate_peers)) {
-		$input_errors[] = gettext("Max candidate pool peers must be a number between 4 and $max_candidate_peers");
+	if (!is_numericint($_POST['ntpmaxpeers']) ||
+	    ($_POST['ntpmaxpeers'] < $min_candidate_peers) ||
+	    ($_POST['ntpmaxpeers'] > $max_candidate_peers)) {
+		$input_errors[] = gettext("Max candidate pool peers must be a number between $min_candidate_peers and $max_candidate_peers");
```
jim-p: You shouldn't use variables inside `gettext()` strings. It should be like this instead:

    sprintf(gettext("Max candidate pool peers must be a number between %d and %d"), $min_candidate_peers, $max_candidate_peers);

src/usr/local/www/services_ntpd.php
```diff
@@ -335,7 +338,7 @@ function build_interface_list() {
 	'Max candidate pool peers',
 	'number',
 	$pconfig['ntpmaxpeers'],
-	['min' => 4, 'max' => $max_candidate_peers]
+	['min' => $min_candidate_peers, 'max' => $max_candidate_peers]
 ))->setHelp('Maximum candidate NTP pool peers. Remember that many servers inside pools are provided by volunteers, ' .
 	'all you will gain from a higher number is extra load on the volunteer time servers. (Default: 5).');
```
jim-p: Reword to avoid addressing the user directly ("you").

Perhaps: 
> Maximum number of candidate peers in the NTP pool. This value should be set low enough to provide sufficient alternate sources while not contacting an excessively large number of peers. Many servers inside public pools are provided by volunteers, and a large candidate pool places unnecessary extra load on the volunteer time servers for little to no added benefit. (Default: 5).

PR #4226, opened 2020-03-10
Add static mapping count to DHCP Server interface tabs. Issue #9282

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/9282
- [ ] Ready for review

services - > DHCP Server > Interface
need a counter that count add static mapping in "DHCP Static Mappings for this Interface"

This PR adds static mappings counter

### DIFF
--- a/src/usr/local/www/services_dhcp.php
+++ b/src/usr/local/www/services_dhcp.php
@@ -1636,7 +1636,14 @@ if (!is_numeric($pool) && !($act == "newpool")) {
 ?>
 
 <div class="panel panel-default">
-	<div class="panel-heading"><h2 class="panel-title"><?=gettext("DHCP Static Mappings for this Interface")?></h2></div>
+<?php
+	if (is_array($a_maps) && (count($a_maps) > 0)) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_dhcp.php
```diff
@@ -1636,7 +1636,14 @@ function build_pooltable() {
 ?>
 
 <div class="panel panel-default">
-	<div class="panel-heading"><h2 class="panel-title"><?=gettext("DHCP Static Mappings for this Interface")?></h2></div>
+<?php
+	if (is_array($a_maps) && (count($a_maps) > 0)) {
+		$title = gettext("DHCP Static Mappings for this Interface (total: " . count($a_maps) . ")");
```
jim-p: You shouldn't use variables in `gettext()` strings. It should be run through `sprintf()` to replace the value. Something like this:

    $title = sprintf(gettext('DHCP Static Mappings for this Interface (total: %d)'), count($a_maps));

vktg: ok, fixed

PR #4229, opened 2020-03-12
Add missing whitespace

mkochenough:
- [ ] Ready for review

This is an annoying bug when using the german translation. I don't filed a redmine issue yet. Is this really needed in this simple case?


sbeaver-netgate:
This issue must be resolved in the Zanata translation platform. pfSense.po files are overwritten so this change would be lost. I will take care of that. Thanks.

https://www.netgate.com/blog/pfsense-software-translations-with-zanata.html

sbeaver-netgate:
Zanata translation has been updated

PR #4230, opened 2020-03-12
IPsec PH2 bypass mode. Issue #3329

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/3329
- [ ] Ready for review

We should have the ability in Phase 2 to negate the action ("none" in the SPD) so that specific traffic can be made to not enter an IPsec tunnel.
These entries would not need to have any encryption options chosen, only the networks defined.

I decided not to create two 'IPv4 Bypass' and 'IPv6 Bypass modes' - this can confuse users and the drop-down selection overload
The IP protocol of the interface subnet is selected depending on the IP type of remoteid.

It creates separate 'bypassX' entries for each rule to work properly


src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -427,9 +441,11 @@
 
 $localid_help_tunnel  = "Local network component of this IPsec security association.";
 $localid_help_vti     = "Local point-to-point IPsec interface tunnel network address.";
+$localid_help_bypass  = "Local network that bypassed from IPsec processing.";
```
jim-p: Change `bypassed from` to `bypasses`.
Add ` for this IPsec connection` to the end since the bypass is setup as specific to this peer, not all peers.

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -427,9 +441,11 @@
...
 $localid_help_mobile  = "Network reachable by mobile IPsec clients.";
 $remoteid_help_tunnel = "Remote network component of this IPsec security association.";
 $remoteid_help_vti    = "Remote point-to-point IPsec interface tunnel network address.";
+$remoteid_help_bypass = "Remote network that bypassed from IPsec processing.";
```
jim-p: Change `bypassed from` to `bypasses`.
Add ` for this IPsec connection` to the end since the bypass is setup as specific to this peer, not all peers.

vktg: fixed

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -427,9 +441,11 @@
 
 $localid_help_tunnel  = "Local network component of this IPsec security association.";
 $localid_help_vti     = "Local point-to-point IPsec interface tunnel network address.";
+$localid_help_bypass  = "Local network that bypasses from IPsec processing for this IPsec connection.";
```
jim-p: Remove "from"

src/usr/local/www/vpn_ipsec_phase2.php
```diff
@@ -427,9 +441,11 @@
...
 $localid_help_mobile  = "Network reachable by mobile IPsec clients.";
 $remoteid_help_tunnel = "Remote network component of this IPsec security association.";
 $remoteid_help_vti    = "Remote point-to-point IPsec interface tunnel network address.";
+$remoteid_help_bypass = "Remote network that bypasses from IPsec processing for this IPsec connection.";
```
jim-p: Remove "from"

vktg: Removed,
But it’s true - if you add bypass rule, it's bypassed for all peers

vktg:
Need to decide where to put this control in this case
Extra tab on top, after 'Advanced Settings' ?

jim-p:
If it's always global then putting it under Advanced Settings should be OK. I'd rather it not have its own tab since that is more likely for it to be misused/misunderstood.

A rowhelper in its own section at the bottom should be fine.

In that case you should also combine it with the existing LAN bypass function as separate children underneath it.

vktg:
Done

I deleted `$scconf['connections']['bypass']['remote_addrs'] = "127.0.0.1";` because bypass works for all peers

swanctl.conf example:
```
connections {
	bypass {
		children {
			bypasslan {
				local_ts = 192.168.3.0/24
				remote_ts = 192.168.3.0/24
				mode = pass
				start_action = trap
			}
			rule0 {
				local_ts = 10.123.0.0/24
				remote_ts = 10.155.155.0/24
				mode = pass
				start_action = trap
			}
		}
	}
	con1000 {
...
```

src/etc/inc/ipsec.inc
```diff
@@ -1668,18 +1668,41 @@ function ipsec_setup_bypasslan() {
 	if (!empty($bypassnets)) {
 		$bypass = implode(',', $bypassnets);
 		$scconf['connections']['bypass'] = array();
-		/* Prevents the connection from being considered for remote peers */
-		$scconf['connections']['bypass']['remote_addrs'] = "127.0.0.1";
```
jim-p: This is absolutely required so that traffic matching the bypass TS does not ever try to initiate or be matched to a specific remote peer.

https://wiki.strongswan.org/projects/strongswan/wiki/UsableExamples#Passthrough-policy

src/etc/inc/ipsec.inc
```diff
@@ -1668,18 +1668,41 @@ function ipsec_setup_bypasslan() {
...
+		$scconf['connections']['bypass']['children']['bypasslan'] = array();
+		$scconf['connections']['bypass']['children']['bypasslan']['local_ts'] = $bypass;
+		$scconf['connections']['bypass']['children']['bypasslan']['remote_ts'] = $bypass;
+		$scconf['connections']['bypass']['children']['bypasslan']['mode'] = "pass";
+		$scconf['connections']['bypass']['children']['bypasslan']['start_action'] = "trap";
 	}
 	return;
 }
 
+function ipsec_setup_extrabypass() {
```
jim-p: Why make an extra function?

If they are both only ever called together, just use one function since it's all setting up the same section.

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -68,6 +70,33 @@
...
+		if (isset($_POST["source{$x}"]) && isset($_POST["destination{$x}"]) &&
+		    isset($_POST["srcmask{$x}"]) && isset($_POST["dstmask{$x}"])) {
+			$source = htmlspecialchars($_POST["source{$x}"]) . '/' . htmlspecialchars($_POST["srcmask{$x}"]);
+			$destination = htmlspecialchars($_POST["destination{$x}"]) . '/' . htmlspecialchars($_POST["dstmask{$x}"]);
+			$rule = array();
+			$rule['source'] = htmlspecialchars($_POST["source{$x}"]);
+			$rule['srcmask'] = htmlspecialchars($_POST["srcmask{$x}"]);
+			$rule['destination'] = htmlspecialchars($_POST["destination{$x}"]);
+			$rule['dstmask'] = htmlspecialchars($_POST["dstmask{$x}"]);
+			$bypassrules['rule'][] = $rule;
```
jim-p: This could all be done in one assignment

    $bypassrules['rule'][] = array(
        ...
    );

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -68,6 +70,33 @@
...
+	for ($x = 0; $x < 99; $x++) {
+		if (isset($_POST["source{$x}"]) && isset($_POST["destination{$x}"]) &&
+		    isset($_POST["srcmask{$x}"]) && isset($_POST["dstmask{$x}"])) {
+			$source = htmlspecialchars($_POST["source{$x}"]) . '/' . htmlspecialchars($_POST["srcmask{$x}"]);
+			$destination = htmlspecialchars($_POST["destination{$x}"]) . '/' . htmlspecialchars($_POST["dstmask{$x}"]);
+			$rule = array();
+			$rule['source'] = htmlspecialchars($_POST["source{$x}"]);
+			$rule['srcmask'] = htmlspecialchars($_POST["srcmask{$x}"]);
+			$rule['destination'] = htmlspecialchars($_POST["destination{$x}"]);
+			$rule['dstmask'] = htmlspecialchars($_POST["dstmask{$x}"]);
```
jim-p: 1. The validation should happen before the values are stored.
2. Why run all of these through `htmlspecialchars()` here? If they get printed in the error, run that through `htmlspecialchars()` if they fail validation before using them in the error message. If they pass validation they wouldn't need encoding since we are certain they are valid IP addresses/subnets.

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -354,17 +403,92 @@ function maxmss_checked(obj) {
...
+$numrows = count($item) -1;
+$counter = 0;
+
+$numrows = count($pconfig['bypassrules']['rule']) -1;
+
+foreach ($pconfig['bypassrules']['rule'] as $rule) {
+	$source = $rule['source'];
+	$srcmask = $rule['srcmask'];
+	$destination = $rule['destination'];
+	$dstmask = $rule['dstmask'];
```
jim-p: Why assign these to new variables only to use them once? Seems unnecessary

src/etc/inc/ipsec.inc
vktg: Ok, thanks for info

All fixed

Danilo-Z:
The functionality works great, it bypasses defined networks/hosts fine. The only objection is that I couldn't delete the defined entry. I could only disable the bypass option.

vktg:
> The functionality works great, it bypasses defined networks/hosts fine. The only objection is that I couldn't delete the defined entry. I could only disable the bypass option.

That seems `rowhelper` issue, same behavior with the "Additional BOOTP/DHCP Options" entries on the DHCP Server configuration page

PR #4231, opened 2020-03-12
Password confirmation when exporting encrypted backup file. Issue #10301

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10301
- [ ] Ready for review

redmine:

> I would highly recommend to implement password confirmation so you have to insert the password two times when exporting an encrypted pfsense backup file. Or alternatively the option to show the password before submitting it.
> It would be a mess in case I really needed to restore the config but got an invalid password because I did a mistake when creating it.

This PR adds password confirmation

### DIFF
--- a/src/usr/local/www/diag_backup.php
+++ b/src/usr/local/www/diag_backup.php
@@ -157,8 +157,8 @@ if ($_POST) {
 	if ($mode) {
 		if ($mode == "download") {
 			if ($_POST['encrypt']) {
-				if (!$_POST['encrypt_password']) {
-					$input_errors[] = gettext("A password for encryption must be supplied and confirmed.");
+				if (!$_POST['encrypt_password'] || ($_POST['encrypt_password'] != $_POST['encrypt_password_confirm'])) {
+					$input_errors[] = gettext("Supplied password and confirmation do not match.");
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


sbeaver-netgate:
We already have a mechanism for automatically inserting password/confirmation fields in the GUI. Simply specify form type = "password". See: https://gitlab.netgate.com/pfSense/pfSense/blob/master/src/usr/local/www/services_pppoe_edit.php#L463

vktg:
hm, I don't see any password confirmation field in my browser on the services_pppoe_edit.php user password

sbeaver-netgate:
pfSense_localdomain_-_Services__PPPoE_Server__Edit

jim-p:
PPPoE user list would be different since it's a rowhelper, but other places have confirmation.

sbeaver-netgate:
radius secret uses the same input type

Steve
--

vktg:
oh, something with gitlab link before
now I see, 
fixed

src/usr/local/www/diag_backup.php
```diff
@@ -157,7 +157,7 @@ function check_and_returnif_section_exists($section) {
 	if ($mode) {
 		if ($mode == "download") {
 			if ($_POST['encrypt']) {
-				if (!$_POST['encrypt_password']) {
+				if (!$_POST['encrypt_password'] || ($_POST['encrypt_password'] != $_POST['encrypt_password_confirm'])) {
 					$input_errors[] = gettext("A password for encryption must be supplied and confirmed.");
```
sbeaver-netgate: Suggest: "Supplied password and confirmation do not match."

vktg: np, replaced

PR #4232, opened 2020-03-13
Exclude unsupported interfaces from DHCP Relay. Issue #10341

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10341
- [ ] Ready for review

Such pseudo-interfaces as VTI, GIF, GRE, L2TP, PPTP, PPP, PPPoE and OpenVPN is unsupported by DHCP Relay:
```
Mar 13 09:03:40    dhcrelay    65228    Unsupported device type 131 for "ipsec2000" 
Mar 13 09:04:14    dhcrelay    32263    Unsupported device type 240 for "gif0" 
Mar 13 09:22:04    dhcrelay    53447    Unsupported device type 131 for "gre0" 
Mar 13 09:24:10    dhcrelay    10853    Unsupported device type 53 for "l2tp1" 
```

They must be excluded from interface selection list,
service must also avoid use them for auto-generated upstream interface list:
https://github.com/pfsense/pfsense/blob/d09e19adf4253251dc2aa6d9edc3043e69096d4f/src/etc/inc/services.inc#L1823

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -7162,4 +7162,14 @@ function create_interface_list() {
 	return($iflist);
 }
 
+function is_pseudo_interface($inf) {
+	$psifs = array('ovpn', 'ipsec', 'l2tp', 'pptp', 'gif', 'gre', 'ppp', 'pppoe');
+	foreach ($psifs as $pif) {
+		if (substr($inf, 0, strlen($pif)) == $pif) {
+			return true;
+		}
+	}
+	return false;
+}
+
 ?>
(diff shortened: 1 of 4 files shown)


vktg:
+ removed link_interface_to_bridge($dhcrelayif) tests
to resolve issue https://redmine.pfsense.org/issues/7778

tested on 2.5.0.a.20200312.1334 - bridges/bridge members can be upstream/downstream interfaces

stephenw10:
DHCP relay can work with OpenVPN but only in TAP mode. I have set that up for at least one customer. This will beak that.

jim-p:
> DHCP relay can work with OpenVPN but only in TAP mode. I have set that up for at least one customer. This will beak that.

OpenVPN was already excluded by https://redmine.pfsense.org/issues/8443

PR #4235, opened 2020-03-15
DHCPv6 service Dynamic DNS fix. Issue #10346

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10346
- [ ] Ready for review

1) The DHCPv6 Server & RA page is not allow to enter IPv6 address in DDNS Server IP field (IPv4?)
2) The DHCPv6 Server & RA page DDNS Domain Key name may only contain the characters a-z, A-Z 0-9, '-' and '_'
'=', '@', '.', '^', etc. are special characters and dhcpd doesn't start if config contains it
3) The DynDNS secret must be enclosed in quotes (“secret”), otherwise you will get error message:
```
/services_dhcpv6.php: The command '/usr/local/sbin/dhcpd -6 -user dhcpd -group _dhcp -chroot 
/var/dhcpd -cf /etc/dhcpdv6.conf -pf /var/run/dhcpdv6.pid vtnet2' returned exit code '1', the output 
was 'Internet Systems Consortium DHCP Server 4.4.1 Copyright 2004-2018 Internet Systems 
Consortium. All rights reserved. For info, please visit https://www.isc.org/software/dhcp/ 
/etc/dhcpdv6.conf line 27: partial base64 value left over: 3. \x09secret 123; ^ Configuration file 
errors encountered -- exiting If you think you have received this message due to a bug rather than a 
configuration issue please read the section on submitting bugs on either our web page at 
www.isc.org or in the README file before submitting a bug. These pages explain the proper 
process and the information we find helpful for debugging. exiting.'

```

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1278,7 +1278,7 @@ EOD;
 		$dhcpdifs[] = get_real_interface($dhcpif);
 		if ($newzone['domain-name']) {
 			if ($need_ddns_updates) {
-				$newzone['dns-servers'] = array($dhcpifconf['ddnsdomainprimary']);
+				$newzone['dns-servers'] = array($dhcpifconf['ddnsdomainprimary'], $dhcpifconf['ddnsdomainsecondary']);
 				$newzone['ddnsdomainkeyname'] = $dhcpifconf['ddnsdomainkeyname'];
 				$newzone['ddnsdomainkeyalgorithm'] = $dhcpifconf['ddnsdomainkeyalgorithm'];
 				$newzone['ddnsdomainkey'] = $dhcpifconf['ddnsdomainkey'];
@@ -1350,9 +1350,9 @@ function dhcpdzones($ddns_zones) {
 			$primary = $zone['dns-servers'][0];
 			$secondary = empty($zone['dns-servers'][1]) ? "" : $zone['dns-servers'][1];
 
-			// Make sure we aren't using any invalid or IPv6 DNS servers.
-			if (!is_ipaddrv4($primary)) {
-				if (is_ipaddrv4($secondary)) {
+			// Make sure we aren't using any invalid servers.
+			if (!is_ipaddr($primary)) {
+				if (is_ipaddr($secondary)) {
 					$primary = $secondary;
 					$secondary = "";
 				} else {
@@ -1363,9 +1363,15 @@ function dhcpdzones($ddns_zones) {
 			// We don't need to add zones multiple times.
 			if ($zone['domain-name'] && !in_array($zone['domain-name'], $added_zones)) {
 				$dhcpdconf .= "zone {$zone['domain-name']}. {\n";
-				$dhcpdconf .= "	primary {$primary};\n";
+				if (is_ipaddrv4($primary)) {
+					$dhcpdconf .= "	primary {$primary};\n";
+				} else {
+					$dhcpdconf .= "	primary6 {$primary};\n";
+				}
 				if (is_ipaddrv4($secondary)) {
 					$dhcpdconf .= "	secondary {$secondary};\n";
+				} elseif (is_ipaddrv6($secondary)) {
+					$dhcpdconf .= "	secondary6 {$secondary};\n";
 				}
 				if ($zone['ddnsdomainkeyname'] <> "" && $zone['ddnsdomainkey'] <> "") {
 					$dhcpdconf .= "	key {$zone['ddnsdomainkeyname']};\n";
@@ -1375,9 +1381,15 @@ function dhcpdzones($ddns_zones) {
 			}
 			if ($zone['ptr-domain'] && !in_array($zone['ptr-domain'], $added_zones)) {
 				$dhcpdconf .= "zone {$zone['ptr-domain']} {\n";
-				$dhcpdconf .= "	primary {$primary};\n";
+				if (is_ipaddrv4($primary)) {
+					$dhcpdconf .= "	primary {$primary};\n";
+				} else {
+					$dhcpdconf .= "	primary6 {$primary};\n";
+				}
 				if (is_ipaddrv4($secondary)) {
 					$dhcpdconf .= "	secondary {$secondary};\n";
+				} elseif (is_ipaddrv6($secondary)) {
+					$dhcpdconf .= "	secondary6 {$secondary};\n";
 				}
 				if ($zone['ddnsdomainkeyname'] <> "" && $zone['ddnsdomainkey'] <> "") {
 					$dhcpdconf .= "	key {$zone['ddnsdomainkeyname']};\n";
@@ -1550,7 +1562,7 @@ EOD;
 		    !empty($dhcpv6ifconf['ddnsdomain'])) {
 			$newzone = array();
 			$newzone['domain-name'] = $dhcpv6ifconf['ddnsdomain'];
-			$newzone['dns-servers'][] = $dhcpv6ifconf['ddnsdomainprimary'];
+			$newzone['dns-servers'] = array($dhcpv6ifconf['ddnsdomainprimary'], $dhcpv6ifconf['ddnsdomainsecondary']);
 			$newzone['ddnsdomainkeyname'] = $dhcpv6ifconf['ddnsdomainkeyname'];
 			$newzone['ddnsdomainkey'] = $dhcpv6ifconf['ddnsdomainkey'];
 			$ddns_zones[] = $newzone;
@@ -1559,7 +1571,7 @@ EOD;
 				foreach ($ptr_zones as $ptr_zone) {
 					$reversezone = array();
 					$reversezone['ptr-domain'] = $ptr_zone;
-					$reversezone['dns-servers'][] = $dhcpv6ifconf['ddnsdomainprimary'];
+					$reversezone['dns-servers'] = array($dhcpv6ifconf['ddnsdomainprimary'], $dhcpv6ifconf['ddnsdomainsecondary']);
 					$reversezone['ddnsdomainkeyname'] = $dhcpv6ifconf['ddnsdomainkeyname'];
 					$reversezone['ddnsdomainkey'] = $dhcpv6ifconf['ddnsdomainkey'];
 					$ddns_zones[] = $reversezone;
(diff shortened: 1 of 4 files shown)


src/etc/inc/services.inc
```diff
@@ -1274,7 +1274,7 @@ function dhcpdkey($dhcpifconf) {
 		$algorithm = empty($dhcpifconf['ddnsdomainkeyalgorithm']) ? 'hmac-md5' : $dhcpifconf['ddnsdomainkeyalgorithm'];
 		$dhcpdconf .= "key {$dhcpifconf['ddnsdomainkeyname']} {\n";
 		$dhcpdconf .= "	algorithm {$algorithm};\n";
-		$dhcpdconf .= "	secret {$dhcpifconf['ddnsdomainkey']};\n";
+		$dhcpdconf .= "	secret \"{$dhcpifconf['ddnsdomainkey']}\";\n";
```
jim-p: The value inside should have double quotes escaped, or else you end up with another situation like https://redmine.pfsense.org/issues/10295

src/usr/local/www/services_dhcpv6.php
```diff
@@ -313,8 +313,11 @@ function dhcpv6_apply_changes($dhcpdv6_enable_changed) {
 	if ($_POST['ddnsupdate'] && !is_domain($_POST['ddnsdomain'])) {
 		$input_errors[] = gettext("A valid domain name must be specified for the dynamic DNS registration.");
 	}
-	if ($_POST['ddnsupdate'] && !is_ipaddrv4($_POST['ddnsdomainprimary'])) {
-		$input_errors[] = gettext("A valid primary domain name server IPv4 address must be specified for the dynamic domain name.");
+	if ($_POST['ddnsupdate'] && !is_ipaddrv6($_POST['ddnsdomainprimary'])) {
```
jim-p: Does this have to be IPv6? It's a DNS server address, it shouldn't have to match the type of address the daemon is handling. It should probably accept either type in either page (IPv4 or IPv6 on DHCPv4 and DHCPv6). May need testing to confirm that it works in both of those cases.

src/usr/local/www/services_dhcpv6.php
```diff
@@ -800,8 +803,8 @@ function dhcpv6_apply_changes($dhcpdv6_enable_changed) {
 	'ddnsdomainprimary',
 	'DDNS Server IP',
 	$pconfig['ddnsdomainprimary'],
-	'V4'
-))->setHelp('Enter the primary domain name server IPv4 address for the dynamic domain name.');
+	'V6'
+))->setHelp('Enter the primary domain name server IPv6 address for the dynamic domain name.');
```
jim-p: If the validation test gets changed as suggested, change this to match.

src/etc/inc/services.inc
vktg: This is incorrect too.
dnssec-keygen should be used, 
dhcpd.conf(5):
>       You should choose your own secret key, of course.  The ISC BIND 9  dis-
>        tribution  comes	 with  a  program  for	generating  secret keys	called
>        dnssec-keygen.  If you are using	BIND 9's dnssec-keygen,	the above  key
>        would be	created	as follows:
> 
> 	    dnssec-keygen -a HMAC-MD5 -b 128 -n	USER DHCP_UPDATER
> 
>        The  key	 name, algorithm, and secret must match	that being used	by the
>        DNS server.

I tried to use hash_hmac(), but it needs testing

src/usr/local/www/services_dhcpv6.php
vktg: True,
it uses primary6 and secondary6 for IPv6 addresses,
see dhcpd.conf(5):
```
       The primary statement specifies the IP address of the name server whose
       zone  information  is to	be updated.  In	addition to the	primary	state-
       ment there are also the primary6	, secondary and	secondary6 statements.
       The  primary6  statement	specifies an IPv6 address for the name server.
```

src/usr/local/www/services_dhcpv6.php
vktg: changed to "BOTH"

src/etc/inc/services.inc
jim-p: That would be to generate a new key, but in this case the user would be entering a key generated elsewhere (on the actual name server) so it has to be a text input.

vktg: all done
not it fixes DynDNS key using strict base64_decode()

vktg:
> I wonder if it is OK to mix IPv4 and IPv6 (e.g. `primary` with `secondary6`). If not we may have to protect against that. Or maybe if there is one IPv4 and one IPv6 should it be `primary` and `primary6` instead? Needs some testing to determine.

I was checked code, and found that there is only one case when more than one DNS server is used:
https://github.com/pfsense/pfsense/blob/2c9e1e43f3619af39dc12c67830cc63a08ac8829/src/etc/inc/services.inc#L780

another question,-
for some reason ddnsdomainprimary is interpreted as an array::
`$newzone['dns-servers'] = array($dhcpifconf['ddnsdomainprimary']);`
https://github.com/pfsense/pfsense/blob/2c9e1e43f3619af39dc12c67830cc63a08ac8829/src/etc/inc/services.inc#L1281
but WebGUI doesn't allow you to enter more than one DNS server.
Should it?

jim-p:
> another question,-
> for some reason ddnsdomainprimary is interpreted as an array::
> 
> https://github.com/pfsense/pfsense/blob/2c9e1e43f3619af39dc12c67830cc63a08ac8829/src/etc/inc/services.inc#L1281

I'm not sure why that is, but it is similar to `dnsserver` which is marked as a list tag which would be treated as an array by the XML parser. Perhaps in the past it overlapped that but eventually changed and that workaround didn't get removed.

> but WebGUI doesn't allow you to enter more than one DNS server.
> Should it?

If the DHCP daemon supports using more than one server we should accommodate it, if it's feasible.

vktg:
Now DHCP/DHCPv6 can use two DDNS servers,
A DDNS pair can have a different IP address family - the dhcpd daemon starts normally

but I found that most of the DDNS options for static entries are ignored:
```
if (isset($sm['ddnsupdate'])) {
					if (($sm['ddnsdomain'] <> "") && ($sm['ddnsdomain'] != $dhcpifconf['ddnsdomain'])) {
						$smdnscfg .= "		ddns-domainname \"{$sm['ddnsdomain']}\";\n";
					}
					$smdnscfg .= "		ddns-update-style interim;\n";
				}
```
https://github.com/pfsense/pfsense/blob/b943d20dcd9a580c18ce804b47f512855272f1dd/src/etc/inc/services.inc#L1183

src/usr/local/www/services_dhcp.php
```diff
@@ -386,14 +387,25 @@ function validate_partial_mac_list($maclist) {
...
+			$input_errors[] = gettext("A valid secondary domain name server IP address must be specified for the dynamic domain name.");
+		}
+		if (!$_POST['ddnsdomainkeyname'] || !$_POST['ddnsdomainkeyalgorithm'] || !$_POST['ddnsdomainkey']) {
+			$input_errors[] = gettext("A valid domain key name, algorithm and secret must be specified.");
+		}
+		if (preg_match('/[^A-Za-z0-9\-_]/', $_POST['ddnsdomainkeyname'])) {
+			$input_errors[] = gettext("The domain key name may only contain the characters a-z, A-Z, 0-9, '-' and '_'");
+		}
+		if ($_POST['ddnsdomainkey'] && !base64_decode($_POST['ddnsdomainkey'], true)) {
+			$input_errors[] = gettext("The domain key secret must be Base64 encoded value.");
```
jim-p, line 407: This should either be
* "The domain key secret must be **a** Base64 encoded value."
or
* "The domain key secret must be Base64 encoded."

src/usr/local/www/services_dhcp_edit.php
```diff
@@ -274,12 +274,22 @@ function staticmaps_sort($ifgui) {
...
+			$input_errors[] = gettext("A valid domain name must be specified for the dynamic DNS registration.");
+		}
+		if (!is_ipaddr($_POST['ddnsdomainprimary'])) {
+			$input_errors[] = gettext("A valid primary domain name server IP address must be specified for the dynamic domain name.");
+		}
+		if (preg_match('/[^A-Za-z0-9\-_]/', $_POST['ddnsdomainkeyname'])) {
+			$input_errors[] = gettext("The domain key name may only contain the characters a-z, A-Z, 0-9, '-' and '_'");
+		}
+		if ($_POST['ddnsdomainkey'] && !base64_decode($_POST['ddnsdomainkey'], true)) {
+			$input_errors[] = gettext("The domain key secret must be Base64 encoded value.");
```
jim-p, line 291: See previous note, this should match.

src/usr/local/www/services_dhcpv6.php
```diff
@@ -310,14 +311,25 @@ function dhcpv6_apply_changes($dhcpdv6_enable_changed) {
...
+			$input_errors[] = gettext("A valid secondary domain name server IP address must be specified for the dynamic domain name.");
+		}
+		if (!$_POST['ddnsdomainkeyname'] || !$_POST['ddnsdomainkeyalgorithm'] || !$_POST['ddnsdomainkey']) {
+			$input_errors[] = gettext("A valid domain key name, algorithm and secret must be specified.");
+		}
+		if (preg_match('/[^A-Za-z0-9\-_]/', $_POST['ddnsdomainkeyname'])) {
+			$input_errors[] = gettext("The domain key name may only contain the characters a-z, A-Z, 0-9, '-' and '_'");
+		}
+		if ($_POST['ddnsdomainkey'] && !base64_decode($_POST['ddnsdomainkey'], true)) {
+			$input_errors[] = gettext("The domain key secret must be Base64 encoded value.");
```
jim-p, line 331: See previous note, this should match.

src/usr/local/www/services_ntpd_acts.php
```diff
@@ -0,0 +1,317 @@
+<?php
```
jim-p, line 1: This file doesn't seem to belong to this PR

vktg: all fixed

src/usr/local/www/services_dhcp_edit.php
jim-p: This one didn't get fixed.

vktg:
all fixed now

most DynDNS options for static mappings are ignored by `services_dhcpdv4_configre()`,
it seems this is just copied block code from services_dhcp.php to services_dhcp_edit.php
I'll create another issue for this

PR #4238, opened 2020-03-17
Add units in ntp status page. Issue #2850

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/2850
- [ ] Ready for review

Hi, I am trying to set up a stratum 1 ntp server with a serial gps (if that is what it is called)
and when I am checking the ntp status page, there is no units behind the numbers.

after figuring out that most of the page is the output of ntpq -p, and looking
up the man page online, I see that the polling interval is in seconds,
and delay, offset, jitter is in milliseconds.

It would be great if there was units behind the numbers (s,ms) for simplicity.

man ntpq(1):
...
poll       poll interval (s)

Other values in milliseconds


vktg:
ok, fixed

PR #4243, opened 2020-03-20
Add option to disable NTP server. Issue #3567

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/3567
- [ ] Ready for review

This is resolved copy of the original PR https://github.com/pfsense/pfsense/pull/3627 by @doktornotor 

It uses `<enable></enable>` to check that the service is enabled.
config version bump is used to add this value.

### DIFF
--- a/src/usr/local/www/services_ntpd.php
+++ b/src/usr/local/www/services_ntpd.php
@@ -93,6 +93,7 @@ if ($_POST) {
 	}
 
 	if (!$input_errors) {
+		$config['ntpd']['enable'] = isset($pconfig['enable']) ? 'enabled' : 'disabled';
 		if (is_array($_POST['interface'])) {
 			$config['ntpd']['interface'] = implode(",", $_POST['interface']);
 		} elseif (isset($config['ntpd']['interface'])) {
@@ -221,6 +222,7 @@ function build_interface_list() {
 
 init_config_arr(array('ntpd'));
 $pconfig = &$config['ntpd'];
+$config['ntpd']['enable'] = isset($pconfig['enable']) ? 'enabled' : 'disabled';
 if (empty($pconfig['interface'])) {
 	$pconfig['interface'] = array();
 } else {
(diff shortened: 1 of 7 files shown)


src/usr/local/www/services_ntpd.php
```diff
@@ -43,14 +43,7 @@
 }
 
 if (empty($config['ntpd']['interface'])) {
-	if (is_array($config['installedpackages']['openntpd']) && is_array($config['installedpackages']['openntpd']['config']) &&
```
vktg: Can this be deleted?
Seems like legacy code, not actual for 2.5

jim-p: There are a few other references around to the old `openntpd` code around in various places. It has probably been gone long enough to remove at this point. Or we could setup some upgrade code to handle this upgrade rather than doing it here or anywhere else.

Either way though I'd prefer to see that in a separate issue+PR rather than combined with this change.

rbgarga:
@vktg please rebase your fork to resolve conflicts

vktg:
all fixed

src/usr/local/www/services_ntpd.php
```diff
@@ -221,6 +226,11 @@ function build_interface_list() {
 
 init_config_arr(array('ntpd'));
 $pconfig = &$config['ntpd'];
+if (!isset($pconfig['enable']) || ($pconfig['enable'] == 'enabled')) {
+	$pconfig['enable'] = true;
+} else {
+	$pconfig['enable'] = false;
+}
```
jim-p: This could be simplified:

`$pconfig['enable'] = (!isset($pconfig['enable']) || ($pconfig['enable'] == 'enabled'));`

src/usr/local/www/services_ntpd.php
```diff
@@ -93,6 +93,11 @@
 	}
 
 	if (!$input_errors) {
+		if (isset($pconfig['enable'])) {
+			$config['ntpd']['enable'] = 'enabled';
+		} else {
+			$config['ntpd']['enable'] = 'disabled';
+		}
```
jim-p: This could be simplified:

`$config['ntpd']['enable'] = isset($pconfig['enable']) ? 'enabled' : 'disabled';`

vktg: thanks,
all fixed

PR #4246, opened 2020-03-21
Ignore events on LAGG members. Issue #10365

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10365
- [ ] Ready for review

The LAGG interface member is cxl0 and cxl1,
if one of these interfaces is down, it triggers check_reload_status:
```
check_reload_status: Linkup starting cxl0
check_reload_status: Linkup starting cxl0
check_reload_status: Reloading filter
check_reload_status: Reloading filter
```

there is code to ignore such events on the BRIDGE interfaces:
https://github.com/pfsense/pfsense/blob/892d8a10810092f47bc417639a9d74346fe7bb28/src/etc/rc.linkup#L74

This PR adds the same code for LAGG members checking,
and fixes LAGG Ports field on the Status / Interfaces page.

### DIFF
--- a/src/etc/rc.linkup
+++ b/src/etc/rc.linkup
@@ -79,6 +79,14 @@ function handle_argument_group($iface, $argument2) {
 		return;
 	}
 
+	/* Ignore events on LAGG members without IP address configuration */
+	$lagg_if = link_interface_to_lagg($iface);
+	if (!empty($lagg_if) && empty($ipaddr) && empty($ip6addr)) {
+		log_error("Ignoring link event for LAGG member without IP address configuration");
+
+		return;
+	}
+
 	if ($staticv4 === true && $staticv6 === true) {
 		$friendly = convert_friendly_interface_to_friendly_descr($iface);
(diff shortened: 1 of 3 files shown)


src/etc/rc.linkup
```diff
@@ -79,6 +79,14 @@ function handle_argument_group($iface, $argument2) {
 		return;
 	}
 
+	/* Ignore events on LAGG members when IP is not configured on it */
```
jim-p: "Ignore events on LAGG members without IP address configuration"

src/etc/rc.linkup
```diff
@@ -79,6 +79,14 @@ function handle_argument_group($iface, $argument2) {
...
+	$lagg_if = link_interface_to_lagg($iface);
+	if (!empty($lagg_if) && empty($ipaddr) && empty($ip6addr)) {
+		log_error("Ignoring link event for LAGG member without IP config");
```
jim-p: "Ignoring link event for LAGG member without IP address configuration"

vktg: all fixed

PR #4248, opened 2020-03-23
Remote OpenVPN server proto definition. Issue #10368

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10368
- [ ] Ready for review

When implementing a OpenVPN server and selecting the protocol "udp on ipv4 only" i expect, that that is exaclty what is included in my client configuration file.
Since this is not what is being configured, clients which have ipv6 available will try to connect through ipv6 and since the server is not running on ipv6 they are unable to connect.

What actually is included:
remote example.org 1194 udp

What i expect to be included:
remote exmaple.org 1194 udp4

This PR adds protocol definition.

TODO: same for OpenVPN client export

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1139,7 +1139,8 @@ function openvpn_reconfigure($mode, $settings) {
 		$conf .= "management {$g['openvpn_base']}/{$mode_id}/sock unix\n";
 
 		// The remote server
-		$conf .= "remote {$settings['server_addr']} {$settings['server_port']}\n";
+		$remoteproto = strtolower($settings['protocol']);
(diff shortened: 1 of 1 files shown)


vktg:
from openvpn(8):

--remote host [port] [proto]
              Remote host name or IP address.  On the client,  multiple  --re‐
              mote  options may be specified for redundancy, each referring to
              a different OpenVPN server.  Specifying  multiple  --remote  op‐
              tions  for  this  purpose  is a special case of the more general
              connection-profile feature.  See the <connection>  documentation
              below.

              The  OpenVPN client will try to connect to a server at host:port
              in the order specified by the list of --remote options.

              proto indicates the protocol to use when connecting with the re‐
              mote, and may be "tcp" or "udp".

              For  forcing  IPv4 or IPv6 connection suffix tcp or udp with 4/6
              like udp4/udp6/tcp4/tcp6.

PR #4251, opened 2020-03-26
DHCP Domain trailing dot validation. Issue #8054

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8054
- [ ] Ready for review

If you add a trailing dot on the domain name in the DHCP static mapping "domain name" or System / General / Domain, this is accepted. The DNS Resolver retrieves this domain name from a lease and adds a dot: the configuration file ends up with 2 trailing dots and unbound fails to parse its config file if "DHCP Registration" or "Static DHCP" is enabled.

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -998,8 +998,11 @@ function is_hostname($hostname, $allow_wildcard=false) {
 }
 
 /* returns true if $domain is a valid domain name */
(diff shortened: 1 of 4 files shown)


vktg:
Done
`function is_domain($domain, $allow_wildcard=false, $trailing_dot=true);`

PR #4252, opened 2020-03-26
Clear DNS dhcpleases entries. Issue #8981

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/8981
- [ ] Ready for review

Uncheck the DHCP registration box in Unbound setting GUI should clear the contents of /var/unbound/dhcpleases_entries.conf to stop resolving these hostnames.

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -622,9 +622,16 @@ function system_dhcpleases_configure() {
 		}
 		@unlink($pidfile);
 		mwexec("/usr/local/sbin/dhcpleases -l {$g['dhcpd_chroot_path']}/var/db/dhcpd.leases -d {$config['system']['domain']} -p {$g['varrun_path']}/{$dns_pid} {$unbound_conf} -h {$g['etc_path']}/hosts");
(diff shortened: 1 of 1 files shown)


src/etc/inc/system.inc
```diff
@@ -622,9 +622,14 @@ function system_dhcpleases_configure() {
...
-	} elseif (isvalidpid($pidfile)) {
-		sigkillbypid($pidfile, "TERM");
-		@unlink($pidfile);
+	} else {
+		if (isvalidpid($pidfile)) {
+			sigkillbypid($pidfile, "TERM");
+			@unlink($pidfile);
+		}
+		if (file_exists("{$g['unbound_chroot_path']}/dhcpleases_entries.conf")) {
+			mwexec("truncate -s 0 {$g['unbound_chroot_path']}/dhcpleases_entries.conf");
```
rbgarga: It's better to use PHP functions when it's possible, in this case `ftruncate()` could be used

vktg: ok, done

PR #4253, opened 2020-03-28
DHCP Relay send/listen on same interface. Issue #10381

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10381
- [x] Ready for review

Required to OpenVPN clients to get DHCP address from internal DHCP server across the bridge.

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1883,8 +1883,11 @@ function services_dhcrelay_configure() {
 	}
 	$srvifaces = array_unique($srvifaces);
 
-	/* The server interface(s) should not be in this list */
-	$dhcrelayifs = array_diff($dhcrelayifs, $srvifaces);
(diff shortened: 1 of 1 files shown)


vktg:
https://redmine.pfsense.org/issues/10381?next_issue_id=10380&prev_issue_id=10382#note-11:
> However, after reading that DHCP is a special case on pfSense, it is allowed with hidden rules that are activated when DHCP is used on an interface. (https://forum.netgate.com/topic/90384/help-clarify-my-understanding-of-the-net-link-bridge-pfil-tunables-please/3) I added firewall rule on LAN interface, allowing UDP on port 67 and 68 to pass and now DHCP is working.
> MacOS OpenVPN clients are receiving DHCP offer from servers and get IP address without dhcrelay running. So I have found "a workaround" or finally have a proper configuration, whatever.

PR #4254, opened 2020-03-28
Bug #10386

kiokoman:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10386
- [ ] Ready for review

### DIFF
--- a/src/usr/local/www/system.php
+++ b/src/usr/local/www/system.php
@@ -250,8 +250,8 @@ if ($_POST) {
 	$_POST['timeservers'] = preg_replace('/[[:blank:]]+/', ' ', $_POST['timeservers']);
 	$_POST['timeservers'] = trim($_POST['timeservers']);
 	foreach (explode(' ', $_POST['timeservers']) as $ts) {
-		if (!is_domain($ts) && (!is_ipaddrv6($ts))) {
-			$input_errors[] = gettext("A NTP Time Server name may only contain the characters a-z, 0-9, '-', '.' and ':'.");
+		if (!is_domain($ts) && (!is_ipaddr($ts))) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system.php
```diff
@@ -250,8 +250,8 @@ function is_timezone($elt) {
 	$_POST['timeservers'] = preg_replace('/[[:blank:]]+/', ' ', $_POST['timeservers']);
 	$_POST['timeservers'] = trim($_POST['timeservers']);
 	foreach (explode(' ', $_POST['timeservers']) as $ts) {
-		if (!is_domain($ts)) {
-			$input_errors[] = gettext("A NTP Time Server name may only contain the characters a-z, 0-9, '-' and '.'.");
+		if (!is_domain($ts) && (!is_ipaddrv6($ts))) {
```
jim-p: Why not use `is_ipaddr()`, should we not also allow bare IPv4 addresses if we're allowing IPv6? -- `is_domain()` may happen to allow IPv4 addresses because of its nature, but expanding the latter test makes the behavior more clear in the code.

src/usr/local/www/system.php
```diff
@@ -250,8 +250,8 @@ function is_timezone($elt) {
...
+			$input_errors[] = gettext("A NTP Time Server name may only contain the characters a-z, 0-9, '-', '.' and ':'.");
```
jim-p: We should probably reword this entirely since it's a bit misleading. It should say "NTP Time Server names must be valid domain names, IPv4 addresses, or IPv6 addresses".

src/usr/local/www/system.php
kiokoman: because it does not work with 0.it.pool.ntp.org or 1.it.pool.ntp.org, that field is automatically generated with data from the ntp server and if someone put host name instead of ip it does not work, the error message is the same

PR #4255, opened 2020-03-31
Generate more complex IPsec PSK

Hobby-Student:
- [x] Ready for review

I'm always generating my own IPsec PSK seperately. This change adds more chars to the key. I never had any issues with these chars in place. Perhaps we could some more which are known to be safe across all sorts of devices?

(I think, redmine issue is not needed...?)


jim-p:
The existing method leveraging sha224 is sufficient. The additions add unnecessary complexity. You are always free to make your own, but if you are that concerned about security, stop using PSK.

Hobby-Student:
ok, understand what you mean.

> stop using PSK
already coding to use road warriors with multiple phase 1 and different authentication. currently testing and pull request will follow in few days

PR #4256, opened 2020-04-01
[IPsec] Multiple phase 1 for mobile clients

Hobby-Student:
- [ ] Ready for review

This is the first attempt to allow multiple phase 1 for mobile clients. I first made the changes to 2.4.5 and tested on my pfsense. Need to get another instance for testing with new code structure.
strongswan itself is handling these changes fine.

Before closing this early stage: Any suggestions for code sections to look at would be nice. I really think this could bring great benefit to pfSense.

Things I'm thinking of:
- **VPN->IPsec->Pre-Shared Keys** should be renamed, because users are now attached to the corresponding phase 1 (mobile).
- GUI needs some more changes to create multiple phase 1 for mobile (currently I just made a duplicate of the first one in the config.xml)

Will create redmine issue on the weekend.

### DIFF
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -80,10 +80,10 @@ $ipsec_log_cats = array(
 
 global $ipsec_identifier_list;
 $ipsec_identifier_list = array(
+	'none' => array('desc' => '', 'mobile' => true),
 	// 'ipv4' => array('desc' => gettext('IPv4 address'), 'mobile' => true),
 	// 'ipv6' => array('desc' => gettext('IPv6 address'), 'mobile' => true),
-	// 'rfc822' => array('desc' => gettext('RFC822'), 'mobile' => true),
-	'none' => array('desc' => '', 'mobile' => true),
+	'rfc822' => array('desc' => gettext('RFC822'), 'mobile' => true),
 	'email' => array('desc' => gettext('E-mail address'), 'mobile' => true),
 	'userfqdn' => array('desc' => gettext('User Fully Qualified Domain Name'), 'mobile' => true)
 	// 'fqdn' => array('desc' => gettext('Fully Qualified Domain Name'), 'mobile' => true),
@@ -204,7 +204,8 @@ $p1_authentication_methods = array(
 global $ipsec_preshared_key_type;
 $ipsec_preshared_key_type = array(
 	'PSK' => 'PSK',
-	'EAP' => 'EAP'
+	'EAP' => 'EAP',
+	'RSA' => 'RSA'
 );
 
 global $ipsec_closeactions;
@@ -1603,7 +1604,7 @@ function ipsec_setup_userpools() {
 		$upbase = "mobile-userpool-{$suffix}";
 		/* The full connection name including a reference to the primary
 		 * mobile connection */
-		$upconn = "con-{$upbase} : con-mobile-defaults";
+		$upconn = "con{$mkent['ikeid']}000-{$upbase} : con-mobile-defaults";
 		/* The full pool name including a reference to the primary
 		 * mobile pool */
 		$upname = "{$upbase} : mobile-pool";
@@ -1611,13 +1612,34 @@ function ipsec_setup_userpools() {
 		$scconf['connections'][$upconn] = array();
 		$scconf['pools'][$upname] = array();
 
-		$clientid = ($mkent['ident_type'] == "none") ? "\"{$mkent['ident']}\"" : "{$mkent['ident_type']}:{$mkent['ident']}";
+		switch ($mkent['ident_type']) {
+			case 'email':
+			case 'rfc822':
+			case 'userfqdn':
+				$clientid = "{$mkent['ident_type']}:{$mkent['ident']}";
+				break;
+			default:
+				$clientid = "\"{$mkent['ident']}\"";
+				break;
+		}
+
 		$clienteapid = ($ph1ent['authentication_method'] == "eap-mschapv2") ? $clientid : '%any';
 
 		/* Craft a cloned connection with the ID information to match */
 		$scconf['connections'][$upconn]['remote'] = array();
-		$scconf['connections'][$upconn]['remote']['id'] = $clientid;
-		$scconf['connections'][$upconn]['remote']['eap_id'] = $clienteapid;
+
+		switch ($mkent['type']) {
+			case 'EAP':
+				$scconf['connections'][$upconn]['remote']['id'] = $clientid;
+				$scconf['connections'][$upconn]['remote']['eap_id'] = $clienteapid;
+				break;
+			case 'RSA':
+				$scconf['connections'][$upconn]['remote']['id'] = '"/CN=' . $mkent['ident'] . '/"';
+				break;
+			default:
+				break;
+		}
+
 		$scconf['connections'][$upconn]['pools'] = $upbase;
 
 		/* Craft a cloned pool with pool settings to override for this user */
@@ -1997,12 +2019,13 @@ function ipsec_setup_tunnels() {
 		/* Determine the name of this connection, either con-mobile for
 		 * mobile clients, or a name based on the IKE ID otherwise. */
 		if (isset($ph1ent['mobile'])) {
-			$cname = "con-mobile";
+			$cdefaults = "con-mobile";
+			$cname = "con{$ph1ent['ikeid']}000-mobile";
 			/* Start with common default values */
-			$scconf["{$cname}-defaults"] = $conn_defaults;
+			$scconf["{$cdefaults}-defaults"] = $conn_defaults;
 			/* Array reference to make things easier */
-			$conn =& $scconf["{$cname}-defaults"];
-			$scconf['connections']["{$cname} : {$cname}-defaults"] = array("# Stub to load con-mobile-defaults");
+			$conn =& $scconf["{$cdefaults}-defaults"];
+			$scconf['connections']["{$cname} : {$cdefaults}-defaults"] = array("# Stub to load con-mobile-defaults");
 		} else {
 			$cname = "con{$ph1ent['ikeid']}000";
 			/* Start with common default values */
--- a/src/usr/local/www/vpn_ipsec_keys_edit.php
+++ b/src/usr/local/www/vpn_ipsec_keys_edit.php
@@ -36,16 +36,21 @@ require_once("functions.inc");
 require_once("guiconfig.inc");
 require_once("ipsec.inc");
 require_once("vpn.inc");
+require_once("certs.inc");
 
 init_config_arr(array('ipsec', 'mobilekey'));
 ipsec_mobilekey_sort();
 $a_secret = &$config['ipsec']['mobilekey'];
+
+init_config_arr(array('ipsec', 'phase1'));
+$a_phase1 = &$config['ipsec']['phase1'];
 
 if (is_numericint($_REQUEST['id'])) {
 	$id = $_REQUEST['id'];
 }
 
 if (isset($id) && $a_secret[$id]) {
+	$pconfig['phase1_ikeid'] = $a_secret[$id]['ikeid'];
 	$pconfig['ident'] = $a_secret[$id]['ident'];
 	$pconfig['type'] = $a_secret[$id]['type'];
 	$pconfig['psk'] = $a_secret[$id]['pre-shared-key'];
@@ -70,8 +75,10 @@ if ($_POST['save']) {
 
 	do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
 
-	if (preg_match("/[^a-zA-Z0-9@\.\-]/", $_POST['ident'])) {
-		$input_errors[] = gettext("The identifier contains invalid characters.");
+	if ($_POST['type'] != "RSA") {
+		if (preg_match("/[^a-zA-Z0-9@\.\-]/", $_POST['ident'])) {
+			$input_errors[] = gettext("The identifier contains invalid characters.");
+		}
 	}
 
 	if (array_key_exists($_POST['ident'], $userids)) {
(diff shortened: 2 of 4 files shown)


jim-p:
A few notes:
* This is going to be a lot messier than just the proposed changes. The entire Mobile Clients tab needs to reflect things as well because a user may or may not want those settings shared between Mobile P1s. Each mobile setup would need to support its own pools, auth settings, things like DNS/domains/etc.
* Entries for the PSK tab would still be global unless you made a new Mobile P1 for every user which would severely overcomplicate things.
* You use "RSA" but it shouldn't be RSA, it also supports other types of certificates (EC), so use a more generic name like "Certificate" or "Cert". I already changed all the other places that mentioned RSA previously for this reason.

We could allow more than one mobile P1 to share the same backend settings (pools, etc) but the utility of that seems limited and not really what users expect when they think of supporting multiple mobile IPsec setups.

Hobby-Student:
will rename RSA to something else. In this example I will stick to it.

> This is going to be a lot messier than just the proposed changes. The entire Mobile Clients tab needs to reflect things as well because a user may or may not want those settings shared between Mobile P1s. Each mobile setup would need to support its own pools, auth settings, things like DNS/domains/etc.

I get your point. I would see it a bit different. Example:
**Team A**
Alice -> alice@a.example.com
Bob -> bob@a.example.com
Charlie -> charlie@both.example.com (with PSK)

**Team B**
Charlie -> charlie@both.example.com (with RSA)
Donald -> donald@b.example.com

**IPsec mobile #1**
Phase 1 EAP-MSChapv2 with AES256, SHA512 and DH16
multiple Phase 2 for different subnets (can be same as IPsec mobile #2)

**IPsec mobile #2**
Phase 1 Mutual RSA with AES256, SHA512 and DH18
multiple Phase 2 for different subnets (can be same as IPsec mobile #1)

Team A is associated with IPsec mobile #1
Team B is associated with IPsec mobile #2
Charlie is associated to both

strongswan only connects if user has permission to connect defined via _rightid_ combined with the phase 1. So this _rightid_ has to be unique in each of the mobile phase 1. This way a user can be identified by phase 1 + _rightid_.

I think phase 1 is the first place to plan the IPsec connections and users are the second place to reflect things.

ipsec.secrets (pfSense 2.4.5)
```
 : RSA /var/etc/ipsec/ipsec.d/private/cert-1.key
 : RSA /var/etc/ipsec/ipsec.d/private/cert-3.key
alice@a.example.com : EAP 0s<base64>
bob@a.example.com : EAP 0s<base64>
charlie@both.example.com : EAP 0<base64>
```

ipsec.conf (incomplete, just user extending)(pfSense 2.4.5)
```
conn con1000-mobile-1
	also = con1000-mobile
	eap_identity = email:alice@a.example.com
	rightid = email:alice@a.example.com
	rightsourceip = 1.2.3.0/24

conn con1000-mobile-2
	also = con1000-mobile
	eap_identity = email:bob@a.example.com
	rightid = email:bob@a.example.com
	rightsourceip = 1.2.3.0/24
	rightdns = 1.2.0.1

conn con1000-mobile-3
	also = con1000-mobile
	eap_identity = email:charlie@both.example.com
	rightid = email:charlie@both.example.com
	rightsourceip = 1.2.3.0/24

conn con3000-mobile-1
	also = con3000-mobile
	rightid = "/CN=charlie@both.example.com/"
	rightsourceip = 1.2.4.0/24
```


> Entries for the PSK tab would still be global unless you made a new Mobile P1 for every user which would severely overcomplicate things.

global PSK should be no problem, as long as the _rightid_ is unique for each user associated to phase 1. This way a user could share its PSK over multiple devices connecting through different phase 1. On the other hand, each user device could have its own _rightid_ (other words: one user has multiple accounts).

> You use "RSA" but it shouldn't be RSA, it also supports other types of certificates (EC), so use a more generic name like "Certificate" or "Cert". I already changed all the other places that mentioned RSA previously for this reason.

will do so.

> We could allow more than one mobile P1 to share the same backend settings (pools, etc) but the utility of that seems limited and not really what users expect when they think of supporting multiple mobile IPsec setups.

Perhaps I'm getting it wrong, but I would stick to the "each phase 1 has its own (multiple) phase 2". Sharing between the phases could be done, but I don't think this is the intention or directly supported by strongswan.
Would lead to "each user has its phase 2"?

jim-p:
> I get your point. I would see it a bit different. Example:
> **Team A**
> Alice -> [alice@a.example.com](mailto:alice@a.example.com)
> Bob -> [bob@a.example.com](mailto:bob@a.example.com)
> Charlie -> [charlie@both.example.com](mailto:charlie@both.example.com) (with PSK)
> 
> **Team B**
> Charlie -> [charlie@both.example.com](mailto:charlie@both.example.com) (with RSA)
> Donald -> [donald@b.example.com](mailto:donald@b.example.com)

If you approach this like OpenVPN client-specific overrides, there is a selector on the user which marks the instances in which the user settings are defined. That may be cleaner than trying to make a per-VPN set of users and would keep the current PSK tab design. So long as the user doesn't define the same user+VPN selection twice it should be OK. The main problem I see is that strongSwan will select the auth based on the ID, so the exact same ID type couldn't be used on multiple VPNs with different auth (e.g. user1 on vpn1 with a cert and user1 on vpn2 with a different cert). You'd have to prevent that in input validation but otherwise it might work.

> ipsec.secrets (pfSense 2.4.5)
> ipsec.conf (incomplete, just user extending)(pfSense 2.4.5)

Using this format is a dead end. Look at 2.5.0 which has been fully converted to the swanctl.conf style.

> > We could allow more than one mobile P1 to share the same backend settings (pools, etc) but the utility of that seems limited and not really what users expect when they think of supporting multiple mobile IPsec setups.
> 
> Perhaps I'm getting it wrong, but I would stick to the "each phase 1 has its own (multiple) phase 2". Sharing between the phases could be done, but I don't think this is the intention or directly supported by strongswan.
> Would lead to "each user has its phase 2"?

"Phase 2" is not pools when it comes to mobile. In mobile tunnels a P2 entry only defines the local_ts, the pools handle the remote end. I'm talking about the dynamic addresses assigned to users when they connect and their associated settings defined on the Mobile Clients tab. Typically users are going to expect to be able to use a different address pool with each separate mobile P1. AFAIK it is OK for multiple P1s to share the same pool, so maybe multiple pools could be added later, but it seems best to design it correctly from the start rather than tack it on after.

So a similar tactic to the one I proposed for the PSK tab could be used here. For example, change the Mobile Clients tab to be a list of instances, and then allow each instance to be attached to one or more mobile P1s.

Hobby-Student:
thx for your input 👍 

> Using this format is a dead end. Look at 2.5.0 which has been fully converted to the swanctl.conf style.

was just for explaining the example. Will try to dig into 2.5.0 at the weekend and work on your suggestions next week.

Hobby-Student:
I now have a pfsense 2.5.0 dev machine (latest as of today). I configured IPsec mobile clients and this is my swanctl.conf:

```
connections {
        bypass {
                remote_addrs = 127.0.0.1
                children {
                        bypass {
                                local_ts = 1.2.3.0/24
                                remote_ts = 1.2.3.0/24
                                mode = pass
                                start_action = trap
                        }
                }
        }
}
pools {
        mobile-pool-v4 : mobile-pool {
                addrs = 1.2.4.0/24
        }
}
mobile-pool {
        28679 = "2"
}
secrets {
        private-0 {
                file = /var/etc/ipsec/private/cert-1.key
        }
        eap-1 {
                secret = 0sOTM4NjloZnc3Znc=
                id-0 = andere@mail.address
        }
        eap-2 {
                secret = 0sOTgzcno5OHdlZnVvcnU5
                id-0 = eine@mail.address
        }
}
```

I'm currently try to get an idea of migrating all settings from ipsec.conf to swanctl.conf...
As far as I can see, in the current development state the "connection mobile-pool" is missing? Or am I missing some settings in a different place?

jim-p:
> I'm currently try to get an idea of migrating all settings from ipsec.conf to swanctl.conf...
> As far as I can see, in the current development state the "connection mobile" is missing? Or am I missing some settings in a different place?

Did you add a mobile P1 and set it all up as usual?  Looks like you only setup the Mobile tab.

Hobby-Student:
I think, I configured it. Will get a closer look next week. For now it's just the master as source.
yes, have to change the password 😄 

config.xml
```
<ipsec>
	<client>
		<enable></enable>
		<radiusaccounting>disabled</radiusaccounting>
		<user_source>Local Database</user_source>
		<group_source>none</group_source>
		<pool_address>10.2.0.0</pool_address>
		<pool_netbits>24</pool_netbits>
		<pfs_group>2</pfs_group>
	</client>
	<phase1>
		<ikeid>1</ikeid>
		<iketype>ikev2</iketype>
		<interface>wan</interface>
		<mobile></mobile>
		<protocol>inet</protocol>
		<myid_type>fqdn</myid_type>
		<myid_data>test.local.domain</myid_data>
		<peerid_type>asn1dn</peerid_type>
		<peerid_data></peerid_data>
		<encryption>
			<item>
				<encryption-algorithm>
					<name>aes</name>
					<keylen>128</keylen>
				</encryption-algorithm>
				<hash-algorithm>sha256</hash-algorithm>
				<prf-algorithm>sha256</prf-algorithm>
				<dhgroup>5</dhgroup>
			</item>
		</encryption>
		<rekey_time></rekey_time>
		<reauth_time>28800</reauth_time>
		<over_time></over_time>
		<pre-shared-key></pre-shared-key>
		<private-key></private-key>
		<certref>5e88376fc473d</certref>
		<pkcs11certref></pkcs11certref>
		<pkcs11pin></pkcs11pin>
		<caref></caref>
		<authentication_method>eap-mschapv2</authentication_method>
		<descr><![CDATA[mobile-test-ph1-1]]></descr>
		<nat_traversal>on</nat_traversal>
		<mobike>off</mobike>
		<closeaction></closeaction>
	</phase1>
	<phase2>
		<ikeid>1</ikeid>
		<uniqid>5e883b03c893a</uniqid>
		<mode>tunnel</mode>
		<reqid>1</reqid>
		<localid>
			<type>network</type>
			<address>10.2.4.0</address>
			<netbits>24</netbits>
		</localid>
		<remoteid>
			<type></type>
		</remoteid>
		<protocol>esp</protocol>
		<encryption-algorithm-option>
			<name>aes</name>
			<keylen>128</keylen>
		</encryption-algorithm-option>
		<hash-algorithm-option>hmac_sha256</hash-algorithm-option>
		<pfsgroup></pfsgroup>
		<lifetime>3600</lifetime>
		<pinghost></pinghost>
		<descr><![CDATA[mobile-test-ph1-1-ph2-1]]></descr>
		<mobile></mobile>
	</phase2>
	<mobilekey>
		<ident>eine@mail.address</ident>
		<type>EAP</type>
		<pre-shared-key>983rz98wefuoru9</pre-shared-key>
		<ident_type>email</ident_type>
		<pool_address>10.2.1.0</pool_address>
		<pool_netbits>24</pool_netbits>
		<dns_address>10.2.1.253</dns_address>
	</mobilekey>
	<mobilekey>
		<ident>andere@mail.address</ident>
		<type>EAP</type>
		<pre-shared-key>93869hfw7fw</pre-shared-key>
		<ident_type>userfqdn</ident_type>
		<pool_address>10.2.1.0</pool_address>
		<pool_netbits>24</pool_netbits>
		<dns_address></dns_address>
	</mobilekey>
</ipsec>
```

image
image

Hobby-Student:
little update:
after powering on again, the swanctl.conf is now created with all *pool*s. now i am going to get multiple mobile phase 1...

Hobby-Student:
> If you approach this like OpenVPN client-specific overrides, there is a selector on the user which marks the instances in which the user settings are defined. That may be cleaner than trying to make a per-VPN set of users and would keep the current PSK tab design. So long as the user doesn't define the same user+VPN selection twice it should be OK.

Just for my understanding:
I added "Linked Phase 1" to the "Pre-Shared Keys" (which are enhanced and should be called something like "mobile Users"). This is generating the same effect as "Client specific overrides"? If someone connects to the IPsec Endpoint, strongswan tries to authenticate with these "Pre-Shared Keys"-Users. If there is no match, strongswan uses the "Mobile Clients" settings...

Phase 1 has specific user overrides? -> authenticate and use specific settings (Pre-Shared Keys)
Phase 1 has no specific user overrides? -> use the global settings (Mobile Clients)

I'm curently testing my approach. It works flawless on 2.4.5 and it seems that the port to swanctl.conf is also working on 2.5.0

jim-p:
Sounds like good progress, but the PSK tab isn't only for mobile PSKs so I'm hesitant to rename it entirely. There are use cases where people need to define PSK entries for certain styles of site-to-site setups, too.

Hobby-Student:
understood. what about:

I'm thinking of making the "Mobile Clients" tab more like the "Tunnels". This would divide mobile and site-to-site. I'm not really sure if this would make things easier or more complicated for users.

alternative:
adding a "+ Add mobile P1" to the "Tunnels".

technically my approach is working (in my environment). but before modifying the GUI I should know where to go.

jim-p:
> I'm thinking of making the "Mobile Clients" tab more like the "Tunnels". This would divide mobile and site-to-site. I'm not really sure if this would make things easier or more complicated for users.

I wouldn't duplicate anything between them. Keep the Mobile Clients tab sort of how it is but just support multiple instances of it. So kinda turn the current tab into a list of instances and then the current form into a page editing a single entry. Common settings can stay on the list tab.

> alternative:
> adding a "+ Add mobile P1" to the "Tunnels".

There is already a mobile flag on P1s it's just hidden. We'd just need to add a checkbox or drop-down selector to choose Site-to-Site/Mobile and adjust the visible options/behavior appropriately.

Hobby-Student:
> There is already a mobile flag on P1s it's just hidden. We'd just need to add a checkbox or drop-down selector to choose Site-to-Site/Mobile and adjust the visible options/behavior appropriately.

This sounds like the way to go. will modify GUI accordingly.

rbgarga:
@Hobby-Student do you still plan to work on this change?

Hobby-Student:
@rbgarga yes. had some weird months... will come back to this topic asap

rbgarga:
Changed it to draft until it's ready for a new review

rbgarga:
Closing due to inactivity since Sep 2020

PR #4257, opened 2020-04-02
OpenVPN client userpass is mandatory. Issue #10409

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10409
- [ ] Ready for review

If you create OpenVPN client connection with user authentication,
but don’t enter the password
System hangs on startup with prompt:
`
Syncing OpenVPN settings...Enter Auth Password:
`

### DIFF
--- a/src/usr/local/www/vpn_openvpn_client.php
+++ b/src/usr/local/www/vpn_openvpn_client.php
@@ -709,7 +709,11 @@ if ($act=="new" || $act=="edit"):
 		'Authentication Retry',
 		'Do not retry connection when authentication fails',
 		$pconfig['auth-retry-none']
-	))->setHelp('When enabled, the OpenVPN process will exit if it receives an authentication failure message. The default behavior is to retry.');
+	))->setHelp('When enabled, the OpenVPN process will exit if it receives an authentication failure message. ' .
+		    'The default behavior is to retry.%1$s%2$s%3$s', '<div class="infoblock">',
+		    sprint_info_box(gettext('WARNING: If the remote server requires both a username and a password, ' .
+		    'but only one is filled in, the system will hang on reboot prompting for OpenVPN Client credentials unless ' .
+		    'Authentication Retry is checked.'), 'info', false), '</div>');
 
 	$form->add($section);
 


jim-p:
I seem to recall there was a specific reason we allowed the password to be empty.  There was an auth use case which required it at some point, but I can't find notes about it anywhere now. I do see 7304c0234042868d91ab484d839a8c69087871be (https://redmine.pfsense.org/issues/3633) but that case is blank user but filled in password.

That same issue with being blocked waiting for auth can happen if the username is blank as well as the password, if the server requires user auth, so it's not something we can always prevent. We do add `auth-retry nointeract` which is _supposed_ to prevent this. But the user can override that by checking "Do not retry connection when authentication fails".

clhols:
This happened to me when I copied a OpenVPN client and only changed the host address. The resulting copy doesn't include the password. I didn't realise this and status said the the daemon couldn't be contacted, so I though a reboot would help. Then the router was stuck at startup like in the description here.
I had to buy a soldering iron to fix my null-modem cable, so it took me many hours to find the issue.
But I guess this PR will not fix the copy issue.

vktg:
I read https://redmine.pfsense.org/issues/3633 and, yeah, in some rare cases the password may be empty,
but I think we need to add a note about possible boot problems 

> This happened to me when I copied a OpenVPN client and only changed the host address. The resulting copy doesn't include the password, even though it indicates it during copy. I didn't realise this and status said the the daemon couldn't be contacted, so I though a reboot would help. Then the router was stuck at startup like in the description here.

This is other issue and I can can confirm that if you copy/close OpenVPN settings `auth_pass` is lost, even though it shows asterisks in the password fields
I'll create new redmine issue

jim-p:
The password issue is probably because of the **Confirm** box on the page, which really isn't necessary. That form only needs the password itself, no need to confirm it since it's not being set like a password, but for later (re)use.

rbgarga:

vktg:
Fixed and updated

jim-p (approved):

PR #4259, opened 2020-04-02
DigitalOcean IPv6 DDNS Client to find IPv6 entries when updating. Issue #10390

csobankesmarki:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10390
- [ ] Ready for review

The DigitalOcean IPv6 Dynamic DNS Client could not find the proper IPv6 entry when updating because it is looking for 'A' entries instead of 'AAAA' ones.
This is a single line fix for that.


csobankesmarki:
@jim-p Sure thing. Going to re-send the same pull request soon.

PR #4271, opened 2020-04-13
Aggressive NSEC option. Issue #10449

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10449
- [ ] Ready for review

Very nice feature for DNS optimization, which can reduce the number of queries to authoritative name servers.
See https://tools.ietf.org/html/rfc8198

unbound.conf(5):
```
aggressive-nsec: <yes or no>
              Aggressive  NSEC uses the DNSSEC NSEC chain to synthesize NXDOMAIN and other denials, using information from previous
              NXDOMAINs answers.  Default is no.  It helps to reduce the query rate towards targets that get a very  high  nonexis‐
              tent name lookup rate.
```

### DIFF
--- a/src/usr/local/www/services_unbound_advanced.php
+++ b/src/usr/local/www/services_unbound_advanced.php
@@ -78,6 +78,10 @@ if (isset($config['unbound']['dnsrecordcache'])) {
 	$pconfig['dnsrecordcache'] = true;
 }
 
+if (isset($config['unbound']['aggressivensec'])) {
+	$pconfig['aggressivensec'] = true;
+}
+
 $pconfig['msgcachesize'] = $config['unbound']['msgcachesize'];
 $pconfig['outgoing_num_tcp'] = isset($config['unbound']['outgoing_num_tcp']) ? $config['unbound']['outgoing_num_tcp'] : '10';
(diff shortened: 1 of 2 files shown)


src/usr/local/www/services_unbound_advanced.php
```diff
@@ -341,6 +350,15 @@
...
 ))->setHelp('When enabled, allows unbound to serve one query even with a TTL of 0, if TTL is 0 then new record will be requested in the background when the cache is served to ensure cache is updated without latency on service of the DNS request.');
 
+$section->addInput(new Form_Checkbox(
+	'aggressivensec',
+	'Aggressive NSEC',
+	'Aggressive Use of DNSSEC-Validated Cache',
+	$pconfig['aggressivensec']
+))->setHelp('When enabled, unbound uses the DNSSEC NSEC chain to synthesize NXDOMAIN and other denials, ' .
+	    'using information from previous NXDOMAINs answers. It helps to reduce the query rate towards ' .
+	    'targets that get a very  high  nonexis‐tent name lookup rate.');
```
jim-p, line 360: A couple issues here, looks like from copy/pasting this description. There are extra spaces around `high`, and a hyphen in the middle of `nonexis-tent` which needs to be removed.

vktg: oops! fixed

PR #4274, opened 2020-04-15
Adding DynDNS response logging

hydrian:
When verbose logging is enabled for a dynamic DNS provider, the DynDNS code will now write the HTTP response to the system log. This is extremely useful when troubleshooting issues.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/10459
- [x] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -1325,6 +1325,12 @@
 				$header_size = curl_getinfo($ch, CURLINFO_HEADER_SIZE);
 				$header = substr($response, 0, $header_size);
 				$data = substr($response, $header_size);
+				if($this->_dnsVerboseLog) {
+					foreach(explode(PHP_EOL,$header) as $hv) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/dyndns.class
```diff
@@ -1325,6 +1325,12 @@
 				$header_size = curl_getinfo($ch, CURLINFO_HEADER_SIZE);
 				$header = substr($response, 0, $header_size);
 				$data = substr($response, $header_size);
+				if($this->_dnsVerboseLog) {
```
jim-p, line 1328: Need a space between `if` and `(`

src/etc/inc/dyndns.class
```diff
@@ -1325,6 +1325,12 @@
...
+					foreach(explode(PHP_EOL,$header) as $hv) {
```
jim-p, line 1329: Need a space between `foreach` and `(`, and after the `,` between parameters.

PR #4277, opened 2020-04-16
NTPd GPS baud rate set fix. Issue #7284

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/7284
- [ ] Ready for review

Fix to correctly set GPS baud rate
see https://forum.netgate.com/topic/152369/ntp-gps-settings-serial-port-speed

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -1603,7 +1603,8 @@ function system_ntp_setup_gps($serialport) {
 		return false;
 	}
 
-	init_config_arr(array('ntpd', 'gps', 'speed'));
+	init_config_arr(array('ntpd', 'gps'));
+
 	$gps_device = '/dev/gps0';
 	$serialport = '/dev/'.$serialport;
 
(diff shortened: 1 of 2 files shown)


src/etc/inc/system.inc
```diff
@@ -1669,7 +1668,7 @@ function system_ntp_setup_gps($serialport) {
 	}
 
 	/* Remove old /etc/remote entry if it exists */
-	if (mwexec("/usr/bin/grep -c '^gps0' /etc/remote")) {
+	if (intval(`/usr/bin/grep -c '^gps0' /etc/remote`)) {
```
jim-p, line 1671: This should use some form of exec call, we try to avoid using backticks wherever possible.

src/etc/inc/system.inc
```diff
@@ -1603,7 +1603,6 @@ function system_ntp_setup_gps($serialport) {
 		return false;
 	}
 
-	init_config_arr(array('ntpd', 'gps', 'speed'));
```
jim-p, line 1606: Why remove this?

src/etc/inc/system.inc
vktg: ok, fixed

src/etc/inc/system.inc
vktg: This is incorrect:
there is no array of speeds in config.xml,
and if it's set as an array, this code is not working:
```php
if (!empty($config['ntpd']['gps']['speed']) && array_key_exists($config['ntpd']['gps']['speed'], $speeds)) {
                $gpsbaud = $speeds[$config['ntpd']['gps']['speed']];
}
```
that's why the speed is always set to 4800

jim-p: Then it should still initialize `ntpd` and `gps` before use.

vktg: ok, done
mwexec return code checking also fixed -
grep returns zero if it found something vs. if it hasn't found anything (non-zero)

PR #4278, opened 2020-04-16
DHCPv6 RA show default values in certain fields. Issue #10448

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10448
- [ ] Ready for review

In Services -> DHCPv6 Server & RA -> Router Advertisements there are some fields - Minimum RA interval, Maximum RA interval, Router lifetime - that have no values filled out by default.

It shows default values from `services.inc`
what about Router lifetime, from radvd.conf(5):
> AdvDefaultLifetime seconds
> The lifetime associated with the default router in units of seconds. The maximum value corresponds to 18.2 hours. A lifetime of 0 indicates that the router is not a default router and should not appear on the default router list. The router lifetime applies only to the router's usefulness as a default router; it does not apply to information contained in other message fields or options. Options that need time limits for their information include their own lifetime fields.
> 
> Must be either zero or between MaxRtrAdvInterval and 9000 seconds.
> 
> Default: 3 * MaxRtrAdvInterval (Minimum 1 second).

### DIFF
--- a/src/usr/local/www/services_router_advertisements.php
+++ b/src/usr/local/www/services_router_advertisements.php
@@ -178,7 +178,7 @@ if ($_POST['save']) {
 		if (!is_numericint($_POST['raminrtradvinterval'])) {
 			$input_errors[] = gettext("Minimum advertisement interval must be an integer.");
 		}
-		if ($_POST['raminrtradvinterval'] < "3") {
+		if ($_POST['raminrtradvinterval'] < 3) {
 			$input_errors[] = gettext("Minimum advertisement interval must be no less than 3.");
 		}
 		if ($_POST['ramaxrtradvinterval'] && $_POST['raminrtradvinterval'] > (0.75 * $_POST['ramaxrtradvinterval'])) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_router_advertisements.php
```diff
@@ -193,8 +193,8 @@
 			$input_errors[] = gettext("Maximum advertisement interval must be no less than 4 and no greater than 1800.");
 		}
 	}
-	if ($_POST['raadvdefaultlifetime'] && !is_numericint($_POST['raadvdefaultlifetime'])) {
-		$input_errors[] = gettext("Router lifetime must be an integer between 0 and 9000.");
+	if ($_POST['raadvdefaultlifetime'] && (($_POST['raadvdefaultlifetime'] < "1") || ($_POST['raadvdefaultlifetime'] > "9000"))) {
```
jim-p, line 196: Why are you putting integer values in quotes? If you are comparing numerically and want to force both to the same type, you should cast the strings to int rather than doing this.

jim-p: Also if the max is 3 * `ramaxrtradvinterval` why are you comparing against 9000 and not calculating the max based on `ramaxrtradvinterval`?

vktg: quotes fixed.

3 * `ramaxrtradvinterval` is default, not maximum:
> Must be either zero or between MaxRtrAdvInterval and 9000 seconds.
> 
> Default: 3 * MaxRtrAdvInterval (Minimum 1 second).

PR #4281, opened 2020-04-17
Allow to download log files from WebGUI. Issue #10468

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10468
- [ ] Ready for review

Ability to download separate log files from the Status / System Logs pages.

Useful for troubleshooting.


jim-p:
I don't think this is a good idea as-is. There are way more logs than it offers (all the compressed rotated files) and it would be practically useless just after rotation. Users can already download files a variety of ways, like from diag_command.php, SCP/SFTP, etc.

PR #4282, opened 2020-04-17
Host Resources module is mandatory for PF SNMP module. Issue #10471

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10471
- [ ] Ready for review

BEGEMOT-PF-MIB objects is missing if you don't select "Host Resources" module
selecting only "PF" modules is not enough

"Host Resources" module must be set as mandatory for "PF" modules


vktg:
This PR is not needed
see https://redmine.pfsense.org/issues/10471#note-3

PR #4283, opened 2020-04-18
Port forward dst port Any fix. Issue #7704

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/7704
- [ ] Ready for review

This PR sets the port range from 1 to 65535, the redirect port to 1 and hide 'Redirect target port' input if "Any" is selected


vktg:
or maybe we need to remove "Any" for destination port range?
very rare case

PR #4285, opened 2020-04-18
Add the default values of the TCP and UDP Timeouts on the WebUI. Issue #7362

vktg:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/7362
- [ ] Ready for review

when you go to System --> Advanced --> Firewall & NAT then you have the possibility to modify the different timeouts but you do not know which timeouts are set by default. So it would be usefull to see the default values within the text fields and they should change when you change the "Firewall Optimization Options" - but should not override user specific modifications.

At least it could be helpfull to have a button like "show timeouts" which shows the output of "pfctl -st" or something else which is helpfull for the user/administrator.

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -3574,4 +3574,31 @@ function is_url_hostname_resolvable($url) {
 	return $resolvable;
 }
 
+function get_pf_timeouts () {
+	$pftimeout = array();
+	exec("/sbin/pfctl -st", $pfctlst, $retval);
+	if ($retval == 0) {
+		foreach ($pfctlst as $i => $pfst) {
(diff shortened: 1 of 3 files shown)


src/etc/inc/pfsense-utils.inc
```diff
@@ -3574,4 +3574,19 @@ function is_url_hostname_resolvable($url) {
 	return $resolvable;
 }
 
+function get_pf_timeouts () {
```
jim-p, line 3577: Since you are providing a function for this, it makes more sense for its output to be structured as `$pftimeout[<protocol>][<type>] = <value>` so whatever ends up calling this doesn't have to remember what is in index 1/2/etc.

vktg: Updated -
now it returns array `$pftimeout[$proto][$type][name/keyname/value]`

PR #4295, opened 2020-04-29
fix #10499 dark theme autocomplete popup not readable

Marc05:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10499
- [ ] Ready for review


jim-p:
Redmine link: https://redmine.pfsense.org/issues/10499

PR #4296, opened 2020-04-30
Remove CA prv key fix. Issue #10509

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10509
- [X] Ready for review

https://forum.netgate.com/topic/153020/removing-a-ca-key

If you edit CA and save, the key is still there.

### DIFF
--- a/src/usr/local/www/system_camanager.php
+++ b/src/usr/local/www/system_camanager.php
@@ -247,10 +247,8 @@ if ($_POST['save']) {
 			$ca['descr']  = $pconfig['descr'];
 			$ca['refid']  = $pconfig['refid'];
 			$ca['serial'] = $pconfig['serial'];
-			$ca['crt']	  = base64_encode($pconfig['cert']);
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_camanager.php
```diff
@@ -115,9 +115,7 @@
 		$pconfig['serial'] = $thisca['serial'];
 		$pconfig['trust']  = ($thisca['trust'] == 'enabled');
 		$pconfig['randomserial']  = ($thisca['randomserial'] == 'enabled');
-		if (!empty($thisca['prv'])) {
-			$pconfig['key'] = base64_decode($thisca['prv']);
-		}
+		$pconfig['key'] = base64_decode($thisca['prv']);
```
jim-p, line 118: Why change this test? The key could be empty when it's being read from the config, and if so, that `base64_decode()` will fail.

vktg: Right, fixed

PR #4299, opened 2020-05-02
VPN MSS clamping improvements. Issue #10493

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10493
- [X] Ready for review

This fix allows you to select for which VPN types / IP proto do MSS clamping

with this fix <negate_networks> populated with all types of VPN, can this cause any issues?


vktg:
> We could add 4/6 MSS options to each individual IPsec P2, OpenVPN client/server, etc, to go along with or replace global values.

The issue is that it's not possible to 'bind' scrub to the particular interface, i.e.
```
scrub in ipsec1000 from any to <vpn_networks> max-mss {$maxmss}
scrub out ipsec1000 from <vpn_networks>  to any max-mss {$maxmss}
```
not possible, which leads to mss clamping of unwanted traffic, see https://redmine.pfsense.org/issues/7815

This is why current PR allows you  to see the "true" mss clamping

> but each of these protocols has different amounts of overhead so each one could use a separate v4/v6 MSS value.

jim-p:
You can limit it by subnet to/from, not interface. Each P2, OpenVPN instance, etc, that I mentioned is  a unique subnet. We're limiting them collectively now but there isn't any reason (other than GUI complexity) we couldn't limit them per instance/subnet.

rbgarga:
@vktg please rebase your fork to fix conflicts

rbgarga:
Closing due to inactivity for long time

PR #4303, opened 2020-05-06
L2TP client Shared Secret option. Issue #10531

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10531
- [x] Ready for review

It is not possible to use Shared Secret by L2TP client,
no such field in WebGUI
Some ISP use it.

This PR adds it.

successfully tested with L2TP pfSense server with https://github.com/pfsense/pfsense/pull/4302 patch

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -2576,6 +2576,14 @@ EOD;
 EOD;
 		}
 
+		if (($type == "l2tp") && !empty($ppp['secret'])) {
+			$secret = str_replace('"', '\"', base64_decode($ppp['secret']));
+			$mpdconf .= <<<EOD
+	set l2tp secret "{$secret}"
+
+EOD;
+		}
+
 		if ($type == "pptp" || $type == "l2tp") {
 			$mpdconf .= <<<EOD
 	set {$type} self {$localips[$pid]}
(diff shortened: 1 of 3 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -2573,6 +2573,14 @@ EOD;
 			$mpdconf .= <<<EOD
 	set pppoe iface {$port}
 
+EOD;
+		}
+
+		if (($type == "l2tp") && !empty($ppp['secret'])) {
+			$secret = base64_decode($ppp['secret']);
+			$mpdconf .= <<<EOD
+	set l2tp secret {$secret}
```
jim-p, line 2582: That should probably be enclosed in double quotes and have quotes escaped inside, to be safe.

vktg: Done

src/usr/local/www/interfaces.php
```diff
@@ -2947,6 +2956,19 @@ function build_port_list() {
...
 
+$group = new Form_Group('Shared Secret');
+
+$group->add(new Form_Input(
+	'l2tp_secret',
+	'*Secret',
+	'password',
+	$pconfig['l2tp_secret']
+))->setHelp('L2TP tunnel Shared Secret. Used to authenticate tunnel connection and encrypt ' .
+	    'important control packets avpairs. (Optional)');
```
jim-p, line 2967: change "control packets avpairs" to "control packet contents"

src/usr/local/www/interfaces_ppps_edit.php
```diff
@@ -576,6 +587,21 @@ function build_country_list() {
...
+if ($pconfig['type'] == 'l2tp') {
+	$group = new Form_Group('Shared Secret');
+
+	$group->add(new Form_Input(
+		'secret',
+		'*Secret',
+		'password',
+		$pconfig['secret']
+	))->setHelp('L2TP tunnel Shared Secret. Used to authenticate tunnel connection and encrypt ' .
+       		    'important control packets avpairs. (Optional)');
```
jim-p, line 599: change "control packets avpairs" to "control packet contents"

PR #4304, opened 2020-05-06
Alias clone feature. Issue #6908

vktg:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/6908
- [x] Ready for review

Alias copy/clone icon


vktg:
fixed.

it rewritten aliases ID before
now works fine

PR #4308, opened 2020-05-08
Feature #10392: GRE: Tunnels cannot have IPv6 and IPv4 addresses …

xrm:
…at the same time. This PR tries to change that.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/10392
- [x] Ready for review

This commit (only) patches the GRE part. GIF interfaces have to be patched separately; and while that patch should mostly consist of directly applying this commit to the gif-file/functions as well, I have no way of testing it.

Patch was tested on 2.4.4-RELEASE-p3. Since I did not come across any issues when merging this back onto master, I expect this to work on the latest master as well.

First commit to the pfSense project ever, so I'm neither sure if the "fallback mode for older configurations" is the right way to do it not if the code fulfils the quality standard in all points.


xrm:
Hi jim-p,

I removed the selection box and added a possible upgrade code to upgrade_config.php. I tested the new code (except for the upgrade_config.php-part since I don't know how) on 2.4.5.

There is one small issue, though: if using IPv4 and IPv6 at the same time, the newly added IPv6 gateway shows the buttons to delete and to edit it - something the IPv4 gateway does not. It doesn not seem to me as if this is related to my changes but I wanted to mention it.

Thanks
Sebastian

src/etc/inc/interfaces.inc
```diff
@@ -1128,34 +1128,49 @@ function interface_gre_configure(&$gre, $grekey = "") {
 		$greif = pfSense_interface_create("gre");
 	}
 
+	$tunnel_type = '';
+	if (($gre['tunnel-local-addr'] !== '') || ($gre['tunnel-remote-addr'] !== '')) {
+		$tunnel_type = 'v4';
+	}
+	if (($gre['tunnel-local-addr6'] !== '') || ($gre['tunnel-remote-addr6'] !== '')) {
```
jim-p, line 1135: These should use `!empty()` and `&&` rather than `!== ''` and ``||``. and this will not give you the result you want if the values are unset. For example `null !== ''` is `true`. If either one is null, the test would be true and it would set the wrong type.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -55,38 +59,80 @@
...
+	$pconfig['tunnel-remote-addr6'] = addrtolower($_POST['tunnel-remote-addr6']);
 	$pconfig['remote-addr'] = addrtolower($_POST['remote-addr']);
 
+	$tunnel_type = '';
+	if (($pconfig['tunnel-local-addr'] !== '') || ($pconfig['tunnel-remote-addr'] !== '')) {
+		$tunnel_type = 'v4';
+	}
+	if (($pconfig['tunnel-local-addr6'] !== '') || ($pconfig['tunnel-remote-addr6'] !== '')) {
+		$tunnel_type .= 'v6';
+	}
```
jim-p, line 72: See previous comments about this kind of test.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -55,38 +59,80 @@
...
+	if (((!is_ipaddr($_POST['tunnel-local-addr'])) ||
+	     (!is_ipaddr($_POST['tunnel-remote-addr']))) &&
+	    in_array($tunnel_type, array('v4', 'v4v6'))) {
+		$input_errors[] = gettext("The tunnel local and tunnel remote fields must have valid IPv4 addresses.");
+	}
+
+	if (((!is_ipaddrv6($_POST['tunnel-local-addr6'])) ||
+	     (!is_ipaddrv6($_POST['tunnel-remote-addr6']))) &&
+	     in_array($tunnel_type, array('v6', 'v4v6'))) {
+		$input_errors[] = gettext("The tunnel local and tunnel remote fields must have valid IPv6 addresses.");
```
jim-p, line 101: Since these are IPv6-specific the text should mention the exact labels of the IPv6 versions of these fields.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -55,38 +59,80 @@
...
+		$input_errors[] = gettext("The v6 GRE tunnel subnet must be an integer.");
+	}
+
+	if (in_array($tunnel_type, array('v4', 'v4v6'))) {
+		if (!is_ipaddrv4($_POST['tunnel-local-addr'])) {
+			$input_errors[] = gettext("The GRE Tunnel local address must be IPv4 if the tunnel type is IPv4.");
+		}
 		if (!is_ipaddrv4($_POST['tunnel-remote-addr'])) {
-			$input_errors[] = gettext("The GRE Tunnel remote address must be IPv4 where tunnel local address is IPv4.");
+			$input_errors[] = gettext("The GRE Tunnel remote address must be IPv4 if the tunnel type is IPv4.");
```
jim-p, line 121: No need to mention the tunnel type here, the fields are separate and have distinct names, only mention specifically which field is wrong.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -55,38 +59,80 @@
...
 
-	if (is_ipaddrv6($_POST['tunnel-local-addr'])) {
-		if (!is_ipaddrv6($_POST['tunnel-remote-addr'])) {
-			$input_errors[] = gettext("The GRE Tunnel remote address must be IPv6 where tunnel local address is IPv6.");
+	if (in_array($tunnel_type, array('v6', 'v4v6'))) {
+		if (!is_ipaddrv6($_POST['tunnel-local-addr6'])) {
+			$input_errors[] = gettext("The GRE Tunnel local address must be IPv6 if the tunnel type is IPv6.");
+		}
+		if (!is_ipaddrv6($_POST['tunnel-remote-addr6'])) {
+			$input_errors[] = gettext("The GRE Tunnel remote address must be IPv6 if the tunnel type is IPv6.");
```
jim-p, line 133: See previous note, no need to mention tunnel type, just the specific field names.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -55,38 +59,80 @@
...
 		}
-		if ($_POST['tunnel-remote-net'] > 128 || $_POST['tunnel-remote-net'] < 1) {
+		if ($_POST['tunnel-remote-net6'] > 128 || $_POST['tunnel-remote-net6'] < 1) {
 			$input_errors[] = gettext("The GRE tunnel subnet must be an integer between 1 and 128.");
```
jim-p, line 136: This should reflect the IPv6 version of the field name as well

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -175,32 +227,66 @@ function build_parent_list() {
...
+	$pconfig['tunnel-remote-net6'],
+	array_combine(range(128, 1, -1), range(128, 1, -1))
+))->setHelp('The subnet is used for determining the v6 network that is tunnelled.');
+
+$group->add(new Form_Checkbox(
+	'link3',
+	'Add Static Route',
+	'Add an explicit static route for the remote inner tunnel address/subnet via the local tunnel address',
+	$pconfig['link3']
+));
```
jim-p, line 286: link0/link1/link2 have special meanings to `ifconfig` and GRE. We can't just make up a new value in this case.

The `link1` property is for the interface as a whole -- see https://github.com/pfsense/pfsense/blob/master/src/etc/inc/interfaces.inc#L1311 -- so you can just take the value of `link1` rather than making a new option.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -175,32 +227,66 @@ function build_parent_list() {
...
 	$pconfig['link1']
 ));
 
+$section->add($group);
+
+$group = new Form_Group(gettext('IPv6'));
+
+$group->add(new Form_IpAddress(
+	'tunnel-local-addr6',
+	'*GRE tunnel local v6 address',
```
jim-p, line 264: Just making one comment but it applies to all the fields on this page. Since we're making changes, they should be in "Title Case" and the "GRE" in all of them is redundant (the whole page is GRE...)

So: "*Tunnel Local IPv6 Address". And the IPv4 versions should be changed to match, and explicitly labeled with IPv4, for example: "*Tunnel Local IPv4 Address"

Alternately, consider leading with the address family: "*IPv6 Tunnel Local Address" / "*IPv4 Tunnel Local Address"

jim-p:
One more thing: You need to increase the value of `latest_config` in `/etc/inc/globals.inc` to match the new upgrade code version.

To test that, you can start the PHP shell (`pfSsh.php`) and run `convert_config();` followed by `exec` after updating `globals.inc`.

src/usr/local/www/interfaces_gre_edit.php
xrm: Ah, okay. This then leads me to the question why the link1 property is set when the user "only" requests a static route to be added? Should I use two distinct fields for each static route and set link1 when either of them is requested or should I really just put one option there for the static route and add a v4 as well as a v6 route when it's set?

xrm: I am now only using link1 and set a static route for any active network.

src/etc/inc/interfaces.inc
```diff
@@ -1128,34 +1128,47 @@ function interface_gre_configure(&$gre, $grekey = "") {
...
 	}
 
-	if (isset($gre['link1']) && $gre['link1']) {
-		mwexec("/sbin/route add " . escapeshellarg($gre['tunnel-remote-addr']) . "/" . escapeshellarg($gre['tunnel-remote-net']) . " " . escapeshellarg($gre['tunnel-local-addr']));
+	if (isset($gre['link1'])) {
+		if (in_array($tunnel_type, array('v4', 'v4v6'))) {
+			mwexec("/sbin/route add " . escapeshellarg($gre['tunnel-remote-addr']) . "/" . escapeshellarg($gre['tunnel-remote-net']) . " " . escapeshellarg($gre['tunnel-local-addr']));
+		}
+		if (in_array($tunnel_type, array('v6', 'v4v6'))) {
+			mwexec("/sbin/route -6 add " . escapeshellarg($gre['tunnel-remote-addr6']) . "/" . escapeshellarg($gre['tunnel-remote-net6']) . " " . escapeshellarg($gre['tunnel-local-addr6']));
```
xrm, line 1164: This is a difference from the original version: in my tests, a v6-route could not be added without the -6 parameter. Not sure if this was simply a bug in the original version or if there's something going on that I missed.

src/usr/local/www/interfaces_gre_edit.php
```diff
@@ -97,7 +144,11 @@
 		}
 
 		if (($gre['if'] == $_POST['if']) && ($gre['tunnel-remote-addr'] == $_POST['tunnel-remote-addr'])) {
-			$input_errors[] = sprintf(gettext("A GRE tunnel with the network %s is already defined."), $gre['remote-network']);
+			$input_errors[] = sprintf(gettext("A GRE tunnel with the same IPv4 tunnel network is already defined."));
+			break;
+		}
+		if (($gre['if'] == $_POST['if']) && ($gre['tunnel-remote-addr6'] == $_POST['tunnel-remote-addr6'])) {
+			$input_errors[] = sprintf(gettext("A GRE tunnel with the same IPv6 tunnel network is already defined."));
```
xrm, line 151: The original code outputs `gre['remote-network']` which is not defined. Since the user put in the network themselves anyway (and should therefore know it without the error message telling them), I decided to just output which of the two possibly entered networks (v4/v6) already exists.

xrm:
Thanks for the detailled feedback. I revised the complete wording on the GRE page and it should (hopefully) be more consistent now. I also used the opportunity to improve a few minor things such as limiting the IP-input fields to v4 and v6, respectively.

The link1-variable still confuses me a bit but I changed it as proposed.

src/usr/local/www/interfaces_gre_edit.php
jim-p: > This then leads me to the question why the link1 property is set when the user "only" requests a static route to be added?

Because the `link1` flag has a history in which it altered the internal route search method used for locating the remote endpoint. So we add that route manually to account for how GRE behaved in the past. That behavior has changed over time (See b5baa9e7a17266e794514747e7046eba6acb1db7 for example) but we kept the user-facing option the same to preserve what users expect over time.

Using just the one option is fine as the behavior is defined for the interface as a whole and shouldn't be a per-family option.

src/etc/inc/interfaces.inc
jim-p: Might have been a bug, either `-6` or `-inet6` is necessary to tell the `route` command to expect IPv6 addresses.

src/usr/local/www/interfaces_gre_edit.php
xrm: Ah! I see, thank you for explaining it. :-)

vktg:
works as expected, converting old-style GRE interfaces to new is OK

```
# ifconfig gre0
gre0: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> metric 0 mtu 1476
	description: GREd
	options=80000<LINKSTATE>
	tunnel inet 192.168.3.4 --> 192.168.3.6
	inet 192.168.46.4 --> 192.168.46.6 netmask 0xffffff00
	inet6 fe80::edd:aeff:fe8a:1d00%gre0 prefixlen 64 scopeid 0xd
	inet6 fc00:46::4 --> fc00:46::6 prefixlen 128
	groups: gre
	nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
# ping6 fc00:46::6
PING6(56=40+8+8 bytes) fc00:46::4 --> fc00:46::6
16 bytes from fc00:46::6, icmp_seq=43 hlim=64 time=1.021 ms
16 bytes from fc00:46::6, icmp_seq=44 hlim=64 time=0.944 ms
16 bytes from fc00:46::6, icmp_seq=45 hlim=64 time=7.456 ms
16 bytes from fc00:46::6, icmp_seq=46 hlim=64 time=1.055 ms
16 bytes from fc00:46::6, icmp_seq=47 hlim=64 time=1.038 ms
^C
--- fc00:46::6 ping6 statistics ---
48 packets transmitted, 5 packets received, 89.6% packet loss
round-trip min/avg/max/std-dev = 0.944/2.303/7.456/2.577 ms
# ping 192.168.3.6
PING 192.168.3.6 (192.168.3.6): 56 data bytes
64 bytes from 192.168.3.6: icmp_seq=0 ttl=64 time=0.893 ms
64 bytes from 192.168.3.6: icmp_seq=1 ttl=64 time=1.005 ms
```

It's better to set IPv6 subnet field inactive and add note, 
otherwise 'always 128' subnet can confuse users

something like:
```php
))->setHelp('The subnet is used for determining the IPv6 network needs to be 128 as enforced by kernel.')->setAttribute('disabled', true);
```
tested on two 2.5.0.a.20200512.2320

xrm:
The subnet is used for the route, though.

PR #4309, opened 2020-05-08
DNS/Ping/Traceroute IDN support. Issue #10538

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10538
- [X] Ready for review

Add support for IDN hostnames on the DNS/Ping/Traceroute diagnostics pages.


vktg:
in the same way it's possible to add IDN support to DNS Resolver/IPsec/OpenVPN etc.
but I think it's better to add `idn_to_utf8()` to Form_Input class

PR #4315, opened 2020-05-11
RADIUS authentication via shell/ssh. Implement #10545

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10545
- [X] Ready for review

Works fine, but It seems that we need "Fallback to Local Database auth" checkbox for using pam_unix.so in case if RADIUS server unavailable.

This is initial version of PR, after fixing/approving LDAP authentication can be done the same way:
https://redmine.pfsense.org/issues/8698


src/etc/inc/auth.inc
```diff
@@ -2171,4 +2171,102 @@ function get_user_remote_authsource() {
...
 }
+function set_pam_auth() {
+	global $config;
+
+	$pam_sshd =<<<EOD
+#
+# \$FreeBSD$
+#
+# PAM configuration for the "sshd" service
+#
```
jim-p, line 2182: Do we need to include all of the comments and disabled/commented-out lines? This should probably instead have a header saying it's automatically generated and that it shouldn't be edited. Similar to other config files we generate. (Same goes for the other files in this PR)

vktg: > his should probably instead have a header saying it's automatically generated and that it shouldn't be edited.
Added

> If that is the case, the GUI needs to warn about that.
Added

> There is a `template_user` option to `pam_radius` which may help but that seems like a different problem source. If that path is taken then we'd need to find/detect/something the automatically generated `passwd` entries and show them in the GUI.
Better to use pam LDAP for that

> Also need to make sure this works with `sudo`.
`sudo` pam already includes system template,-
/usr/local/etc/pam.d/sudo:
```
#
# $FreeBSD$
#
# PAM configuration for the "sudo" service
#

# auth
auth		include		system

# account
account		include		system

# session
# XXX: pam_lastlog (used in system) causes users to appear as though
# they are no longer logged in in system logs.
session		required	pam_permit.so

# password
password	include		system
```

vktg: + it seems safe to remove
```
auth		sufficient	pam_opie.so		no_warn no_fake_prompts
auth		requisite	pam_opieaccess.so	no_warn allow_local
```

src/usr/local/www/system_usermanager_settings.php
```diff
@@ -221,7 +226,7 @@
 	'*Authentication Server',
 	$pconfig['authmode'],
 	$auth_servers
-));
+))->setHelp('To allow login with RADIUS credentials, equivalent local users with the expected privileges must be created first.');
```
jim-p, line 229: This note should specifically call out SSH. That isn't needed for GUI logins, just SSH.

src/etc/inc/auth.inc
```diff
@@ -2171,4 +2171,96 @@ function get_user_remote_authsource() {
...
+# session
+session		required	pam_lastlog.so		no_fail
+
+# password
+password        sufficient      pam_radius.so
+password	required	pam_unix.so		no_warn try_first_pass
+
+EOD;
+
+		@file_put_contents("/etc/radius.conf", $radius_conf);
```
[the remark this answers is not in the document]
vktg: done

src/usr/local/www/system_usermanager_settings.php
vktg: SSH and console (pam system)

Danilo-Z:
I've added the patch on: 

```
2.5.0-DEVELOPMENT (amd64)
built on Wed Sep 16 19:03:19 EDT 2020
FreeBSD 12.2-PRERELEASE
```

I was able to ssh to pfSense using FreeRadius as the Authentication Server.
```
joss@bourguignon:~$ ssh dan@192.168.33.20
RADIUS Password:
[2.5.0-DEVELOPMENT][dan@pftest2.ipbgd.office]/home/dan: 
```

Here are the logs:
`Sep 17 09:08:34 	radiusd 	46695 	(0) Login OK: [dan] (from client pfsense port 0 cli 172.21.10.113)
Sep 17 09:08:34 	sshd 	47529 	in openpam_dispatch(): /usr/lib/pam_radius.so.6: no pam_sm_acct_mgmt()
Sep 17 09:08:34 	sshd 	47529 	in openpam_check_error_code(): pam_sm_acct_mgmt(): unexpected return value 2
Sep 17 09:08:34 	sshd 	47142 	Accepted keyboard-interactive/pam for dan from 172.21.10.113 port 61334 ssh2`

PR #4316, opened 2020-05-12
Alias IDN hostnames support. Issue #7255

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/7255
- [X] Ready for review

It saves IDN hostsnames in punnycode


vktg:
import/export IDN support added

PR #4318, opened 2020-05-13
is_proccess_running empty proc fix. Issue #10540

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10540
- [X] Ready for review

is_process_running generates an error when $process is "".

original patch by [Orion Poplawski](https://redmine.pfsense.org/users/4450)

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -45,9 +45,11 @@ function isvalidpid($pidfile) {
 
 function is_process_running($process) {
 	$output = "";
-	exec("/bin/pgrep -anx " . escapeshellarg($process), $output, $retval);
-
-	return (intval($retval) == 0);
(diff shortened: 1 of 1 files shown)


src/etc/inc/util.inc
```diff
@@ -45,9 +45,11 @@ function isvalidpid($pidfile) {
 
 function is_process_running($process) {
 	$output = "";
-	exec("/bin/pgrep -anx " . escapeshellarg($process), $output, $retval);
-
-	return (intval($retval) == 0);
+	if (empty($process)) {
```
jim-p, line 48: Should be `!empty()`, not `empty()`.

vktg: ohh, fixed
too many problems with such a small fix)

PR #4326, opened 2020-05-22
Allow to submit voucher via URL. Implements #1984

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/1984
- [X] Ready for review

In addition to the web form, allow the submission of voucher via URL e.g.

`http://pfsenseip:8002/index.php?zone=zone_name&redirurl=redir_url&voucher=voucher_code`

Considering the numerous forum posts about shortening pfsense vouchers and the proliferation of handheld devices with bar-code scanning apps, it'd be cool to be able to distribute vouchers as a URL stored in QR code.

### DIFF
--- a/src/usr/local/captiveportal/index.php
+++ b/src/usr/local/captiveportal/index.php
@@ -82,7 +82,12 @@ if ((!empty($cpsession)) && (! $_POST['logout_id']) && (!empty($cpcfg['page']['l
 } elseif (!empty($cpsession) && (!isset($_POST['logout_id']) || !isset($config['captiveportal'][$cpzone]['logoutwin_enable']))) {
 	/* If client try to access captive portal page while already connected, 
 		but no custom logout page does exist and logout popup is disabled */	
-	echo gettext("You are connected.");
+	echo gettext("You are connected.<br/>");
+	if ($_GET['redirurl'] || $_POST['redirurl']) {
+		$redirurl = $_GET['redirurl'] ? $_GET['redirurl'] : $_POST['redirurl'];
+		$redirurl = htmlspecialchars($redirurl);
+		echo ("You can proceed to: <a href='{$redirurl}'>{$redirurl}</a>");
+	} 
 	ob_flush();
 	return;
 } elseif ($orig_host != $ourhostname) {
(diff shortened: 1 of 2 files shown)


src/usr/local/captiveportal/index.php
```diff
@@ -82,7 +82,13 @@
 } elseif (!empty($cpsession) && (!isset($_POST['logout_id']) || !isset($config['captiveportal'][$cpzone]['logoutwin_enable']))) {
 	/* If client try to access captive portal page while already connected, 
 		but no custom logout page does exist and logout popup is disabled */	
-	echo gettext("You are connected.");
+	if ($_REQUEST['voucher'] && $_REQUEST['redirurl']) {
+		header("Location: {$redirurl}");
+	} elseif ($_REQUEST['voucher'] && !empty($cpcfg['redirurl'])) {
+		header("Location: {$cpcfg['redirurl']}");
```
jim-p, line 88: Allowing a redirect to a URL based on a GET request like this is a potential path for abuse/exploitation.

By making the user to a POST, we get CSRF protection and other similar assurances about the origin of the request.

If this only happens via POST, then change `$_REQUEST` to `$_POST`. Or make separate variations for `$_POST` and `$_GET` and have the `$_GET` version behave differently (such as landing on a page with the full redirect URL printed out and linked).

src/usr/local/captiveportal/index.php
```diff
@@ -157,8 +163,12 @@
 	captiveportal_logportalauth("unauthenticated", $clientmac, $clientip, "ACCEPT");
 	portal_allow($clientip, $clientmac, "unauthenticated");
 
-} elseif (isset($config['voucher'][$cpzone]['enable']) && $_POST['accept'] && $_POST['auth_voucher']) {
-	$voucher = trim($_POST['auth_voucher']);
+} elseif (isset($config['voucher'][$cpzone]['enable']) && (($_POST['accept'] && $_POST['auth_voucher']) || $_REQUEST['voucher'])) {
+	if (isset($_POST['auth_voucher'])) {
+		$voucher = trim($_POST['auth_voucher']);
+	} else {
+		$voucher = trim($_REQUEST['voucher']);
```
jim-p, line 170: Similar to above, we shouldn't accept a voucher as-is to change the user state when it is supplied via GET.

It could pre-fill the voucher in the login form, but it should not take the value and log the user in automatically.

vktg: Fixed.
Now it pre-fill the voucher in the login form

src/usr/local/captiveportal/index.php
vktg: Fixed
Now it shows "You can proceed to <URL>" link

jim-p (approved):

PR #4328, opened 2020-05-23
Limiters/queues rename/delete fix. Issue #3924

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/3924
- [X] Ready for review

This PR fixes limiters/queues deleting from the fw rules,
and automatically renames limiters/queues in the rules in case of renaming


vktg:
since the queue/limiter can be successfully removed from fw rules using these functions, it seems that this code is no longer needed:
https://github.com/pfsense/pfsense/blob/d2abe7c919eaf0c40b911278b96f9bab4fa0be45/src/usr/local/www/firewall_shaper_vinterface.php#L99

PR #4330, opened 2020-05-24
Status / Queues root queue bandwidth calculation fix. Issue #3381

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/3381
- [X] Ready for review

There is no need to do += for the root queue, 
but only for child queues that are directly under the root queue

unused/outdated code is also deleted


vktg:
Done

PR #4331, opened 2020-05-24
Update interfaces_staticarp_configure()

batnas:
This commit implements code to skip disabling staticarp on boot.
We do this because interfaces created on boot are created without staticarp enabled, therefore we dont need to disable it on interfaces on boot.
This fixes issue #10589

This greatly improves boot times on systems with many interfaces.

Also implemented is a simplification of code to handle static ARP entries.

- [x ] Redmine Issue: https://redmine.pfsense.org/issues/10589
- [x ] Ready for review


rbgarga:
@batnas please rebase and fix conflicts

rbgarga:
Manually merged.  Thanks!

PR #4332, opened 2020-05-24
DynDNS with gateway group restart on failover event. Issue #9435

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/9435
- [X] Ready for review


src/etc/rc.dyndns.update
```diff
@@ -46,6 +46,16 @@ if (empty($argument) || $argument == "all") {
 	}
 	services_dyndns_configure($interface);
 	services_dnsupdate_process($interface);
+
+	/* check if this interface is used by gateway groups,
+	 * see https://redmine.pfsense.org/issues/9435 */
+	$gwgroups = gateway_is_gwgroup_member($argument);
+	if ($gwgroups) {
```
jim-p, line 53: `gateway_is_gwgroup_member()` will always return an array. Doing a `foreach` on an empty array is essentially a no-op anyhow, so this `if` test can be eliminated. Following that, `$gwgroups` can also be eliminated. Do `foreach (gateway_is_gwgroup_member($argument) as $gw) {` instead.

vktg: Thank, done

rschell:
Commit results in PHP errors, see "PHP Errors after latest update (amd64) built on Thu Jun 11 13:02:15 EDT 2020"

PR #4337, opened 2020-05-27
Setting host-uniq for PPPoE. Implements #10597

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10597
- [X] Ready for review

Host-Uniq tag, required by some Internet providers, using the "[AC-Name][Host-Uniq|]Service-Name" syntax. To set a binary Host-Uniq, it must be encoded as a hexadecimal lower-case string and prefixed with "0x", for example "0x6d792d746167" is equivalent to "my-tag".


src/usr/local/www/interfaces.php
```diff
@@ -2837,6 +2848,13 @@ function build_port_list() {
 	$pconfig['provider']
 ))->setHelp('This field can usually be left empty.');
 
+$section->addInput(new Form_Input(
+	'hostuniq',
+	'Host-Uniq',
+	'text',
+	$pconfig['hostuniq']
+))->setHelp('Required by some ISP, usually can be left empty.');
```
jim-p, line 2856: change to "A unique host tag value for this PPPoE client. Leave blank unless a value is required by the service provider."

vktg: Fixed,

now  it uses correct syntax:
`set pppoe service "hostuniq|provider"`
see https://reviews.freebsd.org/D9270#229774
I see the correct host-uniq value in packet capture

also found the old PR https://github.com/pfsense/pfsense/pull/4048 with this feature

PR #4338, opened 2020-05-27
Do not halt on configuration file not found error. Implements #10556

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10556
- [X] Ready for review

After a shutdown/filesystem error I got on boot:

> XML configuration file not found. pfSense cannot continue booting.

then system halted
This is not a very useful behavior in this situation,
With this PR it goes forward and shows the console menu that allow you to make factory reset for example

### DIFF
--- a/src/etc/inc/config.lib.inc
+++ b/src/etc/inc/config.lib.inc
@@ -237,11 +237,12 @@ function parse_config_bootup() {
 			if (!file_exists("{$g['conf_path']}/config.xml")) {
 				echo sprintf(gettext("XML configuration file not found.  %s cannot continue booting."), $g['product_name']) . "\n";
 				unlock($lockkey);
-				mwexec("/sbin/halt");
-				exit;
+				log_error(gettext("Could not find a usable configuration file! Exiting...."));
+				exit(0);
+			} else {
+				log_error("Last known config found and restored.  Please double check the configuration file for accuracy.");
+				file_notice("config.xml", gettext("Last known config found and restored.  Please double check the configuration file for accuracy."), "pfSenseConfigurator", "");
 			}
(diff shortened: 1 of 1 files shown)


vktg:
> Worth a try, though I have some concerns about how the system might behave. Automatically falling back to the default config could be dangerous as well, so this may be the safest path other than halting.

The current behavior does not give you a choice and just halt the system, 
The only way for the user in this case is to boot in single mode and create an empty `config.xml` to avoid system shutdown, and enter the console menu.

jim-p:
> The current behavior does not give you a choice and just halt the system,
> The only way for the user in this case is to boot in single mode and create an empty `config.xml` to avoid system shutdown, and enter the console menu.

It does this out of an abundance of caution. If the configuration is lost, it has no way to know what should be done with any interfaces, daemons, etc. The behavior could be undesirable in some way, depending on the system state. If booting this way only results in a usable console, and no network connectivity, it should be OK.

The concern with defaulting is that we don't know for sure that the default configuration is safe. For example, the interface used as the default "LAN" may not be a network that should be configured with the default LAN settings (192.168.1.x, DHCP, relaxed rules, etc).

But if this doesn't configure anything and skips right to the menu without configuring any network connectivity it should be reasonably safe.

Though maybe instead of an `exit` here it could call `/etc/rc.initial` directly, though it may be too early for that to be viable.

vktg:
> If the configuration is lost, it has no way to know what should be done with any interfaces, daemons, etc. The behavior could be undesirable in some way, depending on the system state. If booting this way only results in a usable console, and no network connectivity, it should be OK.

With the current PR changes, it shows only config file errors and runs `/etc/rc.initial`,
so it should be OK.

PR #4340, opened 2020-05-28
6RD and 6to4 interface MTU set fix. Issue #6377

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/6377
- [X] Ready for review

6rd and 6to4 tunnel MTU is not user editable and is hardcoded to 1280.

This PR automatically sets the tunnel MTU to the interface MTU - 20 (IPv4 header size)


vktg:
Updated,
now it sets MTU correctly if there is no `$wancfg['mtu']`
also added input validation on `interfaces.php` to avoid MTU < 1280

PR #4344, opened 2020-06-01
Add support for freeDNS DynDNS v2 API [Feature #10617]

einichi:
Redmine Issue: https://redmine.pfsense.org/issues/10617

This adds support for freeDNS (afraid.org)'s DynDNS service's more recent API version, which hosts a IPv6 only endpoint, allowing for more reliable IPv6 DynDNS updates.

https://freedns.afraid.org/dynamic/v2/

I had trouble getting IPv6 to update properly using the V1 module IPv6 implementation. Instead of trying to fix the problem, I figured just implement the latest API. The V1 API implementation is left untouched for backwards compatibility.

Tested working for both v4 and v6.


src/etc/inc/dyndns.class
```diff
@@ -219,6 +223,8 @@
 			switch ($dnsService) {
 			case 'freedns':
 			case 'freedns-v6':
+			case 'freedns-v2':
+			case 'freedns-v6-v2':
```
jim-p, line 227: Using `-v6-v2` seems confusing... Maybe this should be `freedns2` and `freedns2-v6`?

einichi: Of course, that's way better.

PR #4350, opened 2020-06-06
Improve Remote Gateway field description for IPSec VPN Phase 1. Implements #7095

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/7095
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/vpn_ipsec_phase1.php
+++ b/src/usr/local/www/vpn_ipsec_phase1.php
@@ -298,12 +298,24 @@ if ($_POST['save']) {
 		foreach ($a_phase1 as $ph1tmp) {
 			if ($p1index != $t) {
 				$tremotegw = $pconfig['remotegw'];
-				if (($ph1tmp['remote-gateway'] == $tremotegw) && !isset($ph1tmp['disabled']) && (!isset($pconfig['gw_duplicates']) || !isset($ph1tmp['gw_duplicates']))) {
+				if (($ph1tmp['remote-gateway'] == $tremotegw) && ($ph1tmp['remote-gateway'] != '0.0.0.0') &&
+				    ($ph1tmp['remote-gateway'] != '::') && !isset($ph1tmp['disabled']) &&
+				    (!isset($pconfig['gw_duplicates']) || !isset($ph1tmp['gw_duplicates']))) {
 					$input_errors[] = sprintf(gettext('The remote gateway "%1$s" is already used by phase1 "%2$s".'), $tremotegw, $ph1tmp['descr']);
 				}
 			}
 			$t++;
 		}
+	}
+
+	if (($pconfig['remotegw'] == '0.0.0.0') || ($pconfig['remotegw'] == '::')) {
+		if (!isset($pconfig['responderonly'])) {
+			$input_errors[] = gettext('The remote gateway "0.0.0.0" or "::" address can only be used with "Responder Only".');
+		}
+		if ($pconfig['peerid_type'] == "peeraddress") {
+			$input_errors[] = gettext('The remote gateway "0.0.0.0" or "::" address can not be used with IP address peer identifier.');
+		}
+
 	}
 
 	if (($pconfig['iketype'] == "ikev1") && is_array($a_phase2) && (count($a_phase2))) {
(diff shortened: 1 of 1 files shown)


src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -732,7 +732,8 @@ function build_eal_list() {
 		'*Remote Gateway',
 		'text',
 		$pconfig['remotegw']
-	))->setHelp('Enter the public IP address or host name of the remote gateway.');
+	))->setHelp('Enter the public IP address or host name of the remote gateway. ' .
+		    'Use 0.0.0.0 to allow connection from any IPv4 address.');
```
jim-p, line 736: Needs more explanation about the limitations of doing that: Can't initiate, can't use the peer IP address for the remote identifier, and may be limited to a single entry.

Though the only one entry may have been address by PR #4173 but I don't know if anyone tested that with `0.0.0.0`. Test to confirm that.

The identifier/PSK issue may have a workaround (define a PSK for `%any` on the PSK tab) but that's probably not something we should encourage.

vktg: > Needs more explanation about the limitations of doing that: Can't initiate, can't use the peer IP address for the remote identifier, and may be limited to a single entry.

Added

> Though the only one entry may have been address by PR #4173 but I don't know if anyone tested that with `0.0.0.0`. Test to confirm that.

Multiple entries with '0.0.0.0' with different IDs works fine

> The identifier/PSK issue may have a workaround (define a PSK for `%any` on the PSK tab) but that's probably not something we should encourage.

No issues were found after adding Mobile IPsec with PSK% any - other site-to-site '0.0.0.0' connections successfully connect

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -732,7 +744,12 @@ function build_eal_list() {
 		'*Remote Gateway',
 		'text',
 		$pconfig['remotegw']
-	))->setHelp('Enter the public IP address or host name of the remote gateway.');
+	))->setHelp('Enter the public IP address or host name of the remote gateway.%1$s%2$s%3$s',
+	    '<div class="infoblock">',
+	    sprint_info_box(gettext('Use \'0.0.0.0\' to allow connection from any IPv4 address or \'::\' ' .
```
jim-p, line 749: `connection` should be plural: `connections`

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -732,7 +744,12 @@ function build_eal_list() {
 		'*Remote Gateway',
 		'text',
 		$pconfig['remotegw']
-	))->setHelp('Enter the public IP address or host name of the remote gateway.');
+	))->setHelp('Enter the public IP address or host name of the remote gateway.%1$s%2$s%3$s',
+	    '<div class="infoblock">',
+	    sprint_info_box(gettext('Use \'0.0.0.0\' to allow connection from any IPv4 address or \'::\' ' . 
+	    'to allow connection from any IPv6 address.' . '<br/>' . 'It supports only Responder mode, ' .
```
jim-p, line 750: `connection` should be plural: `connections`

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -732,7 +744,12 @@ function build_eal_list() {
 		'*Remote Gateway',
 		'text',
 		$pconfig['remotegw']
-	))->setHelp('Enter the public IP address or host name of the remote gateway.');
+	))->setHelp('Enter the public IP address or host name of the remote gateway.%1$s%2$s%3$s',
+	    '<div class="infoblock">',
+	    sprint_info_box(gettext('Use \'0.0.0.0\' to allow connection from any IPv4 address or \'::\' ' . 
+	    'to allow connection from any IPv6 address.' . '<br/>' . 'It supports only Responder mode, ' . 
+	    'can\'t use the peer IP address for the remote identifier, and limited to a single entry.'), 'info', false),
```
jim-p, line 751: Starting with `It supports only Responder mode,` this needs reworded as it's a bit awkward/incorrect grammar.

Perhaps: "Responder Only must be set, Peer IP Address cannot be used for Remote Identifier, and only a single Phase 1 entry may use this Remote Gateway value."

vktg: ok, all fixed
but you can use 0.0.0.0 for multiple PH1 (tested, see above)

jim-p: If it works OK, that can be removed then.

vktg: 'and limited to a single entry.' is already removed

PR #4355, opened 2020-06-13
PRIQ ignore bandwidth. Fixes #10660

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10660
- [X] Ready for review

The priq scheduler does not support bandwidth specification.
This PR removes any bandwidth manipulations from `priq_queue` class

https://www.freebsd.org/cgi/man.cgi?query=pf.conf&sektion=5&apropos=0&manpath=FreeBSD+12.1-RELEASE+and+Ports:
```
bandwidth <bw>
	   Specifies the maximum bitrate to be processed by the	queue.	This
	   value must not exceed the value of the parent queue and can be
	   specified as	an absolute value or a percentage of the parent
	   queue's bandwidth.  If not specified, defaults to 100% of the par-
	   ent queue's bandwidth.  The priq scheduler does not support band-
	   width specification.
```

### DIFF
--- a/src/etc/inc/shaper.inc
+++ b/src/etc/inc/shaper.inc
@@ -1147,7 +1147,7 @@ class altq_root_queue {
 			null,
 			'number',
 			$this->GetBandwidth()
-		));
+		))->setHelp('The PRIQ scheduler does not support bandwidth specification.');
 
 		$group->add(new Form_Select(
 			'bandwidthtype',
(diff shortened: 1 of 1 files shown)


vktg:
but this may be another issue, pf.conf (5):
 bandwidth <bw>
	   The maximum bitrate for all queues on an interface may be specified
	   using the bandwidth keyword.	 The value can be specified as an ab-
	   solute value	or as a	percentage of the interface bandwidth.	When
	   using an absolute value, the	suffixes b, Kb,	Mb, and	Gb are used to
	   represent bits, kilobits, megabits, and gigabits per	second,	re-
	   spectively.	The value must not exceed the interface	bandwidth.  If
	   bandwidth is	not specified, the interface bandwidth is used (**but
	   take note that some interfaces do not know their bandwidth**, or can
	   adapt their bandwidth rates).

vktg:
misunderstanding of the issue, closed

PR #4357, opened 2020-06-17
Bridge with GIF interface bootup fix. Issue #10524

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10524
- [X] Ready for review

1) There is no need to check `$config['interfaces'][$bridge['if']]['ipaddrv6'] != "track6"` when `checkmember == 2`,- all needed interfaces are up before with `$track6_list` and `$delayed_list` foreach loops, and it's safe to exec `interface_bridge_configure($bridge, $checkmember);`

2) unused `foreach ($members as $member)` removed;

3) this code is not needed because the `<bridged>` does not have `<if>` attribute:
```
if (strstr($bridge['if'], '_vip')) {
	continue;
}
```

### DIFF
--- a/src/etc/inc/pkg-utils.inc
+++ b/src/etc/inc/pkg-utils.inc
@@ -1567,21 +1567,26 @@ function update_repos() {
 // Parse the received JSON data and save the custom repository information
 function save_repo($json) {
 	global $repopath, $g;
+
 	$repo = json_decode($json, true);
 
-	if (($repo != NULL) && isset($repo['abi']) && isset($repo['altabi']) && isset($repo['conf']) &&
-	    isset($repo['descr']) && isset($repo['name']) && (strlen($repo['conf']) > 10)) {
+	if (($repo != NULL) && isset($repo['abi']) && isset($repo['altabi']) &&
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -677,30 +677,23 @@ function interfaces_bridge_configure($checkmember = 0, $realif = "") {
...
-				}
-				$members = explode(',', $bridge['members']);
-				foreach ($members as $member) {
-					if (!empty($config['interfaces'][$bridge['if']]) && $config['interfaces'][$bridge['if']]['ipaddrv6'] == "track6") {
-						continue 2;
-					}
+			$ifname = false;
+			foreach ($config['interfaces'] as $intname => $intpar) {
+				if ($intpar['if'] == $bridge['bridgeif']) {
+					$ifname = $intname;
```
rbgarga, line 686: Are you looking for last item or array?  If not, it would be better to add a `break;` here

vktg: thanks, fixed

for some reason PR is closed after I did `git push -f origin`,
new PR https://github.com/pfsense/pfsense/pull/4360

PR #4360, opened 2020-06-18
Bridge with GIF interface bootup fix. Issue #10524

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10524
- [X] Ready for review

1) There is no need to check `$config['interfaces'][$bridge['if']]['ipaddrv6'] != "track6"` when `checkmember == 2`,- all needed interfaces are up before with `$track6_list` and `$delayed_list` foreach loops, and it's safe to exec `interface_bridge_configure($bridge, $checkmember);`

2) unused `foreach ($members as $member)` removed;

3) this code is not needed because the `<bridged>` does not have `<if>` attribute:
```
if (strstr($bridge['if'], '_vip')) {
	continue;
}
```

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -677,30 +677,24 @@ function interfaces_bridge_configure($checkmember = 0, $realif = "") {
 			if (empty($bridge['bridgeif'])) {
 				$bridge['bridgeif'] = "bridge{$i}";
 			}
-			if (!empty($realif) && $realif != $bridge['bridgeif']) {
+			if (!empty($realif) && ($realif != $bridge['bridgeif'])) {
+				continue;
+			}
+			$ifname = false;
+			foreach ($config['interfaces'] as $intname => $intpar) {
+				if ($intpar['if'] == $bridge['bridgeif']) {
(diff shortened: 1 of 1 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -677,30 +677,24 @@ function interfaces_bridge_configure($checkmember = 0, $realif = "") {
...
-			if (!empty($realif) && $realif != $bridge['bridgeif']) {
+			if (!empty($realif) && ($realif != $bridge['bridgeif'])) {
 				continue;
 			}
-
-			if ($checkmember == 1) {
-				/* XXX: It should not be possible no? */
-				if (strstr($bridge['if'], '_vip')) {
-					continue;
-				}
```
jim-p, line 688: At the time this bit of code was added, that field did exist. Looks like it is `bridgeif` now. Though I don't think it's possible for a CARP VIP to get in there so maybe the check isn't necessary anymore. Back when that was added, CARP still worked like an interface and not like an IP alias as it does now.

vktg: Interesting..
Will this PR be approved?

PR #4363, opened 2020-06-19
Allow to change WiFi channel width. Implements #10678

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10678
- [X] Ready for review

Allow to change 802.11n channel width (ht/20, ht/40, etc) from the WebGUI

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -3267,7 +3267,12 @@ function interface_wireless_configure($if, &$wl, &$wlcfg) {
 		if ($wlcfg['channel'] == "0") {
 			$wlcmd[] = "channel any";
 		} else {
-			$wlcmd[] = "channel " . escapeshellarg($wlcfg['channel']);
+			if ($wlcfg['channel_width'] != "0") {
+				$channel_width = ":" . $wlcfg['channel_width'];
+			} else {
+				$channel_width = '';
+			}
+			$wlcmd[] = "channel " . escapeshellarg($wlcfg['channel']) . $channel_width;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/etc/inc/interfaces.inc
```diff
@@ -6799,6 +6804,72 @@ function get_wireless_modes($interface) {
...
 }
 
+function get_wireless_ht_modes($interface) {
+	/* return wireless HT */
+	$wireless_hts_supported = array(0 => gettext('Auto'));
+
+	$cloned_interface = get_real_interface($interface);
+
+	if ($cloned_interface && is_interface_wireless($cloned_interface)) {
+		$chan_list = "/sbin/ifconfig -v {$cloned_interface} list chan";
```
jim-p, line 6814: I know we have reasonable assurance that `$cloned_interface` is valid here but I would like to see that run through `escapeshellarg()` to be safe.

src/etc/inc/interfaces.inc
```diff
@@ -6799,6 +6804,72 @@ function get_wireless_modes($interface) {
...
+		$stack_list = "/usr/bin/awk -F\"Channel \" '{ gsub(/\\*/, \" \"); print \$2 \"\\\n\" \$3 }'";
+		$format_list = "/usr/bin/awk '{print \$5 \",\" \$6 \",\" \$1}'";
+
+		$interface_channels = "";
+		exec("$chan_list | $stack_list | sort -u | $format_list 2>&1", $interface_channels);
```
jim-p, line 6819: `sort` needs a full path

src/etc/inc/interfaces.inc
```diff
@@ -6799,6 +6804,72 @@ function get_wireless_modes($interface) {
...
```
jim-p, line 6844: Why repeat all of this? Make a common function and call it from both places.

vktg: all fixed

PR #4367, opened 2020-06-22
Improve gateways popup. Implements #885

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/885
- [X] Ready for review

Improved:
- Now uses `return_gateways_array()` to correctly show default gateway
- Does not show empty popup
- Uses `disablealiaspopupdetail` config option (maybe not needed?)

### DIFF
--- a/src/usr/local/www/guiconfig.inc
+++ b/src/usr/local/www/guiconfig.inc
@@ -1087,28 +1087,30 @@ function alias_info_popup($alias_id) {
 	return $content;
 }
 
-function gateway_info_popup($gateway) {
-	global $config;
-	init_config_arr(array('gateways', 'gateway_item'));
+function gateway_info_popup($showgw) {
+	global $config, $user_settings;
+
 	init_config_arr(array('gateways', 'gateway_group'));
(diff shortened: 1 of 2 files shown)


src/usr/local/www/guiconfig.inc
```diff
@@ -1087,28 +1087,34 @@ function alias_info_popup($alias_id) {
 	return $content;
 }
 
-function gateway_info_popup($gateway) {
-	global $config;
-	init_config_arr(array('gateways', 'gateway_item'));
+function gateway_info_popup($showgw) {
+	global $config, $user_settings;
+
+	if ($user_settings['webgui']['disablealiaspopupdetail']) {
```
jim-p, line 1093: Why key off this? The alias popup detail option is there to suppress them because if a user has gigantic aliases, the popup windows do not work properly and can affect page rendering and speed.  There wouldn't be any similar issues with gateway popups.

vktg: Ok, removed

PR #4369, opened 2020-06-23
Hide IPsec Pre-Shared Keys non-psk fields. Issue #10532

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10532
- [X] Ready for review

IPsec Mobile PSK client not able specify a user ID, so it's better to hide fields not supported by PSK

it also removes "PSK for any user can be installed using any identifier." infobox, as this can only confuse users
Now the identifier type should always be selected for EAP


vktg:
Fixed
now it only hides the `identify_type`,` virtualip` and `dns_address` fields if PSK type is selected

jim-p (approved):

PR #4375, opened 2020-06-25
Support dynamic interface address for 1:1 NAT. Implements #7705

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/7705
- [X] Ready for review

Allows to use dynamic interface address (like 'WAN address') for 1:1 NAT
Currently IPv4-only

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -1965,7 +1965,13 @@ function filter_nat_rules_generate() {
 
 			$sn = "";
 			$sn1 = "";
-			$target = alias_expand($rule['external']);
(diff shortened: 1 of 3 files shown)


src/etc/inc/filter.inc
```diff
@@ -1965,7 +1965,14 @@ function filter_nat_rules_generate() {
 
 			$sn = "";
 			$sn1 = "";
-			$target = alias_expand($rule['external']);
+			//$target = alias_expand($rule['external']);
```
rbgarga, line 1968: Commented out line should be removed

vktg: fixed

PR #4381, opened 2020-06-30
Block additional Captive Portal Logins. Implements #9432

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/9432
- [X] Ready for review

Today, if Concurrency Logins enabled, if I try to use a voucher used by other, i can use and last client was disconnected.
But, i need invert this.
New user cannot use voucher, and last client stay connected.

based on `1V1D Patch.zip` from https://forum.netgate.com/topic/136995/one-voucher-per-device/128

### DIFF
--- a/src/usr/local/www/services_captiveportal.php
+++ b/src/usr/local/www/services_captiveportal.php
@@ -46,6 +46,7 @@ require_once("guiconfig.inc");
 global $cpzone;
 global $cpzoneid;
 
+$concurrentlogins_list = array("disabled" => "Disabled", "multiple" => "Multiple", "last" => "Last login", "first" => "First login");
 $cpzoneid = 1; /* Just a default */
 $cpzone = $_REQUEST['zone'];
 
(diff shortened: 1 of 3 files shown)


src/usr/local/www/services_captiveportal.php
```diff
@@ -633,13 +651,19 @@ function build_authserver_list() {
 	))->setHelp("If enabled, connected users won't be disconnected during a pfSense reboot.");
 }
 
-$section->addInput(new Form_Checkbox(
+$concurrentlogins_list = array("disabled" => "Disabled", "multiple" => "Multiple", "last" => "Last login", "first" => "First login");
```
jim-p, line 654: Define this up higher, then you can also use the keys when validating rather than having to duplicate that content with another array. Then if it changes in the future, only one place needs updating.

Also, the user-visible parts should be run through `gettext()`.

src/usr/local/www/services_captiveportal.php
```diff
@@ -278,6 +286,10 @@
 		}
 	}
 
+	if (isset($_POST['noconcurrentlogins']) && !in_array($_POST['noconcurrentlogins'], array('disabled', 'multiple', 'last', 'first'))) {
```
jim-p, line 289: See other note about using the array keys of the option list variable for this instead of hardcoding the possible choices in an array.

src/usr/local/www/services_captiveportal.php
```diff
@@ -633,13 +651,19 @@ function build_authserver_list() {
...
-	'Disable Concurrent user logins',
-	$pconfig['noconcurrentlogins']
-))->setHelp('If enabled only the most recent login per username will be active. Subsequent logins will cause machines previously logged in with the ' .
-			'same username to be disconnected.');
+	$pconfig['noconcurrentlogins'],
+	$concurrentlogins_list
+))->setHelp('Disabled : do not allow concurrent logins per username or voucher.<br />'.
+			'Multiple : no restrictions to the number of logins per username or voucher will be applied.<br />'.
+			' Last login : only the most recent login per username or voucher will be granted.'.
+			' Previous logins will be disconnected.<br />'.
```
jim-p, line 664: Put the `<br />` parts in separate strings so they don't get put into gettext strings. Or use the sprintf style. There are various examples on other pages.

src/usr/local/www/services_captiveportal.php
```diff
@@ -633,13 +651,19 @@ function build_authserver_list() {
...
+			' First login : only the first login per username or voucher will be granted.'.
```
jim-p, line 665: Don't put a space before the `:`, and capitalize the first letter after.

vktg: All fixed

PR #4383, opened 2020-07-03
Gateway down policy rule fix. Issue #10716

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10716
- [X] Ready for review

Fixes PBR rules creation when:
- Gateway is down
- Gateway in gateways group is down or disabled (`force_down`)

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -876,6 +876,7 @@ function filter_generate_gateways() {
 	/* Lookup Gateways to be used in filter rules once */
 	$GatewaysList = return_gateways_array();
 	$GatewayGroupsList = return_gateway_groups_array(true);
+	$GatewaysStatus = return_gateways_status(true);
 
 	if (is_array($GatewaysList)) {
 		foreach ($GatewaysList as $gwname => $gateway) {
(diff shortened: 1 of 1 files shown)


stephenw10:
This correctly omits the rule entirely when 'Skip rules when gateway is down` is selected but it does not omit the gateway from the rule in the default setting.
Gateway looks like:
`GWGW2 = "  "`
But the rule still contains it:
`pass  in  quick  on $LAN  $GWGW2 inet proto tcp  from any to 9.9.9.9 tracker 1593470028 flags S/SA keep state  label "USER_RULE"`

stephenw10:
But of course since the value of $GWGW2 is now '  ' that does omit it!

vktg:
Gateways Group 'Trigger Level' also needs to be considered,
with the latest changes in this PR it only checks for 'Member Down' state

vktg:
Updated
Gateways Group 'Trigger Level' and 'Disable Gateway Monitoring Action' check added

stephenw10:
Tested this. Looks good. Gateway is marked down and removed from ruleset and/or rules removed as expected.

PR #4384, opened 2020-07-03
Reboot message after changing Hardware Checksum Offloading setting. Implements #3031

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/3031
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/system_advanced_network.php
+++ b/src/usr/local/www/system_advanced_network.php
@@ -38,6 +38,8 @@ require_once("functions.inc");
 require_once("filter.inc");
 require_once("shaper.inc");
 
+$show_reboot_msg = false;
+$reboot_msg = gettext('Changing the Hardware Checksum setting requires a system reboot.') . '\n\n' . gettext('Reboot now?');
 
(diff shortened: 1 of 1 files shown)


src/usr/local/www/system_advanced_network.php
```diff
@@ -38,6 +38,9 @@
 require_once("filter.inc");
 require_once("shaper.inc");
 
+$show_reboot_msg = false;
+$reboot_msg = gettext('Changing the Hardware Checksum setting requires a system reboot' .
+    '\n\nReboot now?');
```
jim-p, line 43: Since this is a gettext string, it probably shouldn't have `\n\n` hardcoded. Use a placeholder and sprintf to get that in the string instead, or use separate gettext calls:

`gettext('Changing the Hardware Checksum setting requires a system reboot.') . '\n\n' . gettext('Reboot now?')`

Also, just to make sure, the `\n\n` have to be passed through intact to the JavaScript string, right? Otherwise you wouldn't want to put `\n\n` in single quotes as they won't be interpreted as newline characters.

vktg: Thanks, fixed

PR #4385, opened 2020-07-03
Disable "Hardware Checksum Offloading" if VTNET is detected. Implements #10723

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10723
- [X] Ready for review

It would be better to disable "Hardware Checksum Offloading" on first boot if a VM system is detected

### DIFF
--- a/src/etc/inc/pfsense-utils.inc
+++ b/src/etc/inc/pfsense-utils.inc
@@ -541,7 +541,10 @@ function hardware_offloading_applyflags($iface) {
 	$flags_off = 0;
 	$options = pfSense_get_interface_addresses($iface);
 
-	if (isset($config['system']['disablechecksumoffloading'])) {
+	/* disable hardware checksum offloading for VirtIO network drivers,
+	 * see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=165059 */
+	if (isset($config['system']['disablechecksumoffloading']) ||
+	    stristr($iface, "vtnet")) { 
 		if (isset($options['encaps']['txcsum'])) {
 			$flags_off |= IFCAP_TXCSUM;
 		}


vktg:
Updated
Now checks only for QEMU VM

src/etc/rc.bootup
```diff
@@ -89,6 +89,11 @@ if (file_exists("/root/firmware.tgz")) {
 
 /* Triggering of the initial setup wizard after reboot has been requested */
 if (file_exists("{$g['conf_path']}/trigger_initial_wizard_after_reboot")) {
+	/* Disable "Hardware Checksum Offloading" on first boot if QEMU VM system is detected */
+	if (isqemu()) {
+		$config['system']['disablechecksumoffloading'] = true;
+		write_config();
+	}
```
jim-p, line 96: Why store this in the config? Someone could restore a config without it and end up without the setting.

Seems like it would be better to test when the interface settings are applied https://github.com/pfsense/pfsense/blob/79e269c9f2002f780d2fd141595a1cec1683bb12/src/etc/inc/pfsense-utils.inc#L544

And is this necessary for any interface type in qemu, or only vtnet? If it isn't needed for em interfaces, for example, we probably shouldn't do it there.

vktg:
Updated

jim-p (changes requested):
The net change here only appears to disable TSO for vtnet, and doesn't touch checksum offloading. Did this get mixed up with another change maybe?

vktg:
Fixed
Now it disables hardware checksum offloading for vtnet interfaces

jim-p (approved):

jim-p:
It's OK as is but also your last note on Redmine said it affects `ena` interfaces as well, so you might add that here or make another PR so that doesn't get lost.

PR #4396, opened 2020-07-11
IPsec Mobile RADIUS Group authentication. Implements #10748

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10748
- [X] Ready for review

Allows to limit which users have VPN access via groups (RADIUS `Class` attribute) in the WebGUI
see https://wiki.strongswan.org/projects/strongswan/wiki/EAPRadius#Group-selection
and https://wiki.strongswan.org/projects/strongswan/wiki/Fromipsecconf:
ipsec.conf format | swanctl.conf format
-- | --
rightgroups=`<group list>` | connections.<conn>.remote<suffix>.groups=`<group list>`

### DIFF
--- a/src/usr/local/www/vpn_ipsec_mobile.php
+++ b/src/usr/local/www/vpn_ipsec_mobile.php
@@ -45,13 +45,34 @@ $a_phase1 = &$config['ipsec']['phase1'];
 init_config_arr(array('ipsec', 'client'));
 $a_client = &$config['ipsec']['client'];
 
+init_config_arr(array('system', 'group'));
+$a_group = &$config['system']['group'];
+
+$auth_groups = array();
+foreach ($config['system']['group'] as $group) {
+	if (isset($group['priv'])) {
+		foreach ($group['priv'] as $priv) {
+			if (($priv == 'page-all') || ($priv == 'user-ipsec-xauth-dialin')) {
+				if (!empty($group['description'])) {
+					$auth_groups[$group['name']] = $group['description'] . " (" . $group['name'] . ")";
+				} else {
+					$auth_groups[$group['name']] = $group['name'];
+				}
+				continue 2;
+			}
+		}
+	}
+}
+
 if (count($a_client)) {
 
 	$pconfig['enable'] = $a_client['enable'];
 	$pconfig['radiusaccounting'] = ($a_client['radiusaccounting'] == 'enabled');
+	$pconfig['radius_groups'] = $a_client['radius_groups'];
 
 	$pconfig['user_source'] = $a_client['user_source'];
 	$pconfig['group_source'] = $a_client['group_source'];
+	$pconfig['auth_groups'] = $a_client['auth_groups'];
 
 	$pconfig['pool_address'] = $a_client['pool_address'];
 	$pconfig['pool_netbits'] = $a_client['pool_netbits'];
@@ -73,6 +94,11 @@ if (count($a_client)) {
 	
 	if (isset($pconfig['enable'])) {
 		$pconfig['enable'] = true;
+	}
+
+	if ($pconfig['group_source'] == 'enabled') {
+		$pconfig['group_source'] = true;
+		$pconfig['auth_groups'] = $a_client['auth_groups'];
 	}
 
 	if ($pconfig['pool_address'] && $pconfig['pool_netbits']) {
@@ -153,8 +179,8 @@ if ($_POST['save']) {
 
 	/* input validation */
 
-	$reqdfields = explode(" ", "user_source group_source");
-	$reqdfieldsn = array(gettext("User Authentication Source"), gettext("Group Authentication Source"));
+	$reqdfields = explode(" ", "user_source");
+	$reqdfieldsn = array(gettext("User Authentication Source"));
 
 	do_input_validation($_POST, $reqdfields, $reqdfieldsn, $input_errors);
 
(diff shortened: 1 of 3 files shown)


zyv:
Wow, thank you very much for a prompt reaction, the code looks pretty awesome!

I'm totally new to pfSense (SG-5100 up since 2 days), so I'm not sure can be of any help with testing, but I have checked whether `class_group` is already set to `yes` and it is indeed the case:

https://github.com/pfsense/pfsense/blob/21568e753abb092747fddeeda41a9952827b06d1/src/etc/inc/ipsec.inc#L1423

... so contrary to what I originally thought, due to the work already done in c4a9f99a8d79e201b2af5053a095c83bb1a26467 these changes are really everything that is needed to implement #10748.

vktg:
> I'm totally new to pfSense (SG-5100 up since 2 days), so I'm not sure can be of any help with testing, but I have checked whether `class_group` is already set to `yes` and it is indeed the case

This PR is for pfSense 2.5 branch and is not yet approved,
Please wait until it is approved, and if you want to use it with pfSense 2.4.5, you'll need to modify it

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -477,9 +484,22 @@ function login_banner_change() {
...
 
+$group = new Form_Group('RADIUS Groups');
+$group->addClass('radius_gr');
+
+$group->add(new Form_Input(
+	'radius_groups',
+	'RADIUS Groups',
+	'text',
+	htmlspecialchars($pconfig['radius_groups'])
+))->setHelp('Comma separated list of RADIUS groups for the EAP-RADIUS authentication.');
```
jim-p, line 499: For other areas/purposes like GUI auth we require the groups to be configured locally and matched via RADIUS responses. I'd prefer to see this as a multi-select list of existing local groups rather than a free-form box. Perhaps limit the groups that show up (and get used in the backend) to those with appropriate IPsec privileges as well.

zyv: Hmmm, so as a user, I will have to create all groups that I would like to use here locally on the pfSense box, so that I can select them in the picker?

> Perhaps limit the groups that show up (and get used in the backend) to those with appropriate IPsec privileges as well.

As a user, how would a give a local group "appropriate IPsec privileges"? Thanks for the clarification!

jim-p: > Hmmm, so as a user, I will have to create all groups that I would like to use here locally on the pfSense box, so that I can select them in the picker?

Yes. That's what you have to do for other places that use RADIUS and LDAP group permissions like for GUI authentication. It's so groups obtained remotely can be mapped to groups and privileges used locally.

> As a user, how would a give a local group "appropriate IPsec privileges"? Thanks for the clarification!

Create or edit a group, search for "IPsec" and you'll see the user privilege there.

vktg:
Updated,
`ipsec.auth-user.php` is also fixed, should work with Local/LDAP groups too

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -470,15 +502,30 @@ function login_banner_change() {
...
-	'*Group Authentication',
+	'Group Authentication',
+	'Group Authentication',
 	$pconfig['group_source'],
-	array(
-		'none' => gettext('none'),
-		'system' => gettext('system'),
-	)
-))->setHelp('Source');
+))->setHelp('Authenticate users from the groups that have "User - VPN: IPsec with Dialin" and "WebCfg - All pages" privileges.')
```
zyv, line 510: The code says

```php
if (($priv == 'page-all') || ($priv == 'user-ipsec-xauth-dialin'))
```

but the text says

> that have "User - VPN: IPsec with Dialin" **and** "WebCfg - All pages" privileges

I think that an **and** should be an **or** here, if I understand the intention correctly. In any case I wouldn't want to forced to give `WebCfg - All pages` to my VPN users.

jim-p: Thanks to our wonderful English language both are acceptable. While it's less ambiguous to use "or" in that context, it also doesn't say both are required, thus it can be read as "groups that have A and [groups that have] B".

That said, I am inclined to agree that it should be changed to "or".

jim-p: Change this:
> Authenticate users from the groups that have "User - VPN: IPsec with Dialin" and "WebCfg - All pages" privileges.

To this:
> Authenticate members of groups which have either "User - VPN: IPsec with Dialin" or "WebCfg - All pages" privileges.

src/usr/local/www/vpn_ipsec_mobile.php
```diff
@@ -45,13 +45,33 @@
...
+$auth_groups = array();
+foreach ($config['system']['group'] as $group) {
+	if (isset($group['priv'])) {
+		foreach ($group['priv'] as $priv) {
+			if (($priv == 'page-all') || ($priv == 'user-ipsec-xauth-dialin')) {
+				if (!empty($group['description'])) {
+					$auth_groups[$group['name']] = $group['description'] . " (" . $group['name'] . ")";
+				} else {
+					$auth_groups[$group['name']] = $group['name'];
+				}
```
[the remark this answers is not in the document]
vktg: fixed

src/usr/local/www/vpn_ipsec_mobile.php
vktg: Ok, all fixed

PR #4403, opened 2020-07-21
Set net.inet.ip.dummynet sysctl. Fixes #10780

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10780
- [X] Ready for review

### DIFF
--- a/src/etc/inc/shaper.inc
+++ b/src/etc/inc/shaper.inc
@@ -5252,7 +5252,7 @@ function dnpipe_find_nextnumber() {
 }
 
 function filter_generate_dummynet_rules() {
-	global $g, $dummynet_pipe_list;
+	global $config, $g, $dummynet_pipe_list;
 
 	read_dummynet_config();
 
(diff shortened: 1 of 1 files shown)


src/etc/inc/shaper.inc
```diff
@@ -5272,11 +5272,20 @@ function filter_generate_dummynet_rules() {
...
 			mwexec("/sbin/kldload dummynet");
 		}
-		set_sysctl(array(
+		$sysctls = (array(
 				"net.inet.ip.dummynet.io_fast" => "1",
 				"net.inet.ip.dummynet.hash_size" => "256",
 				"net.inet.ip.dummynet.pipe_slot_limit" => $max_qlimit
 		));
+		init_config_arr(array('sysctl', 'item'));
+		if (is_array($config['sysctl']) && is_array($config['sysctl']['item'])) {
```
jim-p, line 5281: After `init_config_arr()` these will always be arrays, so this test is unnecessary.

vktg: fixed

src/etc/inc/shaper.inc
```diff
@@ -5272,11 +5272,20 @@ function filter_generate_dummynet_rules() {
...
 			mwexec("/sbin/kldload dummynet");
 		}
-		set_sysctl(array(
+		$sysctls = (array(
 				"net.inet.ip.dummynet.io_fast" => "1",
 				"net.inet.ip.dummynet.hash_size" => "256",
 				"net.inet.ip.dummynet.pipe_slot_limit" => $max_qlimit
 		));
+		init_config_arr(array('sysctl', 'item'));
+		if (!empty($config['sysctl']['item'])) {
```
jim-p, line 5281: Not that it needs changed now, but for future reference, this test is also unnecessary. It's harmless to run foreach on an empty array, it is effectively a no-op since the body of the statement never executes.

PR #4412, opened 2020-08-01
Corrects spelling of CONTRIBUTING.md

comradekingu:
- [x] Ready for review


jim-p:
Most of these are invalid changes, introducing grammar and other errors. I don't see a compelling reason to change any of these.

comradekingu:
@jim-p My changes are the ones in green on the diff.
As per https://en.wikipedia.org/wiki/Git and https://en.wikipedia.org/wiki/GitHub

jim-p:
I'm aware of how diffs work. You were introducing many other errors along with the change to Git/GitHub capitalization. If you submit another PR that changes nothing else but Git/GitHub we can consider it.

PR #4413, opened 2020-08-02
status.php: Add DNS Resolver configuration. Implements #10635

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10635
- [X] Ready for review


src/usr/local/www/status.php
```diff
@@ -326,6 +326,7 @@ function get_gateway_status() {
 defCmdT("Disk-Contents of conf", "/bin/ls /conf");
 defCmdT("config.xml", "dumpconfigxml");
 defCmdT("DNS-Resolution Configuration", "/bin/cat /etc/resolv.conf");
+defCmdT("DNS-Resolver Configuration", "/bin/cat /var/unbound/unbound.conf");
```
jim-p, line 329: Probably should grab more than just that, but it's a good start.  Maybe at least access_lists.conf and domainoverrides.conf.

Definitely don't need anything from pfBlocker, though.

vktg: Updated!

PR #4427, opened 2020-08-19
Not destroying VTI interfaces when booting before creating a new one. Fixes #10842

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10842
- [X] Ready for review

During the booting process, we call interface_ipsec_vti_configure() from interfaces.inc multiple times :
- From interfaces_configures() (once)
- From vpn_ipsec_configure() (up to twice)
When create the VTI interface we destroy it beforehand if it exists AND the system is not booting. That results in interface creation attempts when it already exists. An error ensues :

`rc.bootup: The command '/sbin/ifconfig 'ipsec1000' create reqid '1000'' returned exit code '1', the output was 'ifconfig: create: bad value'`

And this command takes around 20s to return this error on our hardware which is a long time, especially during the booting process.

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1505,7 +1505,7 @@ function interface_ipsec_vti_configure($ph1ent) {
 			continue;
 		}
 		// Create IPsec interface
-		if (!platform_booting() && does_interface_exist($ipsecif)) {
+		if (does_interface_exist($ipsecif)) {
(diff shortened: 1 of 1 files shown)


jim-p:
This needs a fair amount of testing before merging. I distinctly remember the boot time check being added to prevent another issue, but it's been a while and the git history didn't offer any clues since it's been there since VTI was originally imported.

rbgarga (approved):
The change makes sense to me.  I just wonder why it's destroying and re-creating the same interface during boot. It should just create it once

PR #4428, opened 2020-08-21
prevent embedded button icons from interfering with click events #10846

Marc05:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10846
- [ ] Ready for review


Marc05:
I am able to apply the patch and get it working. I tried both on Firefox 79 and Edge (non-chromium).

I'm curious why it didn't work for you - perhaps cache? Maybe test in incognito after applying the patch.

PR #4429, opened 2020-08-21
Fix array pass by reference in ipsec.auth-user.php

basicmonkey:
Fixes error: PHP Fatal error:  Uncaught Error: Cannot pass parameter 3 by reference in /etc/inc/ipsec.auth-user.php:95 on IPSec Xauth.

My first ever pull request, hopefully all is correct!


jim-p:
Redmine: https://redmine.pfsense.org/issues/10847

PR #4430, opened 2020-08-31
pfSsh svc script fix. Issue #10855

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10855
- [X] Ready for review

Check for `usage()` to allow use `svc` script multiple times by `pfSsh.php`


jim-p:
Seems unnecessary, there are better ways to run multiple playback scripts that are less likely to break.

PR #4436, opened 2020-09-06
Custom IPSEC NAT-T ports. Feature #10870.

f-bor:
This pull request allows to change the client and server ports used by IPSEC NAT-T.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/10870
- [x] Ready for review

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -4383,22 +4383,23 @@ function filter_generate_ipsec_rules($log = array()) {
 			$shorttunneldescr = substr($descr, 0, 35);
 			// don't add "pass out" rules where $rgip is any as it will over-match and often break VPN clients behind the system in multi-WAN scenarios. redmine #5819
 			if ($rgip != " any ") {
-				$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = 500 tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound isakmp\"\n";
+				$ike_out = isset($ph1ent['ikeport']) ? $ph1ent['ikeport'] : 500;
+				$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = {$ike_out} tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound isakmp\"\n";
 			}
+			$ike_in = isset($config['ipsec']['port']) ? $config['ipsec']['port'] : 500;
 			$ipfrules .= <<<EOD
-pass in {$log['pass']} on \${$FilterIflist[$parentinterface]['descr']} $reply_to proto udp from {$rgip} to (self) port = 500 tracker {$increment_tracker($tracker)} keep state label "IPsec: {$shorttunneldescr} - inbound isakmp"
+pass in {$log['pass']} on \${$FilterIflist[$parentinterface]['descr']} $reply_to proto udp from {$rgip} to (self) port = {$ike_in} tracker {$increment_tracker($tracker)} keep state label "IPsec: {$shorttunneldescr} - inbound isakmp"
 
 EOD;
-			/* If NAT-T is enabled, add additional rules */
-			if ($ph1ent['nat_traversal'] != "off") {
-				if ($rgip != " any ") {
-					$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = 4500 tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound nat-t\"\n";
-				}
-				$ipfrules .= <<<EOD
-pass in {$log['pass']} on \${$FilterIflist[$parentinterface]['descr']} $reply_to proto udp from {$rgip} to (self) port = 4500 tracker {$increment_tracker($tracker)} keep state label "IPsec: {$shorttunneldescr} - inbound nat-t"
-
-EOD;
+			if ($rgip != " any ") {
+				$natt_out = isset($ph1ent['nattport']) ? $ph1ent['nattport'] : 4500;
+				$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = {$natt_out} tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound nat-t\"\n";
 			}
+			$natt_in = isset($config['ipsec']['port_nat_t']) ? $config['ipsec']['port_nat_t'] : 4500;
+			$ipfrules .= <<<EOD
+pass in {$log['pass']} on \${$FilterIflist[$parentinterface]['descr']} $reply_to proto udp from {$rgip} to (self) port = {$natt_in} tracker {$increment_tracker($tracker)} keep state label "IPsec: {$shorttunneldescr} - inbound nat-t"
+
+EOD;
 			/* Add rules to allow the protocols in use */
 			if ($prot_used_esp) {
 				if ($rgip != " any ") {
--- a/src/etc/inc/ipsec.inc
+++ b/src/etc/inc/ipsec.inc
@@ -1379,6 +1379,14 @@ function ipsec_setup_strongswan() {
 		$ssconf['charon']['make_before_break'] = "yes";
 	}
 
+	if (isset($config['ipsec']['port'])) {
+		$ssconf['charon']['port'] = $config['ipsec']['port'];
+	}
+
+	if (isset($config['ipsec']['port_nat_t'])) {
+		$ssconf['charon']['port_nat_t'] = $config['ipsec']['port_nat_t'];
+	}
+
 	$ssconf['charon']['syslog'] = array();
 	$ssconf['charon']['syslog']['identifier'] = 'charon';
 	$ssconf['charon']['syslog'][] = "# log everything under daemon since it ends up in the same place regardless with our syslog.conf";
@@ -2130,6 +2138,17 @@ function ipsec_setup_tunnels() {
 		/* TFC Padding */
 		if (isset($ph1ent['tfc_enable'])) {
 			$conn['tfc_padding'] = (isset($ph1ent['tfc_bytes']) && is_numericint($ph1ent['tfc_bytes'])) ? $ph1ent['tfc_bytes'] : "mtu";
+		}
+
+		/* Custom Ports */
+		if (isset($ph1ent['ikeport'])) {
+			/* For a connection without encapsulation, do not set local_port */
+			$conn['remote_port'] = $ph1ent['ikeport'];
+		} elseif (isset($ph1ent['nattport'])) {
+			/* For an encapsulated connection,
+			 * set local_port to the server NAT-T port value or default (4500) */
+			$conn['local_port'] = isset($config['ipsec']['port_nat_t']) ? $config['ipsec']['port_nat_t'] : 4500;
+			$conn['remote_port'] = $ph1ent['nattport'];
 		}
 
 		/* Arrays for P2s/children */
--- a/src/usr/local/www/vpn_ipsec_phase1.php
+++ b/src/usr/local/www/vpn_ipsec_phase1.php
@@ -83,6 +83,8 @@ if (isset($p1index) && $a_phase1[$p1index]) {
 		$pconfig['mobile'] = 'true';
 	} else {
 		$pconfig['remotegw'] = $a_phase1[$p1index]['remote-gateway'];
+		$pconfig['ikeport'] = $a_phase1[$p1index]['ikeport'];
+		$pconfig['nattport'] = $a_phase1[$p1index]['nattport'];
 	}
 
 	if (empty($a_phase1[$p1index]['iketype'])) {
@@ -293,6 +295,26 @@ if ($_POST['save']) {
 		}
 	}
 
+	if ($_POST['ikeport']) {
+		if (!is_port($pconfig['ikeport'])) {
+			$input_errors[] = gettext("The IKE port number is invalid.");
+		}
+	} else {
+		unset($pconfig['ikeport']);
+	}
+
+	if ($_POST['nattport']) {
+		if (!is_port($pconfig['nattport'])) {
+			$input_errors[] = gettext("The NAT-T port number is invalid.");
+		}
+	} else {
+		unset($pconfig['nattport']);
+	}
+
+	if (isset($pconfig['ikeport']) && isset($pconfig['nattport']) && $pconfig['ikeport'] == $pconfig['nattport']) {
+		$input_errors[] = gettext("IKE and NAT-T port numbers must be different.");
+	}
+
 	if ($pconfig['remotegw'] && is_ipaddr($pconfig['remotegw']) && !isset($pconfig['disabled'])) {
 		$t = 0;
 		foreach ($a_phase1 as $ph1tmp) {
@@ -501,6 +523,16 @@ if ($_POST['save']) {
 			$ph1ent['mobile'] = true;
 		} else {
 			$ph1ent['remote-gateway'] = $pconfig['remotegw'];
+			if ( !empty($pconfig['ikeport']) ) {
+				$ph1ent['ikeport'] = $pconfig['ikeport'];
+			} else {
+				unset($ph1ent['ikeport']);
+			}
+			if ( !empty($pconfig['nattport']) ) {
+				$ph1ent['nattport'] = $pconfig['nattport'];
+			} else {
+				unset($ph1ent['nattport']);
+			}
 		}
 
 		$ph1ent['protocol'] = $pconfig['protocol'];
@@ -739,9 +771,11 @@ $section->addInput(new Form_Select(
 ))->setHelp('Select the interface for the local endpoint of this phase1 entry.');
 
 if (!$pconfig['mobile']) {
-	$section->addInput(new Form_Input(
+	$group = new Form_Group('*Remote Gateway');
+
+	$group->add(new Form_Input(
 		'remotegw',
-		'*Remote Gateway',
+		'Remote Gateway',
 		'text',
 		$pconfig['remotegw']
 	))->setHelp('Enter the public IP address or host name of the remote gateway.%1$s%2$s%3$s',
@@ -750,6 +784,25 @@ if (!$pconfig['mobile']) {
 	    'to allow connections from any IPv6 address.' . '<br/>' . 'Responder Only must be set and ' . 
 	    'Peer IP Address cannot be used for Remote Identifier.'), 'info', false),
 	    '</div>');
+	$group->add(new Form_Input(
+	    'ikeport',
+	    'Remote IKE Port',
+	    'number',
+	    $pconfig['ikeport'],
+	    ['min' => 1, 'max' => 65535]
+	))->setHelp('UDP port for IKE on the remote gateway. Leave empty for default automatic behavior (500/4500).');
+	$group->add(new Form_Input(
+	    'nattport',
+	    'Remote NAT-T Port',
+	    'number',
+	    $pconfig['nattport'],
+	    ['min' => 1, 'max' => 65535]
+	))->setHelp('UDP port for NAT-T on the remote gateway.%1$s%2$s%3$s',
+	    '<div class="infoblock">',
+	    sprint_info_box(gettext('If the IKE port is empty and NAT-T contains a value, the tunnel will use only NAT-T.'),
+	    'info', false),
+	    '</div>');
+	$section->add($group);
 }
 
 $section->addInput(new Form_Input(
--- a/src/usr/local/www/vpn_ipsec_settings.php
+++ b/src/usr/local/www/vpn_ipsec_settings.php
@@ -48,6 +48,8 @@ $pconfig['maxmss'] = $config['system']['maxmss'];
 $pconfig['uniqueids'] = $config['ipsec']['uniqueids'];
 $pconfig['ipsecbypass'] = isset($config['ipsec']['ipsecbypass']);
 $pconfig['bypassrules'] = $config['ipsec']['bypassrules'];
+$pconfig['port'] = $config['ipsec']['port'];
+$pconfig['port_nat_t'] = $config['ipsec']['port_nat_t'];
 
 if ($_POST['save']) {
 	unset($input_errors);
@@ -91,6 +93,26 @@ if ($_POST['save']) {
 			$bypassrules['rule'][] = array('source' => $_POST["source{$x}"], 'srcmask' => $_POST["srcmask{$x}"],
 						'destination' => $_POST["destination{$x}"], 'dstmask' => $_POST["dstmask{$x}"]);
 		}
+	}
+
+	if ($_POST['port']) {
+		if (!is_port($pconfig['port'])) {
+			$input_errors[] = gettext("The IKE port number is invalid.");
+		}
+	} else {
+		unset($pconfig['port']);
+	}
+
+	if ($_POST['port_nat_t']) {
+		if (!is_port($pconfig['port_nat_t'])) {
+			$input_errors[] = gettext("The NAT-T port number is invalid.");
+		}
+	} else {
+		unset($pconfig['port_nat_t']);
+	}
+
+	if (isset($pconfig['port']) && isset($pconfig['port_nat_t']) && $pconfig['port'] == $pconfig['port_nat_t']) {
+		$input_errors[] = gettext("IKE and NAT-T port numbers must be different.");
 	}
 
 	$pconfig['bypassrules'] = $bypassrules;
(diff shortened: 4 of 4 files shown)


src/etc/inc/filter.inc
```diff
@@ -4392,10 +4392,18 @@ EOD;
 			/* If NAT-T is enabled, add additional rules */
 			if ($ph1ent['nat_traversal'] != "off") {
 				if ($rgip != " any ") {
-					$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = 4500 tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound nat-t\"\n";
+					$ikeport = 4500;
+					if ( isset($ph1ent['ikeport']) ) {
+						$ikeport = $ph1ent['ikeport'];
+					}
```
jim-p, line 4398: This could be simplified:

```
$ikeport = isset($ph1ent['ikeport']) ? $ph1ent['ikeport'] : 4500;
```

src/etc/inc/filter.inc
```diff
@@ -4392,10 +4392,18 @@ EOD;
...
+					$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = $ikeport tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound nat-t\"\n";
+				}
+				$port_nat_t = 4500;
+				if ( isset($config['ipsec']['port_nat_t']) ) {
+					$port_nat_t = $config['ipsec']['port_nat_t'];
 				}
```
jim-p, line 4404: See simplification note above.

src/etc/inc/ipsec.inc
```diff
@@ -2132,6 +2136,17 @@ function ipsec_setup_tunnels() {
...
 		}
 
+		/* Custom Port */
+		if (isset($ph1ent['ikeport'])) {
+			if (isset($config['ipsec']['port_nat_t'])) {
+				$conn['leftikeport'] = $config['ipsec']['port_nat_t'];
+			}
+			else {
+				$conn['leftikeport'] = 4500;
+			}
```
jim-p, line 2146: Simplify similar to above comments:
```
$conn['leftikeport'] = isset($config['ipsec']['port_nat_t']) ? $config['ipsec']['port_nat_t'] : 4500;
```

jim-p: On master pfSense is using swanctl.conf format, not ipsec.conf, so the directive is `local_port`: https://wiki.strongswan.org/projects/strongswan/wiki/Fromipsecconf

src/etc/inc/ipsec.inc
```diff
@@ -2132,6 +2136,17 @@ function ipsec_setup_tunnels() {
...
+			$conn['rightikeport'] = $ph1ent['ikeport'];
```
jim-p, line 2147: On master pfSense is using swanctl.conf format, not ipsec.conf, so the directive is `remote_port`: https://wiki.strongswan.org/projects/strongswan/wiki/Fromipsecconf

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -750,6 +758,14 @@ function build_eal_list() {
 	    'to allow connections from any IPv6 address.' . '<br/>' . 'Responder Only must be set and ' . 
 	    'Peer IP Address cannot be used for Remote Identifier.'), 'info', false),
 	    '</div>');
+	$group->add(new Form_Input(
+    'ikeport',
+		'Remote NAT-T port',
```
jim-p, line 763: It looks like this is used any time the GUI value is set, so it's not really a "NAT-T" port, but a "Remote IKE Port".

jim-p (changes requested):
I noted a few problems and suggestions, but the main problem I see is that you are coding as if the config is the old ipsec.conf style and not the new swanctl.conf style, so it would generate an invalid configuration.

See https://wiki.strongswan.org/projects/strongswan/wiki/Fromipsecconf and https://wiki.strongswan.org/projects/strongswan/wiki/Swanctlconf

f-bor:
You're right, I wrote this for 2.4.4 and totally missed the new swanctl configuration when porting the changes to master.

I did all the changes requested.

Reading it again, I'm not happy with the end of the wording for phase 1 settings but I'm not sure what would be better "Enter a custom IKE port number **for IPsec**". For IPsec seems useless, maybe "for remote gateway", or maybe just remove it would be better ?

src/etc/inc/ipsec.inc
```diff
@@ -2132,6 +2136,12 @@ function ipsec_setup_tunnels() {
 			$conn['tfc_padding'] = (isset($ph1ent['tfc_bytes']) && is_numericint($ph1ent['tfc_bytes'])) ? $ph1ent['tfc_bytes'] : "mtu";
 		}
 
+		/* Custom Port */
+		if (isset($ph1ent['ikeport'])) {
+			$conn['local_port'] = isset($config['ipsec']['port_nat_t']) ? $config['ipsec']['port_nat_t'] : 4500;
```
jim-p, line 2141: Thinking about this line some more, should it really depend on `ikeport`? Someone may want to change the local port but not remote.

Perhaps instead it should be its own separate test that only sets `local_port` when `port_nat_t` is present in the config without any relationship to the `ikeport` setting

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +94,10 @@
 		}
 	}
 
+	if (isset($_POST['port_nat_t']) && $_POST['port_nat_t'] == '500' ) {
```
jim-p, line 97: This should test using `is_port()` to validate that what was submitted is a valid port value.

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -501,6 +502,11 @@
 			$ph1ent['mobile'] = true;
 		} else {
 			$ph1ent['remote-gateway'] = $pconfig['remotegw'];
+			if ( !empty($pconfig['ikeport']) && $pconfig['ikeport'] != '500' ) {
```
jim-p, line 505: Not exactly here, but up higher in the input validation, this should also test the value of `ikeport` using `is_port()` to ensure the user submitted a sane value.

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -750,6 +758,14 @@ function build_eal_list() {
 	    'to allow connections from any IPv6 address.' . '<br/>' . 'Responder Only must be set and ' . 
 	    'Peer IP Address cannot be used for Remote Identifier.'), 'info', false),
 	    '</div>');
+	$group->add(new Form_Input(
+    'ikeport',
+		'Remote IKE Port',
+		'number',
+		$pconfig['ikeport'],
+		['min' => 1, 'max' => 65535]
+	))->setHelp('Enter a custom IKE port number for IPsec (leave empty for 500/4500 default scheme)');
```
jim-p, line 767: This could be: `UDP port for IKE on the remote gateway. Leave empty for default automatic behavior (500/4500)`

src/etc/inc/ipsec.inc
jim-p: And this may need tested but do you even need to set `local_port` if you changed the value of `port_nat_t` in charon? I could be misreading the docs, since they are ambiguous, but it sounds like if you change `port_nat_t` then that would change the port automatically used by strongSwan when it selects NAT-T, so `local_port` could stay unset as it would be picked automatically.

But that needs testing to confirm.

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -393,6 +405,14 @@ function maxmss_checked(obj) {
 ))->setHelp('Allow crypto(9) jobs to be dispatched multi-threaded to increase performance. ' .
 		'Jobs are handled in the order they are received so that packets will be reinjected in the correct order.');
 
+$section->addInput(new Form_Input(
+  'port_nat_t',
+	'NAT-T port',
+	'number',
+	$pconfig['port_nat_t'],
+	['min' => 1, 'max' => 65535]
+))->setHelp('Enter a custom port number for IPsec server NAT-T.');
```
jim-p, line 414: This would be better: `Local UDP port for NAT-T (Default: 4500)`

jim-p (changes requested):

src/etc/inc/ipsec.inc
f-bor: About setting local_port when remote_port is set, I did tests with the old ipsec.conf and it was required, at least to have a working client role. I will do some tests again with various configurations and let you know the results.

f-bor: I have set up a server listening on port 55555 (port_nat_t=55555).

Then, I tested the following configurations on client, using old ipsec.conf, with the associated results (l=local_port (or leftikeport), r=remote_port, n=port_nat_t):
- (l=unset, r=55555, n=unset) => KO, ike init sent but not accepted, client local=500 (isakmp: parent_sa ikev2_init[I])
- (l=4500, r=55555, n=unset) => working (NONESP-encap: isakmp: parent_sa ikev2_init[I])
- (l=44444, r=55555, n=unset) => KO, no init sent
- (l=0, r=55555, n=unset) => KO, something sent, client local = 500 (isakmp:)

on all, when replying the server was using source port 55555.

Then, I have set up the port_nat_t on the client-side and had the following results:

- (l=unset, r=55555, n=44444) => KO, ike init sent but not accepted, client local=500 (isakmp: parent_sa ikev2_init[I])
- (l=4500, r=55555, n=44444) => KO, no init sent
- (l=44444, r=55555, n=44444) => working (NONESP-encap: isakmp: parent_sa ikev2_init[I])
- (l=0, r=55555, n=44444) => KO, something sent, client local = 500 (isakmp:)

So imo, the conclusion is that the local port has to be set to the same value as the port_nat_t when using remote_port (or 4500 if port_nat_t is not explicitly set).

f-bor: I had a second thought on this.

When the local_port is not set, the client is connecting on the ike port, as set in remote_port.
When the local_port is set (properly), the client is connecting on the nat-t port, as set in remote_port.

So it seems to me the problem is:
- one can also want to change the ike port (on the server side)
- one can also want to connect on a non-standard ike port (from the client side)

This PR was only about the nat-t/encaps port but some more changes to handle the ike port too would make sense. In that case, I would see two ports with the remote gateway address on phase 1, one for ike, one for natt/encaps, but the user could only fill one, depending on whether he wants to connect to one or the other.

Does that sound good to you ?

f-bor: > one for ike, one for natt/encaps, but the user could only fill one, depending on whether he wants to connect to one or the other

Thinking again: both ports may be needed to have correct route-to rules.

So two ports that can be set, with priority on ike:
- if ike port is set, connect on remote ike port, regardless to nat-t port
- if nat-t port is set, connect on remote nat-t port, but only if ike port is not set

f-bor:
With the last commit, there is now the possibility to set a custom IKE port too, for client and server.

Regarding the client:
- if the remote IKE port is set, it is used first to try an unencapsulated connection
- if the remote NAT-T port is also set, the NAT-T value is only used for setting the correct route-to rule (since that port is negotiated with the server)
- if the remote NAT-T port is set but not the IKE one, only the NAT-T port is used to directly do an encapsulated connection

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
...
+	if ($_POST['port_nat_t']) {
+		if (!is_port($pconfig['port_nat_t']))
+			$input_errors[] = gettext("An invalid NAT-T port number has been specified.");
+	}
+	else {
+		unset($pconfig['port_nat_t']);
+	}
+
+	if (isset($pconfig['port']) && isset($pconfig['port_nat_t']) && $pconfig['port'] == $pconfig['port_nat_t']) {
+		$input_errors[] = gettext("IKE and NAT-T port numbers must be differents.");
```
jim-p, line 115: Typo: "differents" should be "different"

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
 		}
 	}
 
+	if ($_POST['port']) {
+		if (!is_port($pconfig['port']))
+			$input_errors[] = gettext("An invalid IKE port number has been specified.");
```
jim-p, line 100: Reword to avoid the passive voice, e.g. "The IKE port number is invalid"

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
...
+	}
+	else {
+		unset($pconfig['port']);
+	}
+
+	if ($_POST['port_nat_t']) {
+		if (!is_port($pconfig['port_nat_t']))
+			$input_errors[] = gettext("An invalid NAT-T port number has been specified.");
```
jim-p, line 108: Reword to avoid the passive voice, e.g. "The NAT-T port number is invalid"

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -750,6 +784,27 @@ function build_eal_list() {
...
+	))->setHelp('UDP port for IKE on the remote gateway. Leave empty for default automatic behavior (500/4500).');
+	$group->add(new Form_Input(
+	    'nattport',
+	    'Remote NAT-T Port',
+	    'number',
+	    $pconfig['nattport'],
+	    ['min' => 1, 'max' => 65535]
+	))->setHelp('UDP port for NAT-T on the remote gateway.%1$s%2$s%3$s',
+	    '<div class="infoblock">',
+	    sprint_info_box(gettext('Leave IKE port empty and set NAT-T port if you wish to directly ' .
```
jim-p, line 802: Reword to avoid directly addressing the user ("you") e,.g, "If the IKE port is empty and NAT-T contains a value, the tunnel will use only NAT-T"

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
...
+	if ($_POST['nattport']) {
+		if (!is_port($pconfig['nattport']))
+			$input_errors[] = gettext("An invalid NAT-T port number has been specified.");
+	}
+	else {
+		unset($pconfig['nattport']);
+	}
+
+	if (isset($pconfig['ikeport']) && isset($pconfig['nattport']) && $pconfig['ikeport'] == $pconfig['nattport']) {
+		$input_errors[] = gettext("IKE and NAT-T port numbers must be differents.");
```
jim-p, line 315: Typo: "differents" should be "different"

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
...
+		if (!is_port($pconfig['ikeport']))
+			$input_errors[] = gettext("An invalid IKE port number has been specified.");
+	}
+	else {
+		unset($pconfig['ikeport']);
+	}
+
+	if ($_POST['nattport']) {
+		if (!is_port($pconfig['nattport']))
+			$input_errors[] = gettext("An invalid NAT-T port number has been specified.");
```
jim-p, line 308: Reword to avoid the passive voice, e.g. "The NAT-T port number is invalid"

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
 		}
 	}
 
+	if ($_POST['ikeport']) {
+		if (!is_port($pconfig['ikeport']))
+			$input_errors[] = gettext("An invalid IKE port number has been specified.");
```
jim-p, line 300: Reword to avoid the passive voice, e.g. "The IKE port number is invalid"

src/etc/inc/filter.inc
```diff
@@ -4382,27 +4382,37 @@ function filter_generate_ipsec_rules($log = array()) {
...
 EOD;
 			}
 			/* Add rules to allow the protocols in use */
 			if ($prot_used_esp) {
 				if ($rgip != " any ") {
-					$ipfrules .= "pass out {$log['pass']} $route_to proto esp from (self) to {$rgip} tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound esp proto\"\n";
+					// don't add "pass out" if NAT-T is forced
+					if ( isset($ph1ent['ikeport']) || !isset($ph1ent['nattport']) ) {
+						$ipfrules .= "pass out {$log['pass']} $route_to proto esp from (self) to {$rgip} tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound esp proto\"\n";
+					}
```
jim-p, line 4415: IKEv1 tunnels can choose to force NAT-T via the nat_traversal option (set to 'force') as well which should also be accounted for here. Though I'm not sure it's worth the effort trying to nail down all the logic just to exclude this ESP rule.

src/etc/inc/filter.inc
```diff
@@ -4382,27 +4382,37 @@ function filter_generate_ipsec_rules($log = array()) {
...
+				}
 			}
+			$ike_in = isset($config['ipsec']['port']) ? $config['ipsec']['port'] : 500;
 			$ipfrules .= <<<EOD
-pass in {$log['pass']} on \${$FilterIflist[$parentinterface]['descr']} $reply_to proto udp from {$rgip} to (self) port = 500 tracker {$increment_tracker($tracker)} keep state label "IPsec: {$shorttunneldescr} - inbound isakmp"
+pass in {$log['pass']} on \${$FilterIflist[$parentinterface]['descr']} $reply_to proto udp from {$rgip} to (self) port = {$ike_in} tracker {$increment_tracker($tracker)} keep state label "IPsec: {$shorttunneldescr} - inbound isakmp"
 
 EOD;
 			/* If NAT-T is enabled, add additional rules */
 			if ($ph1ent['nat_traversal'] != "off") {
```
jim-p, line 4398: While you're here, the `nat_traversal` option has not had an `off` value in ~6yrs. Only valid values are `auto` and `force`, so this test seems useless.

jim-p (changes requested):

src/etc/inc/filter.inc
f-bor: It seems cleaner to have a consistent behavior on "generating or not the rule", so since I was there too and already did half the test, I have added the one on encapsulation not being forced.

src/etc/inc/filter.inc
```diff
@@ -4382,27 +4382,34 @@ function filter_generate_ipsec_rules($log = array()) {
 			/* Add rules to allow IKE to pass */
 			$shorttunneldescr = substr($descr, 0, 35);
 			// don't add "pass out" rules where $rgip is any as it will over-match and often break VPN clients behind the system in multi-WAN scenarios. redmine #5819
-			if ($rgip != " any ") {
-				$ipfrules .= "pass out {$log['pass']} $route_to proto udp from (self) to {$rgip} port = 500 tracker {$increment_tracker($tracker)} keep state label \"IPsec: {$shorttunneldescr} - outbound isakmp\"\n";
+			if ($rgip != " any " ) {
```
rbgarga, line 4385: Remove extra space added before `)`

src/etc/inc/ipsec.inc
```diff
@@ -2132,6 +2140,18 @@ function ipsec_setup_tunnels() {
 			$conn['tfc_padding'] = (isset($ph1ent['tfc_bytes']) && is_numericint($ph1ent['tfc_bytes'])) ? $ph1ent['tfc_bytes'] : "mtu";
 		}
 
+		/* Custom Ports */
+		if (isset($ph1ent['ikeport'])) {
+			/* we want to initiate an ike connection without encaps, so we don't set local_port */
+			$conn['remote_port'] = $ph1ent['ikeport'];
+		}
+		else if (isset($ph1ent['nattport'])) {
```
rbgarga, line 2148: } and else must be in the same line

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
 		}
 	}
 
+	if ($_POST['ikeport']) {
+		if (!is_port($pconfig['ikeport']))
+			$input_errors[] = gettext("The IKE port number is invalid.");
+	}
+	else {
```
rbgarga, line 302: } and else must be in the same line

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
 		}
 	}
 
+	if ($_POST['ikeport']) {
+		if (!is_port($pconfig['ikeport']))
+			$input_errors[] = gettext("The IKE port number is invalid.");
```
rbgarga, line 300: by pfSense style, always add { and } even if it's a single statement

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
...
+	}
+	else {
+		unset($pconfig['ikeport']);
+	}
+
+	if ($_POST['nattport']) {
+		if (!is_port($pconfig['nattport']))
+			$input_errors[] = gettext("The NAT-T port number is invalid.");
+	}
+	else {
```
rbgarga, line 310: } and else must be in the same line

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -293,6 +295,26 @@
...
+		if (!is_port($pconfig['ikeport']))
+			$input_errors[] = gettext("The IKE port number is invalid.");
+	}
+	else {
+		unset($pconfig['ikeport']);
+	}
+
+	if ($_POST['nattport']) {
+		if (!is_port($pconfig['nattport']))
+			$input_errors[] = gettext("The NAT-T port number is invalid.");
```
rbgarga, line 308: by pfSense style, always add { and } even if it's a single statement

src/usr/local/www/vpn_ipsec_phase1.php
```diff
@@ -739,9 +771,11 @@ function build_eal_list() {
 ))->setHelp('Select the interface for the local endpoint of this phase1 entry.');
 
 if (!$pconfig['mobile']) {
-	$section->addInput(new Form_Input(
+	$group = new Form_Group('*Remote Gateway');
+
+	$group->add(new Form_Input(
 		'remotegw',
-		'*Remote Gateway',
+		'Remote Gateway',
```
rbgarga, line 778: Remote Gateway is a mandatory config item, please add `*` back

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
 		}
 	}
 
+	if ($_POST['port']) {
+		if (!is_port($pconfig['port']))
+			$input_errors[] = gettext("The IKE port number is invalid.");
+	}
+	else {
```
rbgarga, line 102: } and else must be in the same line

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
...
+		unset($pconfig['port']);
+	}
+
+	if ($_POST['port_nat_t']) {
+		if (!is_port($pconfig['port_nat_t']))
+			$input_errors[] = gettext("The NAT-T port number is invalid.");
+	}
+	else {
```
rbgarga, line 110: } and else must be in the same line

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
 		}
 	}
 
+	if ($_POST['port']) {
+		if (!is_port($pconfig['port']))
+			$input_errors[] = gettext("The IKE port number is invalid.");
```
rbgarga, line 100: by pfSense style, always add { and } even if it's a single statement

src/usr/local/www/vpn_ipsec_settings.php
```diff
@@ -93,6 +95,26 @@
...
+	}
+	else {
+		unset($pconfig['port']);
+	}
+
+	if ($_POST['port_nat_t']) {
+		if (!is_port($pconfig['port_nat_t']))
+			$input_errors[] = gettext("The NAT-T port number is invalid.");
```
rbgarga, line 108: by pfSense style, always add { and } even if it's a single statement

rbgarga (changes requested):

src/etc/inc/ipsec.inc
jim-p: Also, use `elseif`, not `else if`: https://docs.netgate.com/pfsense/en/latest/development/developer-style-guide.html#general-rules

src/usr/local/www/vpn_ipsec_phase1.php
jim-p: This looks OK, the `*` is now on the group (Right side, line 774), not just this one option.

jim-p (changes requested):

src/etc/inc/filter.inc
```diff
@@ -4382,27 +4382,34 @@ function filter_generate_ipsec_rules($log = array()) {
...
+				// don't add "pass out" if NAT-T is forced
+				if ( isset($ph1ent['ikeport']) || !isset($ph1ent['nattport']) ) {
```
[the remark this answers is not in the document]
f-bor: I will remove this test as well as the one for the esp but this is not exactly the same kind of forcing. When forcing with the port fields, the port 500 won't be used at all by the tunnel, so the rule is not needed. When forcing with the NAT Traversal option, the nat detection is faked but the port 500 is still used to initiate the connection. That's why I didn't add the test on nat_traversal there.

Since I'm here, I have a question about the "resolve conversation" buttons: who should mark each item resolved ? me or the reviewer ?

I also took a good look at the style guide and developer style guide (thx), I'll try to do a better job next time.

src/usr/local/www/vpn_ipsec_phase1.php
rbgarga: Oh, right.  I've missed that

rbgarga (approved):

jim-p (approved):

PR #4431, opened 2020-08-31
Backup/restore Captive Portal UsedMACs. Implements #10856

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10856
- [X] Ready for review

Backup/restore `/var/db/captiveportal_usedmacs_<cpzone>.db` files used by "Pass-through credits per MAC address" feature

PR #4439, opened 2020-09-12
Backup Captive Portal DB files. Implements #10868

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10868
- [X] Ready for review

Backup/Restore Captive Portal DB files `/var/db/captiveportal<zone>.db` to keep connected users after restore (if "Preserve connected users across reboot" option is enabled)

improved #4431

### DIFF
--- a/src/etc/inc/web/backup.inc
+++ b/src/etc/inc/web/backup.inc
@@ -121,39 +121,35 @@ function restore_rrddata() {
 	}
 }
 
-function restore_captiveportal_data($type='voucher') {
+function restore_xmldatafile($type='voucher') {
 	global $config, $g;
-	if ($type == 'captiveportal_usedmacs') {
-		$conf = 'captiveportal';
-	} else {
-		$conf = $type;
-	}
-	foreach ($config[$conf]["{$type}data"]["{$type}dbfile"] as $db) {
-		$cpdata_file = "{$g['vardb_path']}/{$db['filename']}";
-		if (file_put_contents($cpdata_file, gzinflate(base64_decode($db['dbdata']))) === false) {
-			log_error(sprintf(gettext("Cannot write %s"), $cpdata_file));
+	foreach ($config[$type]["{$type}data"]["xmldatafile"] as $file) {
+		$basename = basename($file['filename']);
+		$xmldata_file = "{$g['vardb_path']}/{$basename}";
+		if (file_put_contents($xmldata_file, gzinflate(base64_decode($file['data']))) === false) {
+			log_error(sprintf(gettext("Cannot write %s"), $xmldata_file));
 			continue;
 		}
 	}
 }
 
-function captiveportal_data_xml($tab=false, $type='voucher') {
+function backup_xmldatafile($tab=false, $type='voucher') {
 	global $g;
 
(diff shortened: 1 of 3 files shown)


vktg:
generic function `backup_xmldatafile` and `restore_xmldatafile` are used
they can be used in the future to backup/restore any files (like dhcp leases)

src/usr/local/www/diag_backup.php
```diff
@@ -121,26 +121,21 @@ function restore_rrddata() {
...
-		$conf = 'captiveportal';
-	} else {
-		$conf = $type;
-	}
-	foreach ($config[$conf]["{$type}data"]["{$type}dbfile"] as $db) {
-		$cpdata_file = "{$g['vardb_path']}/{$db['filename']}";
-		if (file_put_contents($cpdata_file, gzinflate(base64_decode($db['dbdata']))) === false) {
-			log_error(sprintf(gettext("Cannot write %s"), $cpdata_file));
+	foreach ($config[$type]["{$type}data"]["xmldatafile"] as $file) {
+		$xmldata_file = "{$g['vardb_path']}/{$file['filename']}";
```
jim-p, line 127: For safety, this should use `basename()` on the contents of the filename to ensure it doesn't write to the wrong place accidentally (e.g. someone manually edits their config wrong). Otherwise it could have relative path references, subdirectories, etc, which doesn't appear to be necessary or intended.

jim-p (changes requested):

src/usr/local/www/diag_backup.php
vktg: Thanks for info, fixed

jim-p (approved):

rbgarga:

vktg:
Fixed

rbgarga (changes requested):
@vktg I noted you are changing back lots of places from using local `$post` variable to original `$_POST`.  Please review them

src/etc/inc/web/backup.inc
```diff
@@ -214,17 +210,17 @@ function execPost($post, $postfiles, $json = true) {
...
+						$name = "{$_POST['backuparea']}-{$name}";
 					} else if ($post['backuparea'] === "voucher") {
-						$data = backup_config_section($post['backuparea']);
-						$voucher_data_xml = captiveportal_data_xml(false, 'voucher');
+						$data = backup_config_section($_POST['backuparea']);
+						$voucher_data_xml = backup_xmldatafile(false, 'voucher');
 						$closing_tag = "</voucher>";
 						$data = str_replace($closing_tag, $voucher_data_xml . $closing_tag, $data);
-						$name = "{$post['backuparea']}-{$name}";
+						$name = "{$_POST['backuparea']}-{$name}";
```
sbeaver-netgate, line 223: Please use the $post array passed to this function. Do not assume that $_POST is valid. There are some circumstances where it is not. The same is true for $_FILES.

rbgarga (approved):

src/etc/inc/web/backup.inc
vktg: Ok, fixed

sbeaver-netgate (approved):

PR #4442, opened 2020-09-12
DHCPv6 Static Mapping fix. Issue #10882

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10882
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/services_dhcpv6.php
+++ b/src/usr/local/www/services_dhcpv6.php
@@ -37,6 +37,8 @@ require_once("guiconfig.inc");
 require_once("filter.inc");
 
 function dhcpv6_apply_changes($dhcpdv6_enable_changed) {
+	global $config, $g;
+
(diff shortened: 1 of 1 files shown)


src/usr/local/www/services_dhcpv6.php
```diff
@@ -37,6 +37,8 @@
 require_once("filter.inc");
 
 function dhcpv6_apply_changes($dhcpdv6_enable_changed) {
+	global $config;
```
rbgarga, line 40: It also uses global `$g` at line 46

rbgarga (changes requested):

src/usr/local/www/services_dhcpv6.php
vktg: Added

jim-p (approved):

PR #4443, opened 2020-09-13
DNS Resolver multi IP for host overrides. Implements #10896

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10896
- [X] Ready for review

Allow to use comma-separated list of IPs for host overrides


rbgarga (approved):

jim-p (approved):
Seems OK but this used to work by adding more entries with the same hostname+domain. At some point the input validation started rejecting that. I can see benefits of doing it either way, though.

PR #4445, opened 2020-09-13
Add vxlans

stephenw10:
Exclude vxlan interfaces from the interface mismatch check at boot.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/10898
- [x] Ready for review

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1342,12 +1342,14 @@ function interface_vxlan_configure(&$vxlan, $vxlankey = "") {
 		log_error(gettext("could not bring parentif up -- variable not defined -- interface_vxlan_configure()"));
 	}
 
-	if (platform_booting() || !(empty($vxlan['vxlanif']))) {
-		pfSense_interface_destroy($vxlan['vxlanif']);
-		pfSense_interface_create($vxlan['vxlanif']);
+	if (!(empty($vxlan['vxlanif']))) {				//Interface us defined.
+		if (!platform_booting()) {
+			pfSense_interface_destroy($vxlan['vxlanif']);	//Calling this after boot implies we are editing an existing interface so destroy it first.
+		}
+		pfSense_interface_create($vxlan['vxlanif']);		//Create the interfaces at boot.
 		$vxlanif = $vxlan['vxlanif'];
 	} else {
-		$vxlanif = pfSense_interface_create("vxlan");
+		$vxlanif = pfSense_interface_create("vxlan");		//The interface is undefined, we are creating a new one. Next interface name is returned.
 	}
 
 	$ifconfigcmd = "/sbin/ifconfig " . escapeshellarg($vxlanif) . " vxlanid " .
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1351,10 +1351,10 @@ function interface_vxlan_configure(&$vxlan, $vxlankey = "") {
 	}
 
 	$ifconfigcmd = "/sbin/ifconfig " . escapeshellarg($vxlanif) . " vxlanid " .
-	    escapeshellarg($vxlan['vxlanid']) . " vxlandev " . escapeshellarg($realif);
+	    escapeshellarg($vxlan['vxlanid']);
 
 	if (is_mcast($vxlan['remote-addr'])) {
-		$ifconfigcmd .= " vxlangroup " . escapeshellarg($vxlan['remote-addr']);
+		$ifconfigcmd .= " vxlangroup " . escapeshellarg($vxlan['remote-addr']) . " vxlandev " . escapeshellarg($realif);
 	} else {
 		$ifconfigcmd .= " vxlanremote " . escapeshellarg($vxlan['remote-addr']);
 	}
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -1342,7 +1342,7 @@ function interface_vxlan_configure(&$vxlan, $vxlankey = "") {
 		log_error(gettext("could not bring parentif up -- variable not defined -- interface_vxlan_configure()"));
 	}
 
-	if (!(empty($vxlan['vxlanif']))) {				//Interface us defined.
+	if (!(empty($vxlan['vxlanif']))) {				//Interface is defined.
 		if (!platform_booting()) {
 			pfSense_interface_destroy($vxlan['vxlanif']);	//Calling this after boot implies we are editing an existing interface so destroy it first.
 		}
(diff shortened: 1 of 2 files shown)


stephenw10:
Added fixes for: https://redmine.pfsense.org/issues/10899

Tested here seems to work expected. The interface is still brought up twice at boot which seems unnecessary but doesn't appear to hurt anything:
`Configuring WAN interface...
vtnet0: link state changed to UP
vxlan0: Ethernet address: 58:9c:fc:10:cb:30
vxlan0: link state changed to UP
done.
Configuring VXLAN0 interface...done.
vxlan0: link state changed to DOWN
vxlan0: Ethernet address: 58:9c:fc:10:cb:30
vxlan0: link state changed to UP
Configuring LAN interface...
vtnet1: link state changed to UP`

src/etc/inc/interfaces.inc
```diff
@@ -1342,7 +1342,7 @@ function interface_vxlan_configure(&$vxlan, $vxlankey = "") {
 		log_error(gettext("could not bring parentif up -- variable not defined -- interface_vxlan_configure()"));
 	}
 
-	if (!platform_booting() && !(empty($vxlan['vxlanif']))) {
+	if (platform_booting() || !(empty($vxlan['vxlanif']))) {
```
rbgarga, line 1345: I didn't understand the reason for this change.  Could you please elaborate a bit?

stephenw10: I confess the exact logic here eludes me! I tested it that was as that's how gif and gre interfaces are configured and that seems to work.
With the original logic the interface is created twice at boot with the same vxlanid and subsequent vxlans use sequential numbers rather that that defined in the config. Like:

`Configuring WAN interface...
vtnet0: link state changed to UP
vxlan0: Ethernet address: 58:9c:fc:10:cb:30
vxlan0: link state changed to UP
done.
Configuring VXLAN0 interface...done.
vxlan1: Ethernet address: 58:9c:fc:10:ff:9f
vxlan1: network identifier 12345 already exists in this socket
vxlan1: network identifier 12345 already exists in this socket
Configuring LAN interface...
vtnet1: link state changed to UP
done.
vxlan2: Ethernet address: 58:9c:fc:10:ff:99
vxlan2: link state changed to UP`

jim-p: Seems similar to what we were seeing with VTI ipsec interfaces, might need similar logic change to what was done there. Basically only destroy it if the interface already exists. And maybe also test if it exists before creating.

jim-p (changes requested):

src/etc/inc/interfaces.inc
rbgarga: @stephenw10 could you try to replace this change by:
```
if (does_interface_exist($vxlan['vxlanif'])) {
```
And see if it has the same effect?

stephenw10: It behaves differently but still incorrectly. The first interface is created at boot as expected though it is still brought UP then DOWN then back UP again.
The second interface is created as vxlan1 though the config is for vxlan3 resulting in a spurious interface in the GUI Interfaces > Assign.
```Configuring IPsec VTI interfaces...done.
Configuring WAN interface...
vtnet0: link state changed to UP
vxlan0: Ethernet address: 58:9c:fc:10:cb:30
vxlan0: link state changed to UP
done.
Configuring VXLAN0 interface...done.
vxlan0: link state changed to DOWN
vxlan0: Ethernet address: 58:9c:fc:10:cb:30
vxlan0: link state changed to UP
Configuring LAN interface...
vtnet1: link state changed to UP
done.
vxlan1: Ethernet address: 58:9c:fc:10:ff:9f
vxlan1: link state changed to UP```

From config:
```	<vxlans>
		<vxlan>
			<if>wan</if>
			<ipaddr></ipaddr>
			<remote-addr>172.21.16.187</remote-addr>
			<vxlanid>12345</vxlanid>
			<vxlanlocalport></vxlanlocalport>
			<vxlanremoteport></vxlanremoteport>
			<vxlanttl></vxlanttl>
			<vxlanlearn></vxlanlearn>
			<descr><![CDATA[Test VXLAN]]></descr>
			<vxlanif>vxlan0</vxlanif>
		</vxlan>
		<vxlan>
			<if>lan</if>
			<ipaddr></ipaddr>
			<remote-addr>192.168.180.56</remote-addr>
			<vxlanid>54322</vxlanid>
			<vxlanlocalport></vxlanlocalport>
			<vxlanremoteport></vxlanremoteport>
			<vxlanttl></vxlanttl>
			<vxlanlearn></vxlanlearn>
			<descr><![CDATA[Test VXLAN 3]]></descr>
			<vxlanif>vxlan3</vxlanif>
		</vxlan>
	</vxlans>```

**Comment by stephenw10** [comment]
Having reviewed this logic further I believe this is the correct way to create this. All the other sub-interface types are created like this except VTI.
The if state provides two ways to create the interface:
`		pfSense_interface_destroy($vxlan['vxlanif']);
		pfSense_interface_create($vxlan['vxlanif']);
		$vxlanif = $vxlan['vxlanif']`
or
`		$vxlanif = pfSense_interface_create("vxlan");`

There is a clue in what is intended in the bridge interface code. The second way, when the of statement does not match, should only ever be hit when creating a new vxlan at run time when the interface name is then returned by pfSense_interface_create as the next available vxlanX. At all other times it should use the first method. 
That does seem to mean it unnecessarily destroys a non-existent interface at boot but that doesn't seem to be a problem. We could add code to prevent that.

The interface coming going down and back up additionally I was seeing earlier appears to be because that was added on WAN and the WAN interface is brought up before vxlans are configured. vxlans on other parent interfaces do not show that behaviour.

**Review by jim-p — APPROVED** [review]

**Comment by stephenw10** [comment]
Yes in fact preventing the interfaces being destroyed at boot allows them to be configured fully which was not happening previously but I had not spotted. The description and MTU values were not being applied:
```
vxlan1: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1450
	description: VXLAN3
	options=80000<LINKSTATE>
	ether 58:9c:fc:10:ff:9f
	inet6 fe80::5a9c:fcff:fe10:ff9f%vxlan1 prefixlen 64 scopeid 0x8
	groups: vxlan
	vxlan vni 54321 local 172.21.16.167:4789 remote 172.21.16.222:4789
	media: Ethernet autoselect (autoselect <full-duplex>)
	status: active
	nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
```
There is still an issue here though. The interfaces on WAN do come up correctly using that code but errors are logged:
```
rc.bootup: The command '/sbin/ifconfig 'vxlan0' vxlanid '12345'  vxlanremote '172.21.16.187' vxlanlocal '172.21.16.167' vxlanlearn '  returned exit code '1', the output was 'ifconfig: VXLAN_CMD_SET_VNI:  Device busy'
```

rbgarga:
Manually merged.  Thanks!!!

PR #4446, opened 2020-09-14
Exclude VXLAN interface from check list. Fixes #10898

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10898
- [X] Ready for review

exclud vxlan from the list of interfaces to check like other sub-interface types


vktg:
Fix already exist: https://github.com/pfsense/pfsense/pull/4445

PR #4447, opened 2020-09-14
Pushover Notification Support. Implements #10495

overtninja:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10495
- [x] Ready for review

### DIFF
--- a/src/etc/inc/notices.inc
+++ b/src/etc/inc/notices.inc
@@ -521,7 +521,7 @@ function notify_via_pushover($message, $force = false) {
 		if ($result['status']) {
 			unset($err_msg);
 		} else {
-			$err_msg = sprintf(gettext("Failed to send Pushover notification. Error received was: {$result['errors']['0']}"));
+			$err_msg = sprintf(gettext("Failed to send Pushover notification. Error received was: %s"), $result['errors']['0']);
 			log_error($err_msg);
 		}
 	} else {
--- a/src/etc/inc/notices.inc
+++ b/src/etc/inc/notices.inc
@@ -40,7 +40,8 @@ $smtp_authentication_mechanisms = array(
 */
 global $pushover_sounds;
 $pushover_sounds = array(
-	'pushover' => 'Pushover (default)',
+	'devicedefault' => 'Device Default',
+	'pushover' => 'Pushover',
 	'bike' => 'Bike',
 	'bugle' => 'Bugle',
 	'cashregister' => 'Cash Register',
@@ -504,6 +505,9 @@ function notify_via_pushover($message, $force = false) {
 			return gettext("Unable to test Pushover notification without both API Key & User Key set");
 		}
 		return;
+	}
+	if (strcasecmp($config['notifications']['pushover']['sound'], 'devicedefault') == 0) {
+		unset($config['notifications']['pushover']['sound']);
 	}
 
 	$url = "https://api.pushover.net/1/messages.json";
--- a/src/etc/inc/notices.inc
+++ b/src/etc/inc/notices.inc
@@ -38,6 +38,31 @@ $smtp_authentication_mechanisms = array(
  *	'MD5' => 'MD5'
  *	'CRAM-MD5' => 'CRAM-MD5'
 */
+global $pushover_sounds;
+$pushover_sounds = array(
+	'pushover' => 'Pushover (default)',
+	'bike' => 'Bike',
+	'bugle' => 'Bugle',
+	'cashregister' => 'Cash Register',
+	'classical' => 'Classical',
+	'cosmic' => 'Cosmic',
+	'falling' => 'Falling',
+	'gamelan' => 'Gamelan',
+	'incoming' => 'Incomming',
+	'intermission' => 'Intermission',
+	'magic' => 'Magic',
+	'mechanical' => 'Mechanical',
+	'pianobar' => 'Piano Bar',
+	'siren' => 'Siren',
+	'spacealarm' => 'Space Alarm',
+	'tugboat' => 'Tug Boat',
+	'alien' => 'Alien (long)',
+	'climb' => 'Climb (long)',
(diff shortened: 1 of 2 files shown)


src/etc/inc/notices.inc
```diff
@@ -38,15 +38,40 @@ $smtp_authentication_mechanisms = array(
  *	'MD5' => 'MD5'
  *	'CRAM-MD5' => 'CRAM-MD5'
 */
+global $pushover_sounds;
```
rbgarga, line 41: global declarations must happen only inside functions

src/etc/inc/notices.inc
```diff
@@ -461,10 +486,67 @@ function curl_post_telegram($url) {
...
+ * RESULT
+ *   returns NULL if message was sent
+ ******/
+
+function notify_via_pushover($message, $force = false) {
+	global $config;
+	if ((!isset($config['notifications']['pushover']['enabled']) && (!$force))
+	    || !$config['notifications']['pushover']['apikey']
+	    || !$config['notifications']['pushover']['userkey']) {
+		if ($force) {
```
rbgarga, line 503: the main if checks for `!$force` so this condition will never be satisfied

src/etc/inc/notices.inc
```diff
@@ -461,10 +486,67 @@ function curl_post_telegram($url) {
...
+		"retry" => $config['notifications']['pushover']['retry'],
+		"expire" => $config['notifications']['pushover']['expire'],
+		"message" => "{$config['system']['hostname']}.{$config['system']['domain']}\n{$message}"
+	);
+	$result = json_decode(curl_post_pushover($url, $data), true);
+	if (is_array($result)) {
+		if ($result['status']) {
+			unset($err_msg);
+		} else {
+			$err_msg = sprintf(gettext("Failed to send Pushover notification. Error received was: {$result['errors']['0']}"));
```
rbgarga, line 524: You must use `%s` on gettext parameter and the pass the variable to sprintf 2nd parameter, otherwise it makes no sense to use sprintf

rbgarga (changes requested):

src/etc/inc/notices.inc
jim-p: This is OK in recent PHP versions, it works better to declare a variable global "globally" before use like this.

src/usr/local/www/system_advanced_notifications.php
```diff
@@ -186,7 +205,43 @@
...
 		}
 
-		if (!$input_errors && !$testsmtp && !$testtelegram) {
+		// Pushover
+		$config['notifications']['pushover']['enabled'] = ($_POST['enable_pushover'] == "yes") ? true : false;
+		$config['notifications']['pushover']['apikey'] = $_POST['pushoverapikey'];
+		$config['notifications']['pushover']['userkey'] = $_POST['pushoveruserkey'];
+
+		if (preg_replace("/[^A-Za-z0-9]/", "", $config['notifications']['pushover']['apikey']) !== $config['notifications']['pushover']['apikey']) {
+			$input_errors[] = gettext("API keys are case-sensitive, 30 characters long, and will contain the character set A-Z,a-z,0-9");
```
jim-p, line 214: Change "will contain" here to "can only use"

src/usr/local/www/system_advanced_notifications.php
```diff
@@ -186,7 +205,43 @@
...
+		}
+		if (preg_replace("/[^A-Za-z0-9]/", "", $config['notifications']['pushover']['userkey']) !== $config['notifications']['pushover']['userkey']) {
+			$input_errors[] = gettext("User keys are case-sensitive, 30 characters long, and will contain the character set A-Z,a-z,0-9");
```
jim-p, line 217: Change "will contain" here to "can only use"

src/usr/local/www/system_advanced_notifications.php
```diff
@@ -411,6 +476,75 @@
...
+	$form->add($section);
+
+	$section = new Form_Section('Pushover');
+
+	$section->addInput(new Form_Checkbox(
+		'enable_pushover',
+		'Enable Pushover',
+		'Enable Pushover Notifications',
+		$pconfig['enable_pushover']
+		))->setHelp('Check this option to enable Pushover notifications. <br>You will need to create an API key and provide your user key. <a href="https://pushover.net/api#registration" target="_blank">API documentation.</a>');
```
jim-p, line 488: Reword to avoid addressing the user ("you", "your", etc): https://docs.netgate.com/pfsense/en/latest/references/style-guide.html#what-to-avoid

src/usr/local/www/system_advanced_notifications.php
```diff
@@ -411,6 +476,75 @@
...
+
+	$section->addInput(new Form_Input(
+		'pushoverapikey',
+		'API Key',
+		'text',
+		$pconfig['pushoverapikey'],
+		['placeholder' => 'azGDORePK8gMaC0QOYAMyEEuzJnyUi']
+	))->setHelp('Enter your API key required to authenticate with the Pushover API server. <a href="https://pushover.net/apps/build" target="_blank">Create API key here.</a>');
```
jim-p, line 496: Reword to avoid addressing the user ("you", "your", etc): https://docs.netgate.com/pfsense/en/latest/references/style-guide.html#what-to-avoid

jim-p (changes requested):

src/etc/inc/notices.inc
overtninja: Can you take a second look? This is the same logic as the telegram notification. It works for me.

rbgarga: You are right, I misread that

jim-p (approved):

overtninja:
Sorry, made a small change to allow a device's default notification tone to play, rather than being overridden.

jim-p (approved):

PR #4452, opened 2020-09-22
Backup/restore DHCP v4/v6 leases. Implements #10910

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10910
- [X] Ready for review

* Allows to backup/restore DHCP v4/v6 leases;
* Adds '$backuppath' array, to add any extra data files for other services;
* Fixes `$savemsg` variable, making it global, making `print_info_box($savemsg, 'success')` to work correctly (`backup.inc` regression);

### DIFF
--- a/src/etc/inc/web/backup.inc
+++ b/src/etc/inc/web/backup.inc
@@ -115,9 +115,11 @@ function restore_rrddata() {
 
 function restore_xmldatafile($type='voucher') {
 	global $config, $g;
+
 	foreach ($config[$type]["{$type}data"]["xmldatafile"] as $file) {
 		$basename = basename($file['filename']);
-		$xmldata_file = "{$g['vardb_path']}/{$basename}";
+		$dirname = dirname($g['backuppath'][$type]);
+		$xmldata_file = "{$dirname}/{$basename}";
 		if (file_put_contents($xmldata_file, gzinflate(base64_decode($file['data']))) === false) {
 			log_error(sprintf(gettext("Cannot write %s"), $xmldata_file));
 			continue;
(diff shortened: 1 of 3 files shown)


src/etc/inc/web/backup.inc
```diff
@@ -39,6 +39,12 @@ require_once("pkg-utils.inc");
 
 $rrddbpath = "/var/db/rrd";
 $rrdtool = "/usr/bin/nice -n20 /usr/local/bin/rrdtool";
+$backuppath = array(
+	'captiveportal' => "/var/db/captiveportal*.db",
+	'dhcpd' => "/var/dhcpd/var/db/dhcpd.leases",
+	'dhcpdv6' => "/var/dhcpd/var/db/dhcpd6.leases",
+	'voucher' => "/var/db/voucher_*.db"
+);
```
sbeaver-netgate, line 47: Consider keeping these definitions in /etc/inc/globals.inc. That way they could be accessed via the $g[] global variable and there would be no need to keep track of them in multiple places. It would also avoid the need for another global in this function, which I would like to avoid wherever possible. Globals tend to break the separation of displa/logic.

src/etc/inc/web/backup.inc
```diff
@@ -158,7 +166,7 @@ function check_and_returnif_section_exists($section) {
 }
 
 function execPost($post, $postfiles, $ui = true) {
-	global $config, $g;
+	global $config, $g, $backuppath, $savemsg;
```
sbeaver-netgate, line 170: Consider revising this function to return:

`array("input_errors" => $input_errors, "savemsg" => $savemsg)`

Again, that would avoid a messy global.

vktg: Ok! Fixed

sbeaver-netgate (approved):

PR #4454, opened 2020-09-22
PPPoE Server secondary RADIUS server fixes. Issue #10926

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10926
- [X] Ready for review

* Fixes using Secondary RADIUS server;
* Fixes `secret2` -> `secret` renaming for the secondary RADIUS server;
* Additional input validation and filling of the `usernames` array - otherwise it will be cleared if the input validation fails;

### DIFF
--- a/src/etc/inc/vpn.inc
+++ b/src/etc/inc/vpn.inc
@@ -194,8 +194,20 @@ EOD;
 				if (isset($pppoecfg['radius']['server']['acctport'])) {
 					$radiusacctport = $pppoecfg['radius']['server']['acctport'];
 				}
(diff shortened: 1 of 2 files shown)


jim-p (changes requested):
Not that it was working before, but if someone had entered a second RADIUS server, wouldn't this lose the shared secret they entered into the configuration before? We should probably either keep the secret for the second RADIUS server as `secret2` or add upgrade code to change it to `secret`. Unless I'm missing something, which is entirely possible.

vktg:
Agree, `secret2` is reverted

jim-p (approved):

PR #4457, opened 2020-09-23
Bridge interface Captive Portal validation. Issue #6528

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/6528
- [X] Ready for review

Captive Portal cannot be used on a bridge member interface and vise versa

### DIFF
--- a/src/usr/local/www/interfaces_bridge_edit.php
+++ b/src/usr/local/www/interfaces_bridge_edit.php
@@ -184,6 +184,18 @@ if ($_POST['save']) {
 		$input_errors[] = gettext("At least one member interface must be selected for a bridge.");
 	}
 
+	if (is_array($_POST['members']) && is_array($config['captiveportal'])) {
+		foreach ($_POST['members'] as $member) {
+			foreach ($config['captiveportal'] as $cp) {
+				if (in_array($member, explode(',', $cp['interface']))) {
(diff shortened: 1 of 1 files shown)


jim-p (approved):

src/usr/local/www/interfaces_bridge_edit.php
```diff
@@ -184,6 +184,20 @@ function is_aoadv_used($pconfig) {
 		$input_errors[] = gettext("At least one member interface must be selected for a bridge.");
 	}
 
+	if (is_array($_POST['members']) && is_array($config['captiveportal'])) {
+		foreach ($_POST['members'] as $member) {
+			foreach ($config['captiveportal'] as $cp) {
+				foreach (explode(',', $cp['interface']) as $cpint) {
+					if ($member == $cpint) {
```
rbgarga, line 191: This last foreach + if could be replaced by a single statement like this:

`if (in_array($member, explode(',', $cp['interface']))`

rbgarga (changes requested):

src/usr/local/www/interfaces_bridge_edit.php
vktg: Thanks, replaced

PR #4458, opened 2020-09-24
Use user DN for RFC2307 membership search (updated). Issue #9527

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/9527
- [X] Ready for review

Adds checkbox to use full user DN for group membership search, i.e.
`(member=CN=Username,CN=Users,DC=example,DC=com)`
instead of
`(member=Username)`

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1272,11 +1272,6 @@ function ldap_get_groups($username, $authcfg) {
 		$ldapauthcont = $authcfg['ldap_authcn'];
 		$ldapnameattribute = strtolower($authcfg['ldap_attr_user']);
 		$ldapgroupattribute = strtolower($authcfg['ldap_attr_member']);
-		if (isset($authcfg['ldap_rfc2307'])) {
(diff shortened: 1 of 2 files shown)


jim-p (approved):
The default behavior appears to be OK now at least. I don't have a setup which needs the optional/alternate behavior to test against, however.

vktg:
> The default behavior appears to be OK now at least. I don't have a setup which needs the optional/alternate behavior to test against, however.

non-RFC2307 code is not changed,

all changes between this PR and https://github.com/pfsense/pfsense/pull/4366 is:
old PR: https://github.com/pfsense/pfsense/pull/4366/files#diff-647553ae2000651d4860da0cbf861caeR1572
new PR: https://github.com/pfsense/pfsense/pull/4458/files#diff-647553ae2000651d4860da0cbf861caeR1573

vktg:
Updated: tabs&spaces fix

jim-p:
> non-RFC2307 code is not changed,

Right, and now also the default RFC 2307 behavior is OK, where it wasn't before on the previous attempt.

What I was saying is I don't have a setup to test the new "RFC 2307 Use DN for username search" option in this PR

PR #4459, opened 2020-09-25
IPsec PH1 creation fix. Issue #9592

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/9592
- [X] Ready for review

Fixes regression of https://github.com/pfsense/pfsense/pull/4190 -
if PH1 is created without VTI, it always uses the empty name `con`,
i.e. `/var/etc/ipsec/swanctl.conf`:
```
connections {
	...
	con {
```
leaving only the last non-VTI PH1 entry in `swanctl.conf`

This PR uses "safe" >32768 (x00000) numbers for non-VTI connections


jim-p (approved):

rbgarga:
Already merged but because of a mistake not detected.  Closing.

PR #4464, opened 2020-10-01
Same RADIUS ACL parser for IPsec/OpenVPN. Implements #10469

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/10469
- [X] Ready for review

Uses the same Cisco-AVPair parser for both OpenVPN and IPsec

also fixes removing of RADIUS attributes in `ipsec.auth-user.php`:
```
- $userGroups = getUserGroups($username, $authcfg, $attributes = array());
+ $userGroups = getUserGroups($username, $authcfg, $attributes);
```


rbgarga:
I'm going to merge it as is for now but I would like to ask you both @vktg and @jim-p. Don't you think these functions should be in vpn.inc instead of util.inc?  If you agree we can move it later

PR #4470, opened 2020-10-03
Ability to clear all DHCP leases at once. Implements #7406

vktg:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/7406
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -74,6 +74,15 @@ if (($_POST['deleteip']) && (is_ipaddr($_POST['deleteip']))) {
 	header("Location: status_dhcp_leases.php?all={$_REQUEST['all']}");
 }
 
+if ($_POST['cleardhcpleases']) {
+	killbyname("dhcpd");
+	sleep(2);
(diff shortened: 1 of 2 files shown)


src/usr/local/www/status_dhcp_leases.php
```diff
@@ -74,6 +74,16 @@
 	header("Location: status_dhcp_leases.php?all={$_REQUEST['all']}");
 }
 
+if ($_POST['cleardhcpleases']) {
+	$leasesfile = "{$g['dhcpd_chroot_path']}/var/db/dhcpd.leases";
+
+	killbyname("dhcpd");
+	unlink_if_exists($leasesfile);
```
jim-p, line 81: Might need a small sleep here. Since it's being killed nicely, it will try to flush the lease file contents.

It should also remove `dhcpd.leases~` which is a temporary copy of the leases file the daemon can also use as a backup when it rewrites the lease DB.

To avoid using a single-use variable like this, you could do both at once:

```
unlink_if_exists("{$g['dhcpd_chroot_path']}/var/db/dhcpd.leases*");
```

vktg: Updated
also added the same code for DHCPv6 leases page

jim-p (approved):

PR #4482, opened 2020-10-20
Support omitting Cloudflare Zone ID for dynamic DNS

cpick:
In May, Cloudflare
[improved their API](https://community.cloudflare.com/t/bug-zone-detail-by-name-requires-zone-list-permission/128042/14)
so that a token scoped to a particular zone now has permission to
`GET https://api.cloudflare.com/client/v4/zones/?name=<specific_zone_name_here>`.

As a result, the zone id can be queried without increasing permissions.
If the username is blank, query the zone id.

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10992
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -235,6 +235,10 @@
 				break;
 			case "cloudflare-v6":
 			case "cloudflare":
+				if (!$dnsPass) $this->_error(4);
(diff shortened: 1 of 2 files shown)


netpok:
Looks great to me but I have not tested it.

jim-p (approved):

PR #4491, opened 2020-12-17
Fix separators moving when multiple filters deleted - Fix #9887

dsmackie:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/9887
- [x] Ready for review

`$ridx` was being subtracted as each filter was deleted resulting in separators becoming logically under subsequent filters.
I can not see why this subtraction was required since adding a filter does not require an increase of `$ridx`.

History likely dictates why it was required so strike this pull request if I am way off the mark.

Local testing has identified no ill effects from this change.

### DIFF
--- a/src/usr/local/www/firewall_nat.php
+++ b/src/usr/local/www/firewall_nat.php
@@ -138,11 +138,10 @@ if (isset($_POST['del_x']) && have_natpfruleint_access($natent['interface'])) {
 
 	/* delete selected rules */
 	if (is_array($_POST['rule']) && count($_POST['rule'])) {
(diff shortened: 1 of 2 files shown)


dsmackie:
Hold off on this. I've come across another condition where there is still an issue.

dsmackie:
Both [#6801](https://redmine.pfsense.org/issues/6801) and [#9887](https://redmine.pfsense.org/issues/9887) now work as expected.

dsmackie:
Fixed for both Firewall and NAT.

sbeaver-netgate (approved):

PR #4495, opened 2021-01-04
Update MBT-4420/2220 Platform id string

gkrzystek:
With 1.0 bios Intel changed indentifications tring for Turbot
That patch fixes Minnowboard detection on system level

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9242
- [x] Ready for review

To fix that issue 100% installer script need to be updated (factory.sh) but i din't found it in this repo


rbgarga:
Applied a different solution to keep code standards.  Thanks!

PR #4496, opened 2021-01-19
CaptivePortal: Redirect back to Login Page on  Logout

nraven777:
Currently (i.e when a custom logout page is present) when a user clicks on logout , a window with the  logout message is shown which closes automatically after a few seconds ,  this is both visually unappealing and can close the browser if the user only has the pfsense logout tab open , it therefore  would be a good move to redirect the user back to the login page when they click on logout.

Redmine Issue : https://redmine.pfsense.org/issues/11264

### DIFF
--- a/src/usr/local/captiveportal/index.php
+++ b/src/usr/local/captiveportal/index.php
@@ -126,23 +126,7 @@ if ($macfilter || isset($cpcfg['passthrumacadd'])) {
 }
 
 if ($_POST['logout_id']) {
-	echo <<<EOD
-<html>
-<head><title>Disconnecting...</title></head>
-<body bgcolor="#435370">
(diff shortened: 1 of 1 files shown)


jim-p:

Thanks!

nraven777:
Hey @jim-p  ,  could you please review this pull request.

rbgarga (approved):

src/usr/local/captiveportal/index.php
```diff
@@ -126,23 +126,7 @@
...
-<!--
-setTimeout('window.close();',5000) ;
--->
-</script>
-</body>
-</html>
-
-EOD;
-
+	header("Location: index.php?zone=".$cpzone);
```
jim-p, line 129: This should go at the end after the other functions have been executed.

jim-p (changes requested):

PR #4497, opened 2021-01-26
Order DNS provider instructions alphabetically

Bojan023:
Hi! I stumbled upon the weird order of these instructions a couple of times meanwhile, I think it makes sense to order them alphabetically.

- [x] Redmine Issue: minor textual change, so none
- [x] Ready for review


jim-p (changes requested):
Several lines have a mix of leading tabs and spaces, when it should be tabs only. They are easy to see here in the diff here on GitHub.

Bojan023:
You are right, no idea how these slipped into the initial commit. Fixed by last commit!

Bojan023:
Good find, no idea how these slipped into the commit.. Fixed!

Bojan023:
@jim-p can you check this PR? Thanks.

rbgarga (changes requested):
@Bojan023 you are sorting items but you are also renaming Route53 to Azure in the same commit.  Could you please split it into 2 different requests?

PR #4501, opened 2021-02-20
openssl list digest algorithm parameter changed in /etc/inc/openvpn.inc

mschiegl:
At least in OpenSSL version 1.1.1i-freebsd, used by pfsense 2.5, there is no longer a "list-message-digest-algorithms" parameter. It has been replaced by "list -digest-algorithms".
The old parameter results in an error <Invalid command 'list-message-digest-algorithms'; type "help" for a list> and may even cause an endless loop on startup/migration.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/11500
- [x] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -494,7 +494,7 @@ function openvpn_validate_curve($curve) {
 /* Obtain the list of digest algorithms supported by openssl and their alternate names */
 function openvpn_get_openssldigestmappings() {
 	$digests = array();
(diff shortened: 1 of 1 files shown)


rbgarga (approved):

jim-p:

Thanks!

jim-p (approved):
The change is good but it must have a Redmine entry associated with it.

mschiegl:
Redmine entry created (https://redmine.pfsense.org/issues/11500) and cross referenced.

PR #4506, opened 2021-03-01
Fix iftop experimental traffic fetcher, unify and improve output style

Ashus:
I've made a series of changes to the iftop experimental traffic fetcher.

Changes overview:
- multiple requests were sent to browser and they were pending, because script execution was blocking responses - now it is launched in the background and response is taken from file
- formatting was unified with output from rate (k letter size, rounding to two decimals)
- output view is better readable, aligned right and Megabits and Gigabits have their number in bold font
- the second commit improves accuracy. Iftop returns only entries that fit to console screen making results inaccurate if you have 15 connections from the same IP. Now source or destination is grouped and counted internally by iftop. This logic is applied only for cases you watch LAN interface (any interface with DHCP server enabled).

- [x] Ready for review

### DIFF
--- a/src/usr/local/bin/iftop_parser.sh
+++ b/src/usr/local/bin/iftop_parser.sh
@@ -46,7 +46,7 @@ if [ -f $pid_file ]; then
         fi
 else
         echo -n $$ > $pid_file
-        $IFTOP -nNb -i $1 -s 3 -o 2s -t 2>> /dev/null | $AWK -f $awk_script > ${cache_file}.tmp
+        $IFTOP -nNb -i $1 -s 2 -o 2s -t 2>> /dev/null | $AWK -f $awk_script > ${cache_file}.tmp
         $CAT ${cache_file}.tmp > $cache_file
         $CAT $cache_file
         $RM $pid_file
--- a/src/usr/local/bin/iftop_parser.sh
+++ b/src/usr/local/bin/iftop_parser.sh
@@ -1,15 +1,21 @@
 #!/bin/sh
 
 # Usage
-if [ $# -ne 1 -a $# -ne 2 ]; then
-        echo "Usage : $0 iface"
+if [ $# -ne 2 ]; then
+        echo "Usage : $0 iface {all|hidesource|hidedestination}"
+        exit
+fi
+
+if [ "$2" != "all" ] && [ "$2" != "hidesource" ] && [ "$2" != "hidedestination" ]; then
+        echo "Usage : $0 iface {all|hidesource|hidedestination}"
         exit
 fi
 
 # files paths
-pid_file=/var/run/iftop_${1}.pid
-cache_file=/var/db/iftop_${1}.log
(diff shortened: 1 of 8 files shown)


jim-p:

Thanks!

Ashus:
Redmine issue created and linked it with this PR.

jim-p (changes requested):
Do these various configuration files have to be in `/usr/local/bin/`? They should be in `/usr/local/etc/` if they must be hardcoded like that. If they could be generated automatically they would go in `/var/etc/` instead.

Adding config files in `/usr/local/bin/` seems like a bad idea to me if we can avoid it.

Ashus:
Of course the config files don't belong there, thanks for that, it was a quick solution for POC when I began with changes and I forgot to move them. I pushed some additional changes to tidy up temporary files. Also the parser now doesn't need to return the data, so it was a bit simplified.

src/usr/local/bin/iftop_parser.sh
```diff
@@ -1,17 +1,26 @@
...
         exit
 fi
 
 # files paths
-pid_file=/var/run/iftop_${1}.pid
-cache_file=/var/db/iftop_${1}.log
+pid_dir=/var/run/iftop
+pid_file=$pid_dir/${1}_${2}.pid
+data_dir=/var/tmp/iftop
+data_file=$data_dir/${1}_${2}.txt
```
jim-p, line 18: `$data_dir` should be `${data_dir}`

jim-p (changes requested):

src/usr/local/bin/iftop_parser.sh
```diff
@@ -1,17 +1,26 @@
...
+if [ "$2" != "all" ] && [ "$2" != "hidesource" ] && [ "$2" != "hidedestination" ]; then
+        echo "Usage : $0 iface {all|hidesource|hidedestination}"
         exit
 fi
 
 # files paths
-pid_file=/var/run/iftop_${1}.pid
-cache_file=/var/db/iftop_${1}.log
+pid_dir=/var/run/iftop
+pid_file=$pid_dir/${1}_${2}.pid
```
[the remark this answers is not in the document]
Ashus: Fixed.

src/usr/local/bin/iftop_parser.sh
Ashus: Fixed.

jim-p (approved):

vktg:
@Ashus  Please rebase the PR and resolve conflicts

Ashus:
Sorry, but I lost all motivation rebasing this due to the fact, that I cannot use current version in my configuration anymore and lot of changes has been done.
This [bug](https://redmine.pfsense.org/issues/11470) affects me and therefore I need to remain on pfSense 2.4.5 because of Wireguard tunnels, wherein shaping must be done separately. It was easy to port my changes to codebase of 2.5.0, but that was a year ago.

PR #4508, opened 2021-03-12
Include wildcard setting in dyndns refresh

thomasloven:
Every 25th day, my dyndns forcibly updates, and the wildcard entry is gone.
Updating it manually brings it back.

This should fix that, as far as I can tell. Though I must admit I have not tested it live for 25 days.

- [X] Redmine Issue: https://redmine.pfsense.org/issues/11667
- [X] Ready for review


thomasloven:
Done. Thank you.

PR #4513, opened 2021-04-07
Support hiding interface groups via special tag

theonemcdonald:
This supports hiding interface groups from the UI by setting `<hide>yes</hide>`

Redmine Issue: https://redmine.pfsense.org/issues/11790

### DIFF
--- a/src/usr/local/www/interfaces_groups.php
+++ b/src/usr/local/www/interfaces_groups.php
@@ -82,7 +82,7 @@ display_top_tabs($tab_array);
 					</tr>
 				</thead>
 				<tbody>
-<?php foreach ($a_ifgroups as $i => $ifgroupentry): ?>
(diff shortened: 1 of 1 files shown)


src/usr/local/www/interfaces_groups.php
```diff
@@ -82,7 +82,7 @@
 					</tr>
 				</thead>
 				<tbody>
-<?php foreach ($a_ifgroups as $i => $ifgroupentry): ?>
+<?php foreach ($a_ifgroups as $i => $ifgroupentry): if (isset($ifgroupentry['hide']) && $ifgroupentry['hide'] != 'yes'): ?>
```
jim-p, line 85: The `if` part should be on its own line

jim-p (changes requested):

theonemcdonald:
I say we can probably close this?, as much more thought has gone into this since I originally wrote it. This is just one part of a larger package integration problem, though this might be an easy solution during the interim while something more permanent is worked on.

PR #4514, opened 2021-04-14
Add DynDNS provider dy.fi

raphendyr:
I looked into implementing this old and country specific dynamic DNS service. It uses a very limited set of update features. It even doesn't support updates to ipv6 addresses (only via web interface).

This PR is ready for feedback, but it's waiting for other PRs to be handled first. In addition to normal feedback, I would also like to get feedback on the following topics:

- [x] How should the new entries in switch cases be sorted? Based on how related they are to other code or alphabetically. (see blocking PRs)
- [x] Should identical show/hide blocks be merged? If yes, should it be in a separate commit? (see blocking PRs)

Status:

- [x] Blocked by #4516 
- [x] Blocked by #4523 
- [x] Blocked by #4527
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12090
- [x] I have run this in production enough long time to see if breaks...
- [x] Ready for review


src/usr/local/www/services_dyndns_edit.php
```diff
@@ -599,22 +584,6 @@ function setVisible(service) {
...
 				break;
-			case "domeneshop":
-		        case "domeneshop-v6":
-				hideGroupInput('domainname', true);
-				hideInput('resultmatch', true);
-				hideInput('updateurl', true);
-				hideInput('requestif', true);
-				hideCheckbox('curl_ipresolve_v4', true);
-				hideCheckbox('curl_ssl_verifypeer', true);
-				hideInput('username', false);
```
jim-p, line 610: This line for `username` is here for `domenshop` but it's not in your combined version below.

jim-p (changes requested):
> How should the new entries in switch cases be sorted? Based on how related they are to other code or alphabetically.

We've tried to keep the list alphabetical.

> Should identical show/hide blocks be merged? If yes, should it be in a separate commit?

If they are 100% identical, then it could be considered, but that should be a separate PR entirely since it's not related to adding a new provider. It's code optimization.

src/usr/local/www/services_dyndns_edit.php
raphendyr: I was thinking about that. If I get the logic correctly, then I think there is a bug. There is only one provider, which hides the `username` and none of the providers shows it again, so it should stay hidden. I should test that hypothesis.

If that is the case, then I think the code should hide all fields in one place and then show those that are specific for the provider. Or at least have some "defaults" hidden/shown and only difference per provider. I'll look into this a bit soon.

Any case, I revert merging from this PR and do that separately.

raphendyr:
> We've tried to keep the list alphabetical.

Ok. I'll recheck that my additions are in the best possible location based on that. I think few of the switches had cases in relatively random order. Would a fix (sorting cases) be acceptable as a different PR? I'm not super interested to mark those lines for me (in git blame), but keeping the code base clean would be beneficial.

Thanks for the feedback!

jim-p:
> Ok. I'll recheck that my additions are in the best possible location based on that. I think few of the switches had cases in relatively random order. Would a fix (sorting cases) be acceptable as a different PR? I'm not super interested to mark those lines for me (in git blame), but keeping the code base clean would be beneficial.

If it's for a the new code you are submitting, it can be in the same PR. But if you are changing the order of some other provider(s) then it should be an isolated PR so it can be considered separately.

raphendyr:
I updated this PR to be based on #4523. The commit adding the provider can be cherry-picked or I can rebase this PR when that cleanup PR is merged.

rbgarga:
@raphendyr #4523 was merged.  Please rebase your fork

raphendyr:
I'm also working on a feature where the cache time can be defined in settings for the manual provider. I will make that PR against the redmine task currently linked in this PR and I will create another redmine issue for the sprcific dy.fi feature for this PR.

I most likely will get that done within few days, now that the refactprong PR was merged.

src/etc/inc/dyndns.class
```diff
@@ -1814,6 +1831,32 @@
...
+				case 'dyfi':
+					// see specification at https://www.dy.fi/page/specification
+					if (preg_match('/badauth/i', $data)) {
+						$status = $status_intro . $error_str . gettext("User Authorization Failed");
+					} else if (preg_match('/nohost/i', $data)) {
+						$status = $status_intro . $error_str . gettext("No such host");
+					} else if (preg_match('/notfqdn/i', $data)) {
+						$status = $status_intro . $error_str . gettext("Not A FQDN!");
+					} else if (preg_match('/badip/i', $data)) {
+						$status = $status_intro . $error_str . gettext("Invalid IP. IP Address submitted is improperly formatted or is a private IP address or is on a blacklist.");
```
raphendyr, line 1843: I wanted to use an existing string, so there wouldn't be a new translation. However, this error happens also when the client IP is not located in Finland based on geo-ip database. Technically it's a blacklist, but mentioning the country could be better for the user?

raphendyr:
I created a dedicated redmine issue for this, as the original issue was more about the custom cache age.

rbgarga:
@raphendyr #4527 was already merged.  Please rebase your fork and fix conflicts

raphendyr:
Rebased. I noticed that I had used case-sensitive sorting in a previous PR, so I fixed that in the commit 691d0458d90df5c6c84d39f0fcedd23160358cb7 . It's small change in a comment string, so I thought it could be fine here.

I noticed it as #4528 had not sorted this field with a computer.

Additionally, I see that recent PRs didn't notice the comments like https://github.com/pfsense/pfsense/blob/5a8d60131c7badad1aba2caa60ecb7d6944cb7e2/src/etc/inc/dyndns.class#L531

raphendyr:
@jim-p could you re-review this, thank you.

PR #4516, opened 2021-04-22
Refactor hide logic for fields on DynDNS edit page

raphendyr:
The Dynamic DNS edit page included a hide/show logic for username field only for few providers, which meant that if a provider with the hide logic was selected, then the field was lost for all others too.

This PR (or specifically one commit) refactors that logic, so that the default hide/show state is set always and only fields with different states are changed per service.

In addition, last commit sorts the prociders, so that they are in alphabetical order. I can drop this commit, if that change is not wanted or shouldn't be included in the PR.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/11840
- [x] Manually tested that same fields are shown before and after the changes
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -493,10 +493,24 @@ print($form);
 events.push(function() {
 
 	function setVisible(service) {
+		// First, set default visibility and then modify per service, if needed.
+		hideGroupInput('domainname', true);
+		hideInput('resultmatch', true);
+		hideInput('updateurl', true);
+		hideInput('requestif', true);
+		hideCheckbox('curl_ipresolve_v4', true);
+		hideCheckbox('curl_ssl_verifypeer', true);
+		hideInput('username', false); // show by default
+		hideInput('host', false); // show by default
+		hideInput('mx', false); // show by default
+		hideCheckbox('wildcard', false); // show by default
+		hideCheckbox('proxied', true);
+		hideInput('zoneid', true);
+		hideInput('ttl', true);
+
 		switch (service) {
 			case "custom" :
 			case "custom-v6" :
-				hideGroupInput('domainname', true);
 				hideInput('resultmatch', false);
 				hideInput('updateurl', false);
 				hideInput('requestif', false);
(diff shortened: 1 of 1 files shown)


jim-p (changes requested):
The sorting and whitespace and other changes should be a separate PR, let's keep this one to just the show/hide changes and nothing else.

raphendyr:
> The sorting and whitespace and other changes should be a separate PR, let's keep this one to just the show/hide changes and nothing else.

Sounds good!

raphendyr:
> The sorting and whitespace and other changes should be a separate PR, let's keep this one to just the show/hide changes and nothing else.

First, should I make a PR for the whitespace or is that something we should just skip? If I create a PR, do I create a redmine issue too?

Second, is the sorting something that should be done, so should I make a PR of that?

raphendyr:
I wanted to convince my self, that the change shouldn't break anything, so I verified that added and removed lines match (i.e., there are no such lines removed, which wouldn't have a matching line added).

```sh
diff \
  <(git show 5479afd0f9|grep '^-\s\+'|sed 's/^-\s*//'|sort -u) \
  <(git show 5479afd0f9|grep '^+\s\+'|sed -e 's,//.*$,,' -e 's/^\+\s*//' -e 's/\s*$//' -e '/^$/d' |sort -u)
```
Above yields empty output, as it's supposed to. Or wiht `-y` flag for the diff, it yields two identical lists of line contents.

sbeaver-netgate:
Suggest you make a separate PR that contains both whitespace and sorting. Accompany it with a redmine ticket stating "Compliance with pfSense style guide" or similar.

raphendyr:
> Suggest you make a separate PR that contains both whitespace and sorting. Accompany it with a redmine ticket stating "Compliance with pfSense style guide" or similar.

Check. I'll do that once this one is merged. On that note, is there something I'm missing, which is blocking the merge?

rbgarga:
@raphendyr #4523 was merged.  Please rebase your work

raphendyr:
@rbgarga I think this PR was manually merged in https://github.com/pfsense/pfsense/commit/0e4202b77b98e748a321547ee39a2026baacb70b

This was my first PR, so I wasn't sure if it was supposed to stay open (e.g. close at release), hence I didn't say anything.

PR #4519, opened 2021-05-07
Dynamic DNS: support non-Oracle IP Check providers

JamesTheAwesomeDude:
# What this does

Resolves https://forum.netgate.com/topic/158230 by supporting the following IP Check service providers:

## Standard providers

Simply use any of these directly:

 - [Google Domains](https://domains.google.com/checkip)
 - [icanhazip by Major Hayden](https://icanhazip.com/)
 - [Namecheap](https://dynamicdns.park-your-domain.com/getip)
 - [No-IP (Vitalwerks)](http://ip1.dynupdate.no-ip.com/)
 - [ipify by Randall Degges](https://api.ipify.org/)
 - [nsupdate.info by Thomas Waldmann](https://www.nsupdate.info/myip)
 - **all** standard providers not listed above

## Non-standard providers

You can now use these, by entering these special URLs:

 - [FreeDNS (Afraid.org)](https://freedns.afraid.org/dynamic/check.php)
   ```
   https://freedns.afraid.org/dynamic/check.php#/%28%5CS%2B%29%24/m
   ```
 - [Hurricane Electric](https://checkip.dns.he.net/)
   ```
   https://checkip.dns.he.net/#/%28%5B%5E%28%29%5D%2B%29%24/
   ```
 - [ip6.me (Dullies Internet Exchange)](https://ip6.me/api/)
   ```
   https://ip6.me/api/#/%24IPv%5Cd%2B%2C%28%5B%5E%2C%5D%2B%29/
   ```
 - [Dyn (Oracle)](http://checkip.dyndns.org/)  
   grandfathered in the default 😉
 - Any other provider with a jenky or complex response: URL-encode a regex to parse it — store the IP in the **first captured group** — then slap it onto the URL as a fragment. (Remember the leading slash, trailing slash, and any flags. `$` only matches end-of-line if you include `m`!)

# What I changed

I removed the hard-coded regex for parsing IP Check service responses and replaced it with one dynamically loaded from the URL's fragment. The default provider, Oracle, had its regex embedded in its URL in the default settings.


JamesTheAwesomeDude:
https://redmine.pfsense.org/issues/11899

May 7, 2021 8:07:00 AM Jim P ***@***.***>:

src/etc/inc/services.inc
```diff
@@ -2214,7 +2214,7 @@ function dyndnsCheckIP($int) {
 		$ip_result_page = curl_exec($ip_ch);
 		curl_close($ip_ch);
 		$ip_result_decoded = urldecode($ip_result_page);
-		preg_match('=(?:^\\s*|Current IP Address: )([^\\s<>]+)(\\s*$|</body>)=siU', $ip_result_decoded, $matches);
+		preg_match('=(?:^\\s*|(?:Current|Your) IP Address(?: is)? ?: )([^\\s<>]+)(\\s*$|</body>)=siU', $ip_result_decoded, $matches);
```
JamesTheAwesomeDude: Another potential result, not addressed by this commit:
```text
Your Proxy IPaddress is : 203.0.113.0.157(via 198.51.100.104)
```

```suggestion
  		preg_match('=(?:^\\s*|(?:Current|Your(?: Proxy)?) IP ?Address(?: is)? ?: )([^\\s<>]+)(\\s*$|</body>)=siU', $ip_result_decoded, $matches);
```

jim-p:
Since the regex complexity is growing pretty fast, you might also consider allowing each Check IP service entry to define its own specific regex that could be used in place of the default.

Then additional providers could be added without trying to support everyone under the sun with a single complex regex.

JamesTheAwesomeDude:
Indeed; that would have the added benefit of not needing *any* pattern-matching for providers categorized as "standard" above.

I'll code that tomorrow.

src/etc/inc/services.inc
```diff
@@ -2214,7 +2214,7 @@ function dyndnsCheckIP($int) {
 		$ip_result_page = curl_exec($ip_ch);
 		curl_close($ip_ch);
 		$ip_result_decoded = urldecode($ip_result_page);
-		preg_match('=Current IP Address: (.*)</body>=siU', $ip_result_decoded, $matches);
```
JamesTheAwesomeDude, line 2217: Note to self: refactor this to split out the regexes per-provider

Should I store a [mapping](https://www.php.net/manual/en/language.types.array.php) of known per-provider regexes for the oddballs as a constant booger in here, or have that passed in via fragment parameter or directive?

jr-clark:
What about just grabbing the first IP address in the response from the provider?

preg_match('/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/', $ip_result_decoded, $matches);

and then using
 
$matches [0];

JamesTheAwesomeDude:
@jr-clark You've already failed for IPv6 😉

Frankly, I'd *ideally* do-away-with the regexing altogether and remove all possibility for screwing it up by only supporting external providers that expose a standardized API; but Netgate has a partnership with Oracle until they finally cut off Dyn's life-support, so the precedent is set otherwise.

-----

Incidentally, I do not actually expect the regex to grow from its current state much further, given that only those 4 titans: Oracle, Afraid.org, Hurricane Electric, and Dullies have any claim to special coding; any other providers or DDNS services -- as evidenced by Hayden, Degges, and Waldmann's projects linked in the OP, as well as Namecheap and innumerable others -- can trivially stand up one without imposing any special requirements whatsoever (beyond, perhaps, TLS support) on the client.

Even doing any regex at all is really just special treatment for those specific providers.

jr-clark:
Hmmm... You know the ip and type (v4 or v6) of address that you are checking. preg_match the provider response for the appropriate ip type?

Is there a doc somewhere that describes the standardized API for "Whats My IP"?

JamesTheAwesomeDude:
I've just deployed these changes to my live system and can confirm they seem to work.

src/etc/inc/services.inc
```diff
@@ -2214,7 +2215,7 @@ function dyndnsCheckIP($int) {
 		$ip_result_page = curl_exec($ip_ch);
 		curl_close($ip_ch);
 		$ip_result_decoded = urldecode($ip_result_page);
```
JamesTheAwesomeDude, line 2217: Also, it's worth noting that this `urldecode` call is both unnecessary and inappropriate, as no such service would return URL-encoded content as part of a response body. One should simply pass `$ip_result_page` into `$preg_match` directly, and not "decode" the IP response with a codec which would never be used in that context.

(This sounds pedantic, but would theoretically cause problems in the event that a user is doing DDNS on their device's IPv6 link-local address.)

JamesTheAwesomeDude:
Confirmation with a spoofed endpoint also functional:

```
http://ix.io/3mgI#/%28%5CS%2B%29%24/m
```

Screenshot from 2021-05-09 02-53-05

JamesTheAwesomeDude:
I've just realized that the documentation will need to be updated, too…
Screenshot from 2021-05-09 02-59-25

JamesTheAwesomeDude:
I will need help with the `.po` and `.mo` files. I'm not familiar with gettext nor with any of the languages involved to any useful degree.

(Should we even bother mentioning support for regex-parsing of these responses in the documentation?)

raphendyr:
I think it would make more sense to add the regex as a new field. That way people wouldn't need to encode the regex at all, but could just write it in the field.

Additionally, list of known providers could be provided, like with the Dynamic DNS service. In that case, the current feature could be a custom provider. However, that kind of change might be out of a scope of a single PR.

PR #4521, opened 2021-05-20
Fix return logic on sigkillbypid

theonemcdonald:
Not sure why this was returning 1 for success and 0 for a failure? I checked through the remaining codebase and this PR doesn't seem to impact anything else.

- [x] https://redmine.pfsense.org/issues/11940
- [x] Ready for review

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -57,7 +57,7 @@ function isvalidproc($proc) {
 }
 
 /* sigkill a process by pid file, and wait for it to terminate or remove the .pid file for $waitfor seconds */
-/* return 1 for success and 0 for a failure */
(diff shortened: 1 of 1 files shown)


rbgarga:
It returns same `kill` command would return on a shell call and there are places where it's used (e.g. `system_syslogd_start()`) that expects it to return this way.  This change would break at least this place but possibly others since all places calling `killbypid()` would also be affected.  If you want to work on a complete solution I suggest to replace it to return boolean values and include a detailed check on all possible places it's used on both main repo and FreeBSD-ports to make sure change is not breaking anything

theonemcdonald:
I'm going to close this, will open a new PR if/when something more substantial is done here.

PR #4526, opened 2021-06-23
Cisco-AVPair + Framed-IP-Address: correcting clientip

fl0l0u:
Workaround to substitute Framed-IP-Address value in Cisco-AVPair ACL's where {clientip} is used

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12076
- [x] Ready for review

### DIFF
--- a/src/etc/inc/util.inc
+++ b/src/etc/inc/util.inc
@@ -3427,9 +3427,8 @@ function parse_cisco_acl_rule($rule, $devname, $dir, $proto) {
 		$index++;
 		if ((($proto == 'inet') && (is_ipaddrv4(trim($rule[$index])) || (trim($rule[$index]) == "{clientip}"))) ||
 		    (($proto == 'inet6') && (is_ipaddrv6(trim($rule[$index])) || (trim($rule[$index]) == "{clientipv6}")))) {
-			if($GLOBALS['attributes']['framed_ip']) {
-				$framed_ip = $GLOBALS['attributes']['framed_ip'];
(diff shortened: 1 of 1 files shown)


src/etc/inc/util.inc
```diff
@@ -3427,7 +3427,12 @@ function parse_cisco_acl_rule($rule, $devname, $dir, $proto) {
 		$index++;
 		if ((($proto == 'inet') && (is_ipaddrv4(trim($rule[$index])) || (trim($rule[$index]) == "{clientip}"))) ||
 		    (($proto == 'inet6') && (is_ipaddrv6(trim($rule[$index])) || (trim($rule[$index]) == "{clientipv6}")))) {
-			$tmprule .= "from {$rule[$index]} ";
+			if($GLOBALS['attributes']['framed_ip']) {
```
jim-p, line 3430: Needs a space between `if` and `(`

src/etc/inc/util.inc
```diff
@@ -3427,7 +3427,12 @@ function parse_cisco_acl_rule($rule, $devname, $dir, $proto) {
...
+				$framed_ip = $GLOBALS['attributes']['framed_ip'];
```
jim-p, line 3431: Is `$framed_ip` used somewhere else? If not, no need to create it, just use `$GLOBALS['attributes']['framed_ip']` directly in the string below without making the single use variable.

jim-p (changes requested):

jim-p (approved):

PR #4527, opened 2021-06-25
Add option to set max cache age for the custom dynamic DNS provider

raphendyr:
Add ability to manually set an integer for the max cache age, i.e, set a duration (in days) for how often the Dynamic DNS entry is always updated. This is required for some providers or setting really huge number here, would kind of disable the force update.

Currently, the field is visible only for the custom provider (v4 and v6).

- [x] Redmine Issue: https://redmine.pfsense.org/issues/9092
- [x] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -197,7 +197,7 @@
 		function __construct($dnsService = '', $dnsHost = '', $dnsDomain = '', $dnsUser = '', $dnsPass = '',
 					$dnsWildcard = 'OFF', $dnsProxied = false, $dnsMX = '', $dnsIf = '', $dnsBackMX = '',
 					$dnsServer = '', $dnsPort = '', $dnsUpdateURL = '', $forceUpdate = false,
-					$dnsZoneID ='', $dnsTTL='', $dnsResultMatch = '', $dnsRequestIf = '',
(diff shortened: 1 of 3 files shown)


raphendyr:
I guess I kind of don't need to think about the translations for the strings? Anything I should take care about that?

jim-p (approved):

jim-p:
> I guess I kind of don't need to think about the translations for the strings? Anything I should take care about that?

You don't need to worry about translating them, no. They'll get picked up and added to the translation site eventually and then contributors can add translations for the new strings later.

src/etc/inc/dyndns.class
```diff
@@ -338,11 +338,15 @@
 			$this->_dnsUpdateURL = $dnsUpdateURL;
 			$this->_dnsResultMatch = $dnsResultMatch;
 			$this->_dnsRequestIf = get_failover_interface($dnsRequestIf);
+			if ($dnsMaxCacheAge !== "") {
+				$this->_dnsMaxCacheAgeDays = (int)$dnsMaxCacheAge;
+			} else {
+				$this->_dnsMaxCacheAgeDays = 25;
```
raphendyr, line 344: This is written with `if..else`, because #4514 will add `switch` here.

@jim-p to your info, I changed this part after your review.

jim-p (approved):

PR #4530, opened 2021-07-06
Exclude revoked certs from expiration notification. Fix Bug #12109

Alexilmarranen:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12109
- [x] Ready for review


Alexilmarranen:
The changes have been tested (2.5.1) for:
1. Expired Certificates and not revoked certificates - alert (as expected)
2. Expired and revoked certificates - no alert (as expected)
3. Not expired and not revoked - no alert (as expected)

jim-p (changes requested):
That is not a sufficient test. There needs to be a lot more logic here to figure out if the certificate could still be in use. We can't assume that from its presence it just one CRL. That is, the certificate may be revoked from one CRL but not others, or used in other ways.

For example: The admin may have two OpenVPN servers using the same CA, but each with a different CRL. Or they may only have a CRL on one server, not all of them. Or used in IPsec with a CRL but no CRL in OpenVPN. And so on.

We have to err on the side of caution here.

One possible way around it would be an option on system_advanced_notifications.php which controls whether or not notifications are generated for expired revoked certificates. So long as it defaults to enabled, that would be OK. Then if an admin decides they don't want those, they can be suppressed.

Alexilmarranen:
ok, php is not my strong point, but I will try to implement it. It may take a while - don't think I'm lost

src/etc/inc/certs.inc
```diff
@@ -2072,7 +2072,7 @@ function cert_analyze_lifetime($expiredays) {
 		/* Not expiring soon */
 		$lrclass = 'normal';
 		$expstring = gettext("%d %s until expiration");
-	} elseif ($expiredays > 0) {
```
Alexilmarranen, line 2075: This comparison does not generate an alert on the last day before certificate expiration.

src/etc/inc/certs.inc
```diff
@@ -2321,7 +2321,7 @@ function cert_notify_expiring() {
 		/* Fetch and analyze expiration */
 		$expiredays = cert_get_lifetime($cert, true);
 		/* If the result is null, then the lifetime data is missing, so skip the invalid entry. */
-		if ($expiredays == null) {
```
Alexilmarranen, line 2320: Loose comparison on the last day before certificate expiration does not generate an alert.

Alexilmarranen:
I made changes and fixed in a separate commit the wrong comparisons (in my opinion). 

I decided to change the setting - not disabling alerts for revoked certs, but enabling ignore for revoked certs. It seems that for this reason, there should be no problems with updates. Was I correct that there is no need to create a function in upgrade_config.inc?

I tested these fixes on a set of certificates: valid, expired, expired and revoked once, expired and revoked twice, soon expired and revoked once, soon expired and revoked twice. In all cases, when the setting was turned on, I got the expected result.

src/usr/local/www/system_advanced_notifications.php
```diff
@@ -80,6 +80,12 @@
 ))->setHelp('When enabled, the firewall will check CA and Certificate expiration ' .
 	'times daily and file notices when expired or soon-to-expire ' .
 	'entries are detected.');
+$section->addInput(new Form_Checkbox(
+    'revoked_cert_ignore_notify',
+    'Ignore for Revoked Certificate',
```
jim-p, line 85: This is a bit longer than it needs to be, `Ignore Revoked` works better since it won't wrap. It's between two other certificate options and the checkbox description mentions certificates, so no worries about context.

jim-p (changes requested):
One cosmetic change, but otherwise seems good. I gave it a quick test and it worked as expected. With the option off I was notified about an expired+revoked certificate. With the option checked I was not.

src/usr/local/www/system_advanced_notifications.php
Alexilmarranen: Totally agree, fixed it

jim-p (approved):
Looks good. It would be ideal if you could squash these changes into a single commit which mentions the redmine issue, to reduce the number of commits. It's not a hard requirement, though. Thanks!

Alexilmarranen:
> Looks good. It would be ideal if you could squash these changes into a single commit which mentions the redmine issue, to reduce the number of commits. It's not a hard requirement, though. Thanks!

I'm sorry, but I'm afraid to break something, because I'm not very sure about my squash skill) Can you tell me how to do it right? I was pulling from main and not sure how to proceed now

PR #4535, opened 2021-08-02
Bug #12174 - cache gateway_info_popup results for firewall_rules.php page loads

luftegrof:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12174
- [x] Ready for review

Calling gateway_info_popup for each rule with 'gateway' set caused long page loads for rule-sets that make extensive use of policy-based routing.  

This change would cause `firewall_rules.php` to first check an array, `gw_info`, for cached gateway info before resorting to calling the `gateway_info_popup` function.  The full content from each `gateway_info_popup` call is added to the cache.

### DIFF
--- a/src/usr/local/www/firewall_rules.php
+++ b/src/usr/local/www/firewall_rules.php
@@ -823,25 +823,25 @@ foreach ($a_filter as $filteri => $filterent):
 							<?php endif; ?>
 						</td>
 						<td>
-							<?php 
-								/* Cache gateway info for this page load.
-								 * See https://redmine.pfsense.org/issues/12174 */
-								if (isset($filterent['gateway'])) {
+							<?php if (isset($filterent['gateway'])): ?>
+								<?php
+									/* Cache gateway status for this page load.
+									 * See https://redmine.pfsense.org/issues/12174 */
 									if (!is_array($gw_info)) {
 										$gw_info = array();
 									}
 									if (empty($gw_info[$filterent['gateway']])) {
 										$gw_info[$filterent['gateway']] = gateway_info_popup($filterent['gateway']);
 									}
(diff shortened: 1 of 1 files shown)


src/usr/local/www/firewall_rules.php
```diff
@@ -823,11 +823,25 @@ function($type) {
...
+								 * See https://redmine.pfsense.org/issues/12174 */
+								if (isset($filterent['gateway'])) {
+									if (!is_array($gw_info)) {
+										$gw_info = array();
+									}
+									if (empty($gw_info[$filterent['gateway']])) {
+										$gw_info[$filterent['gateway']] = gateway_info_popup($filterent['gateway']);
+									}
+									if (!empty($gw_info[$filterent['gateway']])) {
+										echo sprintf('<span data-toggle="popover" data-trigger="hover focus" title="%s" data-content="%s" data-html="true">', gettext('Gateway details'), $gw_info[$filterent['gateway']]);
```
jim-p, line 837: I'd like to avoid doing a print/echo of HTML like this, it's better to do a similar style to what was removed where the logic is in PHP but the presentation is inline. It is much easier to maintain going forward and less prone to quirks like having to ensure things are quoted exactly right.

jim-p (changes requested):
Logically it looks good but I did have a concern about the style of one part in particular. See my inline comment.

src/usr/local/www/firewall_rules.php
luftegrof: @jim-p , I hope I did that right.  See the latest diff.

I presume you would keep your gateways_status cache as well, so I updated this function call to use it.

jim-p (approved):
Yes, that all looks good now, thanks!

PR #4536, opened 2021-08-10
Alzeebum 10980 - Restores standard /etc/rc.local functionality

alzeebum:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/10980
- [x] Ready for review

Tested this on 2.5.0, but it's a simple enough sledgehammer that it doesn't seem like it can introduce any issues other than potential backwards compatibility.  I can't imagine there are that many people out there using the broken existing rc.local invocation setup though, so IMO the potential for damage is low.


marcos-ng:
Thanks for looking into this. It's been merged with ce83f38a8a51c3abe1291878420627343cf6b4a4 (keeps the .running check).

PR #4539, opened 2021-09-22
Create SECURITY.md

zidingz:
A simple instruction for security researchers.


jim-p:
No need to create it without any meaningful content. If we choose to create one, we'll make one containing all the relevant information and not keep a template around like this.

PR #4542, opened 2021-10-31
DynDNS: make simultaneous update of IP and LegacyIP possible

9Lukas5:
The `_checkIP()` method is only obtaining either the legacyIP as default
or the IP if the flag is set for it.  
For services that would override one entry with the newest on each
update it would not be possible to create entries with both, as this
requires them to be sent in the same request.

For this purpose this adds a new flag to signal a service needs both,
legacyIP and IP, available at the same time and a new field
`_dnsLegacyIP`.  
If the new flag isn't set (which is the default) the `_checkIP()` works
like before. If the new flag is set, the `_dnsIP` field will hold the IP
and the `_dnsLegacyIP` the legacyIP.

To not break the cached-IP handling/having to fool around there as well,
the return of the `_checkIP()` is the legacyIP as default and if both
formats are needed and only IP if the `_useIPv6` is set.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12494
- [x] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -172,6 +172,7 @@
 		var $_dnsDomain;
 		var $_FQDN;
 		var $_dnsIP;
+		var $_dnsLegacyIP;
 		var $_dnsWildcard;
 		var $_dnsProxied;
 		var $_dnsMX;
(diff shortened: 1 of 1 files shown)


marcos-ng:
Closing - see notes on redmine.

PR #4543, opened 2021-10-31
DynDNS: add deSEC IPv4&v6 simultaneos update

9Lukas5:
Leaves the yet included services deSEC and deSEC-v6 untouched.
These are using the automatic ip-recognition from the dns service for
both record types.  
This new service to create/update an entry for the same dns name with
both record types requires the IPs for both records simultaneously,
for which this new service is, leaving the distinct record services in
place for people only want to update a dedicated record.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12495
- [ ] Ready for review

Depends on #4542

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -172,6 +172,7 @@
 		var $_dnsDomain;
 		var $_FQDN;
 		var $_dnsIP;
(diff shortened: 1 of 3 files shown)


marcos-ng:
Closing - see notes on redmine.

PR #4544, opened 2021-11-22
Correct typographical error

alandgri:
Certificates was misspelled

There is no Redmine issue as the guidelines state that is not necessary for minor typographical corrections.
This change is ready to go. I did make the change on a test environment and ensured it does not break.


jim-p:
I fixed it manually, thanks!

PR #4545, opened 2021-12-02
Securely encrypt config file

PhilZ-cwm6:
OpenSSL currently defaults to 10'000 iterations, which is by far a low security standard that was acceptable before 2010.
Nowadays, even on entry level hardware, a count of 100'000 is the minimal.
The impact on import/export of config files is marginal for such an occasional task.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12556
- [x] Ready for review

### DIFF
--- a/src/etc/inc/crypt.inc
+++ b/src/etc/inc/crypt.inc
@@ -39,10 +39,11 @@
 		 *  unless we need to read old data encrypted without it. */
 		$keyder = ($legacy) ? "" : "-pbkdf2";
 		$md = ($legacy) ? "md5" : "sha256";
+		$iter = ($legacy) ? "" : "-iter 300000";
 
 		$output = "";
 		$exitcode = "";
-		exec("/usr/bin/openssl enc {$opt} -aes-256-cbc -in {$file}.dec -out {$file}.enc -pass pass:" . escapeshellarg($pass) . " -salt -md ${md} {$keyder} 2> /dev/null", $output, $exitcode);
+		exec("/usr/bin/openssl enc {$opt} -aes-256-cbc -in {$file}.dec -out {$file}.enc -pass pass:" . escapeshellarg($pass) . " -salt -md ${md} {$keyder} {$iter} 2> /dev/null", $output, $exitcode);
 
 		if (($exitcode == 0) && file_exists("{$file}.enc") && (filesize("{$file}.enc") > 0)) {
 			$result = file_get_contents("{$file}.enc");
--- a/src/etc/inc/crypt.inc
+++ b/src/etc/inc/crypt.inc
@@ -36,10 +36,12 @@
 		file_put_contents("{$file}.dec", $val);
 
 		/* Use PBKDF2 Key Derivation (https://en.wikipedia.org/wiki/PBKDF2)
-		 *  unless we need to read old data encrypted without it. */
+		 *  unless we need to read old data encrypted without it.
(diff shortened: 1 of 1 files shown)


src/etc/inc/crypt.inc
```diff
@@ -39,10 +39,11 @@
 		 *  unless we need to read old data encrypted without it. */
 		$keyder = ($legacy) ? "" : "-pbkdf2";
 		$md = ($legacy) ? "md5" : "sha256";
+		$iter = ($legacy) ? "" : "-iter 300000";
```
jim-p, line 42: We should make this at least 310000 if I'm reading this right:

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

That's the closest I could find to a current recommendation for PBKDF2 iterations.

Would be good to add a note and link to that URL above that line so we have a reference for it to show it wasn't pulled out of thin air. If not that URL, link to the Redmine issue #.

jim-p (changes requested):
Looks promising just needs a small change or two.

Thanks!

src/etc/inc/crypt.inc
PhilZ-cwm6: I updated the commit
Please note that this would make old encrypted backups non recoverable from the GUI !
I am not really sure if a GUI fix should be added for it, but would be out of my scope and time
However, clearly staying with 10000 is a bad design for a firewall

jim-p: If it would cause a failure like that, then we couldn't accept it without the extra accommodation included. There is already an example of it falling back to legacy, but basically it could do this:

* Try to decrypt with the large number of iterations.
* If that fails, try to decrypt with the default number of iterations
* If that fails, use the legacy options

What I would do is add a new parameter to `crypt_data()` which specifies the number of iterations, and then it would be easy to adapt the code to handle the extra choice.

If you aren't able to take that part on I can look into doing it next week.

PhilZ-cwm6: Sure please, as I am not used to pfsense sources, so it would take me too much time running through it
Hope you can follow up and do all the commits you think are needed
Best regards

PhilZ-cwm6:
I updated my comments, the changes will not break import of old backups with a lower iterations count
The only thing that must be tested is that a 350000-500000 iterations count is not too slow on some systems
In my tests, even ARM android devices of 5 years ago can cope quickly with such counts

PhilZ-cwm6:
sorry, I can confirm that decryption currently needs **the exact** iterations count set when it was encrypted, else it will fail

PhilZ-cwm6:
The patch should properly work:
`$iter = ($legacy) ? "" : "-iter 500000";`

So, if legacy is set, the iterations count is left to the current openssl default
Decrypting old backups should work properly with this commit because the iterations count will be properly set for legacy backup files. So old backups must be set as legacy after this update

jim-p:
It's not that simple, when legacy is set, other parameters are also changed which would make it fail. It needs another step between the current code and "legacy" style which uses other modern options and only changes the default iterations.

PhilZ-cwm6:
I see, legacy would drop to md5 hash... and current 2.5 is in between
By the way, the default should be set to 10'000 to avoid breaking the code if openssl changes their default count
Sure the best would be to change the GUi and add a few options like you suggested. Feel free to close this commit and open your own with a full patch regarding all the mentioned limitations

PR #4547, opened 2021-12-11
New states widget and byte sorting #12586

kc1r74p:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/12586
- [X] Ready for review

Side related issue: https://redmine.pfsense.org/issues/6158

Added states widget on basis of Firewall log widget.
Sorted desc on max IO by default.

Added sort for IO on diag_dump_states.php

Result:
grafik


kc1r74p:
Currently playing arround with some more informations, 
however will require 3rd party API and potentially slow down dashboard.

Will not include it in this feature for now. Maybe some opinions ?
grafik

jim-p:
See comments on Redmine.

PR #4549, opened 2021-12-20
Adds missing settings and zero and floating point support for those settings to the limiter scheduler fq_pie and the aqm pie.

hpeters:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/12003
- [ ]


jim-p (approved):
Looks OK at a glance. Too late for this release but we can look at it for the next release.

PR #4551, opened 2022-01-14
add an option to disable auto-creation of static routes for dpinger

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12687
- [x] Ready for review

### Summary

Users may NOT want static routes added for every gateway monitor IP.

This has been discussed before, and I've been running for years with this tiny patch. **_No kittens have died._** 🙀 

The original code was written probably over a decade ago, in the era of `apinger` (anyone remember that?)

`apinger` did not bind to an interface, so the static routes were needed, but I don't think they are anymore now that we've got `dpinger`. There are a lot of footguns when adding routes to anycast IPs (which is the suggested best-practice) since they tend to be DNS servers. I would bet somewhere ~50% of all installs just use 8.8.8.8 etc. So your DNS gets hosed if one of your gateways goes down in Multiwan, defeating the entire purpose.

### Changes

This PR adds a checkbox to System → Advanced → Miscellaneous to disable the auto-added static routes:

You can also disable them per-gateway, by ticking the **Do not add static route** checkbox under System → Routing → Gateway → Edit

### Evidence

1. Directly from the code [`/etc/inc/gwlb.inc`](https://github.com/pfsense/pfsense/blob/ba815f3d219e5bdf404be859e723db2ff0c9258c/src/etc/inc/gwlb.inc#L240):

2. From the [docs](https://docs.netgate.com/pfsense/en/latest/routing/gateway-configure.html):

### Related discussions

- [Changing the monitor IP of a gateway caused all traffic to that IP to go over said gateway, despite not being default - bug? r/PFSENSE](https://www.reddit.com/r/PFSENSE/comments/rcmho8/changing_the_monitor_ip_of_a_gateway_caused_all/)
- [Gateway monitor IPs are being put into the routing table](https://forum.netgate.com/topic/144488/gateway-monitor-ips-are-being-put-into-the-routing-table/19)
- [Setting host as Monitored IP makes it unreachable? | Netgate Forum](https://forum.netgate.com/topic/127191/setting-host-as-monitored-ip-makes-it-unreachable)
- [Bug #2514: static routes for monitor IPs should be removed - pfSense - pfSense bugtracker](https://redmine.pfsense.org/issues/2514)
- [Bug #12608: WireGuard tunnels monitored by dpinger causing system to stop routing completely in certain situations - pfSense Packages - pfSense bugtracker](https://redmine.pfsense.org/issues/12608)

### Testing

- Tested on a 5100 running 2.4.x and 2.5.x, as well as a 6100 running pfS+ 22.01 rc.
- If anyone would like to test this on their system, install the System Patches package and add commit `ad2dd2c1debaf2281eaed685ad8464c027c3b0b1`.

### DIFF
--- a/src/usr/local/www/system_advanced_misc.php
+++ b/src/usr/local/www/system_advanced_misc.php
@@ -304,6 +304,15 @@ $section->addInput(new Form_Checkbox(
 	'down, the rule is created omitting the gateway. This option overrides that '.
 	'behavior by omitting the entire rule instead.');
 
+$section->addInput(new Form_Checkbox(
+	'dpinger_dont_add_static_routes',
+	'Static routes',
+	'Do not add static routes for gateway monitor IP addresses',
+	$pconfig['dpinger_dont_add_static_routes']
+))->setHelp('By default the firewall adds static routes for gateway monitor IP addresses '.
+	'to ensure traffic to the monitor IP address leaves via the correct interface. '.
+	'Enabling this checkbox overrides that behavior.');
+
 $form->add($section);
 $section = new Form_Section('RAM Disk Settings (Reboot to Apply Changes)');
 
--- a/src/usr/local/www/system_gateways_edit.php
+++ b/src/usr/local/www/system_gateways_edit.php
@@ -72,6 +72,7 @@ if (isset($id) && $a_gateways[$id]) {
 	$pconfig['losslow'] = $a_gateways[$id]['losslow'];
 	$pconfig['losshigh'] = $a_gateways[$id]['losshigh'];
 	$pconfig['monitor'] = $a_gateways[$id]['monitor'];
+	$pconfig['dpinger_dont_add_static_route'] = isset($a_gateways[$id]['dpinger_dont_add_static_route']);
 	$pconfig['monitor_disable'] = isset($a_gateways[$id]['monitor_disable']);
 	$pconfig['action_disable'] = isset($a_gateways[$id]['action_disable']);
 	$pconfig['data_payload'] = $a_gateways[$id]['data_payload'];
@@ -222,6 +223,17 @@ $group->add(new Form_Input(
 	'load balancer entries. Use this if the gateway does not respond to ICMP echo '.
 	'requests (pings).');
 $section->add($group);
+
+$section->addInput(new Form_Checkbox(
+	'dpinger_dont_add_static_route',
+	'Static route',
+	'Do not add static route for gateway monitor IP address via the chosen interface',
+	$pconfig['dpinger_dont_add_static_route']
+))->setHelp('By default the firewall adds static routes for gateway monitor IP addresses '.
+	'to ensure traffic to the monitor IP address leaves via the correct interface. '.
+	'Enabling this checkbox overrides that behavior.');
+
+$form->add($section);
 
 $section->addInput(new Form_Checkbox(
 	'force_down',
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -239,18 +239,21 @@ function setup_gateways_monitor() {
 			 * Add static routes for each gateway with their monitor IP
 			 * not strictly necessary but is a added level of protection.
 			 */
-			if (is_ipaddrv4($gateway['gateway']) && $gateway['monitor'] != $gateway['gateway']) {
-				log_error(sprintf(gettext('Removing static route for monitor %1$s and adding a new route through %2$s'), $gateway['monitor'], $gateway['gateway']));
-				if (interface_isppp_type($gateway['friendlyiface'])) {
-					route_add_or_change($gateway['monitor'],
-					    '', $gateway['interface']);
-					system_staticroutes_configure($gateway['friendlyiface']);
-				} else {
-					route_add_or_change($gateway['monitor'],
-					    $gateway['gateway']);
-				}
+			if (!isset($config['system']['dpinger_dont_add_static_routes']) &&
+					!isset($gateway['dpinger_dont_add_static_route'])) {
+				if (is_ipaddrv4($gateway['gateway']) && $gateway['monitor'] != $gateway['gateway']) {
+					log_error(sprintf(gettext('Removing static route for monitor %1$s and adding a new route through %2$s'), $gateway['monitor'], $gateway['gateway']));
+					if (interface_isppp_type($gateway['friendlyiface'])) {
+						route_add_or_change($gateway['monitor'],
+						    '', $gateway['interface']);
+						system_staticroutes_configure($gateway['friendlyiface']);
+					} else {
+						route_add_or_change($gateway['monitor'],
+						    $gateway['gateway']);
+					}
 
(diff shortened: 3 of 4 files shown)


dennypage:
> The original code was written probably over a decade ago, in the era of `apinger` (anyone remember that?)

Yes, I do.  :slightly_smiling_face:

luckman212:
@dennypage Do you have any thoughts on this PR?

jim-p:
Why make this a global option instead of a per-gateway option? If users want to do this selectively, it makes more sense to allow them to disable the behavior on a case-by-case basis rather than one single master on/off that forces them to add manual routes for the cases they do want to have routes.

A gateway option to add the static routes that defaults to enabled would be a lot more flexible, IMO.

luckman212:
@jim-p Thanks for looking. I thought about that but IMO if they want static routes, shouldn't they be adding those on the proper Static Routes page? 



The only use case I could come up with for having it be a per-gateway setting is a PPPoE connection. Because I don't have any to test with, I wasn't sure if the binding was required there for dpinger to do the right thing. Anyway, I'm happy to make the setting per gateway if everyone agrees that's the right thing to do.

jim-p:
It's not that they don't "belong" there necessarily, but if there is an automatic way to do it, it puts an undue burden on the user for it to be "all or nothing" in cases like this. The more gateways they have, the more of an impact this option would have if they wanted to disable the route for just one gateway out of many.

There could still be a global option but a fine-grained way to control it is more user friendly.

luckman212:
Ok, I'll do it.

luckman212:
@jim-p I added a per-gateway setting. So this change can be effected using either the original global override or per-gateway now

luckman212:
Is this ok now? Any other changes you want to see here @jim-p ? Thanks

jim-p:
Might be a while before I can look it over in detail, we're pretty busy with the 22.01/2.6.0 release testing and such at the moment, but I'll check it out when I have some spare cycles.

luckman212:
Ok got it, I was naively hoping it could squeak into 22.01 😊

jim-p:
Ah, nope, we closed the window on new features a while ago, far too close to get anything else in but significant regressions at this point. But the next release is not too far off (22.05).

luckman212:
I've been running 22.05 dev builds now at home (6100) for a week or so. Hoping when the dust from 22.01/2.6 settles this can be considered for a merge!

luckman212:
@jim-p Any issues with this? I don't know when the cutoff is for 22.05 but it seems like it's going to be soon. Could you let me know if you need any more changes?

src/usr/local/www/system_advanced_misc.php
```diff
@@ -304,6 +304,15 @@
 	'down, the rule is created omitting the gateway. This option overrides that '.
 	'behavior by omitting the entire rule instead.');
 
+$section->addInput(new Form_Checkbox(
+	'dpinger_dont_add_static_routes',
+	'Static routes',
+	'Do not add static routes for gateway monitor IPs',
```
jim-p, line 310: Change `IPs` to `IP Addresses`

src/usr/local/www/system_advanced_misc.php
```diff
@@ -304,6 +304,15 @@
...
+	$pconfig['dpinger_dont_add_static_routes']
+))->setHelp('By default, static routes are added for gateway monitor IPs '.
```
jim-p, line 312: Change `IPs` to `IP Addresses`

src/usr/local/www/system_advanced_misc.php
```diff
@@ -304,6 +304,15 @@
...
+	'to ensure traffic to the monitor IP leaves via the correct interface. '.
```
jim-p, line 313: Change `IP` to `IP Address`

src/usr/local/www/system_gateways_edit.php
```diff
@@ -223,6 +224,17 @@
 	'requests (pings).');
 $section->add($group);
 
+$section->addInput(new Form_Checkbox(
+	'dpinger_dont_add_static_route',
+	'Static route',
+	'Do not add static route for gateway monitor IP via the chosen interface',
```
jim-p, line 230: Change `IP` to `IP Address`

src/usr/local/www/system_gateways_edit.php
```diff
@@ -223,6 +224,17 @@
...
+	$pconfig['dpinger_dont_add_static_route']
+))->setHelp('By default, static routes are added for gateway monitor IPs '.
```
jim-p, line 232: Change `IPs` to `IP Addresses`

src/usr/local/www/system_gateways_edit.php
```diff
@@ -223,6 +224,17 @@
...
+	'to ensure traffic to the monitor IP leaves via the correct interface. '.
```
jim-p, line 233: Change `IP` to `IP Address`

src/usr/local/www/system_gateways_edit.php
jim-p, line 232: Change `, static routes are added` to `the firewall adds static routes`

src/usr/local/www/system_advanced_misc.php
jim-p, line 312: Change `, static routes are added` to `the firewall adds static routes`

jim-p (changes requested):
Just some minor grammar nits to correct, looks OK otherwise, thanks!

luckman212:
@jim-p Ok, I think I got all of them with this last commit. Hope you don't mind that I spelled "addresses" with a lowercase "a". Seemed like the right thing to do.

jim-p (approved):

src/usr/local/www/system_gateways_edit.php
```diff
@@ -223,6 +224,17 @@
...
+$section->addInput(new Form_Checkbox(
+	'dpinger_dont_add_static_route',
+	'Static route',
+	'Do not add static route for gateway monitor IP address via the chosen interface',
+	$pconfig['dpinger_dont_add_static_route']
+))->setHelp('By default the firewall adds static routes for gateway monitor IP addresses '.
+	'to ensure traffic to the monitor IP address leaves via the correct interface. '.
+	'Enabling this checkbox overrides that behavior.');
+
+$form->add($section);
```
luckman212, line 236: Sorry, I don't know how this extra `$form->add($section);` snuck in.

luckman212:
@jim-p Sorry for the last minute message here, but as I was working on the v2 of this patch, I realized a critical error slipped through. 

The global `dpinger_dont_add_static_routes` config flag never gets properly set in config.xml with the currently merged version of this patch, because of a missing piece of code in `/usr/local/pfSense/include/www/system_advanced_misc.inc`:

```php
if ($post['dpinger_dont_add_static_routes'] == "yes") {
    $config['system']['dpinger_dont_add_static_routes'] = true;
} else {
    unset($config['system']['dpinger_dont_add_static_routes']);
}
```
Since this is already merged I don't think I can push more changes to it, right? I have the corrected code at #4590 but not sure how to get that pulled in.

PR #4553, opened 2022-01-22
make dashboard display "Inactive" for Hardware crypto instead of empty string when not in use

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12714
- [x] Ready for review

Very small patch that just makes the Dashboard display for **Hardware crypto** a bit friendlier (imo) when there are no actively loaded crypto modules. Instead of a blank row, it will show "Inactive" in line with other sections e.g. Kernel PTI etc.

**without** patch:
image

**with** patch:
image

### DIFF
--- a/src/usr/local/www/includes/functions.inc.php
+++ b/src/usr/local/www/includes/functions.inc.php
@@ -320,7 +320,7 @@ function crypto_accel_get_algs($crypto) {
 		$algs_str .= $alg;
 	}
 
-	return ($algs_str);
+	return (!empty($algs_str) ? $algs_str : gettext("Inactive"));
 }
 
 


luckman212:
Ditto on this one, any issues found?

src/usr/local/www/includes/functions.inc.php
```diff
@@ -320,7 +320,7 @@ function crypto_accel_get_algs($crypto) {
 		$algs_str .= $alg;
 	}
 
-	return ($algs_str);
+	return (!empty($algs_str) ? $algs_str : "Inactive");
```
jim-p, line 323: This should be `gettext("Inactive")` so the string gets localized.

jim-p (changes requested):
Just needs one small change, thanks!

src/usr/local/www/includes/functions.inc.php
luckman212: Thanks @jim-p - I'll remember that next time. I pushed the update.

jim-p (approved):
Tested it both ways (with and without hardware crypto available) and it worked as expected. Looks good.

PR #4554, opened 2022-01-24
Change get_gwgroup_members to include interface in link-local IPv6 ad…

lmcquade:
…dresses (Bug #12721)

- [X] Redmine Issue: https://redmine.pfsense.org/issues/12721
- [X] Ready for review

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -1343,7 +1343,12 @@ function get_gwgroup_members_inner($group, $gateways_status, $gateways_arr, $vip
 						$groupmember = array();
 						$groupmember['gw'] = $member;
 						$groupmember['int'] = $int;
-						$groupmember['gwip'] = $gatewayip;
+						if(is_ipaddrv6($gatewayip) && is_linklocal($gatewayip))
+						{
+							$groupmember['gwip'] = $gatewayip + '%' + $int;
+						} else {
+							$groupmember['gwip'] = $gatewayip;							
+						}
 						$groupmember['weight'] = isset($gateway['weight']) ? $gateway['weight'] : 1;
 						if (is_array($gwvip_arr[$group['name']]) && !empty($gwvip_arr[$group['name']][$member])) {
 							$groupmember['vip'] = $gwvip_arr[$group['name']][$member];


src/etc/inc/gwlb.inc
```diff
@@ -1343,7 +1343,12 @@ function get_gwgroup_members_inner($group, $gateways_status, $gateways_arr, $vip
 						$groupmember = array();
 						$groupmember['gw'] = $member;
 						$groupmember['int'] = $int;
-						$groupmember['gwip'] = $gatewayip;
+						if(is_ipaddrv6($gatewayip) && is_linklocal($gatewayip))
```
jim-p, line 1346: Add a space after `if` ([ref](https://docs.netgate.com/pfsense/en/latest/references/developer-style-guide.html#indent-style))

src/etc/inc/gwlb.inc
```diff
@@ -1343,7 +1343,12 @@ function get_gwgroup_members_inner($group, $gateways_status, $gateways_arr, $vip
...
+						{
```
jim-p, line 1347: Please use ["K&R, BSD KNF variant style" formatting](https://docs.netgate.com/pfsense/en/latest/references/developer-style-guide.html#indent-style).

src/etc/inc/gwlb.inc
```diff
@@ -1343,7 +1343,12 @@ function get_gwgroup_members_inner($group, $gateways_status, $gateways_arr, $vip
...
+							$groupmember['gwip'] = $gatewayip + '%' + $int;
```
jim-p, line 1348: `$int` as you have it here is the internal name of the interface (e.g. `lan`, `wan`, `opt4`) and not what you need, which is the OS level interface (e.g. `igb0`, `ix2`, `igc3`).

You need to run it through something like `get_real_interface($int)` and sanity check the output before using it.

jim-p (changes requested):

src/etc/inc/gwlb.inc
lmcquade: Fixed in 671fd0d

src/etc/inc/gwlb.inc
lmcquade: Fixed in 671fd0d

src/etc/inc/gwlb.inc
lmcquade: I do not believe this is the case.  I added log lines to be sure, and the value of `$int` at this time is `igb0` for me.

jim-p: You are right, I was looking at the field name in the configuration but this section of code retrieved the gateway another way which contains the right value there.

jim-p (approved):

PR #4556, opened 2022-02-05
Code cleaning in filter.inc for pfctl -F call which does not affect existing behavior 

markfrancisonly:
filter_flush_state_table() in filter.inc calls pfctl -F state which is a typo that just happens to work ... on the man page for [pfctl](https://www.freebsd.org/cgi/man.cgi?query=pfctl):

	     -F	nat	   Flush the NAT rules.
	     -F	queue	   Flush the queue rules.
	     -F	rules	   Flush the filter rules.
	     -F	states	   Flush the state table (NAT and filter).
	     -F	Sources	   Flush the source tracking table.
	     -F	info	   Flush the filter information	(statistics that are
			   not bound to	rules).
	     -F	Tables	   Flush the tables.
	     -F	osfp	   Flush the passive operating system fingerprints.
	     -F	all	   Flush all of	the above.

and the docs match the code [pfctl.c](https://github.com/freebsd/freebsd-src/blob/main/sbin/pfctl/pfctl.c)

also, filter_flush_nat_table() in filter.inc is never used and was removed.

note this is my first pr to pfsense, let me know if submission channel process isn't correct,  there isn't an existing redmine "issue" since this isn't a bug but a code clean request


markfrancisonly:
will resubmit once my rep is repaired

PR #4558, opened 2022-02-20
Frontend updates

GChuf:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12838
- [x] Ready for review

Currently, there are a lot of non-minified files (/js/vendor folder), reducing front-end performance. I've minified those files.
I also compressed some png images for smaller download sizes.

There are a couple of updates as well, and removal of obsolete "filebrowser" files.


GChuf:
The four .js files under `pfsense\src\usr\local\www\js` could also be compressed, but since those are maitained by this community, I wasn't sure if you would agree with creating .min files for those as well - since it would mean more work anytime those files would be changed.

Jon8RFC:
Good idea!  The png images can be compressed a little further from 23,816 bytes to 20,732 bytes:
https://drive.google.com/file/d/1SfLR7hkek9VxlUVKy5cnIAuh3z7mVUiY/view?usp=sharing

GChuf:
@Jon8RFC thanks, I see you crushed the files :)
I've added them.

I noticed there were some other pngs inside jquery-ui as well, which I haven't compressed (&crushed) yet.
While I was at it, I also updated jquery-ui to v1.13.1 - according to the changelog there were some security issues fixed and some deprecated APIs removed.

jim-p:
I don't see this getting accepted as is. It is difficult to verify that the content of the files is unmodified compared to the original version. If we were to undertake something like this, the work would have to be done by internal devs, not outside sources.

The web server already employs gzip compression when sending content to clients for JS as well so such tactics are largely unnecessary.

GChuf:
@jim-p I understand it's difficult to verify the integrity of minified .js files. However minification is not all about download and upload, it's also about processing times.
Consider the following:
since the files are cached the majority of time anyway, it's not really about small download size, it's more about slower clients processing the code for drawing graphs and such faster. The server also benefits from smaller initial file sizes (less gzip processing, for example).

How about the other commits? The dependencies updates and image compression? The work has already been done. Let me know what you think.

PR #4560, opened 2022-03-07
Always clean tmp backups & prevent legacy backup

PhilZ-cwm6:
- If backup fails for any reason (exp I/O): prevent creating legacy backups or backups with lower iterations count
- If for some reason the decryption takes too long to finish (CPU overloaded during decryption), and decryption needs to fall back to previous version or legacy, new temp files can be created and old ones will not be removed. This patch ensure that any created temp files are properly deleted

- [X] Redmine Issue: https://redmine.pfsense.org/issues/12897
- [X] Ready for review


jim-p:
I took a slightly different approach, see my comments on Redmine.

PR #4561, opened 2022-03-18
Fix DHCP leases sort

NobleKangaroo:
### Description
Commit c4c730b cleaned up a lot of the DHCP lease code, but unfortunately a result of this cleanup was that DHCP leases and reservations are now sorted backwards:
- DHCP leases are now at the bottom of the list (they were at the top before)
- Higher numbered VLANs are sorted above lower numbered VLANs (they were sorted numerically before)

This PR is a minor change that simply restores the sort order to its original function.

- [ ] Redmine Issue: 
- [X] Ready for review

### Previous sorting: *(restored by this PR)*
image

### Current sort order:
image

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -105,6 +105,8 @@ if ($_REQUEST['order']) {
 	usort($leases['lease'], function($a, $b) {
 		return strcmp($a[$_REQUEST['order']], $b[$_REQUEST['order']]);
 	});
+} else {
+	sort($leases['lease']);
 }
 
 /* only print pool status when we have one */
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -106,7 +106,12 @@ if ($_REQUEST['order']) {
 		return strcmp($a[$_REQUEST['order']], $b[$_REQUEST['order']]);
 	});
 } else {
-	sort($leases['lease']);
+	usort($leases['lease'], function($a, $b) {
+		if ($a['type'] === $b['type']) {
+			return strnatcmp($a['ip'], $b['ip']);
+		}
+		return strcmp($a['type'], $b['type']);
+	});
 }
 
 /* only print pool status when we have one */


NobleKangaroo:
Changed to a more mature sorting algorithm to properly handle IP addresses and sort them in human-readable format (e.g. 10.10.10.20 before 10.10.10.100) with [`strnatcmp`](https://www.php.net/manual/en/function.strnatcmp.php). Dynamic IPs still sorted above static as they were. I believe the sorting algorithm should now effectively be 1:1 with how it was in the past. I added a bunch of test cases to validate this is the case.

image

jim-p:
I'm not seeing the behavior you are seeing before/after the commit you referenced. If I revert that commit, the only difference is whether statics are at the start or end. I don't see anything that sets `$_REQUEST['order']` so that bit of code may be obsolete.

Sorting the array might seem like a good idea but might not scale well for large databases.

The table is JS sortable, though. If you want to sort them just click the header.

Also, changes such as this may seem trivial/minor but they are not. They still need a Redmine as there could be a very real and not insignificant performance impact from it.

NobleKangaroo:
Hi @jim-p,

First of all, thanks for looking into this. Let me preface this by just saying that the intent of this PR was only to restore the behavior that was in place before the aforementioned commit c4c730b.
&nbsp;


> I'm not seeing the behavior you are seeing before/after the commit you referenced. If I revert that commit, the only difference is whether statics are at the start or end.

Sure, it's a minor gripe but the behavior change was not intended, and arose from the removal of the `sort` function ([this line](https://github.com/pfsense/pfsense/commit/c4c730b5aac1305f1a2f132d31f04f227d17f4d7#diff-9e41a688d21300e57b201ec2b08eb697a4dd67f9b5134389608a5cb6ed55019aL310)) and other logic, now leaving the order of the HTML table to be determined directly by the order of the contents in `/var/dhcpcd/var/db/dhcpd.leases`.
&nbsp;


> I don't see anything that sets `$_REQUEST['order']` so that bit of code may be obsolete.

Yeah I saw that `$_REQUEST['order']` bit and couldn't find what was referencing it, so it may just be stale code.
&nbsp;


> Sorting the array might seem like a good idea but might not scale well for large databases.

I don't disagree, but the array was sorted prior to the commit I mentioned. Simply sorting strings is very computationally cheap. It would be different had there been mathematics involved (hashing, etc). There was a sort to begin with, so there should be no consequence in re-implementing a sort to replace the one that was removed.
&nbsp;


> The table is JS sortable, though. If you want to sort them just click the header.

I also don't disagree with you on this one. The JS-sortable table works fine (good job on that), but quite simply it's something extra that we didn't have to do before.
&nbsp;


> Also, changes such as this may seem trivial/minor but they are not. They still need a Redmine as there could be a very real and not insignificant performance impact from it.

Understand. I agree with you there. I can open a Redmine ticket, if you're willing to spend a little time actually looking more into this (as far as performance goes). I'll agree that the scalability of the `sort` should be tested, if nothing else, even though there was a `sort` to begin with.
&nbsp;


Appreciate your looking into this.

jim-p:
> Sure, it's a minor gripe but the behavior change was not intended, and arose from the removal of the `sort` function ([this line](https://github.com/pfsense/pfsense/commit/c4c730b5aac1305f1a2f132d31f04f227d17f4d7#diff-9e41a688d21300e57b201ec2b08eb697a4dd67f9b5134389608a5cb6ed55019aL310)) and other logic, now leaving the order of the HTML table to be determined directly by the order of the contents in `/var/dhcpcd/var/db/dhcpd.leases`.  

It's been three years so it's hard to guess what was intended but not a part of the commit message and redmine comments, but we frequently encounter large lease databases so it's possible that it was intentional for optimization or other reasons. Also for a list of leases it's more ideal to have the dynamic ones first as the statics never change, so they are less relevant to what the user wants to see immediately.

> I don't disagree, but the array was sorted prior to the commit I mentioned. Simply sorting strings is very computationally cheap. It would be different had there been mathematics involved (hashing, etc). There was a sort to begin with, so there should be no consequence in re-implementing a sort to replace the one that was removed.  

That may be true in some cases, but we are dealing with PHP so any extra operations we can remove to help it scale improve it overall.

> I also don't disagree with you on this one. The JS-sortable table works fine (good job on that), but quite simply it's something extra that we didn't have to do before.  

Between the sorting and filtering functions of the page I'm not sure you're using the page the way most users do. If they're looking for something specific, it's easy to filter the list or sort it and get an overview.

> Understand. I agree with you there. I can open a Redmine ticket, if you're willing to spend a little time actually looking more into this (as far as performance goes). I'll agree that the scalability of the `sort` should be tested, if nothing else, even though there was a `sort` to begin with.  

I don't see us changing it back after three years for only one request. You can always add it back as a local patch on your personal systems, though. The system patches package makes it trivial to add it in and auto-apply the change so it's present even after updating.

PR #4562, opened 2022-03-18
Increase max firewall log entries

NobleKangaroo:
### Description
Simple change that performs the following:
- Fixes a minor typo in the module's name
- Increase the max 'number of entries' in firewall logs widget from 20 to 50 (those of us with larger screens appreciate seeing more data)

This PR is a minor change to expand the number of visible log entries.

- [ ] Redmine Issue: 
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/widgets/widgets/log.widget.php
+++ b/src/usr/local/www/widgets/widgets/log.widget.php
@@ -205,7 +205,7 @@ events.push(function(){
 
 	// Create an object defining the widget refresh AJAX call
 	var logsObject = new Object();
-	logsObject.name = "Gateways";
+	logsObject.name = "Firewall Logs";
 	logsObject.url = "/widgets/widgets/log.widget.php";
 	logsObject.callback = logs_callback;
 	logsObject.parms = postdata;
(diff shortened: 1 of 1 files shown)


jim-p:
You can set the number of entries to whatever you like in the widget settings, there is no compelling reason to change the default for everyone on every installation to suit the scenario you describe.

NobleKangaroo:
> You can set the number of entries to whatever you like in the widget settings, there is no compelling reason to change the default for everyone on every installation to suit the scenario you describe.

Hi @jim-p,

I think you missed the point here. That's absolutely not what this does. The default can be set to whatever, but there is a hard limit of 20 imposed by the UI ( in the code that was changed, `max="20"` ). This PR only changes that hard limit to match [the other spot in the code that pulls no more than 50 here](https://github.com/pfsense/pfsense/blob/e01ea791ebb90b4cd24279e42281b6002cbd6ce4/src/usr/local/www/widgets/widgets/log.widget.php#L99).

Hope that clears things up.

jim-p:
You are right I misread where the change was. Sorry about that.

PR #4563, opened 2022-03-19
Improve default sha512 password hashing rounds

PhilZ-cwm6:
By default, sha512 only does 5000 rounds of encryption which is very weak
Increase rounds to 800k to provide same bcrypt default 10 rounds password resiliance

- [X] Redmine Issue: https://redmine.pfsense.org/issues/12962
- [X] Ready for review


PhilZ-cwm6:
Submitted a tested patch on https://redmine.pfsense.org/issues/12863

PR #4564, opened 2022-03-21
Packet capture: add Clear Capture button

PhilZ-cwm6:
When there is a capture file, show a "Clear Capture" button to delete the last capture without having to drop to cli

- [X] Redmine Issue: https://redmine.pfsense.org/issues/12968
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -280,7 +280,6 @@ if ($_POST) {
 			if (file_exists($fp.$fn)) {
 				unlink ($fp.$fn);
 			}
-
 		} elseif ($_POST['stopbtn'] != "") {
 			$action = gettext("Stop");
 			$processes_running = trim(shell_exec("/bin/ps axw -O pid= | /usr/bin/grep tcpdump | /usr/bin/grep {$fn} | /usr/bin/egrep -v '(pflog|grep)'"));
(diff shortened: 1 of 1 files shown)


src/usr/local/www/diag_packet_capture.php
```diff
@@ -297,6 +296,11 @@ function hostport_array_fixer($items,$type) {
 		} elseif ($_POST['downloadbtn'] != "") {
 			//download file
 			send_user_download('file', $fp.$fn);
+		} elseif ($_POST['clearbtn'] != "") {
+			//clear log file
+			if (file_exists($fp.$fn)) {
```
jim-p, line 301: We have a function for this test+action, `unlink_if_exists($filename)`.

src/usr/local/www/diag_packet_capture.php
```diff
@@ -492,6 +496,13 @@ function hostport_array_fixer($items,$type) {
 		'fa-download'
 	))->addClass('btn-primary');
 
+	$form->addGlobal(new Form_Button(
+		'clearbtn',
+		'Clear Log',
```
jim-p, line 501: This isn't a log, it's the previous packet capture. It would be more correct to label this `Clear Last Capture`.

jim-p (changes requested):

src/usr/local/www/diag_packet_capture.php
PhilZ-cwm6: I wanted to be conservative and use same code as the file.
As per your request, I modified the two occurrences, the exiting one and the one I added

src/usr/local/www/diag_packet_capture.php
PhilZ-cwm6: I used that at first, but the button gets too large esthetically...
What about a shorter `Clear Capture` to match the the existing buttons `View Capture` and `Download Capture` ?
I push a separate commit so that you can decide up to where you want to merge it

jim-p: Sure, that's reasonable.

jim-p (approved):

PR #4567, opened 2022-04-01
Packet Capture: add preview results

PhilZ-cwm6:
- allow preview results while a capture is still running
- add a capture summary field
- count: if empty value, properly set it to default 100 instead of 0
- do not pass empty argument to tcpdump when no filter options are posted
- do not check form validity when clicking other buttons than start
- more compact GUI (spannable help tool tips)
- info changes and fixes

- [X] Redmine Issue: https://redmine.pfsense.org/issues/13017
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/diag_packet_capture.php
+++ b/src/usr/local/www/diag_packet_capture.php
@@ -135,6 +135,11 @@ function hostport_array_fixer($items,$type) {
 	}
 }
 
+function tcpdump_get_running_process($f){
+	$processcheck = trim(shell_exec("/bin/ps axw -O pid= | /usr/bin/grep '/usr/sbin/[t]cpdump.*{$f}'"));
+	return $processcheck;
+}
+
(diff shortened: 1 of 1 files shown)


PhilZ-cwm6:
@jim-p 
In the redmine page, I added a patch that applies to the current master branch if it is easier to test/review

crawc:
I found an issue while testing that  I'll comment on...

If you have no OpenVPN servers/clients you get an error on line 171:
`	if (is_array($config['openvpn']["openvpn-{$mode}"])) {` 
Since `$config['openvpn']` is a sting, not an array. I found that if you change it to just...
`	if (is_array($config['openvpn'])) {`
It works as desired.

PhilZ-cwm6:
Fixing it will depend on the fact that devs are interested or not into merging it

marcos-ng:
Implemented with https://redmine.pfsense.org/issues/13382

PR #4569, opened 2022-04-09
More flexible Client-Specific Override options

PhilZ-cwm6:
**Part 1: More flexible Client-Specific Override options GUI changes**
- add individual server overrides for the optional client settings (local routes/gateway, remote routes, DNS, NTP, Netbios...)
- add option to override topology when selecting "Remove All Server Options" (push-reset)
- add option to set gateways (useful when removing routes)
- add option to override Redirect IPv6 Gateway in addition to existing IPv4 option
- move all server override options to the "Override Configuration" group
- hide override options with a "Select Server Overrides" toggle
- disable un-needed overrides when "Remove All Server Options" (push-reset) is selected
- when "Redirect IPv4/6 Gateway" is toggled, the Local Routes fields are hidden/shown correspondingly like for the server edit page

**Small bug fixes**
- fix bug when "0" is entered in DNS/Domain... fields (using !empty() discards 0)
- fix bug: NTP and DNS options: use javascript instead of toggles class because the latter causes inverted results when repeating clicks too quickly

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12522
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/vpn_openvpn_csc.php
+++ b/src/usr/local/www/vpn_openvpn_csc.php
@@ -106,9 +106,9 @@ if (($act == "edit") || ($act == "dup")) {
 		$pconfig['dns_server4'] = $a_csc[$id]['dns_server4'];
 
 		if ($pconfig['dns_server1'] ||
-		    $pconfig['dns_server2'] ||
-		    $pconfig['dns_server3'] ||
-		    $pconfig['dns_server4']) {
+				$pconfig['dns_server2'] ||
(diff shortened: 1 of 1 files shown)


PhilZ-cwm6:
Changes can be previewed here:
https://philz-cwm6.github.io/pfsense_preview/vpn_openvpn_csc_v1dot3_gui.html

PhilZ-cwm6:
Messed the repo, sorry, I will commit a new one

PR #4570, opened 2022-04-09
Make Client-Specific Override options more flexible. Implement #12522

PhilZ-cwm6:
- add individual server overrides for the optional client settings (local routes/gateway, remote routes, DNS, NTP, Netbios...)
- add option to override topology when selecting "Remove All Server Options" (push-reset)
- add option to set gateways (useful when removing routes)
- add option to override Redirect IPv6 Gateway in addition to existing IPv4 option
- move all server override options to the "Override Configuration" group
- hide override options with a "Select Server Overrides" toggle
- disable un-needed overrides when "Remove All Server Options" (push-reset) is selected
- when "Redirect IPv4/6 Gateway" is toggled, the Local Routes fields are hidden/shown correspondingly like for the server edit part

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12522
- [x] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1696,6 +1696,7 @@ function openvpn_delete($mode, $settings) {
 	filter_configure();
 }
 
+// set client specific overrides
 function openvpn_resync_csc($settings) {
 	global $g, $openvpn_tls_server_modes;
 	if (isset($settings['disable'])) {
@@ -1719,8 +1720,40 @@ function openvpn_resync_csc($settings) {
 		$conf .= "push-reset\n";
 	}
 
+	if ($settings['topology_override']) {
+		$conf .= "push \"topology {$settings['topology']}\"\n";
+	}
+
 	if ($settings['remove_route']) {
 		$conf .= "push-remove route\n";
+	}
+
+	if ($settings['remove_iroute']) {
+		$conf .= "push-remove iroute\n";
+	}
+
+	if ($settings['remove_dnsdomain']) {
+		$conf .= "push-remove \"dhcp-option DOMAIN\"\n";
+	}
+
+	if ($settings['remove_dnsservers']) {
+		$conf .= "push-remove \"dhcp-option DNS\"\n";
+	}
+
+	if ($settings['remove_ntpservers']) {
+		$conf .= "push-remove \"dhcp-option NTP\"\n";
+	}
+
+	if ($settings['remove_netbios_ntype']) {
+		$conf .= "push-remove \"dhcp-option NBT\"\n";
+	}
+
+	if ($settings['remove_netbios_scope']) {
+		$conf .= "push-remove \"dhcp-option NBS\"\n";
+	}
+
+	if ($settings['remove_wins']) {
+		$conf .= "push-remove \"dhcp-option WINS\"\n";
 	}
 
 	if ($settings['local_network']) {
@@ -1739,9 +1772,34 @@ function openvpn_resync_csc($settings) {
 		$conf .= openvpn_gen_routes($settings['remote_networkv6'], "ipv6", false, true);
 	}
 
+	// push the ipv4 gateway if specified
+	if (!empty($settings['gateway']) && is_ipaddrv4($settings['gateway'])) {
+		$conf .= "push \"route-gateway {$settings['gateway']}\"\n";
+	}
+
+	/* Currently route-ipv6-gateway is not supported by openvpn
+	if (!empty($settings['gateway']) && is_ipaddrv4($settings['gateway'])) {
+		$conf .= "push \"route-ipv6-gateway {$settings['gateway']}\"\n";
+	}
+	*/
+
+	// Ping override options
+	if ($settings['ping_push']) {
+		$conf .= "push \"ping {$settings['ping_seconds']}\"\n";
+	}
+
+	if ($settings['ping_action_push']) {
+		$action = str_replace("_", "-", $settings['ping_action']);
+		$conf .= "push \"{$action} " .
+			"{$settings['ping_action_seconds']}\"\n";
+	}
+
+	// create client specific dhcp options and gateway redirection
 	openvpn_add_dhcpopts($settings, $conf);
 
+	// custom options are added after all client overrides, and before the tunnel options
 	openvpn_add_custom($settings, $conf);
+
 	/* Loop through servers, find which ones can use this CSC */
 	foreach (config_get_path('openvpn/openvpn-server', []) as $serversettings) {
 		if (isset($serversettings['disable'])) {
--- a/src/usr/local/www/vpn_openvpn_csc.php
+++ b/src/usr/local/www/vpn_openvpn_csc.php
@@ -35,6 +35,8 @@ require_once("pfsense-utils.inc");
 require_once("pkg-utils.inc");
 
 global $openvpn_tls_server_modes;
+global $openvpn_topologies;
+global $openvpn_ping_action, $openvpn_default_keepalive_interval, $openvpn_default_keepalive_timeout;
 
 init_config_arr(array('openvpn', 'openvpn-csc'));
 $a_csc = &$config['openvpn']['openvpn-csc'];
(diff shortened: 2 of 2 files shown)


PhilZ-cwm6:
Preview of changes on a playable page:
https://philz-cwm6.github.io/pfsense_preview/vpn_openvpn_csc_v1dot3_gui.html

If the GUI changes and settings are fine for you as a layout, I can proceed to adding actions in openvpn.inc

luckman212:
@PhilZ-cwm6 sidenote: I love how you posted a working version of the patch! Any tricks for doing that or is it just a copy of the branch on github?

PhilZ-cwm6:
> @PhilZ-cwm6 sidenote: I love how you posted a working version of the patch! Any tricks for doing that or is it just a copy of the branch on github?

I download the page with full scripts using SingleFile chrome extension. I then add it to Github Pages:
https://pages.github.com/

It takes a few minutes for the page to be online, so it can let you think it doesn't work. Your repo must be init with Pages like in the link I posted

PhilZ-cwm6:
About the last upstream fix for the 0 / empty, here's a gist with test code for all exceptions
https://gist.github.com/PhilZ-cwm6/4676368b77fc01d478f42b5c94f26a66

PhilZ-cwm6:
@jim-p 
I pushed the full changes with the actions set in openvpn.inc. That was the easier part as most of them were already there and just needed the settings implementation through GUI
A clickable preview of the page is here:
https://philz-cwm6.github.io/pfsense_preview/vpn_openvpn_csc-v1dot5.html

I fully tested all the functions and client side logs on connection.
The config shown in preview page for example does a `push-reset` followed by all needed actions for a working connection.
This simplifies the user interaction as it is not longer mandatory to enter custom options.
We can also remove selectively server-defined routes or DNS servers and push them to the custom config without a whole push-reset

I posted a patch in redmine

PhilZ-cwm6:
Let me know if there is interest / time in this for a planned merge, and I will commit the changes to resolve conflicts

jadazul:
Hi, this was ever commited? I have a request, but as it's a small one it could be lumped together with this Openvpn client change, please let me know your thoughts.

Currently I have a lot of Openvpn clients configured, but every time I reboot my pfSense all clients are restarted. There's missing an option to select which clients should be restarted at boot. This could be implemented with a checkbox on the client configuration and checking these flags on openvpn.inc (for what I investigated, not an expert and I could be wrong).

Thanks in advance!

PhilZ-cwm6:
I rebased and limited the changes to the client specific overrides
Hope it can be reviewed

Preview link:
https://philz-cwm6.github.io/pfsense_preview/vpn_openvpn_csc_v1dot3_gui.html

marcos-ng:
Nits:
- Netbios -> NetBIOS
- ipv4 / ipv6 -> IPv4 / IPv6
- There's a lot of redundancy in the individual overrides descriptions. Perhaps a list selection (like the Server List) would fit better here, then a single description can be used - e.g. 'A "push-remove" option will be sent to the client for the selected options, removing the respective server-defined option. Client-specific options on this page are not affected.'

Notes:
- I don't think the "Redirect Gateway" setting should be split into individual IPv4/6 options in the GUI. That should be left to the advanced options.
- Local/Remote Routes Settings should be kept in the Tunnel Settings group.
- Some changes were made to these pages (related to NetBIOS/WINS/NBDD) - conflicts need to be resolved.
- I've not looked at the code yet, but it seems like there are breaking changes for current configurations. These should be resolved in `upgrade_config.inc` so previously configured CSOs maintain the same behavior.

PhilZ-cwm6:
@marcos-ng 
I fixed the code as your suggestions.
Before committing, I have 2 questions about your comments:
> * There's a lot of redundancy in the individual overrides descriptions. Perhaps a list selection (like the Server List) would fit better here, then a single description can be used - e.g. 'A "push-remove" option will be sent to the client for the selected options, removing the respective server-defined option. Client-specific options on this page are not affected.'
> 
Is it better now if it looks like attached pic, rather than using a list without detailed info on the commands ?

image

> 
> * I don't think the "Redirect Gateway" setting should be split into individual IPv4/6 options in the GUI. That should be left to the advanced options.

That's the case already in the server options where it shows in GUI. So, if removed from CSOs, it will be a discrepancy between both interfaces, like it was before, with some options not set in GUI for CSOs while available in server

After your feedback, I will look at the `upgrade_config.inc`

PhilZ-cwm6:
I fixed the conflicts and changed as you suggested
- removed redundancy with shorter descriptions
- about your comment: "Redirect Gateway setting should be split into individual IPv4/6 options in the GUI. That should be left to the advanced options."

That's the case already in the server options where it shows in GUI. So, if removed from CSOs, it will be a discrepancy between both interfaces, like it was before, with some options not set in GUI for CSOs while available in server.
- upgrade_config.inc
Sorry, I did not find any issues because all old CSO settings were not changed in new code. Importing old config on new update brings the current CSO as they were configured and they shoudl work. I am not used to the code there, but let me know exactly where are the issues if any.

Thank you

marcos-ng:
Given that the redirect gateway option is split between IPv4/6 in the server settings, I agree with separating the option here as well. Since the previous config behavior is to redirect both IPv4/6, upgrade code is needed to handle `gwredir`. For example, on upgrade:
- Change the id name from `gwredir` to `gwredir4`
- In upgrade_config.inc, check for `gwredir` and enable/disable `gwredir4`/`gwredir6` accordingly

There should be an IPv6 equivalent to the new `IPv4 Gateway` option, and both should be placed above the "Redirect IPv4 Gateway" setting. Additionally, `push_reset` would also enable `server_overrides_enabled` in the upgrade code. There may be other instances of this, I have not checked all options.

Glancing at the code itself, I see some commented out code and it's not clear why it's there.

There are some changes to the fontawesome classes which need to be omitted - keep the original classes as is.

PhilZ-cwm6:
Really sorry, my previous commit, for some reason, missed some file changes and commits to fontawesome classes were tocuhed
I fixed the code properly, and I will look at your suggestions later

PhilZ-cwm6:
> Given that the redirect gateway option is split between IPv4/6 in the server settings, I agree with separating the option here as well. Since the previous config behavior is to redirect both IPv4/6, upgrade code is needed to handle `gwredir`. For example:
> 
> * Change the id name from `gwredir` to `gwredir4`
> * In upgrade_config.inc, check for `gwredir` and enable/disable `gwredir4`/`gwredir6` accordingly
> 

The previous code did properly implement both `gwredir` (ipv4) and `gwredir6` (ipv6) variables in the server GUI to redirect IPv4 and/or IPv6 traffic:
https://github.com/pfsense/pfsense/blob/master/src/etc/inc/openvpn.inc#L802-L807

The options were missing in CSO GUI and were only available as custom options. My patch just adds the existing `gwredir` and `gwredir6` to the CSO GUI code

> 
> There should be an IPv6 equivalent to the new `IPv4 Gateway` option
> 

The current code doesn't support pushing IPv6 gateway even in server:
`OpenVPN doesn't support this yet, clients do not recognize the option fully.`
https://github.com/pfsense/pfsense/blob/master/src/etc/inc/openvpn.inc#L1129-L1134

That's why, I added a commented code for IPv6 gateway `push "route-ipv6-gateway"`
Adding the IPv6 gateway option in CSO GUI will only provide a non working feature
Once available in pfsense, we can just uncomment the code and unviel the GUI option in CSO
When sending a `push-reset`, having the `IPv4 gateway` option in GUI is needed to get the config properly working without advanced custom options

> 
>  and both should be placed above the "Redirect IPv4 Gateway" setting.
> 
I just pushed a commit to follow your recommendation for this

> Additionally, `push_reset` would also enable `server_overrides_enabled` in the upgrade code. There may be other instances of this, I have not checked all options.
> 

That part I am still not sure what you mean as in my tests, I did not find issues to upgrade from old to new configs

> 
> Glancing at the code itself, I see some commented out code and it's not clear why it's there.
> 
I will clean it once a final commit is validated

> 
> There are some changes to the fontawesome classes which need to be omitted - keep the original classes as is.
> 
fixed my commit, sorry

marcos-ng:
Regarding the `topology` option, it states:
> It must match the actual topology specified in server. NOTE: This will perform a "push topology [selection]" without a previous "push-remove topology".

If the pushed topology "must match the actual topology specified in server", then why allow pushing a different one at all? If it must match, then `topology_override` is enough.

> That part I am still not sure what you mean as in my tests, I did not find issues to upgrade from old to new configs

If `push_reset` is enabled in an older config, then there should be an option in `upgrade_config.inc` that would automatically enable `server_overrides_enabled` in the config. Thinking about this some more however, the config doesn't need to track this at all since the option only serves as a display toggle. Regardless, we can instead use a list as mentioned previously - see below for some code to get it started.

```
//==== Form fields
$section->addInput(new Form_Select(
	'reset_options',
	'Reset Server Options',
	$pconfig['reset_options'] ?? 'keep_all', // if empty, default to not reset options
	[
		'keep_all' => 'Keep server options', // default, no config, unsets all
		'push_reset' => 'Reset all options', // unsets specified
		'push_reset_keep_topology' => 'Only keep topology', // unsets specified
		'remove_specified' => 'Remove specified options' // no config, display toggle
	]
))->setHelp('Prevent this client from receiving server-defined client settings. Other client-specific options on this page will supersede these reset options.');
$group = new Form_Group('Reset Options');
$group->addClass('override_list');
$group->add(new Form_Select(
	'override_list',
	null,
	$pconfig['override_list'],
	[
		'remove_route' => 'Local Routes',
		'remove_iroute' => 'Remote Routes',
		'remove_dnsdomain' => 'DNS Domains',
		'remove_dnsservers' => 'DNS Servers',
		'remove_ntpservers' => 'NTP Options',
		'remove_netbios_ntype' => 'NetBIOS Type',
		'remove_netbios_scope' => 'NetBIOS Scope',
		'remove_wins' => 'WINS Options'
	],
	true
))->setHelp('A "push-remove" option will be sent to the client for the selected options, removing the respective server-defined option.');
$section->add($group);

//=== Javascript
function resetoptions_change() {
	hideClass('override_list', ($('#reset_options').find('option:selected').val() != 'remove_specified'));
}
// change handler
$('#reset_options').on('change', function() {
	resetoptions_change();
});
// initial page load
resetoptions_change();
```

PhilZ-cwm6:
> If the pushed topology "must match the actual topology specified in server", then why allow pushing a different one at all? If it must match, then `topology_override` is enough.
> 

Ok, you are right, I verified and it is in fact a client setting only. So, it can be set differently here and override the server setting.
I just should fix the hint text

I wanted to make minimal code changes in `openvpn.inc` and just enhance the current pfsense gui which makes it manadatory to define the topology and gateway manually in custom options to get a working setup

I did not mean to make it a break-proof setup for un-aware users.
That was my initial idea. What are your thoughts about it now that this option can be set by user differently from the server defined one ?

> If `push_reset` is enabled in an older config, then there should be an option in `upgrade_config.inc` that would automatically enable `server_overrides_enabled` in the config. Thinking about this some more however, the config doesn't need to track this at all since the option only serves as a display toggle.
>
Yes, it is not needed. I looked at it and I am quiet sure that the changes won't break old configs. I use same options and I checked that these exceptions are accounted for in the `vpn_openvpn_csc.php` file
I also just tested, and when `push_reset` is checked in old config, it is refelected in new one, same when `remove_route` is checked

> Regardless, we can instead use a list as mentioned previously - see below for some code to get it started.
> 
> ```
> //==== Form fields
> $section->addInput(new Form_Select(
> 	'reset_options',
> 	'Reset Server Options',
> 	$pconfig['reset_options'] ?? ($pconfig['override_list'] ? 'remove_specified' : 'keep_all'), // if empty, default to not reset options
> 	[
> 		'keep_all' => 'Keep server options', // default, unsets config option, unsets all
> 		'push_reset' => 'Reset all options', // unsets all specified, sends push-reset
> 		'push_reset_keep_topology' => 'Only keep topology', // unsets all specified, sends push-reset and server topology
> 		'remove_specified' => 'Remove specified options' // unsets config option, display toggle
> 	]
> ))->setHelp('Prevent this client from receiving server-defined client settings. Other client-specific options on this page will supersede these reset options.');
> $group = new Form_Group('Reset Options');
> $group->addClass('override_list');
> $group->add(new Form_Select(
> 	'override_list',
> 	null,
> 	$pconfig['override_list'], // e.g. ['remove_iroute', 'remove_dnsservers']
> 	[
> 		'remove_route' => 'Local Routes',
> 		'remove_iroute' => 'Remote Routes',
> 		'remove_dnsdomain' => 'DNS Domains',
> 		'remove_dnsservers' => 'DNS Servers',
> 		'remove_ntpservers' => 'NTP Options',
> 		'remove_netbios_ntype' => 'NetBIOS Type',
> 		'remove_netbios_scope' => 'NetBIOS Scope',
> 		'remove_wins' => 'WINS Options'
> 	],
> 	true
> ))->setHelp('A "push-remove" option will be sent to the client for the selected options, removing the respective server-defined option.');
> $section->add($group);
> 
> //=== Javascript
> function resetoptions_change() {
> 
> 	hideClass('override_list', ($('#reset_options').find('option:selected').val() != 'remove_specified'));
> }
> // change handler
> $('#reset_options').on('change', function() {
> 	resetoptions_change();
> });
> // initial page load
> resetoptions_change();
> ```
>
Ok, I will look at it if you really prefer a list
But if you agree, I would drop the `push_reset_keep_topology` option since I clarified this above ?

marcos-ng:
I agree.

marcos-ng:
I've reworked this MR to implement the feature in a simpler way. Give there are significant changes, I'm closing this and updating the redmine. Thanks!

PR #4571, opened 2022-04-18
Update pfSense-dark.css

iZedful:
Matching background/font colors of queue values with dark theme.

- [X] Redmine Issue: https://redmine.pfsense.org/issues/13072
- [X] Ready for review

Before-After


src/usr/local/www/css/pfSense-dark.css
```diff
@@ -418,7 +418,8 @@ button.close {
 }
 
 input, select, textarea {
```
jim-p, line 420: This matches a lot more than the queues page and changes other input fields which makes them difficult to read. A quick example of this is on Diagnostics > Command Prompt.

You'll need to use a different selector to match just the queue output items, possibly including alterations to status_queues.php to make it possible if it isn't already.

jim-p (changes requested):

src/usr/local/www/css/pfSense-dark.css
iZedful: - [x] Done.

Sorry for the number of commits, I should have made drafts and more tests prior submitting.
In short only the css file has changes.

Extra Explanation:
I have tried at first to add "queue" class in the status_queues.php to avoid any color changes in the other pages and it worked as intended.
But I thought it's better to not touch any php file if it's possible to do it from the css side only.
Even though it's a little number of bytes but making everyone loading them if not needed by their theme isn't good in terms of optimizing.
I applied the colors to any text input if it's inside a table...td. However, by searching through the other files using regex, the only thing that conflicts is the rules separator textbox
in this file : \src\usr\local\www\js\pfSenseHelpers.js line 524
But fortunately there is defined id #newsep already so I used it in the css to fix that text box too.

PR #4572, opened 2022-04-22
OpenVPN: expose NBDD servers option and fix gui bugs 

PhilZ-cwm6:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13085
- [x] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -789,6 +789,9 @@ function openvpn_add_dhcpopts(& $settings, & $conf) {
 		if (!empty($settings['nbdd_server1'])) {
 			$conf .= "push \"dhcp-option NBDD {$settings['nbdd_server1']}\"\n";
 		}
+		if (!empty($settings['nbdd_server2'])) {
+			$conf .= "push \"dhcp-option NBDD {$settings['nbdd_server2']}\"\n";
+		}
 	}
 
 	if ($settings['gwredir']) {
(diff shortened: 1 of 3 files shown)


jim-p (changes requested):
There are far too many irrelevant whitespace changes in this. It shouldn't be touching nearly that many lines.

PhilZ-cwm6:
I left the white spaces as they are and force pushed for a cleaner code to review
Again, the last two are optional as described in the redmine issue

jim-p (changes requested):
This is still touching way more than it needs to. We have to limit the scope here so it just the changes for the one option. Please do not change any other code or add functions like this. We need to keep the diff for each PR as minimal as possible. We could consider the other things as separate PRs but this isn't the time for those kinds of changes.

PhilZ-cwm6:
@jim-p 
is the first commit fine at least ? Or you want to only expose NBDD server options without fixing the current WINS servers GUI bug neither the CSC page bug ?

The last one in CSC is really annoying because we end up with a situation where it is impossible to say if the settings will be saved or not. At some time, the NTP server check box is enabled while the server field is hidden !

I can only expose NBDD and keep all current GUI bugs if you prefer, or just remove the last two commits

PhilZ-cwm6:
I removed the two last commits to limit the scope

@jim-p
is the first commit fine at least ? Or you want to only expose NBDD server options without fixing the current WINS servers GUI bug neither the CSC page bug (the changes are really limited in first commit to only what exists elsewhere in code)?

The last one in CSC is really annoying because we end up with a situation where it is impossible to say if the settings will be saved or not. At some time, the NTP server check box is enabled while the server field is hidden ! Since the GUI options are one close to the other, the changes are identical.

I can only expose NBDD and keep all current GUI bugs if you still prefer

jim-p:
Each bug/feature should be a separate Redmine issue and separate PR.

PhilZ-cwm6:
Ok, I'll close and think at it how to submit the nbdd servers broken because of the depending netbios JS function is broken
So, maybe I will push the other fixes first, and this one depending on them

PR #4573, opened 2022-04-22
OpenVPN Server: hide WINS servers when netbios is unchecked

PhilZ-cwm6:
When we check NetBIOS option, and check the Wins servers option, then if Netbios option is unchecked, the WINS servers list was not properly hidden

- [x] Redmine Issue: https://redmine.pfsense.org/issues/13087
- [x] Ready for review


marcos-ng:
Fixed in b57ee7830401697a729ac796e5c09f4c2021ccf0.

PR #4574, opened 2022-04-22
OpenVPN Client overrides: properly hide forms

PhilZ-cwm6:
- in client specific overrides, NTP and DNS options: use javascript instead of toggles class because the latter causes inverted results when repeating clicks too quickly. This also matches the code used in the server page

- [x] Redmine Issue: https://redmine.pfsense.org/issues/13088
- [x] Ready for review


vktg:
Looks good to me.

jim-p:
There was a conflict here still but it was small (just one letter on a line changed case). Rather than wait on it to be fixed, I adjusted it locally and merged it manually after trying it out.

Thanks!

PR #4577, opened 2022-04-26
remove line filename from the  pool section

jzadir:
Remove the filename config single line as it  breaks the conditional dhcp config for pxe;

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/12986
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/services.inc
+++ b/src/etc/inc/services.inc
@@ -1114,9 +1114,7 @@ EOPP;
 						$dhcpdconf .= "		}";
 					}
 					$dhcpdconf .= "\n\n";
-				} elseif (!empty($filename)) {
-					$dhcpdconf .= "		filename \"{$filename}\";\n";
-				}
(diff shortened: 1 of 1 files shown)


jim-p:
There is already a fix on that Redmine issue that should prevent that line from being a problem when it isn't needed. Is the current fix incomplete or not functional? If so, comment that on the Redmine. If you haven't tested the fix on the Redmine yet, please do so before suggesting additional changes.

jzadir:
I've tested the suggested fix and it does not work; this PR is running in production and solved the issue; 
Looking in the history of the file I realize that the cause of the problem was a change in the logic;

marcos-ng:
> Looking in the history of the file I realize that the cause of the problem was a change in the logic;

Is a different change needed? There's likely a reason why the line was included in that case - some additional clarification here would be helpful

PR #4578, opened 2022-04-26
add warning to menu option 14 when it might kill your connection

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13103
- [x] Ready for review

Believe it or not, I fat fingered "13" the other day and typed "14" instead when connected via SSH ... and hit "y" which then locked me out of a firewall. Had to dig out the console cable to get it sorted.

This small patch just checks to see if the current session is via SSH and if so, it prints a warning if the user is about to footgun themselves.

Screenshot:

CleanShot 2022-04-26 at 19 37 04

### DIFF
--- a/src/etc/rc.initial.toggle_sshd
+++ b/src/etc/rc.initial.toggle_sshd
@@ -31,6 +31,12 @@ require_once("filter.inc");
 $fp = fopen('php://stdin', 'r');
 
 if (isset($config['system']['ssh']['enable'])) {
+	if (!empty(getenv('SSH_CONNECTION'))) {
+		echo gettext("!!! WARNING !!!");
+		echo "\n\n";
+		echo gettext("This session is currently established via SSH. Disabling SSH will terminate this session and prevent future SSH connections from any user. Before disabling SSH, ensure that administrators have alternate means of accessing the firewall, such as the GUI or serial console.");
+		echo "\n\n";
+	}
 	echo "SSHD is currently enabled.  Would you like to disable? [y/n]? ";
 	$yn = chop(fgets($fp));
 	if ($yn[0] == "y") {


src/etc/rc.initial.toggle_sshd
```diff
@@ -31,6 +31,21 @@ require_once("filter.inc");
...
+		echo <<<EOL
++------------------------------------------------------------+
+|                                                            |
+| !!! WARNING !!!                                            |
+|                                                            |
+| You are currently connected via SSH! If you disable it,    |
+| you risk locking yourself out of the system. Please ensure |
+| you have another way to access the console if necessary.   |
+|                                                            |
++------------------------------------------------------------+
```
jim-p, line 44: No need to get that fancy with the ASCII art. The text should be run through `gettext()` so it can be translated and aligning it that way is not viable. A line above/below is OK and likely sets it apart enough.

Also, please reword it to avoid addressing the user directly. Perhaps:

> This session is currently established via SSH. Disabling SSH will terminate this session and prevent future SSH connections from any user. Before disabling SSH ensure that administrators have alternate means of accessing the firewall, such as the GUI or serial console.

jim-p (changes requested):

src/etc/rc.initial.toggle_sshd
luckman212: Good points, I actually realized this after testing it from my mobile phone which had a 40 char terminal width. I'll make changes.

luckman212: @jim-p one clarification: the rest of the console menu doesn't appear to be filtered through `gettext()`. So is that really necessary? I see only the changelog that gets saved to config.xml gets interpreted.

jim-p: > @jim-p one clarification: the rest of the console menu doesn't appear to be filtered through `gettext()`. So is that really necessary? I see only the changelog that gets saved to config.xml gets interpreted.

It just hasn't been converted. It's best to add it for new things as we go rather than wait and make the eventual conversion an even larger job. Also it's better if a scary warning gets translated so it's more clearly understood by the user.

src/etc/rc.initial.toggle_sshd
```diff
@@ -31,6 +31,12 @@ require_once("filter.inc");
 $fp = fopen('php://stdin', 'r');
 
 if (isset($config['system']['ssh']['enable'])) {
+	if (!empty(getenv('SSH_CONNECTION'))) {
+		echo (gettext("!!! WARNING !!!"));
```
jim-p, line 35: No need for the extra `()` around the echo contents.

src/etc/rc.initial.toggle_sshd
```diff
@@ -31,6 +31,12 @@ require_once("filter.inc");
...
+		echo "\n\n";
+		echo (gettext("This session is currently established via SSH. Disabling SSH will terminate this session and prevent future SSH connections from any user. Before disabling SSH, ensure that administrators have alternate means of accessing the firewall, such as the GUI or serial console."));
```
jim-p, line 37: No need for the extra `()` around the echo contents.

jim-p (changes requested):

src/etc/rc.initial.toggle_sshd
luckman212: Gotcha. They're gone! Is there a style guide for the project?

jim-p: Yep, there is https://docs.netgate.com/pfsense/en/latest/references/developer-style-guide.html (for development) and also https://docs.netgate.com/pfsense/en/latest/references/style-guide.html for documentation (also should apply to text in the GUI/menus/etc where possible)

luckman212: Nice! Thanks

jim-p (approved):

PR #4580, opened 2022-04-28
Trim MAC addresses when submitting (e.g. on services_dhcp_edit.php)

luckman212:
Small patch to trim MAC address input on POST. This eases copy & paste which sometimes grabs a little extra whitespace on either end. Previously this would cause input validation to fail.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/13109
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/services_captiveportal_mac_edit.php
+++ b/src/usr/local/www/services_captiveportal_mac_edit.php
@@ -93,7 +93,7 @@ if ($_POST['save']) {
 
 	$macfull = explode('/', $_POST['mac']);
 	$macmask = $macfull[1] ? $macfull[1] : false; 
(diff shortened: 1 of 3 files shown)


jim-p (approved):
Looks OK. The pedantic in me wants to say that the `trim()` should be the innermost call not the outermost because it could save the other functions from having to operate on a longer string unnecessarily but since we're talking about short strings here it's not going to save enough cycles to make a significant difference either way.

luckman212:
I actually had that same thought, but somehow this looks easier to read to me, and also thought it's not going to make a measurable difference. If someone is pasting 64k of text into that field, then... well... 🤷‍♂️

PR #4589, opened 2022-05-15
Live refresh for Services dashboard widget and other related improvements

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/TBD
- [x] Ready for review

I noticed the other day that the Services dashboard widget was out of sync with reality. If a service dies or is stopped, the dashboard won't reflect that unless you manually reload the page. Also, there were some other annoyances, like clicking the service control icons causing a full page reload, jumping to the top of the screen etc. I spent some time reworking a few things and made what I think are some good QoL changes.

### Improvements

- live refresh in the Services dashboard widget (adopted Centralized widget refresh system)
- added mutex to lock out stop/start/reload buttons when an action is underway to avoid hysteresis
- actions called from dashboard widget no longer cause the page to "jump" to the top, nor do they require the dashboard to do a full reload (so your graphs etc stay put)
- reworked the way the jQuery selectors are stored and parsed, in what I believe to be a more robust way
- fixed some bugs related to having multiple instances of the Services widget

### Demo

Here's a little video demonstrating how the widget looks & behaves now. This is 2 separate browser windows side by side to show the live updates and mutex handler on the action buttons.

https://user-images.githubusercontent.com/1992842/168460727-f6811523-62d0-41d7-a6b6-bae9399b5cd0.mp4


luckman212:
Forum thread: https://forum.netgate.com/topic/172188/live-refresh-for-dashboard-services-status-widget

luckman212:
Damn. Missed the cutoff again!

luckman212:
Still working fine for me on the latest public 22.05.b.20220531.0600 snap.

Any other testers?

luckman212:
Tested & working on 22.05.r.20220604.1403 rc

PR #4590, opened 2022-06-01
oops. forgot to actually process the dpinger_dont_add_static_routes flag

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12687
- [x] Ready for review

Big apologies for this one. This is a critical fix to #4551. Otherwise the System → Advanced → Miscellaneous → _Do not add static routes for gateway monitor IP addresses_ checkbox is nonfunctional.


jim-p:
I've merged this in, it may or may not make it into 22.05, though. We're building and testing now, but if we have to rebuild it might be possible to include it. If not we can add it into the recommended patches list in the system patches package.

PR #4591, opened 2022-06-02
enhancements to gateway monitor setup / static routing (builds on #4551)

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13242
- [x] Ready for review
---
#### Changes
- fixes orphan static routes hanging around after changing monitor IPs
- dynamically turn global or per-gw dont_add_static_route option on/off without requiring a reboot
- refactor setup_gateways_monitor() to reduce duplicate code
- added note to setHelp area on gw edit to notify user when global dont_add_static_routes option would override the local setting

This is my improved v2 patch for #4551. The main difference is the ability to dynamically remove stale/orphan static routes that were added for gateway monitoring with a new `delete_temporary_static_routes()` helper function. I also tried to re-use some code where possible in the `setup_gateways_monitor()` loop by breaking out the identical sections into separate functions.

I have limited IPv6 capability on my test unit here, so I would love to have some more testers in that area. My basic testing shows it working fine.

I also created a helper script to aid in debugging, see [`dpinger_static_routes.php` gist](https://gist.github.com/luckman212/0fdea1cbdc0a561d781a52c7d34fb60d). That will show all the gateway / static route / related config in realtime so you can watch the changes as you adjust in the GUI.

### To run the helper script:

1. install cmdwatch:
```
pkg add https://pkg.freebsd.org/FreeBSD:12:amd64/latest/All/cmdwatch-0.2.0_2.txz
```
2. download the script:
```
fetch https://gist.githubusercontent.com/luckman212/0fdea1cbdc0a561d781a52c7d34fb60d/raw/ffd321ef196fb1c919dd66700acdd4acc02b3e63/dpinger_static_routes.php
```
then
```
cmdwatch --interval=2 'php -q dpinger_static_routes.php'
```

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -190,6 +190,100 @@ function start_dpinger($gateway) {
 	return mwexec("/usr/local/bin/dpinger {$params} {$gateway['monitor']} >/dev/null");
 }
 
+/* Remove temporary static routes that were dynamically added for dpinger */
+function delete_temporary_static_routes($gateway) {
+	global $config;
+	if (empty($gateway)) {
+		return false;
+	}
+	$ipprotocol = $gateway['ipprotocol'];
+	$rtable = route_table()[$ipprotocol];
+	if (empty($rtable)) {
+		return false;
+	}
+	$gwname = $gateway['name'];
+	$friendly = $gateway['interface'];
+	$ifname = get_real_interface($friendly);
+	$routes_ughs = array_filter($rtable, fn($r) => (
+		$r['flags'] == 'UGHS' &&
+		$r['interface-name'] == $ifname
+	));
+	$staticroutes = get_staticroutes();
+	foreach ($routes_ughs as $r) {
+		$tgt = $r['destination'];
+		if (is_ipaddrv4($tgt)) {
+			$cidr = "{$tgt}/32";
+		} elseif (is_ipaddrv6($tgt)) {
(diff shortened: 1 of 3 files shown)


luckman212:
@jim-p as requested I create a new redmine [#13242](https://redmine.pfsense.org/issues/13242) for this.

franciscodimattia:
On PfSense version 2.6.0, even if "Disable Gateway Monitoring Action" is checked, the interface is shown as Offline in the dashboard and a notification is sent to my email saying that the interface was omited from the load balancing group due to packet loss. Would this pull request fix that issue?

chrcoluk:
Did this make it into 2.7.0 stable? as it seems a important quality of life update, not having the route disabled when unticking the checkbox is messy.

src/etc/inc/gwlb.inc
```diff
@@ -206,6 +300,7 @@ function setup_gateways_monitor() {
 		echo "Setting up gateway monitors...";
 	}
 	$monitor_ips = array();
+	$GLOBALS['inuse_static_routes'][] = array();
```
marcos-ng, line 303: I take it the purpose of this is to avoid removing a static route configured by the user. If so, I don't think it's necessary. When a static route is configured by the user, it stays when dpinger restarts even when "Do not add static route..." is checked on the gateway. Regardless, I see this more of a user configuration issue than something we should worry about in the code.

PR #4592, opened 2022-06-04
add type column to firewall_aliases.php

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13245
- [x] Ready for review

Small QoL addition that adds a Type column to the Alias list views. I was recently cleaning up my aliases and being able to see the type (especially on the "All" tab) was very useful. I also moved the `$types` array into `alias-utils.inc` so it can be re-used.

screenshot

CleanShot 2022-06-04 at 10 20 23@2x

### DIFF
--- a/src/usr/local/pfSense/include/www/alias-utils.inc
+++ b/src/usr/local/pfSense/include/www/alias-utils.inc
@@ -34,6 +34,16 @@ require_once("openvpn.inc");
 init_config_arr(array('aliases', 'alias'));
 $a_aliases = &$config['aliases']['alias'];
 
+$alias_types = array(
+	'host'	=> gettext("Host(s)"),
+	'network' => gettext("Network(s)"),
+	'port' => gettext("Port(s)"),
+	'url' => gettext("URL (IPs)"),
+	'url_ports' => gettext("URL (Ports)"),
+	'urltable' => gettext("URL Table (IPs)"),
+	'urltable_ports' => gettext("URL Table (Ports)"),
+);
+
 function deleteAlias($id, $apply = false) {
 	global $config;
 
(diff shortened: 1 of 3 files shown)


src/usr/local/pfSense/include/www/alias-utils.inc
```diff
@@ -34,6 +34,16 @@ require_once("openvpn.inc");
 init_config_arr(array('aliases', 'alias'));
 $a_aliases = &$config['aliases']['alias'];
 
+$types = array(
```
jim-p, line 37: Rather than making a very generic name here, make it more specific, e.g. `$alias_types` or similar. That way it's more descriptive and less likely to conflict if something else happens to also be included that has its own `$types`.

jim-p (changes requested):

src/usr/local/pfSense/include/www/alias-utils.inc
luckman212: Got it. Patch updated.

jim-p (approved):

src/usr/local/www/firewall_aliases.php
```diff
@@ -145,6 +146,9 @@
 			<td ondblclick="document.location='firewall_aliases_edit.php?id=<?=$i;?>';">
 				<?=htmlspecialchars($alias['name'])?>
 			</td>
+			<td nowrap ondblclick="document.location='firewall_aliases_edit.php?id=<?=$i;?>';">
+				<?=htmlspecialchars($alias_types[$alias['type']])?>
```
jim-p, line 150: This should either be wrapped in a check to ensure it exists (e.g. `array_key_exists()`) or use one of the new methods for safely accessing these types of values, e.g. `array_get_path($alias_types, array_get_path($alias, 'type'), gettext("Not Set"))`. In case for some reason a (very old) alias doesn't have a type value in the configuration.

jim-p (changes requested):

jim-p:
In addition to the unresolved request above, now there are some conflicts that need resolved. Might be easier to redo the PR against the current code than it would be to update it as-is.

jim-p:
Implemented this manually with some changes, so no need for the PR.

Thanks!

PR #4593, opened 2022-06-05
Fix typo in enable

wrt54g:
- [X] Ready for review

I fixed a typo in the word "Enable" in services_dhcp.php


jim-p:
It's not really a typo, that phrasing is common in the GUI and reads as "The checkbox/option enables network booting". But if we decide to change that there are some others on the same page that could be rephrased so no need for a PR for the single letter difference.

PR #4594, opened 2022-06-08
Add warning infobox if duplicate IP is entered in DHCP static map

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13256
- [x] Ready for review

This is an improvement to add a warning back to input validation when a user enters the same IP for multiple DHCP static maps. It still allows this to be done, by adding a pconfig checkbox that the user needs to check. I believe this is the best of both worlds.

#### Example input validation warning
image

#### If you want to really do this, just check the box:
image

_related forum thread: https://forum.netgate.com/topic/172698/why-does-the-ui-allow-duplicate-ip-addresses-different-macs-in-dhcp-static-mappings_

### DIFF
--- a/src/usr/local/www/services_dhcp_edit.php
+++ b/src/usr/local/www/services_dhcp_edit.php
@@ -226,6 +226,11 @@ if ($_POST['save']) {
 			$input_errors[] = gettext("This MAC address or Client identifier already exists.");
 			break;
 		}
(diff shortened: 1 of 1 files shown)


luckman212:
@jim-p I pushed a revised version that is less obtrusive. Just prints a friendly infobox reminder to help avoid unintended footguns. I hope this is acceptable.

PR #4596, opened 2022-06-09
a few updates for the console menu /etc/rc.initial

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13258
- [x] Ready for review

### changes
- add full pathnames to all binaries (before some were and some weren't)
- less forking for process checking, instead of ps | grep, use pgrep
- clean up spaces vs tabs
- remove the text "cheap menu"
- fixup hidden option 100 (links browser)

**n.b.** the `patch` command is failing to automatically patch this via system patches, I think because the open source version differs so much from the pfS+ version that has hardware-specific elements. So for testing, I had to manually patch my 6100 to test it. Not sure if there is a way to get it to cooperate with System Patches, maybe breaking it up into a set of smaller commits, but hopefully these changes are small enough that it can be reviewed without all that extra fuss.

### DIFF
--- a/src/etc/rc.initial
+++ b/src/etc/rc.initial
@@ -24,47 +24,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# make sure the user can't kill us by pressing Ctrl-C,
-# ctrl-z, etc.
-trap : 2
-trap : 3
-trap : 4
+# make sure the user can't kill us by pressing Ctrl-C, ctrl-z, etc.
+trap : INT
+trap : QUIT
+trap : ILL
 
 unset do_sleep
 if [ -f /etc/rc.local ]; then
-	if ! ps auxwww | egrep -q '[r]c.local$'; then
+	if ! /bin/pgrep -qf 'rc.local$'; then
 		echo ">>> Launching rc.local in background..."
-		sh /etc/rc.local &
+		/bin/sh /etc/rc.local &
 		do_sleep=1
 	fi
-	if [ -f /etc/rc.local.running ] && \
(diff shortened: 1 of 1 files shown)


luckman212:
**n.b.** the dashboard may not actually load with the `links` browser ... but I was able to load other pages e.g. google. So, might still be useful to leave it in.

image

image

src/etc/rc.initial
```diff
@@ -151,18 +148,21 @@ case ${opmode} in
...
-		links "http://localhost"
+	protocol=$(/usr/local/sbin/read_xml_tag.sh string system/webgui/protocol)
+	port=$(/usr/local/sbin/read_xml_tag.sh string system/webgui/port)
+	if [ -z "$port" ]; then
+		case $protocol in
+			https) port=443;;
+			*)     port=80;;
+		esac
 	fi
+	links "${protocol}://localhost:${port}"
```
jim-p, line 159: `protocol` could be empty, so it needs an extra test to drop a default value in there if it is (`http`).

jim-p (changes requested):

luckman212:
Didn't know protocol could be empty. Ok I will rework to account for that.

luckman212:
Should be good now @jim-p

luckman212:
Couple more very minor touchups

- removed `11 | 111)` -- why was that there? to catch typos?
- consistency with single vs double quotes
- removed a couple of stray `;`s

src/etc/rc.initial
luckman212: Didn't know protocol could be empty. Ok I will rework to account for that.

jim-p (approved):

PR #4598, opened 2022-06-12
tiny change to better align console menu with longer ifnames

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13268
- [x] Ready for review

small change to add some width and better align things if interface names are longer than just "WAN", "WAN2" etc.

before:
image

after:
image

I wanted to auto size the columns based on the terminal width, but the shell doesn't seem to export the `$COLUMNS` variable that would be required to do that.


marcos-ng:
See https://redmine.pfsense.org/issues/13268#note-2

PR #4599, opened 2022-06-20
add capability to redirect to requested page after login

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13286
- [x] Ready for review

Something that has bugged me for a while now is that if you are logged out of pfSense, and request a "deep" page e.g. https://my.pfsense.lan/services_unbound.php, you will be shown the login page, but after successfully logging in, you are placed on the dashboard instead of the originally requested page. 

This makes bookmarking specific pages impossible.

Here's a small PR that adds this capability.

Tested on 22.05.r.20220617.0613

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -2116,6 +2116,7 @@ function session_auth() {
 				$_SESSION['authsource'] = 'Local Database Fallback';
 			}
 			$_SESSION['Username'] = $_POST['usernamefld'];
+			$_SESSION['redirecturi'] = htmlspecialchars_decode($_POST['redirecturi']);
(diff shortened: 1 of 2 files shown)


jim-p:
This is done on purpose for security reasons. Until the entire GUI is purged of any page that takes action on GET, the user could be tricked into following a link that takes a harmful action.

luckman212:
@jim-p Is there a redmine for the remaining conversion from GET to POST/CSRF in the wC? I was going to offer to help move it along.

luckman212:
Hello random person on the internet who found this somehow 👋 

You may want to check out [v2](https://github.com/pfsense/pfsense/pull/4602) of this patch.

PR #4604, opened 2022-06-22
fix func params for get_dpinger_status() call in gwlb.inc

luckman212:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13295
- [x] Ready for review

There seems to be an error in `gwlb.inc` around line 479, the call to `get_dpinger_status()` has the `$action_disable` arg at position 2, but the function definition needs it at position 3. So, the bool is actually passed as param 2 which is supposed to be an array (`$gways`). Not sure what problems this may have caused, but it seems incorrect.


fislysandi:
#not related to the pull request

hey can somebody fix the documentation? i cant access the website (why did the documentation was moved to that site in the first place since i cant access it). but so far this project looks promising having a literall pc on the go.

jim-p (approved):

PR #4608, opened 2022-06-27
Fix destination IPv6 prefix list not properly built

seyfidin:
Destination IPv6 prefix list is not built properly due to wrongly placed string operator
Removed wrong placed string operator from $sntext.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/13310
- [x] Ready for review


jim-p:
I addressed this in a different way. More info on the Redmine issue.

PR #4500, opened 2021-02-15
Add IPv6 support for Gandi LiveDNS dynamic DNS handler

bitscher:
* Add support for AAAA records with Gandi LiveDNS.
* Simplify the use of the LiveDNS API by using another API call that
handles both creating and updating for a given name and record type
instead of the former two step operation that was fetching the whole
content of the zone file and looked for the relevant record to update.
https://api.gandi.net/docs/livedns/#v5-livedns-domains-fqdn-records-rrset_name-rrset_type
* Remove duplicated code by merging relevant switch statements.
* Fix erroneous input validation for hosts '@.' and '*.' that are
not supported by Gandi LiveDNS and should not be saved in config.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/11420
- [x] Ready for review

PR #4610, opened 2022-08-13
Update status_upnp.php to address Bug #4500

rtadams89:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/4500
- [ ] Ready for review

This implements two things to address UPnP configured NAT rules that do not properly display in the status table:

1. Updates the regex to make the "label" optional
2. Matches on and extracts specific allowed source IPs (as opposed to just matching on literal "any") and displays this extracted value in a new table column.


jim-p (approved):
Tested it and it appears to work fine. Looks good to merge.

jim-p:
Automatic merge failed, said it couldn't find the upstream branch. I merged it by hand, thanks!

PR #4612, opened 2022-08-26
Add formatting to online status in DHCP Status Pages

zeroflow:
Adds color & formatting to host online status

online = bold + green
offline = italic + red

2022-08-26_13h13_24
2022-08-26_13h21_17

- [X] Redmine Issue: https://redmine.pfsense.org/issues/10345
- [X] Ready for review

### DIFF
--- a/src/usr/local/www/status_dhcp_leases.php
+++ b/src/usr/local/www/status_dhcp_leases.php
@@ -293,7 +293,16 @@ foreach ($leases['lease'] as $data):
 						<td><?=gettext("n/a")?></td>
 						<td><?=gettext("n/a")?></td>
 					<? endif; ?>
-					<td><?=htmlspecialchars($data['online'])?></td>
+					<td>
+<?php if ($data['online'] == $online_string):?>
+							<p style="font-weight: bold; color:green;">
+<?php else: ?>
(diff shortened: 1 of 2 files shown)


jim-p:
Looking at it in action I'm not a fan of the red. It implies something bad is happening when it could just be that the system is up but hasn't sent traffic to/through the firewall in a while and the ARP entry expired.

Also using only green and red is not accessible to color blind users so it should have an icon as well.

How about this:

* Green/Online can stay green but the text should say "Active" with an `fa-arrow-up` icon
* Red/Offline should either have *no* color or be yellow and say "Idle/Offline" with an `fa-arrow-down` icon

zeroflow:
That sounds reasonable, I will update with a rework.

zeroflow:
Thanks for the input, this now looks a lot cleaner. 

I did not go for the renaming, as this string is in the translations and the best I could to would be some online translator for everything except english & german. If the new wording is mandatory, I could either try for machine translated texts or simply remove existing (e.g. korean) translations for these messages.

2022-08-26_19h08_24

zeroflow:
Thanks for the advice, it's now updated 

image

jim-p (approved):

PR #4614, opened 2022-08-29
Fix typo in is_URL

brotherla:
Fix undefined symbol exception

`Fatal error: Uncaught Error: Call to undefined function isURL() in /etc/inc/pfsense-utils.inc:2653 Stack trace: #0 /usr/local/pfSense/include/www/alias-utils.inc(330): process_alias_urltable('pfB_PRI1_6_v6', 'urltable', 'https://127.0.0...', 0, true, true) #1 /usr/local/www/firewall_aliases_edit.php(157): saveAlias(Array, '1') #2 {main}`

- [x] Ready for review


marcos-ng:
This PR no longer applies to 2.7/23.01 and should be closed.

PR #4615, opened 2022-09-17
Expand detection of PC Engines APU2 platform to include all variants

blkeller:
Fixes #13498 in Redmine.

This fixes a problem with garbled serial console output in the bootloader that was originally described in PR #3268, but that fix is not getting applied to the newer hardware variants because the platform detection code needs to be updated.

The PC Engines naming conventions are a little confusing here.  The motherboard platform is identified as "APU2" (though sometimes it also appears in lowercase), but within this platform there are hardware variants such as "apu4", "apu6", and even "apu2".  For example, one can be the owner of an apu4 APU2, if you can believe it.

The detection code reads the DMI data of the product name from the SMBIOS.  Historically, that value has been `apu2`, then it was `APU2` for a little while, then it changed back to `apu2`.  As the newer variants have been released, they have reported values such as `apu3`, `apu4`, etc.  The APU2 platform uses coreboot for its firmware, so luckily we can see the source code for the current list of reported SMBIOS values.  See [here](https://github.com/pcengines/coreboot/blob/b693f04134f8bb728462f6904a2c2233e5fb4498/src/mainboard/pcengines/apu2/Kconfig#L41-L47) and [here](https://github.com/pcengines/coreboot/blob/b693f04134f8bb728462f6904a2c2233e5fb4498/src/Kconfig#L856-L862).

This pull request adds support for all of the variants currently listed in the firmware, plus leaves intact the historical `APU2` value for those who might still be on a much older BIOS version.

I currently have physical access to an apu4 variant.  I tested this patch against that hardware, and it fixes the problem with the garbled serial console output on this unit, plus the pfSense dashboard now shows the system as "PC Engines APU2" as I would expect instead of the generic "pfSense".

- [X] Redmine Issue: https://redmine.pfsense.org/issues/13498
- [X] Ready for review

### DIFF
--- a/src/etc/inc/system.inc
+++ b/src/etc/inc/system.inc
@@ -3166,9 +3166,14 @@ function system_identify_specific_platform() {
 				return (array('name' => '1540', 'descr' => 'Super Micro XG-1540'));
 			}
 			break;
-		case 'apu2':
 		case 'APU2':
-			return (array('name' => 'apu2', 'descr' => 'PC Engines APU2'));
+		case 'apu2':
+		case 'apu3':
+		case 'apu4':
+		case 'apu5':
+		case 'apu6':
+		case 'apu7':
+			return (array('name' => 'apu2', 'descr' => "PC Engines APU2 Platform (\"{$product[0]}\" model)"));
 			break;
 		case 'VirtualBox':
 			return (array('name' => 'VirtualBox', 'descr' => 'VirtualBox Virtual Machine'));


RKatTC:
-		case 'apu2':
-		case 'APU2':
-			return (array('name' => 'apu2', 'descr' => 'PC Engines APU2'));

probably a bit more universal - RegEx-Match APU2 to APU9 case-insensitive: 

+        case (preg_match("/apu[2-9]/i", $product[0]) ? true : false):
either: +	return (array('name' => 'apu2', 'descr' => $maker[0]. strtoupper($product[0])));
or:      +	return (array('name' => $product[0], 'descr' => $maker[0]. strtoupper($product[0])));

dont know, if 'name' => $product[0]  would break something

blkeller:
That's not a bad idea, and I considered using a regex when I wrote this, but I ultimately decided against it for a couple of reasons:
- All of the other platform detection code in pfSense uses explicit string matching, with no regexes anywhere, and I didn't want to break from the established pattern without Netgate's blessing.
- I don't want to be overly broad in the detection.  As of writing this, there is no such device as the apu8 or apu9, and I don't want to alter pfSense behavior from the default for a potential device that I can't verify behaves in the same way as the other existing models.

> dont know, if 'name' => $product[0] in return-line would break something

Yes, the "apu2" value is used elsewhere in the codebase to tweak settings specific to the members of this hardware platform.

As far as the user-facing description goes, while the different variants are named things like "apu2e4" and "apu4d2", the entire platform is collectively referred to as "the APU2 platform" as far as I can tell from various places on the PC Engines website, e.g.:
https://www.pcengines.ch/apu2.htm

Given that, I was under the impression that "PC Engines APU2" is still an accurate description even for the "apu4d2" variant.

RKatTC:
Thanx for clarifying that the "apu2" value is used elsewhere in the codebase.

My Main-Intention for changing the description is to have an accurate indicator when updating the firmware as those differ
between APUx's (https://pcengines.github.io/).

marcos-ng:
> Thanx for clarifying that the "apu2" value is used elsewhere in the codebase.
> 
> My Main-Intention for changing the description is to have an accurate indicator when updating the firmware as those differ between APUx's (https://pcengines.github.io/).

This is a good point. I think it's better to have a clearer description in this case.

blkeller:
Good idea.  It's definitely handy to know the more specific variant name when performing a firmware update.

I've updated the description to show both the platform and model, trying to make it very clear which is which:
image

Code updated, rebased, squashed, and ready for re-review, please.

PR #4616, opened 2022-09-26
Update thermal sensor widget

GChuf:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13520
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/css/pfSense-dark.css
+++ b/src/usr/local/www/css/pfSense-dark.css
@@ -394,6 +394,10 @@ div.inputselectcombo span input {
     border-bottom-left-radius: 0;
 }
 
+.progress-bar-success {
+	background-color: #009688 !important;
+}
+
 .no-separator {
     border-bottom: 0!important;
 }
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -667,7 +667,11 @@ events.push(function() {
 			    }
 			}
 
-		    setTimeout(function() { executewidget(); }, 1000);
+			if (ajaxcntr == 0) { //no timeout on first load
+				setTimeout(function() { executewidget(); }, 0);
+			} else {
+				setTimeout(function() { executewidget(); }, 1000);
+			}
 	  	}
 	}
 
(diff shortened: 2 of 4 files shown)


GChuf:
I had to change tsObject.freq = from 5 to 1 for some reason, to keep the graph updating at the regular interval of 5 seconds ... not sure why. If someone understands the widget update loop well enough, please take a look.

marcos-ng:
Please separate bug fixes, new functionality, and whitespace adjustments into their own commits.

At a glance what mainly stands out is the critical/warning temp value changes. I'm not sure these should be changed since 70C is a normal operating temp for many systems and 100C seems like a close enough general critical temp.

GChuf:
@marcos-ng thank you, I've separated the commits.
I agree with you regarding the temperatures - I've left them at default values for now. For the whitespaces/tabs - i think it might be better to create another PR where all instances of whitespaces are replaced ... so I've left that out for now as well.

Let me know what you think.

GChuf:
Hi @marcos-ng
I’ve addressed your feedback a while ago and made the changes. Could you please take another look when you have some time?
Thank you!

marcos-ng:
In effort to include some of these changes in the release, I've added them in as separate commits:
- 4b0deb88ca437dea96d719388a4b9136b49123ee: This change affects all progress bars, not only the thermal widget. Instead of overriding the default theme with `!important`, use the same selector and leave the override behavior to the include order.
- b0c455aa5d3f1c3b1926dbcca0eacbe4f57e3ef6: Improve the thermal widget readability by adding a margin and sorting. The label is kept below the progress bar to match the rest of the GUI (e.g. the disk widget).

As for the widget rendering behavior of "rebuilding" the HTML vs "updating" it: I don't think removing a function definition after it's called is any better than how it's done now. Not to say it's done particularly well as is, but the change is not worth risking potential breakage.

I will close out this PR. Thank you for contributing and I hope the new changes are satisfactory for now. Feel free to create a new PR if you'd like - the widget code could certainly still be improved.

PR #4622, opened 2022-12-15
ARM Support

Daniel3356:
Hi,

More and more modern day CPU are now becoming ARM64. I was wondering that will be grate to make a version compatible with ARM64 computers. 

Thanks for making an ARM version.


jim-p:
This site is not for support, features/issues, or diagnostic discussion.

For assistance in solving problems, please post on the [Netgate Forum](https://forum.netgate.com).

See [Reporting Issues with pfSense Software](https://docs.netgate.com/pfsense/en/latest/development/bug-reports.html) for more information.

PR #4624, opened 2023-01-05
add whitelist for igmpproxy

IgorYbema:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/13840
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/services_igmpproxy_edit.php
+++ b/src/usr/local/www/services_igmpproxy_edit.php
@@ -51,6 +51,7 @@ if (isset($id) && $a_igmpproxy[$id]) {
 	$pconfig['threshold'] = $a_igmpproxy[$id]['threshold'];
 	$pconfig['type'] = $a_igmpproxy[$id]['type'];
 	$pconfig['address'] = $a_igmpproxy[$id]['address'];
+	$pconfig['whitelist'] = $a_igmpproxy[$id]['whitelist'];
 	$pconfig['descr'] = html_entity_decode($a_igmpproxy[$id]['descr']);
 }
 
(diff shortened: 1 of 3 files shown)


src/usr/local/www/services_igmpproxy_edit.php
```diff
@@ -120,6 +142,7 @@
 		//we received input errors, copy data to prevent retype
 		$pconfig['descr'] = $_POST['descr'];
 		$pconfig['address'] = $address;
+		$pconfig['address'] = $whitelist;
```
marcos-ng, line 145: Incorrect assignment.

src/usr/local/www/services_igmpproxy_edit.php
```diff
@@ -252,6 +275,60 @@
 	'fa-plus'
 ))->removeClass('btn-primary')->addClass('btn-success addbtn');
 
+
+$counter = 0;
+$whitelist = $pconfig['whitelist'];
+
+//if ($whitelist == "") {
```
marcos-ng, line 282: Unnecessary comments.

src/usr/local/www/services_igmpproxy_edit.php
```diff
@@ -252,6 +275,60 @@
...
+	$group->add(new Form_Button(
+		'deleterow' . $counter,
+		'Delete',
+		null,
+		'fa-trash'
+	))->removeClass('btn-primary')->addClass('btn-warning');
+
+	$counter++;
+	$section->add($group);
+} // end foreach
```
marcos-ng, line 323: Unnecessary comments.

marcos-ng (changes requested):
- IMO it's preferable to have the GUI refer to the "whitelist" options as "passlist".
- The "Add whitelist" button functionality conflicts with the "Add network" button functionality.

src/usr/local/www/services_igmpproxy_edit.php
IgorYbema: Thank you for pointing this

marcos-ng:
It would be an issue if the config allowed for a separate "passlist" option. Since that is not the case, the GUI can adapt the modern practice. As for the button conflict, the input fields being added/removed can jump between the two options.

PR #4627, opened 2023-02-22
Typo : dhcpc6vpt => dhcp6cvpt

Noelite:
There is a typo in the firewall rules generation code
It's not dhcpc6vpt but dhcp6cvpt in filter.inc

### DIFF
--- a/src/etc/inc/filter.inc
+++ b/src/etc/inc/filter.inc
@@ -3727,7 +3727,7 @@ EOD;
 			// The DHCPv6 client rules ***MUST BE ABOVE BOGONSV6!***  https://redmine.pfsense.org/issues/3395
 			$vlantag = '';
 			if (is_array($vlanprio_values)) {
-				$vlantag = array_get_path($vlanprio_values, config_get_path("interfaces/{$on}/dhcpc6vpt", ''), '');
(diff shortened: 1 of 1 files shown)


src/etc/inc/filter.inc
```diff
@@ -3727,7 +3727,7 @@ EOD;
 			// The DHCPv6 client rules ***MUST BE ABOVE BOGONSV6!***  https://redmine.pfsense.org/issues/3395
 			$vlantag = '';
 			if (is_array($vlanprio_values)) {
-				$vlantag = array_get_path($vlanprio_values, config_get_path("interfaces/{$on}/dhcpc6vpt", ''), '');
+				$vlantag = array_get_path($vlanprio_values, config_get_path("interfaces/{$on}/dhcp66vpt", ''), '');
```
jim-p, line 3730: That's not quite right either. Looks like it was introduced in aea387d30c2a7dc3e94aef80854bb19276a2e8de but the commit here is changing it to a different typo. :-)

I'll commit a fix here shortly and mention you in the commit msg to attribute pointing it out, thanks!

jim-p (changes requested):

PR #4630, opened 2023-03-31
Create 1:1 NAT mappings in DNS resolver

yakatz:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/14208
- [x] Ready for review

### DIFF
--- a/src/etc/inc/unbound.inc
+++ b/src/etc/inc/unbound.inc
@@ -878,21 +878,41 @@ function unbound_generate_zone_data($domain, $hosts, &$added_ptr, $zone_type = "
 	return $zone_data;
 }
 
-function unbound_generate_respip_data($nat_entry) {
-
-	if (is_ipaddrv4($nat_entry['external'])) {
+function unbound_generate_respip_data($nat_rule) {
+	$zone_data = '';
+	$mask = intval($nat_rule['srcmask']);
+	
+	if ($nat_rule['ipprotocol'] == 'inet') {
 		$type = 'A';
-		$external_addr = "{$nat_entry['external']}/32";
-	} else if (is_ipaddrv6($nat_entry['external'])) {
-		$type = 'AAAA';
-		$external_addr = "{$nat_entry['external']}/128";
-	} else {
-		return "";
-	}
-	$redirect_addr = $nat_entry['source']['address'];
+		$external_first_address = "{$nat_rule['external']}/{$mask}";
+		$internal_first_address = "{$nat_rule['src']}/{$mask}";
+		
+		$external_all_addresses = subnetv4_expand($external_first_address);
+		$internal_all_addresses = subnetv4_expand($internal_first_address);
+		
+		$addr_mappings = array_map(function ($k, $v) {
+			return ["{$k}/32", $v];
+		}, $external_all_addresses, $internal_all_addresses);
 
-	$zone_data  = "response-ip: {$external_addr} redirect\n";
-	$zone_data .= "response-ip-data: {$external_addr} \"{$type} {$redirect_addr}\"\n";
+	} elseif ($nat_rule['ipprotocol'] == 'inet6') {
+		$type = 'AAAA';
+		
+		// IPv6 ranges are not supported
+		if ($mask != 128) {
+			return '';
+		}
+
+		$addr_mappings[] = ["{$nat_rule['external']}/128", $nat_rule['src']];
+	} else {
+		return '';
+	}
+
+	// Create a separate entry for each IP in a subnet
+	foreach ($addr_mappings as $m) {
+		list($external_addr, $internal_addr) = $m;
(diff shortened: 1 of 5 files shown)


src/usr/local/www/services_unbound.php
```diff
@@ -435,6 +437,14 @@ function build_if_list($selectedifs) {
 	    'in "tun" mode. The domain in %sSystem: General Setup%s should also be set to the proper value.',
 	    '<a href="system.php">','</a>'));
 
+$section->addInput(new Form_Checkbox(
+	'replace1to1',
+	'1:1 NAT Addresses',
+	'Replace 1:1 NAT IP Addresses in Responses',
+	$pconfig['replace1to1']
+))->setHelp('If this option is set, then when an A record is returned with an IP address from the 1:1 NAT table, the value will be replaced'.
+					' with the internal IP address. This is an alternative to manually creating split-DNS.');
```
jim-p, line 446: It seems like it's worth having a strong note here stating that this activates the `respip` module which inspects and alters DNS responses directly, which can degrade performance.

I could see some users having privacy or performance concerns with that, but so long as there is a warning there it should be fine.

src/usr/local/pfSense/include/www/firewall_nat_1to1.inc
```diff
@@ -229,6 +229,11 @@ function save1to1NATrule($post, $id, $json = false) {
 
 		if (write_config(gettext("Firewall: NAT: 1:1 - saved/edited NAT 1:1 mapping.")) && !$json) {
 			mark_subsystem_dirty('natconf');
+
+			if (isset($config['unbound']['enable']) && isset($config['unbound']['replace1to1'])) {
```
jim-p, line 233: This check should use the newer functions like the one in the next part of the diff:

```php
if (config_path_enabled('unbound') && config_path_enabled('unbound', 'replace1to1')) {
```

src/etc/inc/unbound.inc
```diff
@@ -767,6 +787,25 @@ function unbound_generate_zone_data($domain, $hosts, &$added_ptr, $zone_type = "
...
+	if (is_ipaddrv4($nat_entry['external'])) {
+		$type = 'A';
+		$external_addr = "{$nat_entry['external']}/32";
+	} else if (is_ipaddrv6($nat_entry['external'])) {
+		$type = 'AAAA';
+		$external_addr = "{$nat_entry['external']}/128";
+	} else {
+		return "";
+	}
+	$redirect_addr = $nat_entry['source']['address'];
```
jim-p, line 801: This looks like it would be OK for the basic case when a 1:1 NAT entry is for an IP address set to "Single Host" but when about when it's for a subnet?

If the internal IP address type is set to Network and given a subnet mask, it makes a whole subnet's worth of 1:1 NAT in a single entry (e.g. `x.x.x.0/24 <-> y.y.y.0/24`). Similar case for the interface network macros as they are also networks, not addresses

If you want to restrict this to acting on single addresses only, that would make sense, but it would need to test the internal address type to make sure it's only what you want to allow (single host and maybe the "<interface> Address" macros).

If you do restrict the behavior, be sure to note that in the option help as well.

jim-p (changes requested):

src/etc/inc/unbound.inc
yakatz: I did a significant amount of testing including for large IP ranges and this should work properly.

src/usr/local/www/services_unbound.php
yakatz: Help text improved.

yakatz:
Changes in the second commit:

- I added options on each NAT mapping to override whether it would be included.
- I centralized the logic of which rules would be included so it could potentially be shown in the user interface
- I moved some of the NAT rule processing to `util.inc` so it could be reused in the unbound processing

jim-p (commented):
Seems OK, still has some minor source style/formatting issues but nothing stands out as being a problem. Too late for 24.03 but we can revisit it for 24.07.

yakatz:
> still has some minor source style/formatting issues

I definitely want to submit code that is up to the standards - I read the style guide and I am not sure what needs to be improved. Can you give me anything more specific?

jim-p:
Mostly it's about using K&R style braces/formatting and using parenthesis around every test that has multiple parts in `if` statements, even for single-line `if` statements.

e.g.

bad:
```
if (a == b && c == d) {
```
good:
```
if ((a == b) && (c == d)) {
```

Bad:
```
if (a == b) { something; }
```
Good:
```
if (a == b) {
	something;
}
```

yakatz:
Got it. For the one-line `if (...) { return false; }`, I just thought that looked simpler.
I just pushed a change to all of them.

yakatz:
Rebased again.

PR #4632, opened 2023-04-13
Fixed One.com DynDNS updater not working

AnonymousRetard:
- Login URL is now retrieved from the <form>-tag action attribute
- CURL COOKIEJAR is used as I was unable to make login work otherwise. Previous cookie-extractor regex won't honor cookie domain rules, paths, etc which could be the issue? Temporary cookie file is deleted after process is complete.
- Allow a hostname of "@" to refer to the root domain

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/14276
- [ ] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -1306,53 +1306,41 @@
 					/* see https://redmine.pfsense.org/issues/11293
 					 * and https://redmine.pfsense.org/issues/12352 */
 
+					$cookieFile = "/var/tmp/one.com.cookie";
 					curl_setopt($ch, CURLOPT_URL, "https://www.one.com/admin/");
-					curl_setopt($ch, CURLOPT_HEADER, 1); //return the full headers to extract the cookies
+					curl_setopt($ch, CURLOPT_HEADER, 0);
+					curl_setopt($ch, CURLOPT_COOKIEJAR, $cookieFile);
 					curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1);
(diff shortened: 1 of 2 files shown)


marcos-ng:
It should be possible to correctly extract the cookie without needing to introduce the new behavior of keeping track of files on the system.

AnonymousRetard:
@marcos-ng Probably, although I will probably not spend any time trying to figure it out. When initially developing this I was doing a couple of tests in Postman and I was never able to sign in by manually copying cookies. So I'm pretty sure already it's more than one cookie that will need copying.

Without this change the one.com DynDNS updater doesn't work at all and with this change it still works for me and at least one more guy from the pfSense forums: https://forum.netgate.com/topic/124904/dynamic-dns-one-com/35

I am still running with this change on pfSense 2.7.0 now.

I initially reported this issue to redmine together with what I had found out about how to make it work again and was asked to create a pull request, so I did. My hope was that someone else would pick it up from there.

I'm not really interested in spending more time trying to get this merged or to maintain it. I don't like the fact at all that one.com doesn't have any real API to do these things and that the only way to do it is by trying to scrape the website to find form elements to post to. I only signed up with them because they had a cheap deal at the time and because it looked like acme and the pfsense dyndns updater were supporting them. It was only later I found out that apparent this stopped working a long time ago and we seem to be very few users using this because there doesn't seem to be a lot of interest in fixing it or even complaining that it doesn't work.

When my domain expires I might just look at another provider that has APIs to do these things instead. The Acme package also doesn't work any more to update certificates for one.com: https://github.com/acmesh-official/acme.sh/issues/2103

PR #4635, opened 2023-04-30
Releng 2 7 0 backup file

mikko70:
Hi

Because almost every new pfSense version change config.
Request for config backup file, add config revision to backup file name.
Inside back up file is marking <version>22.9</version> but is easiest found proper file, if 22.9 added in filename. 
Example config-firewall.PFSFW-20230430091137.xml -> 22.9_config-firewall.PFSFW-20230430091137.xml
I'm not sure, if we not need  "091137" part anymore.


jim-p:
This PR is very broken, making many changes in places it shouldn't.

Also, before changing any code or submitting a PR, please propose a change in a feature request on https://redmine.pfsense.org/ if one does not already exist.

Including the configuration revision in the filename isn't a bad idea per se, but it would need to be done in a way that makes it clear it's not the pfSense version. Timestamps are still necessary.

PR #4637, opened 2023-05-07
DHCP Server - Add Checkbox to disable update-optimization in DHCPD

styletronix:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/14294
- [x] Ready for review
 
Added checkbox to DDNS section of DHCP and DHCPv6 to disable update-optimization

Added handling of "update-optimization" in services.inc


marcos-ng:
The description of the option implies that this can be set per-client. If so, the option should be available within static mappings as well.

PR #4645, opened 2023-06-20
Fixed message about adaptive state handling

pypckompsite:
Fixed misleading message regarding adaptive state handling.

States are reduced from 100% to the $scalingfactor value, NOT by the value.

Optionally, this could be fixed by subtracting the scaling factor from 100%.


marcos-ng:
Please see https://docs.netgate.com/pfsense/en/latest/development/pull-request.html

PR #4649, opened 2023-09-18
Captive Portal - Adding OTP authentication option to the portal's authentication options

xNUTx:
I have created a small modification to the captive portal pages so it would be possible to use an OTP as login option for the portal.

This will allow for an ever changing password on the portal but also control over who gets to use it.

- [ ] Ready for review


xNUTx:
fixed for the latest version.

PR #4657, opened 2023-11-14
#14981 Fix Gateway WAN_DHCP6 or WAN_SLAAC pending/Unknown issue

dh377:
Re: [WAN_DHCP6 Pending/Unknown](/topic/159912/wan_dhcp6-pending-unknown)

This is quite an old bug on Comcast/Xfinity. Especially when using XFINITY/xfinitywifi as WAN, those do not propagate IPv6 Prefix through DHCP/RA, and `dhcp6c` or `rtsold` could not properly run gateway setup scripts internally. This could be fixed with making it fetch a default gateway address anyway through `netstat`.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/14981
- [x] Ready for review

### DIFF
--- a/src/etc/inc/interfaces.inc
+++ b/src/etc/inc/interfaces.inc
@@ -5086,6 +5086,10 @@ EOD;
 			}
 		}
 
+		/* A default router file will be written by RTSOLD */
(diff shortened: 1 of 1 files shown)


dh377:
Invalid PR. CRLF line-ending make the gateway setup script not run.

PR #4659, opened 2023-12-02
Fix: local private ipv6 is being used instead of public track interface IPv6 in radvd

Noelite:
- created function is_private_ipv6
- added a check in get_interface_track6ip to skip private v6 address

- [x] Redmine Issue: https://redmine.pfsense.org/issues/15057
- [x] Ready for review


marcos-ng:
There's no need for a new function, `is_v6gua()` already exists. There should also be a fallback to ULA if a GUA isn't found. However, further testing/troubleshooting is needed to understand the issue better before code is merged - see my comments on the redmine.

marcos-ng:
Fixed with a different commit - details on redmine.

PR #4664, opened 2023-12-21
Add option to not use system DNS for DNS Forwarder. Fix #14165

opoplawski:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/14165
- [x] Ready for review


marcos-ng:
Needs some minor rewording, but looks fine to me otherwise. What use case do you have for separating the DNS Forwarder servers from those used by the system?

opoplawski:
Currently we have the forwarder serving DNS to our public/visitor network which we don't want resolving internal IPs.  Thanks for merging!

PR #4665, opened 2024-01-04
Fix for Redmine #15140 by killing invalid syslog states after FRR package is started. 

FreeWillyB:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/15140
- [x] Ready for review

Note: This could probably be improved by checking connectivity to the syslog server instead of a static sleep time of 15 seconds. However, the fix worked during testing, and I'm no longer experiencing the issue.

### DIFF
--- a/src/etc/rc.start_packages
+++ b/src/etc/rc.start_packages
@@ -68,6 +68,54 @@ foreach (config_get_path('installedpackages/package', []) as $pkgid => $package)
 	start_service($internal_name, true);
 	unset($rcfiles[RCFILEPREFIX . strtolower($internal_name) . ".sh"]);
 	echo "done.\n";
+        
+	/* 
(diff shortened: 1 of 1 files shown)


jim-p:
We do not want to hardcode knowledge of specific packages in the base system. The proper way to do this is to setup a plugin style system where the system checks any installed packages to find out if they have some code they want to run at this point, and then the check would go into the package itself where it registers the plugin at install time.

There are a few recent examples of this, such as https://redmine.pfsense.org/issues/14777  (and https://redmine.pfsense.org/issues/14588 for the FRR part)

Also, such behavior should be optional and off by default, the code here doesn't check to make sure the remotes are even involved in routing over FRR, so it could be harmful if they are routed in other ways.

PR #4666, opened 2024-01-23
upnp: Update plugin title/menu to include PCP as supported

Self-Hosting-Group:
Update plugin title and menu option to `UPnP IGD & PCP/NAT-PMP` as `PCP` is supported and also mention `UPnP IGD` as `UPnP` is probably not specific enough.

*Updated title/menu option:*
`UPnP IGD & PCP/NAT-PMP`
*Current title/menu option:*
`UPnP & NAT-PMP`

Port Control Protocol (PCP) is the successor to NAT-PMP and has similar protocol concepts and packet formats, but adds IPv6 support.
*PCP standard:*
https://datatracker.ietf.org/doc/html/rfc6887
https://en.wikipedia.org/wiki/Port_Control_Protocol
*NAT-PMP std. (see 9.1 Simplicity - 9.3 for some diff. to UPnP IGD):*
https://datatracker.ietf.org/doc/html/rfc6886

- [x] Ready for review

### DIFF
--- a/src/etc/inc/priv.defs.inc
+++ b/src/etc/inc/priv.defs.inc
@@ -1004,8 +1004,8 @@ $priv_list['page-status-dns-resolver']['match'] = array();
 $priv_list['page-status-dns-resolver']['match'][] = "status_unbound.php*";
 
 $priv_list['page-status-upnpstatus'] = array();
-$priv_list['page-status-upnpstatus']['name'] = gettext("WebCfg - Status: UPnP Status");
-$priv_list['page-status-upnpstatus']['descr'] = gettext("Allow access to the 'Status: UPnP Status' page.");
+$priv_list['page-status-upnpstatus']['name'] = gettext("WebCfg - Status: UPnP IGD & PCP/NAT-PMP");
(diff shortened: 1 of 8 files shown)


jim-p:
IMO this is unnecessary. It makes it even more confusing/busier/more cluttered than it already is. I don't think there is a need to list every single protocol there when people get the idea from what it already says. Maybe some of the description text could mention it, but it doesn't need to be in every menu entry and page title.

marcos-ng:
Thanks for the feedback. I've clarified the text around UPnP, see commit 429312ff683b5762d0cb6eca69c474121a9dfca1.

marcos-ng:
This goal is to be clear about _purpose_ while keeping things simple. That purpose is "automatic port mapping and rule configuration" which is the "UPnP" part in the references "UPnP Port Mapping", "UPnP protocol", "UPnP service", and "UPnP rules".

marcos-ng:
Thanks for the contribution, Please open a new PR based off the new branch/changes and we'll review one last time before merging.

marcos-ng:
I tried but github won't let it reopen. If this is going to make it into the release, I rather simply open a new one for now.

PR #4672, opened 2024-02-29
Updated much faster log.widget Bug #12673

LouisAtGH:
This widget is to replace the actual log.widget.. Loading the log in one go does speed up the widget about 20 times. I left debug statements in the code (for now) so that it is easy to folow what is happening. For more detail see Bug #12673: Firewall Logs Widget fails to update at intervals below 5 seconds. - pfSense - pfSense bugtracker

- [Bug #12673] Redmine Issue: https://redmine.pfsense.org/issues/NNNN
- [ ] Ready for review


jim-p:
The PR was creating a new file instead of updating the existing one and had some whitespace/formatting issues but I cleaned it up and committed it manually. See 1586f580e6ce8f39cb235b0bdc14d470501f7714 which should show up on GitHub shortly.

Thanks!

PR #4673, opened 2024-03-08
Add description to WOL log message

PhilZ-cwm6:
When WOL is sent using mac url or power-on button, display the server description in output log message

- [X] Redmine Issue: https://redmine.pfsense.org/issues/15323
- [X] Ready for review


jim-p:
This is the wrong approach and introduces security concerns by trusting input that could be manipulated by the user. This is the sort of usage that can introduce problems such as cross-site scripting vulnerabilities, especially since the value is taken from POST data and placed directly into the HTML output without any encoding.

If the description is stored in a WOL entry in the configuration, then the function to send the WOL packet should search through the WOL entries there for one matching the MAC+Interface and then take the description stored in config.xml -- it should not take the description from POST or pass it around that way.

PR #4676, opened 2024-03-25
fix: using first or last ipv6 address in VIP alias causes error

Noelite:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/15361
- [x] Ready for review

IPv6 addresses should not be checked for network/broadcast addresses


marcos-ng:
Closing this as it's already been addressed.

PR #4684, opened 2024-05-08
statishe -> statische

Presskopp:


Presskopp:
also added flags -> flags 'translation' instead of 'Fahnen' which here makes no sense

PR #4689, opened 2024-07-12
UPnP IGD & PCP: Simplify service wording

Self-Hosting-Group:
and change menu/title to `UPnP IGD & PCP` as newer PCP protocol is supported

- Replace the frequently repeated protocols term 9 times with `Service`
- Rename `UPnP & NAT-PMP Rules` to `Active Service Port Maps` and `Port` to `Ext Port`, add `Source Port` and remove `Ext IP` as always `any` on the status page
- Use only `UPnP IGD & PCP`, including the newer IETF protocol, as the full term will get automatic truncated (see screenshot) on mobile to just `NAT-PMP`, but use `PCP/NAT-PMP` in the actual setting
- Use `UPnP IGD` as `UPnP` alone is not specific enough as UPnP is also used as a MediaServer
- Add help text: `Enable automatic port mapping service`

### Proposed service settings UI

pfsense-igd-pcp-service-settings

*[Full screenshot](https://github.com/pfsense/pfsense/assets/155233284/ec498ee4-a254-4cda-8ce6-ca16dc0d0a3d)*

### Current service settings UI

pfsense-igd-pcp-service-settings-old

### Proposed status UI

pfsense-igd-pcp-status

### Unexpected truncation on mobile

pfsense-wrapping

Port Control Protocol (PCP) is the successor to NAT-PMP and has similar protocol concepts and packet formats, but adds IPv6 support.

*PCP standard:*
https://datatracker.ietf.org/doc/html/rfc6887
https://en.wikipedia.org/wiki/Port_Control_Protocol
*NAT-PMP std. (see 9.1 Simplicity - 9.3 for some diff. to UPnP IGD):*
https://datatracker.ietf.org/doc/html/rfc6886
*Port Mapping Protocols Overview:*
https://github.com/Self-Hosting-Group/wiki/wiki/Port-Mapping-Protocols-Overview

### DIFF
--- a/src/etc/inc/priv.defs.inc
+++ b/src/etc/inc/priv.defs.inc
@@ -1004,8 +1004,8 @@ $priv_list['page-status-dns-resolver']['match'] = array();
 $priv_list['page-status-dns-resolver']['match'][] = "status_unbound.php*";
 
 $priv_list['page-status-upnpstatus'] = array();
-$priv_list['page-status-upnpstatus']['name'] = gettext("WebCfg - Status: UPnP Status");
-$priv_list['page-status-upnpstatus']['descr'] = gettext("Allow access to the 'Status: UPnP Status' page.");
+$priv_list['page-status-upnpstatus']['name'] = gettext("WebCfg - Status: UPnP IGD & PCP");
+$priv_list['page-status-upnpstatus']['descr'] = gettext("Allow access to the 'Status: UPnP IGD & PCP' page.");
 $priv_list['page-status-upnpstatus']['match'] = array();
 $priv_list['page-status-upnpstatus']['match'][] = "status_upnp.php*";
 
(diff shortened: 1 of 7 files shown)


marcos-ng:
If this is ready to merge, please remove the draft tag.

jim-p (approved):

jim-p:
I'd just pushed a fix for that, might not have showed up yet since it take a bit to sync all the commits around.

jim-p:
> @jim-p: Thanks for the merge. Can the documentation team be pinged to update the following page? See: https://docs.netgate.com/pfsense/en/latest/services/upnp.html

I handle the majority of those updates. Things like that will usually be updated with a new release and not before, so they get staged in a release branch and then deployed along with the new release. That way the docs don't get ahead of what users have available. There are occasional exceptions like for big features we want people to test in beta stages but in most cases they are held back.

That said, any docs change requests or reminders of that nature should be filed here: https://redmine.pfsense.org/projects/pfsense-docs/issues

PR #4691, opened 2024-07-19
Fix OpenVPN CSC options. Fix #12522

PhilZ-cwm6:
Also add override/remove options for Redirect Gateways and for Block Outside DNS.

- [X] Redmine Issue: https://redmine.pfsense.org/issues/12522
- [X] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1678,39 +1678,42 @@ function openvpn_resync_csc($settings) {
 	}
 
 	// Reset options
-	if (isset($settings['override_options']) && ($settings['override_options'] == 'push_reset')) {
-		if (isset($settings['keep_minimal'])) {
-			$auto_config_gateway4 = true;
-			$auto_config_gateway6 = true;
-			$auto_config_topology = true;
-		}
-		$conf .= "push-reset\n";
-	}
-	if (!empty($settings['remove_options'])) {
-		foreach (explode(',', $settings['remove_options']) as $option) {
-			if (isset($settings['keep_minimal']) && ($option == 'remove_route')) {
+	if (isset($settings['override_options'])) {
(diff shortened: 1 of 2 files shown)


marcos-ng:
The feedback/testing is appreciated - thanks!

I'm not clear on what you mean with:
> the push-remove options were not processed properly

The new remove options and order look good.

The `keep_minimal` config item is meant to be a flag. When it's unchecked, it gets removed from the config. When checked, it gets added as an empty config item. The proposed changes would change that flag behavior. I do see that the value is not checked correctly when the page loads though, good catch!

The only reason to move `push_reset` into `override_options` is if `override_options` were to be used for more than one setting. Since the GUI code can determine the correct `override_options` selection without introducing a new config item, then `push_reset` can stay where it is and upgrade code isn't needed for backwards compatibility.

With the above in mind, I'm thinking 012e36e6b87cf63743e644179a15e67f38d94f77 is good. What do you think?

PhilZ-cwm6:
> The feedback/testing is appreciated - thanks!
> 
> I'm not clear on what you mean with:
> 
> > the push-remove options were not processed properly
> 
> The new remove options and order look good.
> 
> The `keep_minimal` config item is meant to be a flag. When it's unchecked, it gets removed from the config. When checked, it gets added as an empty config item. The proposed changes would change that flag behavior. I do see that the value is not checked correctly when the page loads though, good catch!
> 
> The only reason to move `push_reset` into `override_options` is if `override_options` were to be used for more than one setting. Since the GUI code can determine the correct `override_options` selection without introducing a new config item, then `push_reset` can stay where it is and upgrade code isn't needed for backwards compatibility.
> 
> With the above in mind, `012e36e6b87cf63743e644179a15e67f38d94f77` should be good. What do you think?

I finished a full testing. Your last commit is still broken
I will post the details and a fixc in a few minutes

PR #4692, opened 2024-07-20
More revisions to OpenVPN CSC. Fix #12522

PhilZ-cwm6:
More revisions to OpenVPN CSC. Fix #12522
Fix push reset was not applied
Fix some flags were always saved to config

- [X] Redmine Issue: https://redmine.pfsense.org/issues/12522
- [X] Ready for review

### DIFF
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1678,15 +1678,14 @@ function openvpn_resync_csc($settings) {
 	}
 
 	// Reset options
-	if (isset($settings['override_options']) && ($settings['override_options'] == 'push_reset')) {
+	if (isset($settings['push_reset'])) {
 		if (isset($settings['keep_minimal'])) {
 			$auto_config_gateway4 = true;
 			$auto_config_gateway6 = true;
 			$auto_config_topology = true;
 		}
 		$conf .= "push-reset\n";
-	}
-	if (!empty($settings['remove_options'])) {
+	} elseif (!empty($settings['remove_options'])) {
 		foreach (explode(',', $settings['remove_options']) as $option) {
 			if (isset($settings['keep_minimal']) && ($option == 'remove_route')) {
 				$auto_config_gateway4 = true;
--- a/src/usr/local/www/vpn_openvpn_csc.php
+++ b/src/usr/local/www/vpn_openvpn_csc.php
@@ -304,17 +304,14 @@ if ($_POST['save']) {
 	if (!$input_errors) {
 		$csc = array();
 
-		if (isset($pconfig['keep_minimal'])) {
-			$csc['keep_minimal'] = true;
-		}
 		// Handle "Reset Server Options" and "Reset Options"
-		if (($pconfig['override_options'] == 'remove_specified')) {
-			// If no options are specified, keep the default behavior.
-			if (!empty($pconfig['remove_options'])) {
-				$csc['remove_options'] = implode(',', $pconfig['remove_options']);
-			}
-		} elseif ($pconfig['override_options'] == 'push_reset') {
+		if ($pconfig['override_options'] == 'push_reset') {
 			$csc['push_reset'] = true;
+		} elseif (($pconfig['override_options'] == 'remove_specified') && !empty($pconfig['remove_options'])) {
+			$csc['remove_options'] = implode(',', $pconfig['remove_options']);
+		}
+		if (isset($pconfig['keep_minimal']) && (isset($csc['push_reset']) || isset($csc['remove_options']))) {
+			$csc['keep_minimal'] = true;
 		}
 
 		if (is_array($pconfig['server_list'])) {
@@ -507,11 +504,9 @@ if ($act == "new" || $act == "edit"):
 		]
 	))->setHelp('Prevent this client from receiving server-defined client settings. Other client-specific options on this page will supersede these options.');
 
-	$group = new Form_Group('Remove Options');
-	$group->addClass('remove_options');
-	$group->add(new Form_Select(
+	$section->addInput(new Form_Select(
 		'remove_options',
-		null,
+		'Remove Options',
 		$pconfig['remove_options'],
 		[
 			'remove_route' => 'Local Routes & Gateways',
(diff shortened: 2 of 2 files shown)


marcos-ng:
Now I see the issues with `push_reset` are because I missed a line in `openvpn.inc` when changing the behavior back to avoid the upgrade code.

As for `keep_minimal`, I conflated the hide*() js code with disableInput(). The latter results in the field being omitted from $_POST. This we can fix on the csc page by moving the condition check; and since we're doing that anyway, we can avoid duplicating work by moving the `override_options` checks too.

Lastly, `remove_options` doesn't need to be part of a group section, so we can clean that up too. Once you verify the following patch, feel free to update your branch accordingly and I'll merge this PR.

```
diff --git a/src/etc/inc/openvpn.inc b/src/etc/inc/openvpn.inc
index 12799d3d29..c52722d0e5 100644
--- a/src/etc/inc/openvpn.inc
+++ b/src/etc/inc/openvpn.inc
@@ -1678,7 +1678,7 @@ function openvpn_resync_csc($settings) {
 	}
 
 	// Reset options
-	if (isset($settings['override_options']) && ($settings['override_options'] == 'push_reset')) {
+	if (isset($settings['push_reset'])) {
 		if (isset($settings['keep_minimal'])) {
 			$auto_config_gateway4 = true;
 			$auto_config_gateway6 = true;
diff --git a/src/usr/local/www/vpn_openvpn_csc.php b/src/usr/local/www/vpn_openvpn_csc.php
index 50f9aad181..392d68d740 100644
--- a/src/usr/local/www/vpn_openvpn_csc.php
+++ b/src/usr/local/www/vpn_openvpn_csc.php
@@ -304,17 +304,19 @@ if ($_POST['save']) {
 	if (!$input_errors) {
 		$csc = array();
 
-		if (isset($pconfig['keep_minimal'])) {
-			$csc['keep_minimal'] = true;
-		}
 		// Handle "Reset Server Options" and "Reset Options"
-		if (($pconfig['override_options'] == 'remove_specified')) {
-			// If no options are specified, keep the default behavior.
-			if (!empty($pconfig['remove_options'])) {
-				$csc['remove_options'] = implode(',', $pconfig['remove_options']);
+		if (!empty($pconfig['override_options']) && ($pconfig['override_options'] != 'default')) {
+			if (isset($pconfig['keep_minimal'])) {
+				$csc['keep_minimal'] = true;
+			}
+			if ($pconfig['override_options'] == 'remove_specified') {
+				// If no options are specified, keep the default behavior.
+				if (!empty($pconfig['remove_options'])) {
+					$csc['remove_options'] = implode(',', $pconfig['remove_options']);
+				}
+			} elseif ($pconfig['override_options'] == 'push_reset') {
+				$csc['push_reset'] = true;
 			}
-		} elseif ($pconfig['override_options'] == 'push_reset') {
-			$csc['push_reset'] = true;
 		}
 
 		if (is_array($pconfig['server_list'])) {
@@ -507,11 +509,9 @@ if ($act == "new" || $act == "edit"):
 		]
 	))->setHelp('Prevent this client from receiving server-defined client settings. Other client-specific options on this page will supersede these options.');
 
-	$group = new Form_Group('Remove Options');
-	$group->addClass('remove_options');
-	$group->add(new Form_Select(
+	$section->addInput(new Form_Select(
 		'remove_options',
-		null,
+		'Remove Options',
 		$pconfig['remove_options'],
 		[
 			'remove_route' => 'Local Routes & Gateways',
@@ -529,8 +529,7 @@ if ($act == "new" || $act == "edit"):
 			'remove_wins' => 'WINS Options'
 		],
 		true
-	))->setHelp('A "push-remove" option will be sent to the client for the selected options, removing the respective server-defined option.');
-	$section->add($group);
+	))->addClass('remove_options')->setHelp('A "push-remove" option will be sent to the client for the selected options, removing the respective server-defined option.');
 
 		'keep_minimal',
@@ -948,7 +947,7 @@ events.push(function() {
 
 	function remove_options_change() {
 		hideCheckbox('keep_minimal', ($('#override_options').find('option:selected').val() == 'default'));
-		hideClass('remove_options', ($('#override_options').find('option:selected').val() != 'remove_specified'));
+		hideMultiClass('remove_options', ($('#override_options').find('option:selected').val() != 'remove_specified'));
 	}
 
 	// ---------- Click checkbox handlers ---------------------------------------------------------

```

PhilZ-cwm6:
Looks good to me except 1 point:
`keep_minimal` is still saved in your patch when `override_options` is set to `remove_specified` and `override_options` is left empty. The default behaviour is properly saved but the `keep_minimal` flag is preserved.

Also, a smal commit in inc file: `elseif` instead of `if` as `push_reset` and `override_options` can never be set at the same time

I'm pushing your patch and my small changes. Let me know if you want me to squash the commits

marcos-ng:
Good catch. In that case we can simplify things a bit by doing this instead:
```
		// Handle "Reset Server Options" and "Reset Options"
		if ($pconfig['override_options'] == 'push_reset') {
			$csc['push_reset'] = true;
		} elseif (($pconfig['override_options'] == 'remove_specified') && !empty($pconfig['remove_options'])) {
			$csc['remove_options'] = implode(',', $pconfig['remove_options']);
		}
		if (isset($pconfig['keep_minimal']) && (isset($csc['push_reset']) || isset($csc['remove_options']))) {
			$csc['keep_minimal'] = true;
		}
```
The rest looks good. Please rebase the branch, make the appropriate changes, and squash the PR to one commit.

PhilZ-cwm6:
Still your simplification won't work properly. I though at it in all ways ending with that simple commit I did.

```
if (isset($pconfig['keep_minimal']) && (isset($csc['push_reset']) || isset($csc['remove_options']))) {
	$csc['keep_minimal'] = true;
}
```
This will cause `keep_minimal` to still be preserved if `remove_options` are left empty, while it shouldn't (falls to default)

Also, in PHP, we should check for unset or null variable before testing it in a condition to avoid warning and maybe future deprecation

So, the code 

```
if ($pconfig['override_options'] == 'push_reset') {
	$csc['push_reset'] = true;
} elseif (($pconfig['override_options'] == 'remove_specified') && !empty($pconfig['remove_options'])) {
	$csc['remove_options'] = implode(',', $pconfig['remove_options']);
}
```
Will have to be:
```
if (!empty($pconfig['override_options'])) {
	if ($pconfig['override_options'] == 'push_reset') {
		$csc['push_reset'] = true;
	} elseif (($pconfig['override_options'] == 'remove_specified') && !empty($pconfig['remove_options'])) {
		$csc['remove_options'] = implode(',', $pconfig['remove_options']);
	}
}
```

Ending with a more complicated code.
Since the conditions are really intricated in this scenario, I think simplifying it too much will impact readability and hierachical logic

If you agree, I rebase and squash current code

Thank you again for all your followup on this

PhilZ-cwm6:
You're right, I didn't notice the check `isset($csc['remove_options'])` was only true if some option is selected. Confused it with pconfig
I will rebase as soon as I test it in openvpn client
Thank you

PhilZ-cwm6:
Looks good to me know as per your code suggestions

PR #4694, opened 2024-08-14
Fixes #14289, #10000 and #15593, plus various code improvements.

introzen:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/14289
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/15593
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/10000
- [ ] Ready for review

Allows dyndns apex records for Name.com and Azure via "@" in the Host input field. 

Unfortunately the Name.com API seems to accept both "@" and empty strings for the Host field while it is only possible to create "empty string"-records in their web application. This means that sending an "@" in the API request creates duplicate records in the Name.com DNS table. In this PR, a workaround solution is provided by replacing "@" with an empty string before saving & updating the record.

Various code improvements.

### DIFF
--- a/src/usr/local/www/services_dyndns_edit.php
+++ b/src/usr/local/www/services_dyndns_edit.php
@@ -347,7 +347,7 @@ $group->add(new Form_Input(
 ));
 
 $group->setHelp('Enter the complete fully qualified domain name. Example: myhost.dyndns.org%1$s' .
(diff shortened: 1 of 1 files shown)


marcos-ng:
Thank you for the contribution. I've implemented this slightly differently for better readability. See:
- 88e4113ba319eba2d50cb523729c85eca438d470
- 60a5fd09e14bf64aca79efd9a27a42164e872c90
- 33626a940a92ba93c6f21c6241603c41a8e07bbd

PR #4695, opened 2024-08-23
UPnP IGD & PCP: Improve descriptions

Self-Hosting-Group:


marcos-ng:
Regarding:

> Set the maximum upload speed in kbit/s, informational only.

What is meant by `informational only`?

marcos-ng:
Thank you for explaining. I think we can leave the interface column since it provides a quick way to know which WAN interface the mapping was set up for (e.g. in case of failover gateways).

That max speed description is certainly misleading. Perhaps we can clarify that more by saying "Report a maximum upload speed (kbit/s)."

marcos-ng:
Looks good. Remove the draft flag from the PR and perhaps simplify the commit message a bit - like "UPnP IGD & PCP: improve descriptions". Should be good to merge after.

systemcrash:
How about `What is my maximum download speed in kbit/s?`

otherwise `Report maximum download speed in kbit/s as`

marcos-ng:
I suggested "report" because we're reporting to clients of the max speed. That also implies it's informational only.

PR #4696, opened 2024-08-27
Remove unnecessary "(SSH only)" from logout option in console menu - Redmine Issue #15705

lriley2020:
- [x] Ready for review

This PR removes the "(SSH only)" part of the "0) Logout" option on the console menu. This is because the logout option on the menu is not exclusive to people connecting to the console over SSH, and having "(SSH only)" potentially creates confusion.

Password protection for the console menu can in fact be enabled (System/Advanced/Admin Access/Console Options/Password protect the console menu), which secures the physical console as well as remote SSH connections. I have tested and ensured that the "0) Logout" option does still work as expected on a physical console when console password protection is enabled.

I assumed this change falls under a "very minor typo/wording fix", so didn't create a bug entry on Redmine, although I would be happy to do so if this is neccesary :)

### DIFF
--- a/src/etc/rc.initial
+++ b/src/etc/rc.initial
@@ -65,7 +65,7 @@ fi
 
 # display console menu
 echo ""
-echo " 0) Logout (SSH only)                  9) pfTop"
+echo " 0) Logout / Disconnect SSH            9) pfTop"
 echo " 1) Assign Interfaces                 10) Filter Logs"
 echo " 2) Set interface(s) IP address       11) Restart GUI"
 echo " 3) Reset admin account and password  12) PHP shell + ${product_label} tools"


jim-p:
This would still warrant a Redmine since it's not really a typo, it's changing a menu option in a non-trivial way and it would need a Redmine entry to be certain it appears in the release notes.

You are correct, though, the `0` menu option always logs the user out no matter how they logged in. Even in the case of an open console they are logged out, but autologin logs them right back in and ends up redrawing the menu so they may not realize it was a logout/login event.

What might need some more thought is that pressing `Enter` without providing input has the same effect as pressing `0`. And SSH, unlike the other paths to the menu, gets disconnected, so it's still a special case. But there may not be enough room to fit all that in while being clear about what happens. Maybe one of these:

- `0) Logout / Disconnect SSH`
- `0/Enter) Logout / Disconnect SSH`
- `0/Empty) Logout / Disconnect SSH`

Or something else along those lines.

lriley2020:
Just created an issue on Redmine :)

marcos-ng:
Update the commit message to `Clarify console menu option 0 description. Implement #15705` and change the menu text to `0) Logout / Disconnect SSH`. Should be good to merge after.

lriley2020:
Have just made those changes now :+1:

marcos-ng:
Please squash the commits as well - thanks!

lriley2020:
Should be okay now I think!

PR #4697, opened 2024-09-14
Widget refresh rework

GChuf:
- [ x ] Redmine Issue: https://redmine.pfsense.org/issues/NNNN
- [ x ] Ready for review


GChuf:
@marcos-ng can you take a look? More detailed explaination in redmine. Thank you!

marcos-ng:
Should be good to merge after the following is addressed:

- Drop the refresh interval commit - I will adjust them separately for CE/Plus. 
- Keep the `ajaxcntr` condition as is - no need to change it.
- Avoid the single-use variable.
- Log errors to the browser console.
- Fix the whitespace.

```
diff --git a/src/usr/local/www/index.php b/src/usr/local/www/index.php
index 4293d25018..ce5037ec9a 100644
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -545,8 +545,6 @@ function set_widget_checkbox_events(checkbox_panel_ref, all_none_button_id) {
 // ---------------------Centralized widget refresh system -------------------------------------------
 // These need to live outside of the events.push() function to enable the widgets to see them
 var ajaxspecs = new Array();	// Array to hold widget refresh specifications (objects )
-var ajaxidx = 0;
-var ajaxmutex = false;
 var ajaxcntr = 0;
 
 // Add a widget refresh object to the array list
@@ -615,8 +613,6 @@ events.push(function() {
 	ajaxtimeout = false;
 
 	function make_ajax_call(wd) {
-		ajaxmutex = true;
-
 		$.ajax({
 			type: 'POST',
 			url: wd.url,
@@ -636,38 +632,30 @@ events.push(function() {
 					}
 				}
 
-				ajaxmutex = false;
 			},
 
 			error: function(e){
-//				alert("Error: " + e);
-				ajaxmutex = false;
+				console.log("Error: " + e);
 			}
 		});
 	}
 
+	function execute_ajax_call(item){
+		if ((ajaxcntr % item.freq) === 0) {
+			make_ajax_call(item);
+		}
+	}
+
 	// Loop through each AJAX widget refresh object, make the AJAX call and pass the
 	// results back to the widget's callback function
 	function executewidget() {
-		if (ajaxspecs.length > 0) {
-			var freq = ajaxspecs[ajaxidx].freq;	// widget can specify it should be called freq times around the loop
-
-			if (!ajaxmutex) {
-				if (((ajaxcntr % freq) === 0) && (typeof ajaxspecs[ajaxidx].callback === "function" )) {
-				    make_ajax_call(ajaxspecs[ajaxidx]);
-				}
-
-			    if (++ajaxidx >= ajaxspecs.length) {
-					ajaxidx = 0;
+		ajaxspecs.forEach(execute_ajax_call);
 
-					if (++ajaxcntr >= 4096) {
-						ajaxcntr = 0;
-					}
-			    }
-			}
+		if (++ajaxcntr == 4096) {
+			ajaxcntr = 0;
+		}
 
-		    setTimeout(function() { executewidget(); }, 1000);
-	  	}
+		setTimeout(function() { executewidget(); }, 1000);
 	}
 
 	// Kick it off

```

GChuf:
Thank you, everything was addressed. Good catch on the single use variable.
I hope I understood you correctly about the ajaxcntr condition. I reverted the change from `(++ajaxcntr == 4096)` back to `if (++ajaxcntr >= 4096)`.

PR #4698, opened 2024-09-15
Thermal sensor widget improvement

GChuf:
- [ x ] Redmine Issue: https://redmine.pfsense.org/issues/NNNN
- [ x ] Ready for review


GChuf:
@marcos-ng this is an improved fix for the thermal sensor widget "rebuilding" every time. I fixed the "firstTime" variable along with some other optimizations. Examples of current and fixed behavior described in redmine. Please take a look, thank you!

src/usr/local/www/widgets/widgets/thermal_sensors.widget.php
```diff
@@ -136,9 +136,9 @@ function getBoolValueFromConfig(&$configArray, $valueKey, $defaultValue, $widget
 	//handle checkboxes separately
 	saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_show_raw_output");
 	saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_show_full_sensor_name");
-	saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_pulsate_warning");
-	saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_pulsate_critical");
 	saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_show_fahrenheit");
+	//saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_pulsate_warning");
+	//saveGraphDisplaySettings($user_settings, $_POST, "thermal_sensors_widget_pulsate_critical");
```
marcos-ng, line 141: Remove the code instead of commenting it out and change the commit message to:
```
Remove old code references in thermal widget

Originally removed in https://redmine.pfsense.org/issues/5369
```

src/usr/local/www/widgets/javascript/thermal_sensors.js
```diff
@@ -117,37 +120,19 @@ function buildThermalSensorsDataGraph(thermalSensorsData, tsParams, widgetKey) {
...
-			criticalTemp = tsParams.pchCriticalTempThreshold;
-		} else { //assuming sensor is for a zone, check Zone Threshold config settings
-			warningTemp = tsParams.zoneWarningTempThreshold;
-			criticalTemp = tsParams.zoneCriticalTempThreshold;
-		}
-
-		if (!tsParams.showFullSensorName) {
-			sensorName = getSensorFriendlyName(sensorName);
-		}
-
```
marcos-ng, line 150: It seems like these should be removed from `buildThermalSensorsDataGraph()` instead.

marcos-ng (changes requested):
Squash the "fix" and "optimize" commits and set the commit message to:
```
Improve thermal widget refresh code. Implement #15728

Only replace values instead of entire DOM elements.
```

src/usr/local/www/widgets/javascript/thermal_sensors.js
```diff
@@ -23,17 +23,20 @@
 var warningTemp = 9999;
 var criticalTemp = 100;
 var widgetUnit = 'C';
-ajaxBusy = false;
+var firstTime = true;
 
-function buildThermalSensorsData(thermalSensorsData, widgetKey, tsParams, firstTime) {
+
```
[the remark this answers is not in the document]
GChuf: Removed.

src/usr/local/www/widgets/javascript/thermal_sensors.js
GChuf: I think it's best to calculate these values when "building", and avoid unnecessary work when "updating". When a user updates the thresholds, this results in a full page reload anyway. As long as the user has only one tab open and doesn't change settings in another tab, the thresholds will always be up to date if we only calculate them once at the beginning.

src/usr/local/www/widgets/widgets/thermal_sensors.widget.php
GChuf: Removed and ammended the commit.

PR #4703, opened 2024-10-12
Speed up MBUF Usage command in system information widget

GChuf:
- [ ] Ready for review

### DIFF
--- a/src/usr/local/www/includes/functions.inc.php
+++ b/src/usr/local/www/includes/functions.inc.php
@@ -37,12 +37,7 @@ function get_stats($sitems = array()) {
 	$stats['datetime'] = (!in_array('current_datetime', $sitems)) ? update_date_time() : '';
 	$stats['cpufreq'] = (!in_array('cpu_type', $sitems)) ? get_cpufreq() : '';
 	$stats['load_average'] = (!in_array('load_average', $sitems)) ? get_load_average() : '';
-	if (!in_array('mbuf_usage', $sitems)) {
-		get_mbuf($stats['mbuf'], $stats['mbufpercent']);
-	} else {
-		$stats['mbuf'] = '';
-		$stats['mbufpercent'] = '';
-	}
+	$stats['mbuf_usage'] = (!in_array('mbuf_usage', $sitems)) ? get_mbuf() : '';
 	$stats['statepercent'] = (!in_array('state_table_size', $sitems)) ? get_pfstate(true) : '';
 	$stats = join("|", $stats);
 	return $stats;
@@ -127,11 +122,11 @@ function get_hwtype() {
 	return;
 }
 
-function get_mbuf(&$mbuf, &$mbufpercent) {
-	$mbufs_output=trim(`/usr/bin/netstat -mb | /usr/bin/grep "mbuf clusters in use" | /usr/bin/awk '{ print $1 }'`);
-	list($mbufs_current, $mbufs_cache, $mbufs_total, $mbufs_max) = explode("/", $mbufs_output);
-	$mbuf = "{$mbufs_total}/{$mbufs_max}";
-	$mbufpercent = ($mbufs_max > 0) ? round(($mbufs_total / $mbufs_max) * 100, 0) : "NA";
+function get_mbuf() {
+	$mbufs_output=trim(`/usr/bin/vmstat -z | /usr/bin/grep mbuf_cluster | /usr/bin/grep -v debugnet | /usr/bin/awk '{print $3,$4,$5}'`);
+	list($mbufs_max,$mbufs_used, $mbufs_free) = explode(",", $mbufs_output);
+	$mbufs_total = (int) $mbufs_used + (int) $mbufs_free;
+	return $mbufs_total . "/" . $mbufs_max;
 }
 
 function get_temp() {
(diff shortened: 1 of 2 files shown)


src/usr/local/www/includes/functions.inc.php
```diff
@@ -127,11 +122,11 @@ function get_hwtype() {
 	return;
 }
 
-function get_mbuf(&$mbuf, &$mbufpercent) {
-	$mbufs_output=trim(`/usr/bin/netstat -mb | /usr/bin/grep "mbuf clusters in use" | /usr/bin/awk '{ print $1 }'`);
-	list($mbufs_current, $mbufs_cache, $mbufs_total, $mbufs_max) = explode("/", $mbufs_output);
-	$mbuf = "{$mbufs_total}/{$mbufs_max}";
-	$mbufpercent = ($mbufs_max > 0) ? round(($mbufs_total / $mbufs_max) * 100, 0) : "NA";
+function get_mbuf() {
+	$mbufs_output=trim(`/usr/bin/vmstat -z | /usr/bin/grep mbuf_cluster | /usr/bin/grep -v debugnet | /usr/bin/awk '{print $3,$4,$5}'`);
```
jim-p, line 126: This does not work properly in several cases. The output is not guaranteed to have spaces between columns.

For example:
```
: /usr/bin/vmstat -z | /usr/bin/grep mbuf_cluster | /usr/bin/grep -v debugnet 
mbuf_cluster:          2048,1000000,   12028,    1434,  144434,   0,   0,   0

: /usr/bin/vmstat -z | /usr/bin/grep mbuf_cluster | /usr/bin/grep -v debugnet | /usr/bin/awk '{print $3,$4,$5}'
12027, 1435, 144434,
```

If you want to try going that route, call `vmstat -z --libxo=json` and process that output to find the exact values you want accurately. Or to narrow that a bit:

```
: /usr/bin/vmstat -z --libxo=json | /usr/local/bin/jq '."memory-zone-statistics".zone.[] | select(.name=="mbuf_cluster")'
{
  "name": "mbuf_cluster",
  "size": 2048,
  "limit": 1000000,
  "used": 12028,
  "free": 1434,
  "requests": 144620,
  "fail": 0,
  "sleep": 0,
  "xdomain": 0
}
```

jim-p (changes requested):

src/usr/local/www/includes/functions.inc.php
GChuf: My bad, you're right - I noticed this one when testing but the fix slipped away somewhere in between the testing and commiting.
One idea was to use comma delimiter with awk, but then whitespaces become potentially problematic (output needs to be trimmed). I'll take a look at json output and try to find the cleanest/fastest solution.

GChuf:
Updated the code.
It might not look as clean. I'm using `sub(/^ +/, "", $2)` to cleanup leading whitespaces, and I'm using comma as delimiter for awk.

The other option was:
`/usr/bin/vmstat -z --libxo=json | /usr/local/bin/jq -r '."memory-zone-statistics".zone.[] | select(.name=="mbuf_cluster") | "\(.total) \(.used) \(.free)"'`
After testing on my system, using json and jq was approximately twice slower (but both options are fast).

jim-p:
The libxo method is necessary to ensure accuracy, and is the only method that will be accepted. Anything else is just guessing and making assumptions about the format which could be wrong. It would require a lot more testing on a lot more hardware/networks when all such concerns are eliminated by using libxo which is designed for this purpose.

GChuf:
Understood - changed the code.
Using this now:
`/usr/bin/vmstat -z --libxo=json | /usr/local/bin/jq -r '."memory-zone-statistics".zone.[] | select(.name=="mbuf_cluster") | "\(.limit) \(.used) \(.free)"'`

src/usr/local/www/includes/functions.inc.php
```diff
@@ -127,11 +122,11 @@ function get_hwtype() {
 	return;
 }
 
-function get_mbuf(&$mbuf, &$mbufpercent) {
-	$mbufs_output=trim(`/usr/bin/netstat -mb | /usr/bin/grep "mbuf clusters in use" | /usr/bin/awk '{ print $1 }'`);
-	list($mbufs_current, $mbufs_cache, $mbufs_total, $mbufs_max) = explode("/", $mbufs_output);
-	$mbuf = "{$mbufs_total}/{$mbufs_max}";
-	$mbufpercent = ($mbufs_max > 0) ? round(($mbufs_total / $mbufs_max) * 100, 0) : "NA";
+function get_mbuf() {
+	$mbufs_output=trim(`/usr/bin/vmstat -z --libxo=json | /usr/local/bin/jq -r '."memory-zone-statistics".zone.[] | select(.name=="mbuf_cluster") | "\(.limit) \(.used) \(.free)"'`);
```
jim-p: The libxo output has changed slightly so this no longer works on dev snapshots. The output is now inside a `vmstat` container so the `jq` string needs to start with `."vmstat"`:

```
/usr/bin/vmstat -z --libxo=json | /usr/local/bin/jq -r '."vmstat"."memory-zone-statistics".zone.[] | select(.name=="mbuf_cluster") | "\(.limit) \(.used) \(.free)"'
```

jim-p (changes requested):
Needs one more small change since the vmstat command output format changed, but otherwise looks/works OK once that adjustment is made.

PR #4711, opened 2024-10-28
Shell command

GChuf:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/15798
- [x] Ready for review

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,6 @@
 {
     "name": "pfsense/pfsense",
     "require": {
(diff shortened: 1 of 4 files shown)


jim-p:
This doesn't seem like it's saving anything. Yes, it eliminates a dependency, but calling exec directly carries more/different risks than a library that attempts to do so in a safer manner with isolated arguments. It may only be used here at the moment, but that is unlikely to be the case going forward.

PR #4712, opened 2024-10-28
Remove unnecessary functions.inc.php and functions.php imports

GChuf:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/15799
- [x] Ready for review

### DIFF
--- a/src/usr/local/www/index.php
+++ b/src/usr/local/www/index.php
@@ -44,7 +44,6 @@ ob_start(null, "1000");
 
 ## Load Essential Includes
 require_once('guiconfig.inc');
-require_once('functions.inc');
 require_once('notices.inc');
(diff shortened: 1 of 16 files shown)


jim-p:
Has every function of every widget been tested with this change to ensure it doesn't break anything? While it is true that `functions.inc` has been deprecated, it includes several other files which may still be required by these widgets.

This seems like a lot of risk for little/no reward.

GChuf:
I've tested all the widgets myself. I'm actively using most of these widgets on my "active" pfsense installation and I have had no problems anywhere after applying the change.
If it's deprecated, it makes sense to me to start phasing it out, even it there's some extra work involved.

Someone else should test the changes as well. Another option would be to check if all the functions are still properly defined, with some kind of automated script ...

PR #4715, opened 2024-11-06
nologdefaultpass is not translated to logdefaultpass, #15820

antoncxx:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/15820
- [x] Ready for review


marcos-ng:
This would require bumping the config version and adding upgrade code to handle previous configs. I don't think it's worth doing now for a semantics issue. It'd be best to rewrite those settings to allow for more granular control over the default rules (e.g. include control over the snort2c rules).

PR #4717, opened 2024-11-18
UPnP IGD & PCP: Simplify UI and finalise wording and descriptions update

Self-Hosting-Group:
- Simplify UI by creating advanced settings, including STUN configuration, and move many options there, and finalise updated wording and descriptions
- Wording update: clear -> delete all port maps
- Keep ports numeric on status page by using pfctl with `-P`
- No need to enter STUN port and update example servers
  Include new server name (`stun.counterpath.com`), the previous one was a CNAME to it, and provide a documented server on the default port

This PR completes committed changes #4689 and #4695 for the UPnP IGD & PCP/NAT-PMP autonomous port mapping service.

### Settings UI:

pfsense-settings
_[Screenshot 2.7.3 release](https://github.com/user-attachments/assets/4cf59c42-565b-4d92-8243-7ebeec58369c)_

### Status UI:

pfsense-status
_[Screenshot 2.7.3 release](https://github.com/user-attachments/assets/957c6088-adc5-44ac-8898-ae06db1455bd)_

The **Port Control Protocol (PCP)** is the successor to NAT-PMP, has similar protocol concepts and packet formats, but adds support for **IPv6 port maps** and options/extensions. For more information, see:
**Port Mapping Protocols Overview and Comparison 2024 - About UPnP IGD & PCP/NAT-PMP**
https://github.com/Self-Hosting-Group/wiki/wiki/Port-Mapping-Protocols-Overview

### DIFF
--- a/src/usr/local/pkg/miniupnpd.xml
+++ b/src/usr/local/pkg/miniupnpd.xml
@@ -105,8 +105,8 @@
 				STUN Server, either hostname or IP address.</br></br>
 
 				Some public STUN servers:</br>
-				stun.xten.com</br>
-				stun.l.google.com (on non standard port 19302)</br>
+				stun.counterpath.com</br>
+				stun.cloudflare.com</br>
 				]]>
 			</description>
 		</field>
--- a/src/usr/local/pkg/miniupnpd.inc
+++ b/src/usr/local/pkg/miniupnpd.inc
@@ -121,7 +121,7 @@
 			if (!$post['stun_host'] || (!is_ipaddrv4($post['stun_host']) && !is_hostname($post['stun_host']))) {
 				$input_errors[] = 'A valid IP address or hostname for \'STUN Server\' must be specified';
 			}
-			if (!$post['stun_port'] || !is_port($post['stun_port'])) {
+			if ($post['stun_port'] && !is_port($post['stun_port'])) {
 				$input_errors[] = 'A valid port number for \'STUN Port\' must be specified';
 			}
 		}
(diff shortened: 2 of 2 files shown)


src/usr/local/pkg/miniupnpd.xml
```diff
@@ -140,9 +140,7 @@
 				<![CDATA[
 				STUN Server, either hostname or IP address.</br></br>
 
-				Some public STUN servers:</br>
-				stun.xten.com</br>
-				stun.l.google.com (on non standard port 19302)</br>
+				Some public STUN servers: stun.cloudflare.com and stun.counterpath.com</br>
```
marcos-ng, line 143: Personally I'd prefer to leave this to the online docs rather than including it in the GUI. Since it's already there however, keep the same format / line breaks.

src/usr/local/www/status_upnp.php
```diff
@@ -61,7 +61,7 @@
 ?>
 
 <div class="panel panel-default">
-	<div class="panel-heading"><h2 class="panel-title"><?=gettext("Active Service Port Maps")?></h2></div>
+	<div class="panel-heading"><h2 class="panel-title"><?=gettext("Active UPnP IGD &amp; PCP/NAT-PMP Port Maps")?></h2></div>
```
marcos-ng, line 64: Better to go with:
```
<?=htmlentities(gettext("Active UPnP IGD & PCP/NAT-PMP Port Maps"))?>
```

marcos-ng (changes requested):
This won't make it to 24.11 unfortunately.

marcos-ng:
The changes look good to me. A redmine issue will need to be opened to update the docs:
https://docs.netgate.com/pfsense/en/latest/services/upnp.html

Separate the port requirement change to its own commit, and put the text change in with the "wording update" commit.

marcos-ng:
> A comment on the previous one?

I'm not clear, what are you referring to? Regarding the docs update, I can create a redmine for it once this is in.

marcos-ng:
I don't believe the change will affect any existing systems. Removing the option would require additional changes, hence it's best to leave it for now.

marcos-ng:
Yes, though we don't have a date for that yet.

jim-p:
We'll take those into consideration for potential future changes, but for now it's OK as it is.

PR #4720, opened 2024-12-07
Feature : Add mTLS feature to LDAP Authentencation servers

Inocious:
- [X] Feature
- [X] Ready for review


jim-p:
LDAP client certificate support is already implemented in Plus: https://docs.netgate.com/pfsense/en/latest/usermanager/ldap.html

This code would cause a conflict with that implementation.

Neustradamus:
It is possible to add in opensource code?

PR #4726, opened 2025-02-03
Update Gandi livedns with PAT instead of API Key

JasperSurmont:
- [X] Redmine Issue: https://redmine.pfsense.org/issues/15258
- [X] Ready for review

Gandi has stopped supporting API keys and instead moved to PAT keys which requires the Bearer attribute in the Authorization header instead of the Apikey attrubite.

Read more [here](https://api.gandi.net/docs/authentication/)


marcos-ng:
Thanks! Changes added with https://github.com/pfsense/pfsense/commit/e329a2fbb202af69901d1c2dba7f216f81956a40.

PR #4729, opened 2025-03-21
Don't unset v4 aliases when invoked with inet6.  Fixes #16103.

strix-technica:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/16103
- [ ] Ready for review


marcos-ng:
Done with https://github.com/pfsense/pfsense/commit/e7f3a037364743a1124c589136e947ce98c1b9ed

PR #4730, opened 2025-04-03
UPnP IGD & PCP: Add `Allow Third-Party Mapping` option

Self-Hosting-Group:
To allow adding port maps for non-requesting IP addresses, helps
https://redirect.github.com/opnsense/plugins/pull/3727

- [X] Redmine Issue: https://redmine.pfsense.org/issues/16273
- [X] Ready for review

### DIFF
--- a/src/usr/local/pkg/miniupnpd.xml
+++ b/src/usr/local/pkg/miniupnpd.xml
@@ -35,13 +35,13 @@
 		<field>
 			<name>Service Settings</name>
 			<type>listtopic</type>
-			<enablefields>enable_upnp,enable_natpmp,ext_iface,iface_array,download,upload,overridewanip,upnpqueue,logpackets,sysuptime,permdefault</enablefields>
+			<enablefields>enable_upnp,enable_natpmp,ext_iface,iface_array,download,upload,overridewanip,allow_third_party_mapping,upnpqueue,logpackets,sysuptime,permdefault</enablefields>
 		</field>
 		<field>
 			<fielddescr>Enable</fielddescr>
 			<fieldname>enable</fieldname>
 			<type>checkbox</type>
-			<enablefields>enable_upnp,enable_natpmp,ext_iface,iface_array,download,upload,overridewanip,upnpqueue,logpackets,sysuptime,permdefault</enablefields>
+			<enablefields>enable_upnp,enable_natpmp,ext_iface,iface_array,download,upload,overridewanip,allow_third_party_mapping,upnpqueue,logpackets,sysuptime,permdefault</enablefields>
 			<description>Enable port mapping service</description>
 			<sethelp>
 				<![CDATA[
@@ -191,6 +191,16 @@
 		<field>
 			<name>Advanced Settings</name>
 			<type>listtopic</type>
+		</field>
+		<field>
+			<fielddescr>Allow Third-Party Mapping</fielddescr>
+			<fieldname>allow_third_party_mapping</fieldname>
+			<description>
+				<![CDATA[
+				Allow adding port maps for non-requesting IP addresses
+				]]>
+			</description>
+			<type>checkbox</type>
 		</field>
 		<field>
 			<fielddescr>Download Speed</fielddescr>
--- a/src/usr/local/pkg/miniupnpd.inc
+++ b/src/usr/local/pkg/miniupnpd.inc
@@ -236,8 +236,10 @@
 					$config_text .= "system_uptime=yes\n";
 				}
 
-				/* set secure_mode */
-				$config_text .= "secure_mode=yes\n";
+				if ($upnp_config['allow_third_party_mapping'] == 'on') {
+					$config_text .= "secure_mode=no\n";
+					$config_text .= "pcp_allow_thirdparty=yes\n";
+				}
 
 				/* set webgui url */
 				$webgui_config = config_get_path('system/webgui');


fivemtebex:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemhostingserver:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

scriptforfivem:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemserverhostings:
This is a well-structured codebase. I appreciate the clean separation of concerns.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -286,9 +295,9 @@
 				criteria. The entries only control access for IPv4 clients, they do not apply
 				to IPv6 clients.
 				<br/><br/>
-				Format: <tt>[allow or deny] [ext port or range] [int ipaddr or ipaddr/CIDR] [int port or range]</tt>
+				Format: <tt>(allow|deny) (ext port|range) (int IP|IP/netmask) (int port|range)</tt>
```
marcos-ng, line 298: The current text is fine since this is more a matter of preference and the current text may be clearer to some.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -286,9 +295,9 @@
...
 				<br/>
-				Example: <tt>allow 1024-65535 192.168.0.0/24 1024-65535</tt>
+				Example: <tt>allow 1024-65535 192.168.1.0/24 1024-65535</tt>
```
marcos-ng, line 300: The current text is fine since this is simply an example and isn't meant to follow any system defaults.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -209,7 +218,7 @@
 			<fieldname>upload</fieldname>
 			<description>
 				<![CDATA[
-				Value to report when clients query the maximum link upload speed (Kbit/s).
+				Value to report when clients query the maximum link upload speed (kbit/s).
```
marcos-ng, line 221: The current text is fine since the configuration is generated by converting the GUI-provided Kb/s value into kb/s.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -192,12 +191,22 @@
...
+			</description>
+			<type>checkbox</type>
+		</field>
 		<field>
 			<fielddescr>Download Speed</fielddescr>
 			<fieldname>download</fieldname>
 			<description>
 				<![CDATA[
-				Value to report when clients query the maximum link download speed (Kbit/s).
+				Value to report when clients query the maximum link download speed (kbit/s).
```
marcos-ng, line 209: The current text is fine since the configuration is generated by converting the GUI-provided Kb/s value into kb/s.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -192,12 +191,22 @@
 			<name>Advanced Settings</name>
 			<type>listtopic</type>
 		</field>
+		<field>
+			<fielddescr>Allow Third-Party Mapping</fielddescr>
```
marcos-ng, line 195: UPnP is inherently insecure but I'm not sure this functionality is worth exacerbating the issue.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -72,7 +71,7 @@
 			<sethelp>
 				<![CDATA[
 				Enables support for client requests using Port Control Protocol (PCP) and its
-				compatible predecessor NAT Port Mapping Protocol (NAT-PMP).
+				predecessor NAT Port Mapping Protocol (NAT-PMP).
```
marcos-ng, line 74: The current text is fine given that, AFAIK, PCP supports NAT-PMP as well.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -26,22 +26,21 @@
 			<description>
 				<![CDATA[
 				Universal Plug and Play Internet Gateway Device (UPnP IGD) and Port Control Protocol (PCP) are
-				network protocols which allow local software and devices to configure each other when attaching
-				to a network. This includes autonomously creating dynamic NAT rules to redirect and pass
-				incoming connections from remote hosts.
+				network protocols which allow local software and devices to autonomously create dynamic NAT
```
marcos-ng, line 29: The current text is fine given that clients may change their configuration as well, e.g. as part of the reported download/upload values

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -58,7 +57,7 @@
 			<sethelp>
 				<![CDATA[
 				Enables support for client requests using Universal Plug and Play Internet
-				Gateway Device (UPnP IGD).
+				Gateway Device (UPnP IGD) Control Protocol.
```
marcos-ng, line 60: Either applies so the current text is fine.

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -161,7 +160,7 @@
 				Example public STUN servers:
 				<ul>
 					<li>stun.counterpath.com</li>
-					<li>stun.cloudflare.com</li>
+					<li>stun.nextcloud.com</li>
```
marcos-ng, line 163: > it was recently discovered that the daemon needs certain features and doesn't check for their support, leading to incorrect STUN CGNAT check results and disabling the service.

Is there a link to this discussion / more details?

src/usr/local/pkg/miniupnpd.xml
marcos-ng: You're right.
> Bitrates reported by daemon in bits per second

https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/miniupnpd.conf#L130C3-L130C49

src/usr/local/pkg/miniupnpd.xml
marcos-ng: > A PCP-only gateway receiving a NAT-PMP request (identified by the first octet being zero) will interpret the request as a version mismatch. Normal PCP processing will emit a PCP response that is compatible with NAT-PMP, without any special handling by the PCP server.

https://datatracker.ietf.org/doc/rfc6887/

It sounds to me like leaving "compatible" in the pfSense GUI is fine.

src/usr/local/pkg/miniupnpd.xml
marcos-ng: Thanks. This seems fine to me given miniupnpd has updated their default:
https://github.com/miniupnp/miniupnp/commit/3f0316584ab71876e89f401ee193a63c79bcf00e

src/usr/local/pkg/miniupnpd.xml
marcos-ng: Fair enough. This should be opened as a separate feature request with an example / details of its use case.

src/usr/local/pkg/miniupnpd.xml
marcos-ng: Right. I interpret the sentence that follows as stating what "normally" happens in the case of a version mismatch.

src/usr/local/pkg/miniupnpd.xml
marcos-ng: I see, thanks. Let's keep the current text in this case.

marcos-ng:
> Only this would have to be set: `$config_text .= "ext_perform_stun=allow-filtered\n";`

This shouldn't be needed provided that the STUN server works as intended. Either way, let's keep this PR for text-only changes since behavior changes are unlikely to make it in 2.8.0.

src/usr/local/pkg/miniupnpd.xml
marcos-ng: OK.

marcos-ng:
> Am I right in thinking that the 2.3.9 daemon update will be in 2.8.0, main branch?

Unfortunately that happened after 2.8.0 was already branched. As of now both devel and 2.8.0 are on 2.3.7 hence most likely 2.8.0 will be released with its current version. It's possible for the package to be updated post-release though that's normally reserved for critical issues. I'd expect it to be in the release after instead, e.g. 2.8.1.

> However, the final solution I'm thinking of would require the daemon to open and close these ports for the test itself, without range restriction.

Yes, the tests should automatically open those ports in pf like it would normally for port mappings.

> But adding `Allow Third-Party Mapping`?

Indeed it should be its own PR with its own redmine that can be considered post 2.8.0, e.g. 2.8.1. Thanks for the link, the reasoning given there makes sense:
> My reasoning for this is to allow upnp port forwarding of kubernetes services where the client requesting the port forward will not have the same IP as that in the request.

marcos-ng:
Sounds fine to me.

marcos-ng:
Yep. Preferably feature implementations should have their own commit and be separated from unrelated changes, hence the separate redmine and PR from simple text changes (which can potentially be merged without an associated redmine).

src/usr/local/www/status_upnp.php
```diff
@@ -50,9 +50,7 @@
 	print_info_box($savemsg, 'success');
 }
 
-if (!config_get_path('installedpackages/miniupnpd/config/0/iface_array') ||
-    !config_path_enabled('installedpackages/miniupnpd/config/0')) {
-
+if (!config_get_path('installedpackages/miniupnpd/config/0/enable')) {
```
marcos-ng, line 53: The check here should be:
```
(config_get_path('installedpackages/miniupnpd/config/0/enable') != 'on')
```

src/usr/local/pkg/miniupnpd.xml
```diff
@@ -111,16 +110,16 @@
 			<type>info</type>
 			<description>
 				<![CDATA[
-				This service <strong>must</strong> be able to locate its routable public IP
-				address to function properly. If this device is behind NAT, port forwarding may
+				This service <strong>must</strong> be able to locate its routable public IPv4
+				address to function properly. If this device is behind CGNAT, port forwarding may
```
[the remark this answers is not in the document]
marcos-ng: Let's keep the current text.

marcos-ng:
I would like redmine feature request created specifically for the feature commit and a PR dedicated for that. It can remain on this PR and the rest of the changes moved to a different one, or vice versa. For the text changes, I'm referring to the currently unresolved conversations on this PR.

I appreciate the work, thank you!

src/usr/local/pkg/miniupnpd.xml
marcos-ng: Unless it's shown that pfSense/miniupnpd won't handle NAT-PMP clients when only PCP is enabled, let's keep the "compatible" part.

src/usr/local/pkg/miniupnpd.xml
marcos-ng: Remove this change.

marcos-ng:
- Remove the changes mentioned in this PR along with the new changes in the service-fixes branch (the `IPv4` and `<default_value>3478</default_value>` diffs).
- Create a redmine issue for each PR here https://redmine.pfsense.org/projects/pfsense then reference the ID in the commit title; e.g. for this PR make it "UPnP IGD & PCP: Add Allow Third-Party Mapping UI option. Implement #00000". Make sure each PR has commits squashed.
- Updates to the public documentation are handled on redmine here https://redmine.pfsense.org/projects/pfsense-docs

marcos-ng:
> the STUN port has not been necessary since last year's PR 
I forgot about that; it's fine then. Just make sure to include a note about this in the commit message.

Self-Hosting-Group:
Any news about this PR or https://redmine.pfsense.org/issues/16273?

marcos-ng:
The consensus seems to be that the security implications of this feature are not worth the benefit.

PR #4731, opened 2025-05-13
auth.inc: Fixing TLS Setup for LDAP Authentication

jakobsen-lrz:
- [ ] Redmine Issue: https://redmine.pfsense.org/issues/15060
- [ ] Ready for review

I changed the `ldap_setup_caenv ` function, to set the `LDAP_OPT_X_TLS_REQUIRE_CERT ` option globally, and moved calls of `ldap_setup_caenv ` to before calls of `ldap_connect`, as is required. According to https://www.php.net/manual/en/function.ldap-set-option.php, all TLS Options must be set globally, before calling `ldap_connect`.

I think this causes some weirdness when having multiple LDAP servers, with different certificates, as described similarly in the Redmine issue.

Thanks.

### DIFF
--- a/src/etc/inc/auth.inc
+++ b/src/etc/inc/auth.inc
@@ -1088,6 +1088,9 @@ function ldap_test_connection($authcfg) {
 		return false;
 	}
 
+	/* Setup CA environment if needed. */
+	ldap_setup_caenv($authcfg);
+
 	/* connect and see if server is up */
 	$error = false;
 	if (!($ldap = ldap_connect($ldapserver))) {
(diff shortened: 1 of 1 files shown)


jim-p:
Unless something changed in PHP, this isn't likely to be a viable fix. I've tried this method in the past and it broke in other ways. It's worth checking again, however, it may take some time to setup a proper test environment.

jim-p:
With the patch applied it's still extremely inconsistent. It works maybe 1 out of 3 tries when switching between two LDAP servers (one with a cert from a custom self-signed CA, the other with an ACME cert). Sometimes it also fails without switching CAs.

It will log one of two errors:
```
/diag_authentication.php: ERROR! Could not bind to LDAP server ldapselfca. Please check the bind credentials. Error Message: error:0A000086:SSL routines::certificate verify failed (self-signed certificate in certificate chain). 
```
```
/diag_authentication.php: ERROR! Could not bind to LDAP server ldapacme. Please check the bind credentials. Error Message: error:0A000086:SSL routines::certificate verify failed (unable to get local issuer certificate). 
```

And yet if I try again after a few moments the exact same connection works, and then maybe fails again right after.

I've been fighting with this code for many, many years. I can't recall everything I've tried, but some of the history is here: https://redmine.pfsense.org/issues/9417

jakobsen-lrz:
There is this [php issue](https://github.com/php/php-src/issues/12081) in which the user says, setting it using putenv and ldap_set_option works. Have you tried that aswell?

jakobsen-lrz:
I moved the change to the logging of error messages into a seperate Pull Request.

jim-p:
We used `putenv` before and it was worse. You had to completely restart PHP-FPM+nginx to get it to switch between CAs and other issues.

See the commits on https://redmine.pfsense.org/issues/9417 for more history. That PHP bug is also linked on there.

jakobsen-lrz:
Ill close this Pull Request.

jim-p:
We really appreciate the effort -- I, in particular, would love to see this working after it's been a known problem for so long.

The main method of drawing out the problem is having two separate LDAP servers which each have different CA certificates, and at least one of them is *not* globally trusted. When doing that, it has to change the CA info before connecting but because of the way PHP/LDAP libraries/environments work, the settings either end up not fully taking effect or stomping on other concurrent queries. If the variables passed into the PHP LDAP library worked properly it should be a non-issue, so it may just need more pushback upstream with better test cases.

There are some potential workarounds but none that preserve the security of validating that the server CA matches a specific entry for one (or more) different LDAP servers.

PR #4733, opened 2025-05-15
Add option to control ALTQ support for VTNET NICs

jakobsen-lrz:
This changes the description of the hnaltqenable option, to reflect this change. And the generation of tunables to use the new `hw.vtnet.altq_disable` tunable.

I co-opted the hnaltqenable option under `system/advanced/networking`, and the config.xml option `system/hn_altq_enable` to implement this, because i did not want to change the format of config files used, and because it should be impossible for HN and VTNET NICs to exist on the same pfsense.

I hope this is fine, as the [change ](https://github.com/freebsd/freebsd-src/pull/1679) was merged in some time ago, in the upstream freebsd repository, but isnt included yet in the pfsense/freebsd-src repository.

- [ ] Redmine Issue: https://redmine.pfsense.org/issues/16166
- [ ] Ready for review


jim-p:
We are in the middle of a release cycle at the moment so we have not synchronized the source with upstream recently. Once we have released CE 2.8.0 and Plus 25.03 and pulled in updated upstream source we can take a look at this.

jakobsen-lrz:
Do you synchronize the sources before the next release? So will pfsense CE 2.8.0 be based on the FreeBSD 15 Release, which isnt released before December? Even if this PR isnt added included in the pfsense CE 2.8.0 release, would i still like to use the new tunable.

jim-p:
During development cycles we periodically sync with FreeBSD-CURRENT/HEAD not a specific FreeBSD release.

We mark the commit at which we sync upstream on this page, though for in-development versions like Plus 25.07 that won't be fully updated until just before release: https://docs.netgate.com/pfsense/en/latest/releases/versions.html#x

chrcoluk:
This is great you considering the request, , if its after 2.8.0 please put out a minor update for CE after its in.

jim-p:
The `hw.vtnet.altq_disable` tunable is present now on internal 2.9 testing snapshots but there are other problems we are working on from the upstream src merge that prevent those snapshots from working properly yet. As such, this will also need to wait until after we get Plus 25.03 out so we can dedicate more resources to addressing issues with the latest upstream changes. But that shouldn't be too much more time at its current stage of its release cycle.

jakobsen-lrz:
Hello, can we get this MR merged in soon? If so i would fix the Conflicts.

jakobsen-lrz:
I have implemented the changes you requested.

PR #4736, opened 2025-06-11
rrd bump support to 200gbps from 20gbps

steven-agg:
anything over 20gbps would return null in the monitoring. this increases to 200gbps.

- [X] Redmine Issue: https://redmine.pfsense.org/issues/16257
- [X] Ready for review


spnkr:
+1 / bump

PR #4737, opened 2025-06-17
UPnP IGD & PCP: Fix status page and improve help

Self-Hosting-Group:
- Fix status page to only show when service is enabled, corrects 63d6bb4
- Remove default STUN port as no more needed by bf31326
- Only propose compatible STUN servers in help, use the newer wording from RFC 5389 and improve help slightly
- Arrange `Advanced Settings` options and add `IGD Settings`
- Propose docs.netgate.com updates in https://github.com/Self-Hosting-Group/pfsense/tree/doc-updates

- [X] Ready for review

### DIFF
--- a/src/usr/local/pkg/miniupnpd.xml
+++ b/src/usr/local/pkg/miniupnpd.xml
@@ -120,7 +120,7 @@
 				address, the service will refuse to map any ports without additional
 				configuration.
 				<br/><br/>
-				If the <strong>External Interface</strong> is behind unrestricted NAT with
+				If the <strong>External Interface</strong> is behind unrestricted endpoint-independent (1:1) NAT with
 				incoming traffic forwarded from upstream without any filtering, this service can
 				still function so long as it can locate its public IP address.
 				<br/><br/>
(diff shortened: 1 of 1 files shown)


marcos-ng:
- Remove the "Arrange `Advanced Settings` options and add `IGD Settings`" change. It will just get in the way of accepting and merging the PR :)
- Remove the "Fix status page" change. There's more that should be done there; I'll create a redmine and commit specifically for this fix. Also FWIW the issue existed before the commit referenced in the message.
- With the above changes we won't need a redmine so you can set the commit title to "Improve UPnP IGD & PCP help text".

With the above done and no new changes added this PR will be ready to merge.

marcos-ng:
Yes - it will be a "smaller" release. Let's not reorder the GUI fields since that would necessitate additional work elsewhere and would delay merging the already-accepted changes for the next release.

marcos-ng:
"UPnP IGD & PCP: Improve help text" is fine - make sure to squash the commit.

marcos-ng:
CE 2.8.1 is based off of 2.8.0 and its changes are limited. Unfortunately the new version won't be included there but it will be available after 2.8.1.

PR #4740, opened 2025-08-19
Update dyndns.class to support sending the IP address and preserving …

Godwottery:
Simpler version of https://github.com/pfsense/pfsense/pull/4543 and https://github.com/pfsense/pfsense/pull/4542

This one sets the IP reported to the API to that detected by the pfsense device, and preserves the non-updated IP.

Per the DeSEC documentation: 
https://desec.readthedocs.io/en/latest/dyndns/update-api.html

'''It is recommended to always specify query string parameters for both IP address types. If your device does not have both types, use preserve for the non-existing address. You can also use the empty value (which will delete any existing such address).'''

This adds the detected IP, and sets the other IP (v6 if updating v4, and v4 if updating v6) to be preserved.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/12494
- [x] Redmine Issue: https://redmine.pfsense.org/issues/12495
- [x] Ready for review

### DIFF
--- a/src/etc/inc/dyndns.class
+++ b/src/etc/inc/dyndns.class
@@ -1817,11 +1817,11 @@
 					curl_setopt($ch, CURLOPT_POSTFIELDS, json_encode($body));
 					break;
 				case 'desec':
-					curl_setopt($ch, CURLOPT_URL, 'https://update.dedyn.io/?hostname=' . $this->_dnsHost);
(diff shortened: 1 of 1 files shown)


marcos-ng:
Changes look good. Please squash the commits and update the commit message. You can use the title of the redmine - also include the issue number. For example: "Preserve other record types when updating IPv4 or IPv6 using deSEC DDNS. Fix #12495".

PR #4751, opened 2026-05-10
Feature: Add MFA framework + hooks to OpenVPN (Redmine #12546)

krystakadlec:
- [#12546] Redmine Issue: https://redmine.pfsense.org/issues/12546
- [x] Ready for review

This PR implements a framework for supporting MFA providers natively within pfSense and necessary hooks for its integration in OpenVPN authentication. This addresses ticket #12546 for 2FA support to Local Database Authentication.

**Description**
Currently, adding 2FA directly to pfSense services requires either relying on external RADIUS infrastructure or hardcoding specific solutions into core files. This PR creates an abstraction layer to solve this:

* **Generic MFA Provider Pool:** Establishes a standard where external packages can register their authentication scripts into a global `$config['mfa_providers']` array.
* **OpenVPN GUI Integration:** Updates `vpn_openvpn_server.php` with UI elements. Administrators can now enable secondary authentication on a per-server basis and select from installed MFA providers. Includes input validation for preventing incorrect configurations.
* **Authentication Hooks:** Modifies `openvpn.auth-user.php` to step into the authentication process after a successful primary authentication. If enabled, it dynamically includes the selected MFA provider's file and executes its validation function.
* **Client IP Tracking:** Updates `ovpn_auth_verify` and `ovpn_auth_verify_async` to pass `untrusted_ip` and `untrusted_ip6` arguments to the PHP backend. This ensures the MFA provider has access to the client's IP address (for location-based policies or logging).

**Implementation Note**
Please note that this PR implements only the core framework and core hooks for OpenVPN. It does not contain any specific provider logic. This serves as the prerequisite preparation. A separate pfSense package utilizing this new infrastructure is prepared and submitted at https://github.com/pfsense/FreeBSD-ports/pull/1440


joe16joe16:
Great to see movement on this — native MFA support has been requested for years (#12546) and the RADIUS/FreeRADIUS route is significant overkill for most single-firewall setups.

One question on scope: this PR wires the MFA provider framework into OpenVPN's auth-user.php hook. Is there a plan to also expose the same provider registry to the WebConfigurator local database login (System > User Manager)? For a lot of users, VPN 2FA is nice to have, but securing the admin GUI login itself without standing up FreeRADIUS is the actual pain point. If the registry is generic enough, hooking it into the local user auth path as a follow-up PR would cover both cases with one provider implementation (e.g. TOTP) instead of maintaining separate integrations per service.

It would also be great to see this MFA provider framework extended to the other VPN backends pfSense ships — IPsec (xauth) and especially WireGuard. I'm aware WireGuard's protocol is keypair-based with no built-in username/password or auth-hook step, so a "second factor" there would likely need to sit in front of the tunnel (e.g. gating peer config activation, or a pre-auth portal) rather than inline like OpenVPN's auth-user hook — but if the provider registry from this PR stays generic, that's a solvable follow-up rather than a redesign.

Also curious whether a plain TOTP provider (RFC 6238, no external API) is planned alongside the Duo backend — not everyone wants a cloud dependency for MFA on their firewall.

PR #4752, opened 2026-05-16
fix: [WebGUI] improve filesystem isolation in diag_command downloads

skraft9:
- [ ] Redmine Issue: N/A (Security hardening / CVE-2025-53392)
- [x] Ready for review

#### What does it do?
This PR hardens the diagnostic file download handler in `diag_command.php` to ensure path integrity and resource isolation. 

The current implementation allows for unrestricted filesystem access via the `dlPath` parameter, which can be abused by authenticated users to read sensitive system files outside of the intended diagnostic scope. This change introduces strict path canonicalization and directory safelisting.

**Changes:**
- Implemented `realpath()` to resolve canonical paths and neutralize directory traversal attempts.
- Added a directory safelist restriction, limiting downloads to verified diagnostic locations (`/tmp`, `/var/log`, and `/cf/conf`).
- Integrated security logging via `log_error()` for unauthorized access attempts to improve system auditability.
- Retained existing file upload functionality safely.


jim-p:
The explicit intent of that page is to freely allow all of those actions. Users who do not wish to allow that can limit access to the page. Pretending it can be effectively limited in any meaningful way gives users a false sense of security.

PR #4754, opened 2026-06-01
gwlb.inc patch to account for changes in `dpinger_dont_add_static_route` - redmine 16861

luckman212:
If a gateway monitor IP is set, and then later the `dpinger_dont_add_static_route` option is enabled, the static route is NOT removed.

- [x] Redmine Issue: https://redmine.pfsense.org/issues/16861
- [x] Ready for review

Came up with this patch, which I tested successfully on my 6100 running 26.03.1

It also accounts for the case when the global option is changed from System → Advanced → Miscellaneous

**N.B.** the reason for these few extra changed lines is I couldn't resist the urge to fix the mismatched whitespace just above

image

### DIFF
--- a/src/etc/inc/gwlb.inc
+++ b/src/etc/inc/gwlb.inc
@@ -2663,19 +2663,19 @@ function save_gateway($gateway_settings, $realid = "", $json = false) {
 		$a_gateway_item = config_get_path("gateways/gateway_item/{$realid}");
 		/* NOTE: If gateway ip is changed need to cleanup the old static interface route */
 		if ($gateway_settings['monitor'] != "dynamic" &&
-		    !empty($a_gateway_item) &&
-		    is_ipaddr($a_gateway_item['gateway']) &&
-		    $gateway['gateway'] != $a_gateway_item['gateway'] &&
-		    isset($a_gateway_item["nonlocalgateway"])) {
+			!empty($a_gateway_item) &&
+			is_ipaddr($a_gateway_item['gateway']) &&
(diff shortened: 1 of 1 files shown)


marcos-ng:
I'm closing this PR in favor of a different change (and corresponding credit) - the main difference being the omission of the global preference check because in this case it can conflict with other routing configuration. Handling of the "dpinger_dont_add_static_route" global preference is better left to a separate change and discussion.

For reference the mixed tabs and spaces are part of the project style guide (referenced in the docs). My preference would be to simply apply standard formatting guidelines to everything but that has its own cons. Thanks for the contribution!

luckman212:
Ok @marcos-ng 
Understood- is the other fix publicly viewable? or it's only on the private gitlab?

luckman212:
I found the note in the [Developer Style Guide](https://docs.netgate.com/pfsense/en/latest/references/developer-style-guide.html), in case someone stumbles onto this

CleanShot 2026-06-01 at 15 24 58

marcos-ng:
> Ok @marcos-ng Understood- is the other fix publicly viewable? or it's only on the private gitlab?

I included it in the redmine.

luckman212:
In case someone wants a URL that can be applied in System Patches, I copied marcos' patch from the redmine into a gist:

```
https://gist.githubusercontent.com/luckman212/779fba075440e75537094b9917e36a1d/raw/16dcd1956ad4f5aadbe8bf73487daa706daef12e/gwlb.patch
```

PR #4758, opened 2026-07-23
Add OpenID Connect (SSO) authentication server type for the webConfigurator

cdmx-in:
- [x] Redmine Issue: https://redmine.pfsense.org/issues/16967
- [x] Ready for review

## Summary

Adds a third authentication server type, `oidc`, alongside LDAP and RADIUS, allowing the webConfigurator to authenticate against an OpenID Connect identity provider (Google Workspace, Entra ID, Okta, Authentik, Keycloak).

A provider is configured under **System > User Manager > Authentication Servers** with only an issuer URL, client ID and client secret; endpoints come from OIDC discovery and are cached for an hour.

## Why it is wired into `session_auth()`

OIDC is a redirect flow, not a synchronous credential check, so it cannot be another `*_backed()` function called from `authenticate_user()`. It is driven from `session_auth()` — which already handles the interactive login POST — via `?oauth=start` and `?oauth=callback`. This deliberately avoids adding a new pre-authentication endpoint to the GUI, and lets the successful-login path reuse the existing session setup verbatim.

`authenticate_user()` returns false for an `oidc` server, so a password can never authenticate against one.

## Authorisation

Reuses the existing remote-user model: no local user account is required. `getUserEntry()` already returns a synthetic entry when the auth mode is not Local Auth, and privileges are derived by mapping a configurable group claim onto pfSense groups — the same path as LDAP and RADIUS group mapping, extended in `getUserGroups()`, `get_user_privileges()` and `getAllowedPages()`.

A user whose claims map to no pfSense group receives no privileges.

## Security

- Authorization code flow with PKCE (`S256`); `state` and `nonce` generated per request via `random_bytes()` and validated on return.
- Tokens are fetched over a TLS-verified back channel (`CURLOPT_SSL_VERIFYPEER`/`VERIFYHOST` on), so the ID token is trusted without local JWKS signature verification. `iss`, `aud`, `exp` and `nonce` are all validated. Local signature verification would only be required if tokens were ever accepted via the front channel.
- Logins may be restricted to one or more hosted domains, matched against the `hd` claim or the email domain.

## Optional: disable password login

Adds a **Disable password login** option under **System > User Manager > Settings** so a firewall can be made SSO-only. It is rejected on save unless the selected authentication server is an OIDC server, which prevents configuring a lockout.

It affects only the interactive login form and the local-database fallback in `session_auth()`. XML-RPC (HA sync) authenticates through a separate path and is intentionally unchanged. Recovery from a broken SSO configuration is via console or SSH, consistent with existing lockout scenarios.

## Testing

Verified against live Google Workspace: discovery, the authorization redirect, token exchange, ID-token claim validation and login all succeed end to end, resolving to the expected username with `authsource` `OIDC/<name>`.

Claim validation was additionally exercised with unit assertions covering wrong `aud`, expired token, wrong and empty `nonce`, wrong issuer, disallowed domain, email-domain fallback, and array-valued `aud`.

Not yet exercised against a live provider, and worth reviewer attention:

- **Group claim mapping.** Plain Google OAuth emits no groups claim, so the live login resolved to no groups. The mapping code paths are exercised only by inspection.
- **Domain restriction** end to end. It is covered by the unit assertions above, but the live round trip ran with an empty allow-list.

All changed files pass `php -l`.


roney492:
Any update on this?

PR #4760, opened 2026-09-06
Experimental MBIM support via umb(4) and umbctl(8)

LuKePicci:
- [ ] Redmine Issue: no issue
- [x] Ready for review

Supports recovering of the PDP context while moving, partial reconfiguration on IP address change, full reconfiguration in case modem resets, PIN unlock on full reconfig, one time PUK and general modem info reporting.

Developed and tested on latest public pfSense/pfSense RELENG_2_7_2 branch, needed a backported [umb(4)](https://man.freebsd.org/cgi/man.cgi?query=umb&apropos=0&sektion=4&manpath=FreeBSD+15.0-RELEASE&format=html) driver and [umbctl(8)](https://man.freebsd.org/cgi/man.cgi?query=umbctl&apropos=0&sektion=8&manpath=FreeBSD+15.0-RELEASE&format=html) from FreeBSD 15.0-RELEASE to latest pfSense/FreeBSD-src RELENG_2_7_2 branch kernel tree.

In the example, a working MBIM interface called WWAN from a Sierra Wireless MC7710 dongle.

Notes:
- kernel note: umb driver currently does not support plug'n'play dynamic loading, currently loaded at boot time with umb_load="YES" in loader.conf.local)
- tested HW note 1: the tested MC7710 and possibly other older MBIM dongles do not expose the standard-required Union Descriptor, the umb driver was also patched to support falling back to IAD in such cases. Patch is under review at [usb: umb: locate MBIM data interface using IAD (reviews.freebsd.org/D58641)](https://reviews.freebsd.org/D58641)
- tested HW note:2: an usb quirk is needed in loader.conf.local for MC7710 to make the kernel select the secondary USB composition config when MC7710 is set to a double configuration UD composition. Single configuration UD composition that do only expose MBIM interface is not recommended as it make difficult to move away the dongle from there once set. Single configuration UD composition that do also expose other AT control interfaces together with MBIM are fine too.

## Home dashboard

⚠️ Please note: screenshots captured before renaming UMB to MBIM

### Interfaces 

image

###  Gateways

image

### Interface Assignments

image

### Interface configuration

image
image

## System: Routing

image

## Status: Interfaces

image


LuKePicci:
Just renamed UMB to MBIM as a protocol type. References to upstream umb as kernel driver, real interface names and umbctl are unchanged.

